PR #1, opened 2022-02-14
[FLINK-26078] Kubernetes Operator Prototype

gyfora:
This is the initial prototype of the kubernetes operator. Please feel free to review and leave comments for larger issues but rough edges are expected in this first version.


gyfora:
cc @wangyang0918

examples/basic-checkpoint-ha.yaml
```diff
@@ -0,0 +1,61 @@
...
+apiVersion: flink.apache.org/v1alpha1
+kind: FlinkDeployment
+metadata:
+  namespace: default
+  name: basic-checkpoint-ha-example
+spec:
+  image: flink:1.14.3
+  flinkVersion: 1.14.3
+  flinkConfiguration:
+    taskmanager.numberOfTaskSlots: "2"
```
Grypse, line 28: Task Manager already has slot configuration. Is Flink Configuration redundant?

gyfora: Good catch, the config is definitely redundant here in the yamls and I will clean it up.
In general for application clusters we would not even need this config in either place (TM or config) as it is automatically computed from the parallelism and number of TMs.

For session clusters it is a very important config so I would still keep it under the taskManager section. 

What do you think?

wangyang0918:
@gyfora Thanks for creating the prototype with the first version. I have tested the functionality in a minikube with basic example. It works really well. Great job :)

Except for the implementation details, I find some issues while doing the test step by step.
* It seems the project could not be compiled with java8. Is is an expected limitation?
* I find many overlapping classes of fabric8 `kubernetes-client` when building the project. Because we have a direct dependency with 5.12.1 and a transitive dependency in `flink-kubernetes` with 5.5.0.
* When I delete the JobManager deployment accidentally, it is not recreated again. Is it expected?

README.md
```diff
@@ -1,2 +1,69 @@
...
+In order to use the webhook for FlinkDeployment validation, you must install the cert-manager on the Kubernetes cluster:
+```
+kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v1.7.1/cert-manager.yaml
+```
+
+The webhook can be disabled during helm install by passing the `--set webhook.create=false` parameter or editing the `values.yaml` directly.
+
+## User Guide
+### Create a new Flink deployment
+The flink-operator will watch the CRD resources and submit a new Flink deployment once the CR it applied.
```
martin-g, line 21: `... is applied` ?

README.md
```diff
@@ -1,2 +1,69 @@
...
+
+kubectl delete flinkdep {dep_name}
+```
+
+### Get/List Flink deployments
+Get all the Flink deployments running in the K8s cluster
+```
+kubectl get flinkdep
+```
+Describe a specific Flink deployment to show the status(including job status, savepoint, ect.)
```
martin-g, line 40: s/ect/etc/

README.md
```diff
@@ -1,2 +1,69 @@
...
+When using `minikube tunnel` the rest service is exposed on `localhost:8081`
+```
+> minikube tunnel
+
+> kubectl get services
+NAME                         TYPE           CLUSTER-IP     EXTERNAL-IP   PORT(S)             AGE
+basic-session-example        ClusterIP      None           <none>        6123/TCP,6124/TCP   14h
+basic-session-example-rest   LoadBalancer   10.96.36.250   127.0.0.1     8081:30572/TCP      14h
+```
+The operator pics up the default log and flink configurations from `/opt/flink/conf`. You can put the rest configuration parameters here:
```
martin-g, line 64: s/pics/picks/

gyfora:
@wangyang0918 thanks for the review and the comments:

1. It seems that the operator sdk itself requires minimum java 11 and I dont immediately see a workaround to support java 8 with the current code. While originally we wanted to support java 8 to match Flink, I think if this becomes technically difficult I think we can live with java 11 as this would not affect Flink jobs deployed by the operator (they can still be java 8). So this only really affects the build env and container runtime. What do you think? cc @tweise 

2. We should try to resolve this somehow, maybe by excluding the transitive fabric8 dep from `flink-kubernetes`, will take a look

3. Right now the operator does not recover the flink job if something like this happens (as deleting the deployment is beyond the native integration) but we should defintely support this, I will file a jira for it.

pom.xml
```diff
@@ -0,0 +1,286 @@
...
+        <flink.version>1.14.3</flink.version>
+        <flink.shaded.version>15.0</flink.shaded.version>
+
+        <slf4j.version>1.7.15</slf4j.version>
+        <log4j.version>2.17.1</log4j.version>
+
+        <spotless.version>2.4.2</spotless.version>
+        <awaitility.version>4.1.0</awaitility.version>
+        <it.skip>true</it.skip>
+        <mockito.version>2.21.0</mockito.version>
```
martin-g, line 74: Why such old Mockito version ?

gyfora:
@wangyang0918 The problem with `flink-kubernetes` is that it is not a proper maven dependency but we release a fatjar, this means we cannot really exclude the fabric8 from it because its simply part of the jar.

This is something we should resolve on the Flink side because it doesnt make too much sense the way it is now. What we could do as a temporary workaround is that we repackage it in a new module and remove / relocate the shaded classes so that it does not conflict with the proper fabric8 deps accidentally

pom.xml
gyfora: Not sure but let's upgrade it, I agree :)

pom.xml
```diff
@@ -0,0 +1,285 @@
...
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+    <parent>
+      <groupId>org.apache</groupId>
+      <artifactId>apache</artifactId>
+      <version>20</version>
+    </parent>
```
dannycranmer, line 26: Should we use `flink-parent` here so we can inherit the same quality plugins etc?

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/reconciler/JobReconcilerTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+        // Test stateless upgrade
+        FlinkDeployment statelessUpgrade = TestUtils.clone(deployment);
+        statelessUpgrade.getSpec().getJob().setUpgradeMode(UpgradeMode.STATELESS);
+        statelessUpgrade.getSpec().getFlinkConfiguration().put("new", "conf");
+        reconciler.reconcile("test", statelessUpgrade, config);
+        Mockito.verify(flinkService, times(1))
+                .cancelJob(eq(JobID.fromHexString(JOB_ID)), eq(UpgradeMode.STATELESS), eq(config));
+
+        Mockito.verify(flinkService, times(1))
```
dannycranmer, line 77: Flink code style instructs us to [avoid using Mockito](https://flink.apache.org/contributing/code-style-and-quality-common.html#avoid-mockito---use-reusable-test-implementations). I am not sure whether these rules apply to other repos, but I assume so. Let's clarify this

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkUtils.java
```diff
@@ -0,0 +1,229 @@
...
+                    effectiveConfig.set(
+                            CoreOptions.DEFAULT_PARALLELISM, spec.getJob().getParallelism());
+                }
+            }
+
+            return effectiveConfig;
+        } catch (Exception e) {
+            throw new RuntimeException("Failed to load configuration", e);
+        }
+    }
```
dannycranmer, line 168: This method is very long, suggest splitting it out into smaller methods

pom.xml
```diff
@@ -0,0 +1,285 @@
...
+        <maven-failsafe-plugin.version>3.0.0-M4</maven-failsafe-plugin.version>
+        <maven-resources-plugin.version>3.2.0</maven-resources-plugin.version>
+
+        <operator.sdk.version>2.1.1</operator.sdk.version>
+        <fabric8.version>5.12.1</fabric8.version>
+        <lombok.version>1.18.22</lombok.version>
+
+        <scala.version>2.12</scala.version>
+        <flink.version>1.14.3</flink.version>
+        <flink.shaded.version>15.0</flink.shaded.version>
```
dannycranmer, line 66: Why are we using a newer version for shaded libs?

dannycranmer (commented):
I had a quick skim and added a few questions

pom.xml
gyfora: Flink 1.15 already uses this version, I think by the time we have a first release here that is going to be the current version :)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkUtils.java
gyfora: I completely agree, I would not jump on this straight away but will add a ticket for this so we don't forget.

pom.xml
gyfora: Does that mean that plugins, profiles etc defined in flink-parent are autmatically enabled here also? If that is the case that might make our life more difficult in the long run

Looking at the statefun project that also does not use the flink parent: https://github.com/apache/flink-statefun/blob/master/pom.xml

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/reconciler/JobReconcilerTest.java
gyfora: That's a very good point, I think the codestyle definitely applies here as well. I will try to see how we can change those tests to not use mockito

wangyang0918:
> @wangyang0918 The problem with `flink-kubernetes` is that it is not a proper maven dependency but we release a fatjar, this means we cannot really exclude the fabric8 from it because its simply part of the jar.
> 
> This is something we should resolve on the Flink side because it doesnt make too much sense the way it is now. What we could do as a temporary workaround is that we repackage it in a new module and remove / relocate the shaded classes so that it does not conflict with the proper fabric8 deps accidentally

You are right. The overlapping classes issue need to be fixed in the upstream project Flink. I just remember kubernetes client won't work normally if we shaded `io.fabric8` classes. I am also not sure whether the kubernetes client could work with different `jackson` `okhttp3` `okio` dependencies. That's why `flink-kubernetes` is compiled to a fat jar.

Anyway, this is not a blocker for the this PR to be merged.

gyfora:
@dannycranmer I pushed a commit that reworks the 2 tests that used mockito :) I think this change makes sense and mockito did not add too much value in this case

helm/flink-operator/templates/webhook.yaml
```diff
@@ -0,0 +1,103 @@
...
+  namespace: {{ .Values.operatorNamespace.name }}
+spec:
+  selfSigned: {}
+---
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingWebhookConfiguration
+metadata:
+  annotations:
+    cert-manager.io/inject-ca-from: default/flink-operator-serving-cert
+  name: flink-operator-validating-webhook-configuration
```
bgeng777, line 84: Just want to check if we should add `namespace` for `flink-operator-validating-webhook-configuration` and change `default/flink-operator-serving-cert` to `{{ .Values.operatorNamespace.name }}/flink-operator-serving-cert`

gyfora: The webhook config does not have a namespace (it's global), but you are absolutely right about the cert name, that is a mistake. Fixing now

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkUtils.java
gyfora: https://issues.apache.org/jira/browse/FLINK-26163

helm/flink-operator/templates/webhook.yaml
```diff
@@ -0,0 +1,103 @@
...
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: flink-operator-serving-cert
+  namespace: {{ .Values.operatorNamespace.name }}
+spec:
+  dnsNames:
+  - flink-operator-webhook-service.default.svc
+  - flink-operator-webhook-service.default.svc.cluster.local
```
morhidi, line 52: must be:
flink-operator-webhook-service.{{ .Values.operatorNamespace.name }}.svc
flink-operator-webhook-service.{{ .Values.operatorNamespace.name }}.svc.cluster.local

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/status/JobStatus.java
```diff
@@ -0,0 +1,36 @@
...
+
+/** Status of an individual job within the Flink deployment. */
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+@Builder
+public class JobStatus {
+    private String jobName;
+    private String jobId;
+    private String state;
```
bgeng777, line 33: Should `state` field fall into one enum value of `JobState`? I find the usages of this field include `jobStatus.setState("suspended");` in `JobReconciler.java` and `jobStatus.setState("RUNNING");` in `JobReconcilerTest.java`. The mix use of upper/lower case may be somehow misleading.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/spec/JobSpec.java
```diff
@@ -0,0 +1,41 @@
...
+
+/** Flink job spec. */
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+@Builder
+public class JobSpec {
+    private String jarURI;
+    private int parallelism;
+    private String entryClass;
```
bgeng777, line 34: A small improvement: `entryClassName` may be more accurate as in our upstream flink code, this field is named as `applicationClassName`.

bgeng777:
I follow the developer guide to install it via helm and it works smoothly. Thanks a lot for the awesome work! I just left some minor comments and hope they can help to improve the code :).

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/status/JobStatus.java
gyfora: Good point, I think this could be mapped to the JobState enum, I will fix this tomorrow

gyfora:
@bgeng777 thanks a lot for giving this a try! really appreciate it

pom.xml
tweise: Besides it being the next version, the dependencies do not need to line up exactly with Flink core. It is conceivable that in the future they will diverge in other cases also.

pom.xml
tweise: We don't have to inherit Flink parent, but we could if there is value. On one hand there may be plugin setups that we may benefit from, on the other hand we may inherit things that we don't need or that conflict. I would propose to analyze that outside of this PR.

FuyaoLi2017:
Hi @gyfora , I can't found the CRD file. Based on helm standard, I think it should be added in `helm/flink-operator/crds`. Maybe you missed it?

wangyang0918:
@FuyaoLi2017 The CRD is generated automatically. You will find it once you build the project successfully.

FuyaoLi2017:
For session cluster, maybe we can consider some job level control?

I am not sure if this will be add much value or not. We can have some discussion over this. This is also not top priority for this very first version.

```
  sessionClusterMaintainedJobs:
    type: object
    additionalProperties:
      type: object
      properties:
        artifactoryJarHeader:
          type: array
          nullable: true
          items:
            type: string
        artifactoryJarUri:
          type: string
        savepointGeneration:
          type: integer
          minimum: 0
        parallelism:
          type: integer
          minimum: 1
        allowNonRestoredState:
          type: boolean
          nullable: true
        drainFlag:
          type: boolean
          nullable: true
        savepointsDir:
          type: string
          nullable: true
        fromSavepoint:
          type: string
          nullable: true
        autoSavepointSeconds:
          type: integer
          minimum: 10 # Generally, this value is recommended to be significantly larger than RECONCILE_INTERVAL_MS (the reconcile period) in FlinkApplicationController
        jobAction:
          type: string
          enum: [stop, cancel, run]
        mainArgs:
          type: array
          nullable: true
          items:
            type: string
        entryClass:
          type: string
          nullable: true
```

asardaes:
I don't know if you already considered how to allow users to specify logging configuration, but even if you did, I think it would be counterproductive to redeploy a cluster every time log configuration changes. I personally would like to mount a `logback-console.xml` file from a place I control, but I can't mount it directly into `$FLINK_HOME/conf` because the controller would already mount things there.

I imagine a simple solution would be for me to mount my conf file somewhere else and then inject an environment variable in my template that instructs Flink to use it, and as far as I can tell that only requires a small change [here]( https://github.com/apache/flink/blob/1d564a95581f5b4b1d8ae343676474a4fa570ffa/flink-dist/src/main/flink-bin/bin/flink-console.sh#L101).

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/status/JobStatus.java
gyfora: Opened a jira ticket for this

gyfora:
@FuyaoLi2017 managing multiple jobs especially on a cluster that is intended to be at least partially controlled by user (session cluster) seems to be extremely difficult. Since the user submits the job, it is already difficult to identify a job as multiple jobs can have the same names etc.

I feel that this is beyond what the operator should do at this point.

gyfora:
@asardaes At the moment we allow users to provide a configmap containing the default Flink and logging configuration for log4j and this can be overriden from the deployment spec.

We haven't focused too much effort on making this more flexible yet but we should definitely do soon. I will open a jira ticket for this.

bgeng777:
I want to check that in TaskManagerSpec, there is no field like `replicas` in JobManagerSpec. As a result, when I run the operator in a real k8s cluster and want to add a new TaskManager, I cannot directly update the `basic.yaml` and apply it. Is there any concern for removing this field for TaskManagerSpec?

gyfora:
@bgeng777 with the native integration, Flink controls the number of taskmanagers based on parallelism/task slots. Once we support standalone integration with reactive scaling this will make sense (and we should add it then)

PR #4, opened 2022-02-17
Use github actions for flink-kubernetes-operator CI

wangyang0918:
This PR tries to use the github actions for the CI. We have the following two jobs.
* test_ci, this job will run all the unit tests and integration tests
* e2e_ci, this job will run all the e2e tests
  * The new added e2e test `test_kubernetes_application_ha.sh` will start a Flink application with HA enabled. Wait for successful checkpoints and then kill the JobManager to verify the restoring.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,68 @@
+################################################################################
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+# limitations under the License.
+################################################################################
+
+name: CI
+on:
+  push:
+  pull_request:
+jobs:
+  test_ci:
+    runs-on: ubuntu-latest
+    name: test_ci
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up JDK 11
+        uses: actions/setup-java@v2
+        with:
+          java-version: '11'
+          distribution: 'adopt'
+      - name: Build with Maven
+        run: mvn clean install
+      - name: Start minikube
+        uses: medyagh/setup-minikube@master
+      - name: Install cert-manager
+        run: |
+          kubectl get pods -A
+          kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v1.7.1/cert-manager.yaml
+          kubectl -n cert-manager wait --all=true --for=condition=Ready --timeout=300s pod
+      - name: Build image
+        run: |
+          export SHELL=/bin/bash
+          export DOCKER_BUILDKIT=1
+          eval $(minikube -p minikube docker-env)
+          docker build -f ./Dockerfile -t flink-kubernetes-operator:ci-latest .
+          docker images
+      - name: Start the operator
+        run: |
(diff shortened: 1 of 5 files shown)


wangyang0918:
The CI has been successfully run my personal actions. It will take effect as soon as we merge this PR.

https://github.com/wangyang0918/flink-kubernetes-operator/actions/runs/1856555659

cc @gyfora

gyfora:
I will be away during the day today, I will take a look in the evening :)  

@morhidi / @martonb please review if you have time,

.github/workflows/ci.yml
```diff
@@ -0,0 +1,104 @@
...
+        run: |
+          kubectl get pods -A
+          kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v1.7.1/cert-manager.yaml
+          kubectl -n cert-manager wait --all=true --for=condition=Ready --timeout=300s pod
+      - name: Build image
+        run: |
+          export SHELL=/bin/bash
+          export DOCKER_BUILDKIT=1
+          eval $(minikube -p minikube docker-env)
+          docker build -f ./Dockerfile -t flink-java-operator:ci-latest .
```
morhidi, line 48: nit: we could probably use the flink-kubernetes-operator name here too

.github/workflows/ci.yml
```diff
@@ -0,0 +1,104 @@
...
+          docker images
+      - name: Start the operator
+        run: |
+          helm install flink-operator helm/flink-operator --set image.repository=flink-java-operator --set image.tag=ci-latest
```
morhidi, line 52: nit: we could probably use the flink-kubernetes-operator name here too

.github/workflows/ci.yml
```diff
@@ -0,0 +1,104 @@
...
+        run: |
+          kubectl get pods -A
+          kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v1.7.1/cert-manager.yaml
+          kubectl -n cert-manager wait --all=true --for=condition=Ready --timeout=300s pod
+      - name: Build image
+        run: |
+          export SHELL=/bin/bash
+          export DOCKER_BUILDKIT=1
+          eval $(minikube -p minikube docker-env)
+          docker build -f ./Dockerfile -t flink-java-operator:ci-latest .
```
morhidi, line 92: nit: we could probably use the flink-kubernetes-operator name here too

.github/workflows/ci.yml
```diff
@@ -0,0 +1,104 @@
...
+          docker images
+      - name: Start the operator
+        run: |
+          helm install flink-operator helm/flink-operator --set image.repository=flink-java-operator --set image.tag=ci-latest
```
morhidi, line 96: nit: we could probably use the flink-kubernetes-operator name here too

morhidi (commented):
nice work!

tisonkun (commented):
What about publishing the operator image to ghcr.io also? You can see [this job](https://github.com/tisonkun/engula/blob/517ec981429c28350b0a6bf6085fc96e74990164/.github/workflows/build-image.yml) for example.

mbalassi:
Hi @wangyang0918!

This looks great, I like the light-weight approach of using GitHub actions. As far as I understand this solution is also [free](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#about-billing-for-github-actions) as this is a public repo with moderate load (for the time being).
Since we are deviating from the Azure Pipelines approach of the Flink repo could you please add a short subsection on the CI to the [Developer Guide](https://github.com/apache/flink-kubernetes-operator#developer-guide)?

@tisonkun I like publishing the docker images when the CI succeeds (with some rolling tagging policy) and obviously with stable tags on releases. I prefer having these images on DockerHub similarly to the [flink](https://hub.docker.com/r/apache/flink) repository there. I think this merits its own ticket and a bit of discussion on technical details, would you be so kind to open a ticket under [FLINK-25963](https://issues.apache.org/jira/browse/FLINK-25963)?

tisonkun:
@mbalassi thanks for your reply! Created at https://issues.apache.org/jira/browse/FLINK-26214. We can discuss in details there.

wangyang0918:
cc @mbalassi I have addressed the comments. Please have another look.

tweise:
@wangyang0918 thanks for contributing this! Will there any indicator on the PR itself whether the CI has passed or not?

wangyang0918:
@tweise Yes. But it could not work now because of apache security requirement. I have attached a PR #7 to fix this.

PR #6, opened 2022-02-17
[FLINK-26183] Support metrics using the Flink metric system

morhidi:
- Leverages the existing Metric and Reporter infrastructure of Flink to expose metrics of the Flink operator
- Users are able to access the built in metric reporter plugins
- Users can add their own plugins by building their own operator image or by using init containers
- It exposes standard JVM metrics, and later we can add operator specific metric groups if needed

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,8 +17,8 @@
 ################################################################################
 # Build
 FROM maven:3.8.4-openjdk-11 AS build
-WORKDIR /app
 
+WORKDIR /app
 ENV OPERATOR_DIR=flink-kubernetes-operator
 ENV WEBHOOK_DIR=flink-kubernetes-webhook
 
@@ -44,6 +44,7 @@ ENV WEBHOOK_JAR=flink-kubernetes-webhook-$OPERATOR_VERSION-shaded.jar
 
 COPY --from=build /app/flink-kubernetes-operator/target/$OPERATOR_JAR /
 COPY --from=build /app/flink-kubernetes-webhook/target/$WEBHOOK_JAR /
+COPY --from=build /app/flink-kubernetes-operator/target/plugins /opt/flink/plugins
 
 COPY docker-entrypoint.sh /
 ENTRYPOINT ["/docker-entrypoint.sh"]
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/metrics/KubernetesOperatorMetricGroup.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/metrics/KubernetesOperatorMetricGroup.java
@@ -1,0 +1,81 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.kubernetes.operator.metrics;
+
+import org.apache.flink.configuration.Configuration;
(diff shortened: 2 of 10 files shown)


mbalassi:
Thanks, @morhidi. I find this an elegant solution for adding metrics support, since user have their Flink `MetricsReporter` setups already this leverages that. I am a bit torn on whether we should add all default metric reporter plugins to the base operator image, but I am leaning towards your solution since we also ship them in the Flink image too and they are small.

morhidi:
cc @gyfora @wangyang0918

helm/flink-operator/templates/flink-operator.yaml
```diff
@@ -118,6 +124,9 @@ metadata:
   labels:
     {{- include "flink-operator.labels" . | nindent 4 }}
 data:
+  flink-conf.yaml: |+
```
wangyang0918, line 127: I am suggesting to use the `PrometheusReporter` as default. And maybe we also need to add the prometheus related annotations so that it could integrate with the metrics system easily.

https://stackoverflow.com/questions/53365191/monitor-custom-kubernetes-pod-metrics-using-prometheus

wangyang0918 (commented):
Since the plugin jars are not too big, I am also in favor of baking the plugins in the image by default.

Dockerfile
```diff
@@ -17,8 +17,18 @@
 ################################################################################
 # Build
 FROM maven:3.8.4-openjdk-11 AS build
-WORKDIR /app
 
+WORKDIR /tmp/flink/plugins
+ENV FLINK_VERSION=1.14.3
+RUN --mount=type=cache,target=/root/.m2 \
```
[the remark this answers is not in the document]
tweise: I think this is a great idea. If the extra jar files can be collected using the dependency plugin then the entire /tmp/flink/plugins can be staged from pom.xml and simply copied here.

Dockerfile
```diff
@@ -17,8 +17,18 @@
...
+      mvn dependency:copy -Dartifact=org.apache.flink:flink-metrics-datadog:$FLINK_VERSION -DoutputDirectory=flink-metrics-datadog; \
+      mvn dependency:copy -Dartifact=org.apache.flink:flink-metrics-graphite:$FLINK_VERSION -DoutputDirectory=flink-metrics-graphite; \
+      mvn dependency:copy -Dartifact=org.apache.flink:flink-metrics-influxdb:$FLINK_VERSION -DoutputDirectory=flink-metrics-influxdb; \
+      mvn dependency:copy -Dartifact=org.apache.flink:flink-metrics-prometheus:$FLINK_VERSION -DoutputDirectory=flink-metrics-prometheus; \
+      mvn dependency:copy -Dartifact=org.apache.flink:flink-metrics-slf4j:$FLINK_VERSION -DoutputDirectory=flink-metrics-slf4j; \
+      mvn dependency:copy -Dartifact=org.apache.flink:flink-metrics-statsd:$FLINK_VERSION -DoutputDirectory=flink-metrics-statsd
+
```
tweise, line 30: Do we really need all the reporters? But then, how to decide which one to include or not? I think this is fine and we can come back to it later. It would be really nice though if we can add a section to README how to customize the image, since that will be a frequent use case no matter what we decide here.

tweise:
@morhidi thanks for the contribution! Did you already think about what metrics the operator should expose? That would make for an interesting mailing list thread. There are various metrics that we should track, like reconciliation attempts/failures, count of CRs, etc.

morhidi:
Thanks guys for your comments, I'll look into these today!

helm/flink-operator/templates/flink-operator.yaml
wangyang0918: Make sense to me.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/metrics/KubernetesOperatorMetricGroup.java
```diff
@@ -0,0 +1,48 @@
...
+import org.apache.flink.runtime.metrics.dump.QueryScopeInfo;
+import org.apache.flink.runtime.metrics.groups.AbstractMetricGroup;
+
+/** Flink based operator metric group. */
+public class KubernetesOperatorMetricGroup extends AbstractMetricGroup<AbstractMetricGroup<?>> {
+
+    private static final String GROUP_NAME = "kubernetes-operator";
+
+    private KubernetesOperatorMetricGroup(MetricRegistry registry, String hostname) {
+        super(registry, new String[] {hostname, GROUP_NAME}, null);
```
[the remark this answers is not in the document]
morhidi: @Aitozi I added these variables to the reporter:
metrics.scope.kubernetes-operator: <namespace>.<name>.<host>

PR #7, opened 2022-02-18
[FLINK-26234] Start minikube manually to replace external github actions

wangyang0918:
According to the apache requirement, external github actions are not allowed to used in the workflow. So we need to start minikube manually instead of relying on the existing actions.

See https://github.com/apache/flink-kubernetes-operator/actions/runs/1859307064.


wangyang0918:
cc @gyfora @tweise 
Just like this PR, now we have the CI pre-checks for all the tests.

tweise (approved):
Nice, thanks for fixing this.

PR #9, opened 2022-02-21
[FLINK-26163] Refactor FlinkUtils#getEffectiveConfig into smaller pieces

bgeng777:
1. Refactor `FlinkUtils#getEffectiveConfig` into smaller pieces.
2. Refactor some codes in `TestUtils` and add unit test for `FlinkUtils#getEffectiveConfig`.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkUtils.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkUtils.java
@@ -17,19 +17,9 @@
 
 package org.apache.flink.kubernetes.operator.utils;
 
-import org.apache.flink.configuration.ConfigConstants;
 import org.apache.flink.configuration.Configuration;
-import org.apache.flink.configuration.CoreOptions;
-import org.apache.flink.configuration.DeploymentOptions;
 import org.apache.flink.configuration.GlobalConfiguration;
-import org.apache.flink.configuration.JobManagerOptions;
-import org.apache.flink.configuration.PipelineOptions;
-import org.apache.flink.configuration.TaskManagerOptions;
-import org.apache.flink.kubernetes.configuration.KubernetesConfigOptions;
-import org.apache.flink.kubernetes.configuration.KubernetesDeploymentTarget;
 import org.apache.flink.kubernetes.operator.crd.FlinkDeployment;
-import org.apache.flink.kubernetes.operator.crd.spec.FlinkDeploymentSpec;
-import org.apache.flink.util.StringUtils;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
@@ -37,15 +27,9 @@ import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import io.fabric8.kubernetes.api.model.Pod;
 import io.fabric8.kubernetes.client.KubernetesClient;
-import io.fabric8.kubernetes.client.internal.SerializationUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.io.File;
-import java.io.IOException;
-import java.net.URI;
-import java.nio.file.Files;
-import java.util.Collections;
 import java.util.Iterator;
 
 /** Flink Utility methods used by the operator. */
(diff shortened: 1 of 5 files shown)


bgeng777:
cc @wangyang0918 @gyfora @tweise

wangyang0918:
> 1. Refactor `FlinkUtils#getEffectiveConfig` into smaller pieces.
> 2. Fix the problem that the field of `replicas` in `JobManagerSpec` will not take effect.
> 3. Fix the problem that the basic `podTemplate` in `FlinkDeploymentSpec` will not take effect if JM or TM spec does not specify `podTemplate` on their own.
> 4. Refactor some codes in `TestUtils` and add unit test for `FlinkUtils#getEffectiveConfig`.

* I am suggesting to factor out No.2 into a separate ticket.
* Why we have the No.3 issue? IIUC, currently the pod template should work for both JobManager and TaskManager.
* For No.1, what do you think about introducing a `FlinkConfigBuilder` just like following. It will make things easier to test.
```
/** Build effective flink config from {@link FlinkDeployment} */
public class FlinkConfigBuilder {
    private final FlinkDeployment deploy;
    private final Configuration effectiveConfig;

    public FlinkConfigBuilder(FlinkDeployment deploy) {
        this.deploy = deploy;
        this.effectiveConfig = FlinkUtils.loadConfiguration();
    }

    public FlinkConfigBuilder applyImageToConfig() {
        effectiveConfig.set(KubernetesConfigOptions.CONTAINER_IMAGE, deploy.getSpec().getImage());
        return this;
    }

    // Apply other fields to the config
    ... ...

    public Configuration build() {
        // Set some basic configuration
        final String namespace = deploy.getMetadata().getNamespace();
        final String clusterId = deploy.getMetadata().getName();
        effectiveConfig.setString(KubernetesConfigOptions.NAMESPACE, namespace);
        effectiveConfig.setString(KubernetesConfigOptions.CLUSTER_ID, clusterId);
        return effectiveConfig;
    }
}
```

```
final Configuration effectiveConfig = new FlinkConfigBuilder(flinkApp).applyImageToConfig().apply<...>.build();
```

bgeng777:
1. Sure. It seems that the fix of JM replicas is just some lines of codes so I did it here. I will move it out and leave the work to the separate [jira](https://issues.apache.org/jira/browse/FLINK-26216).
2. My bad. Check the `KUBERNETES_POD_TEMPLATE` field again and I understood it wrongly before.
3. Yes, I also think about using Builder pattern when refactoring. Considering not to change the code style too much, I did not use it initially. But I believe it could make codes more neat. Will update it later.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkUtils.java
```diff
@@ -55,109 +43,19 @@
...
+                    new FlinkConfigBuilder(flinkApp)
+                            .applyFlinkConfiguration()
+                            .applyImage()
+                            .applyImagePullPolicy()
+                            .applyCommonPodTemplate()
+                            .applyIngressDomain()
+                            .applyJobManagerSpec()
+                            .applyTaskManagerSpec()
+                            .applyJobOrSessionSpec()
+                            .build();
```
gyfora, line 57: This could be part of the FlinkConfigBuilder as a static method.

```
public static Configuration buildFor(FlinkDeployment dep) {...}
```

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkUtils.java
```diff
@@ -172,7 +70,7 @@ public static Configuration loadConfiguration(String confDir) {
         return configuration;
     }
 
-    private static String createTempFile(Pod podTemplate) throws IOException {
+    public static String createTempFile(Pod podTemplate) throws IOException {
```
gyfora, line 73: We could move this to the FlinkConfigbuilder and keep as private

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/utils/FlinkConfigBuilderTest.java
```diff
@@ -0,0 +1,175 @@
...
+        flinkDeployment.getSpec().setIngressDomain("test.com");
+        flinkDeployment.getSpec().getJobManager().setPodTemplate(pod1);
+        flinkDeployment.getSpec().getTaskManager().setPodTemplate(pod2);
+        flinkDeployment.getSpec().getJob().setParallelism(2);
+    }
+
+    @Test
+    public void testApplyImage() {
+        Configuration configuration = new FlinkConfigBuilder(flinkDeployment).applyImage().build();
+        System.out.println(configuration);
```
gyfora, line 75: remove sysout please

tweise (approved):
Thanks, great to get some test coverage for this area.

tweise:
Also see #10 for related field changes

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/utils/FlinkConfigBuilderTest.java
bgeng777: Fixed in a02f7185edfdac9e71e470f46b69d5f89c4dacab

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkUtils.java
bgeng777: Moved in a02f7185edfdac9e71e470f46b69d5f89c4dacab

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkUtils.java
bgeng777: Thanks for the advice. Fixed in a02f7185edfdac9e71e470f46b69d5f89c4dacab

gyfora (approved):
Looks good! Thank you :)

Will merge this after breakfast

PR #11, opened 2022-02-22
[FLINK-26260] Support watching specific namespaces only

gyfora:
This PR introduces a new config option for watching a list of specific namespaces for FlinkDeployment resources.
The config can be set in the Helm chart as a list and is passed as an env variable to the operator.

It also includes a change to the webhook configuration name to include the operator namespace. This allows us to have multiple webhook configs in the same cluster pointing to different operators (which is necessary if all of them are watching their own namespace)

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,12 +10,18 @@ The operator is managed helm chart. To install run:
  helm install flink-operator .
 ```
 
+### Validating webhook
+
 In order to use the webhook for FlinkDeployment validation, you must install the cert-manager on the Kubernetes cluster:
 ```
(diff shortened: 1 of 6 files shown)


gyfora:
cc @wangyang0918 @morhidi

wangyang0918 (approved):
Maybe we also update the readme since passing an array to `helm install` is not very natural.

```
helm install flink-operator helm/flink-operator --set image.repository=flink-java-operator --set image.tag=latest --set watchNamespaces={flink-test}
```

gyfora:
Thanks @wangyang0918 I have fixed the generic classes + added a section to the README

wangyang0918:
+1 for merging.

tweise (approved):

PR #13, opened 2022-02-22
[FLINK-26236] Track and cap retries in ReconciliationStatus

morhidi:
- introduce generic error handling 
- distinguish between fatal (e.g. invalid CR) and recoverable errors (e.g. timeout)
- the java-operator-sdk uses exponential backoff strategy to handle recoverable errors (see https://javaoperatorsdk.io/docs/features for further detail)
- fatal errors are handled in the reconcile method itself, with no retry

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
@@ -19,8 +19,9 @@ package org.apache.flink.kubernetes.operator.controller;
 
 import org.apache.flink.configuration.Configuration;
 import org.apache.flink.kubernetes.operator.crd.FlinkDeployment;
-import org.apache.flink.kubernetes.operator.crd.status.FlinkDeploymentStatus;
 import org.apache.flink.kubernetes.operator.crd.status.ReconciliationStatus;
+import org.apache.flink.kubernetes.operator.exception.InvalidDeploymentException;
+import org.apache.flink.kubernetes.operator.exception.ReconciliationException;
 import org.apache.flink.kubernetes.operator.observer.JobStatusObserver;
 import org.apache.flink.kubernetes.operator.reconciler.JobReconciler;
 import org.apache.flink.kubernetes.operator.reconciler.SessionReconciler;
@@ -47,15 +48,14 @@ import java.util.Optional;
 import java.util.concurrent.TimeUnit;
 
 /** Controller that runs the main reconcile loop for Flink deployments. */
-@ControllerConfiguration
+@ControllerConfiguration(generationAwareEventProcessing = false)
 public class FlinkDeploymentController
         implements Reconciler<FlinkDeployment>,
                 ErrorStatusHandler<FlinkDeployment>,
                 EventSourceInitializer<FlinkDeployment> {
     private static final Logger LOG = LoggerFactory.getLogger(FlinkDeploymentController.class);
 
-    public static final int OBSERVE_REFRESH_SECONDS = 10;
-    public static final int RECONCILE_ERROR_REFRESH_SECONDS = 5;
+    public static final int REFRESH_SECONDS = 60;
 
     private final KubernetesClient kubernetesClient;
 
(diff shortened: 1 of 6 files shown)


morhidi:

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
```diff
@@ -178,7 +150,14 @@ private void updateForReconciliationError(FlinkDeployment flinkApp, String err)
     @Override
     public Optional<FlinkDeployment> updateErrorStatus(
             FlinkDeployment flinkApp, RetryInfo retryInfo, RuntimeException e) {
-        LOG.warn("TODO: handle error status");
-        return Optional.empty();
+        LOG.warn(
```
tweise, line 153: The log messages are as expected, but maybe we should also capture the retry info in the status? And add an event when retries are exhausted? Otherwise, that information is only transient in the operator SDK?

tweise:
LGTM, there is room to better capture the error details as commented, which can be added as follow-up. I'm going to make the changes for https://issues.apache.org/jira/browse/FLINK-26261 on top of this.

tweise (approved):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
```diff
@@ -178,7 +150,14 @@ private void updateForReconciliationError(FlinkDeployment flinkApp, String err)
...
+                "attempt count: {}, last attempt: {}",
+                retryInfo.getAttemptCount(),
+                retryInfo.isLastAttempt());
+
+        updateForReconciliationError(
+                flinkApp,
+                (e instanceof ReconciliationException) ? e.getCause().toString() : e.toString());
```
wangyang0918, line 160: Do we need to `getMessage()` here?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
```diff
@@ -94,51 +94,23 @@ public DeleteControl cleanup(FlinkDeployment flinkApp, Context context) {
...
-            if (!err.equals(lastErr)) {
-                // Log new errors on the first instance
-                LOG.error("Error while reconciling deployment change", e);
-                updateForReconciliationError(flinkApp, err);
-                return UpdateControl.updateStatus(flinkApp)
-                        .rescheduleAfter(RECONCILE_ERROR_REFRESH_SECONDS, TimeUnit.SECONDS);
-            } else {
-                return UpdateControl.<FlinkDeployment>noUpdate()
-                        .rescheduleAfter(RECONCILE_ERROR_REFRESH_SECONDS, TimeUnit.SECONDS);
+            boolean successfulObserve = observer.observeFlinkJobStatus(flinkApp, effectiveConfig);
```
wangyang0918, line 100: Maybe this is out the scope of this PR. I am afraid the root cause will be hidden since we will always get a `java.util.concurrent.TimeoutException` when Flink client failed to deploy application.

Will this issue be resolved in FLINK-26261? cc @tweise

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
morhidi: Good question, when tested I got null in the message field if the job was not running. So the toString() seemed to be more informative. At least it gives you the class `java.util.concurrent.TimeoutException`. 
public String toString() {
        String s = getClass().getName();
        String message = getLocalizedMessage();
        return (message != null) ? (s + ": " + message) : s;
}

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
morhidi: Maybe we should consider storing the whole stack trace in the error message.

gyfora: Yes the TimeoutException case should be solved as part of the ticket Thomas is working on. Let's leave this as is now, and we can improve later as we see fit from experience.

PR #14, opened 2022-02-22
[FLINK-26193] Commit generated CRD into Git repo

morhidi:
- adding the CRD into Git repo to better track the schema changes and enable helm installs without compiling the source
- picked `Method 1` from https://helm.sh/docs/chart_best_practices/custom_resource_definitions


morhidi:

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/spec/FlinkDeploymentSpec.java
```diff
@@ -41,5 +41,4 @@
     private JobManagerSpec jobManager;
     private TaskManagerSpec taskManager;
     private JobSpec job;
-    private Map<String, String> logging;
```
wangyang0918, line 44: Do we remove the `logging` field because it is not supported now?

morhidi: yes, also we discussed it with @gyfora that map is not the ideal form here for the logging, we'll have to reintroduce it in another form to support log4j, logback.. configs

wangyang0918: Great.

wangyang0918 (approved):
+1 for merging.

PR #18, opened 2022-02-23
[FLINK-26332]Move the operator env to common utils

Aitozi:
Move the scattered environments variables to the `EnvUtils` and make it clear whether the variable is required

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/EnvUtils.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/EnvUtils.java
@@ -1,0 +1,71 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.kubernetes.operator.utils;
+
+import org.apache.commons.lang3.StringUtils;
+
+/** Util to get value from environments. */
+public class EnvUtils {
+
+    public static final String ENV_FLINK_OPERATOR_CONF_DIR = "FLINK_OPERATOR_CONF_DIR";
+    public static final String ENV_FLINK_CONF_DIR = "FLINK_CONF_DIR";
+    public static final String ENV_WEBHOOK_KEYSTORE_FILE = "WEBHOOK_KEYSTORE_FILE";
+    public static final String ENV_WEBHOOK_KEYSTORE_PASSWORD = "WEBHOOK_KEYSTORE_PASSWORD";
+    public static final String ENV_WEBHOOK_KEYSTORE_TYPE = "WEBHOOK_KEYSTORE_TYPE";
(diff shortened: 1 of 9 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/EnvUtils.java
```diff
@@ -0,0 +1,60 @@
...
+    public static final String ENV_OPERATOR_NAMESPACE = "OPERATOR_NAMESPACE";
+
+    /**
+     * Get the value provided by environments.
+     *
+     * @param key the target key
+     * @param required whether to verify the key is exists.
+     * @return the value or default value provided by environments.
+     */
+    public static String get(String key, boolean required) {
```
morhidi, line 42: We could probably use two separate methods get() and getRequired() here.

Aitozi: I added separate `get` method without `required` parameter

wangyang0918: nit: I also lean to `getRequired()`.

wangyang0918 (approved):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/EnvUtils.java
Aitozi: updated

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/EnvUtils.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    /**
+     * Get the value provided by environments.
+     *
+     * @param key the target key
+     * @param required whether to verify the key is exists.
+     * @return the value provided by environments.
+     */
+    public static String getRequired(String key, boolean required) {
```
wangyang0918, line 52: Hmm. We will not need the second argument `required` now. Right?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/EnvUtils.java
```diff
@@ -0,0 +1,70 @@
...
+        String value = System.getenv().get(key);
+        if (value == null && required) {
```
wangyang0918, line 54: Do we also need to check the empty string here?

flink-kubernetes-webhook/src/main/java/org/apache/flink/kubernetes/operator/admission/FlinkOperatorWebhook.java
```diff
@@ -117,26 +111,19 @@ protected void initChannel(SocketChannel ch) {
...
-        if (StringUtils.isAnyEmpty(keystorePassword, keystoreType)) {
-            throw new RuntimeException(
-                    WEBHOOK_KEYSTORE_PASSWORD
-                            + " and "
-                            + WEBHOOK_KEYSTORE_TYPE
-                            + " must no be empty");
-        }
-
+        String keystorePassword =
+                EnvUtils.getRequired(EnvUtils.ENV_WEBHOOK_KEYSTORE_PASSWORD, true);
```
wangyang0918, line 125: `getRequired` does not check the empty string.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/EnvUtils.java
gyfora, line 52: getRequired should not have a boolean parameter, it is always true (its in the method name)

gyfora (changes requested):

gyfora:
I added a minor comment, also please rebase on top of your config changes and then it should be good to go :)

Aitozi:
I have rebased the master and addressed the comments, Please take a look again, thanks cc @gyfora @wangyang0918

wangyang0918:
LGTM. +1 for merging.

PR #19, opened 2022-02-23
[FLINK-26337]Avoid to load flink conf at each reconcile loop

Aitozi:
It will create FlinkConfigBuilder at every loop of reconcile,  It's not necessary and may bring overhead. The default flink conf and operator conf should load at entry point. If the ConfigMap is updated, then operator should be trigger upgrade to load new config.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/FlinkOperator.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/FlinkOperator.java
@@ -17,6 +17,7 @@
 
 package org.apache.flink.kubernetes.operator;
 
+import org.apache.flink.kubernetes.operator.config.DefaultConfig;
 import org.apache.flink.kubernetes.operator.controller.FlinkControllerConfig;
 import org.apache.flink.kubernetes.operator.controller.FlinkDeploymentController;
 import org.apache.flink.kubernetes.operator.metrics.OperatorMetricUtils;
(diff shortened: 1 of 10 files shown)


gyfora (approved):
Looks good! I guess you will have to rebase this if the environment changes go in first or vice versa :)

wangyang0918:
Since the `flink-conf.yaml` is mounted via ConfigMap, I am afraid we will not get the latest configuration if the ConfigMap is updated. Do we have to restart the operator? Or we expect the `flink-operator-config` should not be hot updated.

Aitozi:
@wangyang0918 Thanks for your question, I have think over it a little. 
Now, the `flink-default-config` and `flink-operator-config` are mounted via ConfigMap, We do not do the config-reload for the operator config now. It's a bit inconsistent. The detect for the config change is not so high efficient and may bring some the problem of inconsistency. 
So I tend not to use the hot update feature of the ConfigMap now. If ConfigMap is updated, It should be restart to take effect.

wangyang0918:
But the ConfigMap hot update is very useful to change log level dynamically. Although we have not configured `monitorInterval` in the log4j2.properties.

Anyway, restarting the K8s operator to make the configurations(including `flink-default-config` and `flink-operator-config`) take effect make sense to me.

Aitozi:
Thanks for your explanation. Later, if we want to use the hot update ability of the ConfigMap do update the `flink-default-config`, Maybe we could also use the mechanism like `monitorInterval` to sync the `flink-conf.yaml`? But we may need to think about the dynamic effect scope.

wangyang0918:
> Thanks for your explanation. Later, if we want to use the hot update ability of the ConfigMap do update the `flink-default-config`, Maybe we could also use the mechanism like `monitorInterval` to sync the `flink-conf.yaml`? But we may need to think about the dynamic effect scope.

Make sense.

PR #21, opened 2022-02-24
[FLINK-26261] Wait for JobManager deployment ready before accessing REST API

tweise:
First part following resource dependencies in the reconciler logic: Wait for the JM deployment to be ready before trying to access the JM. This will be built upon for improving the error detection/handling later on.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
@@ -23,16 +23,15 @@ import org.apache.flink.kubernetes.operator.crd.FlinkDeployment;
 import org.apache.flink.kubernetes.operator.crd.status.ReconciliationStatus;
 import org.apache.flink.kubernetes.operator.exception.InvalidDeploymentException;
 import org.apache.flink.kubernetes.operator.exception.ReconciliationException;
-import org.apache.flink.kubernetes.operator.observer.JobStatusObserver;
+import org.apache.flink.kubernetes.operator.reconciler.BaseReconciler;
 import org.apache.flink.kubernetes.operator.reconciler.JobReconciler;
 import org.apache.flink.kubernetes.operator.reconciler.SessionReconciler;
 import org.apache.flink.kubernetes.operator.utils.FlinkUtils;
 import org.apache.flink.kubernetes.operator.utils.IngressUtils;
 import org.apache.flink.kubernetes.operator.validation.FlinkDeploymentValidator;
+import org.apache.flink.kubernetes.utils.Constants;
 
 import io.fabric8.kubernetes.api.model.apps.Deployment;
-import io.fabric8.kubernetes.api.model.apps.DeploymentSpec;
-import io.fabric8.kubernetes.api.model.apps.DeploymentStatus;
 import io.fabric8.kubernetes.client.KubernetesClient;
 import io.fabric8.kubernetes.client.informers.SharedIndexInformer;
 import io.javaoperatorsdk.operator.api.reconciler.Context;
@@ -50,10 +49,8 @@ import io.javaoperatorsdk.operator.processing.event.source.informer.Mappers;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
-import java.util.concurrent.TimeUnit;
 
 /** Controller that runs the main reconcile loop for Flink deployments. */
 @ControllerConfiguration(generationAwareEventProcessing = false)
@@ -63,33 +60,26 @@ public class FlinkDeploymentController
                 EventSourceInitializer<FlinkDeployment> {
     private static final Logger LOG = LoggerFactory.getLogger(FlinkDeploymentController.class);
 
-    public static final int REFRESH_SECONDS = 60;
-    public static final int PORT_READY_DELAY_SECONDS = 10;
-
     private final KubernetesClient kubernetesClient;
 
     private final String operatorNamespace;
 
     private final FlinkDeploymentValidator validator;
-    private final JobStatusObserver observer;
     private final JobReconciler jobReconciler;
     private final SessionReconciler sessionReconciler;
     private final DefaultConfig defaultConfig;
-    private final HashSet<String> jobManagerDeployments = new HashSet<>();
 
     public FlinkDeploymentController(
             DefaultConfig defaultConfig,
             KubernetesClient kubernetesClient,
             String operatorNamespace,
             FlinkDeploymentValidator validator,
-            JobStatusObserver observer,
             JobReconciler jobReconciler,
             SessionReconciler sessionReconciler) {
         this.defaultConfig = defaultConfig;
         this.kubernetesClient = kubernetesClient;
         this.operatorNamespace = operatorNamespace;
         this.validator = validator;
-        this.observer = observer;
         this.jobReconciler = jobReconciler;
         this.sessionReconciler = sessionReconciler;
     }
@@ -104,7 +94,7 @@ public class FlinkDeploymentController
                 operatorNamespace,
                 kubernetesClient,
                 true);
-        jobManagerDeployments.remove(flinkApp.getMetadata().getSelfLink());
+        getReconciler(flinkApp).removeDeployment(flinkApp);
         return DeleteControl.defaultDelete();
     }
 
@@ -122,14 +112,11 @@ public class FlinkDeploymentController
         Configuration effectiveConfig =
                 FlinkUtils.getEffectiveConfig(flinkApp, defaultConfig.getFlinkConfig());
         try {
-            // only check job status when the JM deployment is ready
-            boolean shouldReconcile =
-                    !jobManagerDeployments.contains(flinkApp.getMetadata().getSelfLink())
-                            || observer.observeFlinkJobStatus(flinkApp, effectiveConfig);
-            if (shouldReconcile) {
-                reconcileFlinkDeployment(operatorNamespace, flinkApp, effectiveConfig);
-                updateForReconciliationSuccess(flinkApp);
-            }
+            UpdateControl<FlinkDeployment> updateControl =
+                    getReconciler(flinkApp)
+                            .reconcile(operatorNamespace, flinkApp, context, effectiveConfig);
+            updateForReconciliationSuccess(flinkApp);
+            return updateControl;
         } catch (InvalidDeploymentException ide) {
             LOG.error("Reconciliation failed", ide);
             updateForReconciliationError(flinkApp, ide.getMessage());
@@ -137,62 +124,10 @@ public class FlinkDeploymentController
         } catch (Exception e) {
             throw new ReconciliationException(e);
         }
-
-        return checkJobManagerDeployment(flinkApp, context, effectiveConfig);
     }
 
-    private UpdateControl<FlinkDeployment> checkJobManagerDeployment(
-            FlinkDeployment flinkApp, Context context, Configuration effectiveConfig) {
-        if (!jobManagerDeployments.contains(flinkApp.getMetadata().getSelfLink())) {
-            Optional<Deployment> deployment = context.getSecondaryResource(Deployment.class);
-            if (deployment.isPresent()) {
-                DeploymentStatus status = deployment.get().getStatus();
-                DeploymentSpec spec = deployment.get().getSpec();
-                if (status != null
-                        && status.getAvailableReplicas() != null
-                        && spec.getReplicas().intValue() == status.getReplicas()
-                        && spec.getReplicas().intValue() == status.getAvailableReplicas()) {
-                    // typically it takes a few seconds for the REST server to be ready
-                    if (observer.isJobManagerReady(effectiveConfig)) {
-                        LOG.info(
-                                "JobManager deployment {} in namespace {} is ready",
-                                flinkApp.getMetadata().getName(),
-                                flinkApp.getMetadata().getNamespace());
-                        jobManagerDeployments.add(flinkApp.getMetadata().getSelfLink());
-                        if (flinkApp.getStatus().getJobStatus() != null) {
-                            // short circuit, if the job was already running
-                            // reschedule for immediate job status check
-                            return UpdateControl.updateStatus(flinkApp).rescheduleAfter(0);
-                        }
-                    }
-                    LOG.info(
-                            "JobManager deployment {} in namespace {} port not ready",
-                            flinkApp.getMetadata().getName(),
-                            flinkApp.getMetadata().getNamespace());
-                    return UpdateControl.updateStatus(flinkApp)
-                            .rescheduleAfter(PORT_READY_DELAY_SECONDS, TimeUnit.SECONDS);
-                } else {
-                    LOG.info(
-                            "JobManager deployment {} in namespace {} not yet ready, status {}",
-                            flinkApp.getMetadata().getName(),
-                            flinkApp.getMetadata().getNamespace(),
-                            status);
-                }
-            }
-        }
-        return UpdateControl.updateStatus(flinkApp)
-                .rescheduleAfter(REFRESH_SECONDS, TimeUnit.SECONDS);
-    }
-
-    private void reconcileFlinkDeployment(
-            String operatorNamespace, FlinkDeployment flinkApp, Configuration effectiveConfig)
-            throws Exception {
-
-        if (flinkApp.getSpec().getJob() == null) {
-            sessionReconciler.reconcile(operatorNamespace, flinkApp, effectiveConfig);
-        } else {
-            jobReconciler.reconcile(operatorNamespace, flinkApp, effectiveConfig);
-        }
+    private BaseReconciler getReconciler(FlinkDeployment flinkDeployment) {
+        return flinkDeployment.getSpec().getJob() == null ? sessionReconciler : jobReconciler;
     }
 
     private void updateForReconciliationSuccess(FlinkDeployment flinkApp) {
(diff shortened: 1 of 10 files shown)


tweise:
@gyfora PTAL

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
```diff
@@ -109,6 +119,22 @@ public DeleteControl cleanup(FlinkDeployment flinkApp, Context context) {
...
+
+        if (!jobManagerDeployments.containsKey(flinkApp)) {
+            return context.getSecondaryResource(Deployment.class)
+                    .map(
+                            deployment -> {
+                                LOG.info(
+                                        "JobManager deployment {} in namespace {} is ready",
+                                        flinkApp.getMetadata().getName(),
+                                        flinkApp.getMetadata().getNamespace());
+                                jobManagerDeployments.put(flinkApp, deployment);
```
gyfora, line 131: Shouldn't we check and put into the jobManagerDeployments befre we try to observe? It seems now we introduce an extra refesh between every reconcile and observe.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
```diff
@@ -88,17 +92,23 @@ public DeleteControl cleanup(FlinkDeployment flinkApp, Context context) {
                 operatorNamespace,
                 kubernetesClient,
                 true);
+        jobManagerDeployments.remove(flinkApp);
```
gyfora, line 95: Should we also remove from this map when we cancel the flink job in the Flink service?

gyfora: Otherwise after an upgrade the observer might hit the same issue

gyfora (changes requested):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
tweise: We do check before we observe. The deployment check is executed once, it does not introduce an extra refresh. It also works when the operator is restarted and finds existing deployments.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
tweise: Good catch, will look into that tomorrow.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
gyfora: Oh I think I see what you mean. Do you mean that we check after the initial reconcolitiation? (right after the job was deployed).

My suggestion was to move before line 108 just to be safe, but maybe it's not necessary

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
```diff
@@ -109,6 +119,22 @@ public DeleteControl cleanup(FlinkDeployment flinkApp, Context context) {
         } catch (Exception e) {
             throw new ReconciliationException(e);
         }
+
+        if (!jobManagerDeployments.containsKey(flinkApp)) {
```
gyfora, line 123: Maybe we could extract the checking and loading of the deployment from the secondaryResource into a new method in the controller to keep the main reconcile flow as simple as possible

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
```diff
@@ -144,7 +170,16 @@ private void updateForReconciliationError(FlinkDeployment flinkApp, String err)
         //        return List.of(new PerResourcePollingEventSource<>(
         //                new FlinkResourceSupplier, context.getPrimaryCache(), POLL_PERIOD,
         //                FlinkApplication.class));
```
gyfora, line 172: I think we can remove these comments now that we have some logic here :)

gyfora (changes requested):

gyfora:
I wonder if caching the deployment is even necessary, it adds complexity and can lead to cases where the deployment disappears and we do not notice. The reconcile loops should not run too often for this to stress kubernetes, what do you think @tweise ?

morhidi:
I tested your PR locally and hit this error, when deleting the CR @tweise did you notice?

```
2022-02-24 11:44:24,277 o.a.f.k.o.c.FlinkDeploymentController [INFO ] [default.basic-example] Cleaning up application cluster basic-example
Exception in thread "pool-3-thread-4" java.lang.IllegalStateException: Cached custom resource must be present at this point
	at io.javaoperatorsdk.operator.processing.event.EventProcessor.lambda$isCacheReadyForInstantReconciliation$1(EventProcessor.java:248)
```

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
morhidi: We could also move the whole logic into JobStatusObserver passing the context object.

morhidi:
> I wonder if caching the deployment is even necessary, it adds complexity and can lead to cases where the deployment disappears and we do not notice. The reconcile loops should not run too often for this to stress kubernetes, what do you think @tweise ?

I agree with @gyfora here. The refresh loop is 1 min at the moment, shouldn't be a big deal.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
tweise: It would scatter the resource watching logic. I would prefer to keep it here since it needs to work in tandem with `prepareEventSources`

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
tweise: The current implementation checks job status only after the deployment was observed. I think my earlier reply was misleading.

tweise: Note that that there is other logic in `JobStatusObserver` that will exit when not ready to watch, confusingly returning true. I did not want to touch that in the PR. My goal is to implement the terminal error check when the deployment never shows up, for example due to invalid service account.

tweise:
> I wonder if caching the deployment is even necessary, it adds complexity and can lead to cases where the deployment disappears and we do not notice. The reconcile loops should not run too often for this to stress kubernetes, what do you think @tweise ?

We actually don't need to cache the whole objects for this purpose. However, I would like to limit the checking for secondary resources unless really necessary as that may become a scalability concern. How frequently we check the job status should ideally be independent from this logic.

tweise:
> I tested your PR locally and hit this error, when deleting the CR @tweise did you notice?
> 
> ```
> 2022-02-24 11:44:24,277 o.a.f.k.o.c.FlinkDeploymentController [INFO ] [default.basic-example] Cleaning up application cluster basic-example
> Exception in thread "pool-3-thread-4" java.lang.IllegalStateException: Cached custom resource must be present at this point
> 	at io.javaoperatorsdk.operator.processing.event.EventProcessor.lambda$isCacheReadyForInstantReconciliation$1(EventProcessor.java:248)
> ```

I also see it now. It is related to the deployment event watching as it goes away when disabling `prepareEventSources`

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
```diff
@@ -109,6 +118,19 @@ public DeleteControl cleanup(FlinkDeployment flinkApp, Context context) {
...
+        if (!jobManagerDeployments.contains(flinkApp.getMetadata().getSelfLink())) {
+            Optional<Deployment> deployment = context.getSecondaryResource(Deployment.class);
+            if (deployment.isPresent()) {
+                LOG.info(
+                        "JobManager deployment {} in namespace {} is ready",
+                        flinkApp.getMetadata().getName(),
+                        flinkApp.getMetadata().getNamespace());
+                jobManagerDeployments.add(flinkApp.getMetadata().getSelfLink());
+                // reschedule for immediate job status check
+                return UpdateControl.updateStatus(flinkApp).rescheduleAfter(0);
```
tweise, line 131: We will still need a delay here as even after the pod is ready it still takes a bit for the REST server to listen at the port. This is how I currently see the job status check still hitting a timeout. BTW @gyfora job status checks blocking the operator threads isn't so nice!

morhidi: Try @ControllerConfiguration(generationAwareEventProcessing = true)

morhidi: In this case, the reconcile loop will kick off after 60 seconds. Can be a simple workaround until we find a solution for this

gyfora, line 131: could we move this whole thing into a `boolean checkAndCacheDeployment(...)` -or something similar- method?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
gyfora: I also agree with @morhidi that this probably fits best into the observer itself because that is the component responsible for checking the running jobs already. The observer could also prepare the event sources that we would just pass in the controller.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
tweise: I moved it into into separate method for now. Will see how it can be arranged better when I have the error/timeout check in place.

tweise: The annotation does not seem to change anything wrt the original error. There is a separate reschedule interval for this condition, which in absent more sophisticated readiness check I have set to 5s for now - that seems to avoids the timeout exception.

gyfora:
@wangyang0918 if you have some time to look at this we could use a bit of extra feedback :)

wangyang0918:
I am getting to this PR now.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
```diff
@@ -145,11 +190,16 @@ private void updateForReconciliationError(FlinkDeployment flinkApp, String err)
...
-        //        return List.of(new PerResourcePollingEventSource<>(
-        //                new FlinkResourceSupplier, context.getPrimaryCache(), POLL_PERIOD,
-        //                FlinkApplication.class));
-        return Collections.emptyList();
+        // reconcile when job manager deployment and REST API are ready
+        SharedIndexInformer<Deployment> deploymentInformer =
+                kubernetesClient
+                        .apps()
+                        .deployments()
+                        .inAnyNamespace()
```
wangyang0918, line 198: Maybe we do not need to watch all the namespaces if `FLINK_OPERATOR_WATCH_NAMESPACES` configured.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
```diff
@@ -145,11 +190,16 @@ private void updateForReconciliationError(FlinkDeployment flinkApp, String err)
...
+                        .withLabel("type", "flink-native-kubernetes")
```
wangyang0918, line 199: Using constants in `org.apache.flink.kubernetes.utils.Constants` could make us aware of upstream changes.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
```diff
@@ -114,6 +126,39 @@ public DeleteControl cleanup(FlinkDeployment flinkApp, Context context) {
...
+                DeploymentSpec spec = deployment.get().getSpec();
+                if (status != null
+                        && status.getAvailableReplicas() != null
+                        && spec.getReplicas().intValue() == status.getReplicas()
+                        && spec.getReplicas().intValue() == status.getAvailableReplicas()) {
+                    LOG.info(
+                            "JobManager deployment {} in namespace {} is ready",
+                            flinkApp.getMetadata().getName(),
+                            flinkApp.getMetadata().getNamespace());
+                    jobManagerDeployments.add(flinkApp.getMetadata().getSelfLink());
```
wangyang0918, line 148: Do we need to remove the current flinkApp from cache `jobManagerDeployments` when the replicas is not enough? For example, the JobManager crashed for a while.

wangyang0918 (commented):
I think this PR is a good step to avoid unnecessary reconciliation. But it seems that I still find some `java.util.concurrent.TimeoutException` when accessing the rest service even after `PORT_READY_DELAY_SECONDS`. Maybe my minikube is a little slow or JobManager does not have enough resources(0.5 cpu).

tweise:
@gyfora this is working pretty well now, I also added a port readiness check. But the current implementation with control logic scattered over job status observer and reconcilers makes it pretty hard to implement this in a cleaner way. I'm working on some refactoring and hopefully that will lead to simplification. Hope to have it done soon to avoid larger merge conflicts.

gyfora:
Sounds good @tweise !

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
```diff
@@ -114,6 +126,39 @@ public DeleteControl cleanup(FlinkDeployment flinkApp, Context context) {
         } catch (Exception e) {
             throw new ReconciliationException(e);
         }
+
+        return checkDeployment(flinkApp, context);
+    }
+
+    private UpdateControl<FlinkDeployment> checkDeployment(
+            FlinkDeployment flinkApp, Context context) {
+        if (!jobManagerDeployments.contains(flinkApp.getMetadata().getSelfLink())) {
```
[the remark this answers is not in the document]
tweise: It is a unique reference that avoids caching entire resource objects and duplicate key construction. Why should the field not exist? Did you encounter it somewhere?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
tweise: Maybe, the job status check would fail in that case and we could handle it there. We can address this as follow-up. There is a lot of work that needs to go into the reconciliation logic based on what I noticed while working on this PR.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
tweise: I would prefer that. Need to figure out how to get the controller config from the context and set multiple namespaces.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
tweise: Refactored the logic and `JobStatusObserver` can be removed / consolidated with JobReconciler in a follow-up.

tweise:
@gyfora I think after the refactor this looks much better now. Minor improvements can be dealt with as follow-up. I think there will be significant work necessary to make the reconciliation more robust. I'm wondering if it should be implemented as a state machine. The job upgrade part needs to be looked at. What happens when the previous job never came up due to a code issue and user pushes an update to fix the problem. This should be possible and we need to have test coverage for it also.

tweise:
> I think this PR is a good step to avoid unnecessary reconciliation. But it seems that I still find some `java.util.concurrent.TimeoutException` when accessing the rest service even after `PORT_READY_DELAY_SECONDS`. Maybe my minikube is a little slow or JobManager does not have enough resources(0.5 cpu).

I changed the delay to 10s and added a port readiness check. The 10s may still not be enough in some situations for the REST service to get ready (after the port is ready). That is covered by the retry though.

gyfora:
I will have to take a look again after the refactor, but there is also in the ticket opened in the meantime: https://issues.apache.org/jira/browse/FLINK-26377

cc @Aitozi

gyfora (approved):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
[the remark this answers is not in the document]
wangyang0918: I do not find the `selfLink` in my minikube with version `v1.23.0`. Could we simply use the `metadata.name` here?

image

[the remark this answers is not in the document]
tweise: I see, it appears that selflink was deprecated. We need a unique identifier across namespaces. `UID` should do the trick.

PR #22, opened 2022-02-24
[FLINK-26141] Support last-state upgrade mode

wangyang0918:
This PR tries to support last-state upgrade mode. 

Given that the HA is enabled, if we want to upgrade the job with latest successful checkpoint, we could simply destroy the K8s deployment. All the HA related ConfigMaps will be retained. Then we start the job with same cluster-id and it should recover from the latest retained checkpoint.

[1]. https://nightlies.apache.org/flink/flink-docs-master/docs/deployment/ha/overview/#high-availability-data-lifecycle
[2]. https://nightlies.apache.org/flink/flink-docs-master/docs/deployment/ha/kubernetes_ha/#high-availability-data-clean-up

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/JobReconciler.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/JobReconciler.java
@@ -25,7 +25,6 @@ import org.apache.flink.kubernetes.operator.crd.spec.JobSpec;
 import org.apache.flink.kubernetes.operator.crd.spec.JobState;
 import org.apache.flink.kubernetes.operator.crd.spec.UpgradeMode;
 import org.apache.flink.kubernetes.operator.crd.status.JobStatus;
-import org.apache.flink.kubernetes.operator.exception.InvalidDeploymentException;
 import org.apache.flink.kubernetes.operator.service.FlinkService;
 import org.apache.flink.kubernetes.operator.utils.IngressUtils;
 import org.apache.flink.runtime.jobgraph.SavepointConfigOptions;
@@ -80,23 +79,21 @@ public class JobReconciler {
                     upgradeFlinkJob(flinkApp, effectiveConfig);
                 }
                 if (desiredJobState.equals(JobState.SUSPENDED)) {
-                    if (upgradeMode == UpgradeMode.STATELESS) {
-                        cancelJob(flinkApp, effectiveConfig);
-                    } else {
-                        suspendJob(flinkApp, effectiveConfig);
-                    }
+                    printCancelLogs(upgradeMode, flinkApp.getMetadata().getName());
+                    cancelJob(flinkApp, upgradeMode, effectiveConfig);
                 }
             }
-            if (currentJobState == JobState.SUSPENDED) {
-                if (desiredJobState == JobState.RUNNING) {
-                    if (upgradeMode == UpgradeMode.STATELESS) {
-                        deployFlinkJob(flinkApp, effectiveConfig, Optional.empty());
-                    } else if (upgradeMode == UpgradeMode.SAVEPOINT) {
-                        restoreFromLastSavepoint(flinkApp, effectiveConfig);
-                    } else {
-                        throw new InvalidDeploymentException(
-                                "Only savepoint and stateless strategies are supported at the moment.");
-                    }
+            if (currentJobState == JobState.SUSPENDED && desiredJobState == JobState.RUNNING) {
+                if (upgradeMode == UpgradeMode.STATELESS) {
+                    deployFlinkJob(flinkApp, effectiveConfig, Optional.empty());
+                } else if (upgradeMode == UpgradeMode.SAVEPOINT) {
+                    restoreFromLastSavepoint(flinkApp, effectiveConfig);
+                } else if (upgradeMode == UpgradeMode.LAST_STATE) {
+                    final String savepointLocation =
+                            flinkApp.getStatus().getJobStatus().getSavepointLocation();
+                    // Upgrade mode changes from savepoint -> last-state
+                    deployFlinkJob(
+                            flinkApp, effectiveConfig, Optional.ofNullable(savepointLocation));
                 }
             }
         }
@@ -116,7 +113,8 @@ public class JobReconciler {
     private void upgradeFlinkJob(FlinkDeployment flinkApp, Configuration effectiveConfig)
             throws Exception {
         LOG.info("Upgrading running job");
-        Optional<String> savepoint = cancelJob(flinkApp, effectiveConfig);
+        final Optional<String> savepoint =
+                cancelJob(flinkApp, flinkApp.getSpec().getJob().getUpgradeMode(), effectiveConfig);
         deployFlinkJob(flinkApp, effectiveConfig, savepoint);
     }
 
@@ -126,17 +124,20 @@ public class JobReconciler {
         deployFlinkJob(flinkApp, effectiveConfig, Optional.of(jobStatus.getSavepointLocation()));
     }
 
-    private Optional<String> suspendJob(FlinkDeployment flinkApp, Configuration effectiveConfig)
-            throws Exception {
-        LOG.info("Suspending {}", flinkApp.getMetadata().getName());
-        return cancelJob(flinkApp, UpgradeMode.SAVEPOINT, effectiveConfig);
-    }
-
-    private Optional<String> cancelJob(FlinkDeployment flinkApp, Configuration effectiveConfig)
-            throws Exception {
-        LOG.info("Cancelling {}", flinkApp.getMetadata().getName());
-        UpgradeMode upgradeMode = flinkApp.getSpec().getJob().getUpgradeMode();
-        return cancelJob(flinkApp, upgradeMode, effectiveConfig);
+    private void printCancelLogs(UpgradeMode upgradeMode, String name) {
+        switch (upgradeMode) {
+            case STATELESS:
+                LOG.info("Cancelling {}", name);
+                break;
+            case SAVEPOINT:
+                LOG.info("Suspending {}", name);
+                break;
+            case LAST_STATE:
+                LOG.info("Cancelling {} with last state retained", name);
+                break;
+            default:
+                throw new RuntimeException("Unsupported upgrade mode " + upgradeMode);
+        }
     }
 
     private Optional<String> cancelJob(
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
@@ -17,6 +17,7 @@
 
 package org.apache.flink.kubernetes.operator.service;
 
+import org.apache.flink.annotation.VisibleForTesting;
 import org.apache.flink.api.common.JobID;
 import org.apache.flink.client.cli.ApplicationDeployer;
 import org.apache.flink.client.deployment.ClusterClientFactory;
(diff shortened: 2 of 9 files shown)


wangyang0918:
cc @tweise @gyfora

wangyang0918:
cc @tweise @gyfora

gyfora:
Looks very nice, I will review this in the afternoon!

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
```diff
@@ -123,6 +127,15 @@ public void submitSessionCluster(FlinkDeployment deployment, Configuration conf)
                                     .get(1, TimeUnit.MINUTES);
                     savepointOpt = Optional.of(savepoint);
                     break;
+                case LAST_STATE:
```
Aitozi, line 130: nit: Can we rename this mode to `CHECKPOINT`/`LAST_CHECKPOINT` ? It looks more natural with Flink term.

Besides , I have a question here: what's the difference between the `STATELESS` and `LAST_STATE` it seems we can deleteCluster for both. I don't quite get the current implementation that `STATE_LESS` only invoke the `clusterClient.cancel()`, will the resource eventually released in `STATE_LESS` ?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
```diff
@@ -123,6 +127,15 @@ public void submitSessionCluster(FlinkDeployment deployment, Configuration conf)
...
+                    if (!HighAvailabilityMode.isHighAvailabilityModeActivated(conf)) {
+                        throw new InvalidDeploymentException(
```
Aitozi, line 132: This check/validation can be done in advance, not in the `cancel` phase. We may move this to the [shared validation logic](https://issues.apache.org/jira/browse/FLINK-26136) later ?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/JobReconciler.java
```diff
@@ -86,22 +86,17 @@ public void reconcile(
...
-                        suspendJob(flinkApp, effectiveConfig);
-                    }
+                    printCancelLogs(upgradeMode, flinkApp.getMetadata().getName());
+                    cancelJob(flinkApp, upgradeMode, effectiveConfig);
                 }
             }
             if (currentJobState == JobState.SUSPENDED) {
                 if (desiredJobState == JobState.RUNNING) {
-                    if (upgradeMode == UpgradeMode.STATELESS) {
+                    if (upgradeMode == UpgradeMode.STATELESS
```
Aitozi, line 95: nit: I think the logic here is a bit mixed up with three `if` nested

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/JobReconciler.java
```diff
@@ -138,17 +134,20 @@ private void restoreFromLastSavepoint(FlinkDeployment flinkApp, Configuration ef
...
-        LOG.info("Suspending {}", flinkApp.getMetadata().getName());
-        return cancelJob(flinkApp, UpgradeMode.SAVEPOINT, effectiveConfig);
-    }
-
-    private Optional<String> cancelJob(FlinkDeployment flinkApp, Configuration effectiveConfig)
-            throws Exception {
-        LOG.info("Cancelling {}", flinkApp.getMetadata().getName());
-        UpgradeMode upgradeMode = flinkApp.getSpec().getJob().getUpgradeMode();
-        return cancelJob(flinkApp, upgradeMode, effectiveConfig);
+    private void printCancelLogs(UpgradeMode upgradeMode, String name) {
```
Aitozi, line 137: Can this log integrate with the  `org.apache.flink.kubernetes.operator.service.FlinkService#cancelJob` ?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
gyfora: LAST_STATE means either last checkpoint or savepoint whatever was the latest state of the job. This is a stateful upgrade mode that does not trigger a savepoint.

STATELESS means that there won't be any savepoints taken for upgrade and the job will be restored without state.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
gyfora: Good point, let's keep it here I will open a PR soon with the extracted validation logic. I will move it there once this PR is merged

gyfora (approved):
Very nice job @wangyang0918 , the code and tests look great. Tested this locally as well and seems to work exactly as expected.

One cornerncase we might want to cover (but we can also leave this for later and just open a ticket) is if the user suspends a job with savepoint upgrademode and then changes the upgrade mode to last-state when resubmitting it to running state.

In this case with the current logic the job would simply start from an empty state which is incorrect. So what we would need is if the upgrade mode changes from savepoint -> last-state when resuming a job we just simply use the savepoint restore logic.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
Aitozi: It seems the `LAST_STATE` only restore from the last checkpoint, It relies on the HA data. The `SAVEPOINT` can restore from the explicit savepoint path. So I think directly naming the `LAST_STATE` to `CHECKPOINT` or `LAST_CHECKPOINT` will more explicit. Not insist on it :).

> Besides , I have a question here: what's the difference between the STATELESS and LAST_STATE it seems we can deleteCluster for both. I don't quite get the current implementation that STATE_LESS only invoke the clusterClient.cancel(), will the resource eventually released in STATE_LESS ?

For this question, I have figured it out that, `cancel` will clean up the HA data, `deleteCluster` will keep the HA data

tweise: `LAST_STATE` should restore from checkpoint or savepoint, whichever is more recent.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
tweise: Maybe we need some documentation on how this should work eventually? There could be a recent checkpoint, even when HA is not enabled. That could be queried through the REST API. Though that path hinges on the availability of the REST API, it would still allow for the upgrade.

gyfora: The rest API is much trickier as there is no guarantee that it is available once the job is stopped if something happens to the process. I think it is a fair requirement that last_state only works with ha enabled now, and maybe we can improve this in the long run.

gyfora: I think for production use-cases this won't be a problem anyways

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/JobReconciler.java
wangyang0918: If we failed to execute `getClusterClient()` in `FlinkService#cancelJob`, the logs will not show up. So I suggest to keep it here.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
wangyang0918: Yes. The pre-check could be done in webhook validation logic.

I agree with @gyfora that relying on REST API is not stable enough. Maybe FLINK-24113 is useful in such case. Since enabling the K8s HA for Flink cluster does not bring any burden, I think most of the users is running Flink applications with HA enabled in production environment. Let's support last-state upgrade without HA in the future if necessary.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/JobReconciler.java
wangyang0918: Nice catch. We could save a `if` here.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
wangyang0918: If we trigger a savepoint, I think it will also increase the checkpoint counter and store in the HA ConfigMaps/ZKNodes. So we do not need to differentiate checkpoint/savepoint here.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/JobReconciler.java
Aitozi: Get it

wangyang0918:
> One cornerncase we might want to cover (but we can also leave this for later and just open a ticket) is if the user suspends a job with savepoint upgrademode and then changes the upgrade mode to last-state when resubmitting it to running state.
> 
> In this case with the current logic the job would simply start from an empty state which is incorrect. So what we would need is if the upgrade mode changes from savepoint -> last-state when resuming a job we just simply use the savepoint restore logic.

This is a valid concern. I will fix this and add a unit test.

Aitozi (approved):

wangyang0918:
cc @gyfora @Aitozi I have pushed a `fixup!` commit to address the comments. Please have a look.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/JobReconciler.java
```diff
@@ -86,23 +86,21 @@ public void reconcile(
...
+                if (upgradeMode == UpgradeMode.STATELESS) {
+                    deployFlinkJob(flinkApp, effectiveConfig, Optional.empty());
+                } else if (upgradeMode == UpgradeMode.SAVEPOINT) {
+                    restoreFromLastSavepoint(flinkApp, effectiveConfig);
+                } else if (upgradeMode == UpgradeMode.LAST_STATE) {
+                    final String savepointLocation =
+                            flinkApp.getStatus().getJobStatus().getSavepointLocation();
+                    // Upgrade mode changes from savepoint -> last-state
+                    deployFlinkJob(
+                            flinkApp, effectiveConfig, Optional.ofNullable(savepointLocation));
```
Aitozi, line 103: IMO, we may need to clean the `savepointLocation` after we cancel the job with `LAST_STATE`, Otherwise, if we have done once `SUSPEND` at `SAVEPOINT` mode, this path persistent in the `JobStatus` It will always be the first choice  in `LAST_STATE` 
 upgrade mode.

About the original case posted by @gyfora 

>  if the user suspends a job with savepoint upgrademode and then changes the upgrade mode to last-state when resubmitting it to running state.
In this case with the current logic the job would simply start from an empty state which is incorrect

I have a question here, why it will start from an empty state? It will start from the latest checkpoint from HA data, It only start from an empty state when there have not generated one checkpoint.

wangyang0918: If the user suspends a job with savepoint upgrademode, the HA ConfigMaps will be cleaned up as well.

wangyang0918: > IMO, we may need to clean the `savepointLocation` after we cancel the job with `LAST_STATE`, Otherwise, if we have done once `SUSPEND` at `SAVEPOINT` mode, this path persistent in the `JobStatus` It will always be the first choice in `LAST_STATE` upgrade mode.

Having the savepoint path in the `JobStatus` is harmless. If we specify both savepoint and HA config options, Flink will tries to pick the latest one. We will introduce manual or periodical savepoint in the future, I think having the savepoint path in the `JobStatus` will make more senses then.

Aitozi: > If we specify both savepoint and HA config options, Flink will try to pick the latest one.

I have no detailed background knowledge for this, I will take a look for it then. Make sense to me, thanks for explanation! +1 for this 👍

gyfora (approved):
Looks good @wangyang0918 , please go ahead and merge it! :)

wangyang0918:
Merging this PR and will create a new ticket to move the sanity check into validator.

PR #23, opened 2022-02-24
[FLINK-26178] Use the same enum for expected and observed jobstate 

bgeng777:
- Rename the `state` field in `JobStatus` to `jobState` and set its type to `JobState` enum.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/status/JobStatus.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/status/JobStatus.java
@@ -17,6 +17,8 @@
 
 package org.apache.flink.kubernetes.operator.crd.status;
 
+import org.apache.flink.kubernetes.operator.crd.spec.JobState;
+
 import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.Data;
@@ -30,7 +32,7 @@ import lombok.NoArgsConstructor;
 public class JobStatus {
     private String jobName;
     private String jobId;
-    private String state;
+    private JobState state;
     private String updateTime;
     private String savepointLocation;
 }
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
@@ -95,7 +95,7 @@ public class JobStatusObserver {
         if (newStatus == null) {
             newStatus = createJobStatus(newJob);
         } else {
-            newStatus.setState(newJob.getJobState().name());
+            newStatus.setState(JobState.valueOf(newJob.getJobState().name()));
             newStatus.setJobName(newJob.getJobName());
             newStatus.setJobId(newJob.getJobId().toHexString());
         }
@@ -106,7 +106,7 @@ public class JobStatusObserver {
         JobStatus jobStatus = new JobStatus();
         jobStatus.setJobId(message.getJobId().toHexString());
         jobStatus.setJobName(message.getJobName());
-        jobStatus.setState(message.getJobState().name());
+        jobStatus.setState(JobState.valueOf(message.getJobState().name()));
         return jobStatus;
     }
 }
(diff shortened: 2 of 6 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/status/JobStatus.java
```diff
@@ -30,7 +32,7 @@
 public class JobStatus {
     private String jobName;
     private String jobId;
-    private String state;
```
gyfora, line 33: Why are we renaming this and not other fields? I agree that it is not super consistent we 
jobName, jobId, state, updateTime, savepointLocation

Wouldnt it be better to have simply name, id, state ...

gyfora: Maybe this could be a separate JIRA discussion to improve field naming in the spec/status and do multiple changes together to simplify testing (redeploying CRD etc.)

gyfora (changes requested):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/status/JobStatus.java
gyfora: So to summarize, my suggestion is to not change the name of the field in this PR, but instead let's open a ticket where we gather some improvements to the spec/status instead.

bgeng777: Sure. It makes sense to me to leave the name change to another ticket. The renaming has been rolled back.
The reason of renaming here is to follow the naming convention of `org.apache.flink.runtime.client.JobStatusMessage` of flink. We could discuss if such change is necessary later.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
```diff
@@ -95,7 +95,7 @@ private JobStatus mergeJobStatus(
         if (newStatus == null) {
             newStatus = createJobStatus(newJob);
         } else {
-            newStatus.setState(newJob.getJobState().name());
+            newStatus.setState(JobState.valueOf(newJob.getJobState().name()));
```
gyfora, line 98: This will give an ugly error if the job is not in a running state. Not sure how to go around this or even if it makes sense to use the operator jobstate enum here after all

bgeng777: Hi @gyfora thanks for sharing this point and I think about it as well and I totally agree that there is some error handle missing here.
I discuss it a little in the [jira](https://issues.apache.org/jira/browse/FLINK-26178). I think using the job state enum should be reasonable as this state itself is an enum. I believe the key is [FLINK-26139](https://issues.apache.org/jira/browse/FLINK-26139) has not been resolved properly which makes the error handle in this part is misleading. I would spend some time on FLINK-26139, hoping to propose a state machine to describe the job state transition later.

This PR seems to be done a little early. We may finish it after we resolve FLINK-26139. What's your advice?

gyfora: Yea I agree we need to first figure out how exactly we want to handle the state transitions and react to different job states. I am sorry to have wasted your time with this ticket, I should have thought more about it before opening it.

bgeng777: Oh, it doesn't matter. The work is not a waste at all, I may just need to refine it once we achieve consensus on the state transition. Thanks a lot for the above review and I will close this PR and may reopen it when [FLINK-26139](https://issues.apache.org/jira/browse/FLINK-26139) is done.

bgeng777:
Will start refining this PR once [FLINK-26139](https://issues.apache.org/jira/browse/FLINK-26139) is done.

PR #25, opened 2022-02-26
[FLINK-26356] Use the common method to construct service name

Aitozi:
Use the common method to construct service name. 

Besides, the `RestOptions.ADDRESS` is always generated at the server side we can not get it from the client config. 
Since the flink-operator is always deployed with the flink job in the same cluster, we can directly use the `Servicename.Namespace` to talk to Flink job rest server.
I'am not sure whether need to remove the `config.getString(RestOptions.ADDRESS, xx)`

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
@@ -32,6 +32,7 @@ import org.apache.flink.configuration.Configuration;
 import org.apache.flink.configuration.RestOptions;
 import org.apache.flink.kubernetes.configuration.KubernetesConfigOptions;
 import org.apache.flink.kubernetes.kubeclient.Fabric8FlinkKubeClient;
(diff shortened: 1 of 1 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
```diff
@@ -107,7 +108,11 @@ public void submitSessionCluster(FlinkDeployment deployment, Configuration conf)
         final int port = config.getInteger(RestOptions.PORT);
         final String host =
                 config.getString(
-                        RestOptions.ADDRESS, String.format("%s-rest.%s", clusterId, namespace));
+                        RestOptions.ADDRESS,
```
wangyang0918, line 111: Currently, the `RestOptions.ADDRESS` is always null in effective configuration. So we could simply always use the rest service name here. Right?

Aitozi: Agree. I will follow your suggestion.

wangyang0918 (approved):
+1 for merging.

PR #27, opened 2022-02-28
[FLINK-26328] Control Logging Behavior in Flink Deployments

morhidi:


morhidi:

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkConfigBuilder.java
```diff
@@ -80,6 +83,17 @@ public FlinkConfigBuilder applyFlinkConfiguration() {
         return this;
     }
 
+    public FlinkConfigBuilder applyLogConfiguration() throws IOException {
+        if (spec.getLogConfiguration() != null) {
+            String confDir =
+                    createLogConfigFiles(
+                            spec.getLogConfiguration().get(CONFIG_FILE_LOG4J_NAME),
```
gyfora, line 90: Since only 2 specific file names (keys) will be recognized and used, we should validate that the user did not specify anything else. 

This check could go straight into the validator component.

morhidi: Thanks, @gyfora. Makes sense!

morhidi: @gyfora  Added the requested validation, PTAL.

PR #28, opened 2022-02-28
[FLINK-26336] Call cancel on deletion & clean up configmaps as well

gyfora:
Seperating the original commit from https://github.com/apache/flink-kubernetes-operator/pull/26 to not interfere with the refactor discussion.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/BaseReconciler.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/BaseReconciler.java
@@ -20,18 +20,21 @@ package org.apache.flink.kubernetes.operator.reconciler;
 import org.apache.flink.configuration.Configuration;
 import org.apache.flink.kubernetes.operator.crd.FlinkDeployment;
 import org.apache.flink.kubernetes.operator.service.FlinkService;
+import org.apache.flink.kubernetes.operator.utils.FlinkUtils;
+import org.apache.flink.kubernetes.operator.utils.IngressUtils;
 
 import io.fabric8.kubernetes.api.model.apps.Deployment;
 import io.fabric8.kubernetes.api.model.apps.DeploymentSpec;
 import io.fabric8.kubernetes.api.model.apps.DeploymentStatus;
+import io.fabric8.kubernetes.client.KubernetesClient;
 import io.javaoperatorsdk.operator.api.reconciler.Context;
+import io.javaoperatorsdk.operator.api.reconciler.DeleteControl;
 import io.javaoperatorsdk.operator.api.reconciler.UpdateControl;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.HashSet;
 import java.util.Optional;
-import java.util.concurrent.TimeUnit;
 
 /** BaseReconciler with functionality that is common to job and session modes. */
 public abstract class BaseReconciler {
@@ -42,6 +45,14 @@ public abstract class BaseReconciler {
     public static final int PORT_READY_DELAY_SECONDS = 10;
 
     private final HashSet<String> jobManagerDeployments = new HashSet<>();
+
+    protected final KubernetesClient kubernetesClient;
+    protected final FlinkService flinkService;
+
+    public BaseReconciler(KubernetesClient kubernetesClient, FlinkService flinkService) {
+        this.kubernetesClient = kubernetesClient;
+        this.flinkService = flinkService;
+    }
 
     public boolean removeDeployment(FlinkDeployment flinkApp) {
         return jobManagerDeployments.remove(flinkApp.getMetadata().getUid());
@@ -54,11 +65,8 @@ public abstract class BaseReconciler {
             Configuration effectiveConfig)
             throws Exception;
 
-    protected UpdateControl<FlinkDeployment> checkJobManagerDeployment(
-            FlinkDeployment flinkApp,
-            Context context,
-            Configuration effectiveConfig,
-            FlinkService flinkService) {
+    protected JobManagerDeploymentStatus checkJobManagerDeployment(
+            FlinkDeployment flinkApp, Context context, Configuration effectiveConfig) {
         if (!jobManagerDeployments.contains(flinkApp.getMetadata().getUid())) {
             Optional<Deployment> deployment = context.getSecondaryResource(Deployment.class);
             if (deployment.isPresent()) {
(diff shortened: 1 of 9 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkUtils.java
```diff
@@ -108,21 +111,38 @@ private static void mergeInto(JsonNode toNode, JsonNode fromNode) {
...
+            boolean deleteHaConfigmaps) {
         kubernetesClient
                 .apps()
                 .deployments()
                 .inNamespace(namespace)
                 .withName(clusterId)
                 .cascading(true)
                 .delete();
+
+        if (deleteHaConfigmaps) {
```
Aitozi, line 138: This seems only handle the situation for HA based on ConfigMap. Will this also work for the HA based on ZK ?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/JobDeploymentStatus.java
```diff
@@ -0,0 +1,58 @@
...
+    MISSING;
+
+    public UpdateControl<FlinkDeployment> toUpdateControl(FlinkDeployment flinkDeployment) {
+        switch (this) {
+            case DEPLOYING:
+                return UpdateControl.updateStatus(flinkDeployment)
+                        .rescheduleAfter(REFRESH_SECONDS, TimeUnit.SECONDS);
+            case DEPLOYED_NOT_READY:
+                return UpdateControl.updateStatus(flinkDeployment)
+                        .rescheduleAfter(PORT_READY_DELAY_SECONDS, TimeUnit.SECONDS);
```
Aitozi, line 51: Maybe make this configurable , I will create a ticket for this.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/BaseReconciler.java
```diff
@@ -75,29 +83,45 @@ public boolean removeDeployment(FlinkDeployment flinkApp) {
...
-                        .rescheduleAfter(REFRESH_SECONDS, TimeUnit.SECONDS);
+
+                return JobDeploymentStatus.DEPLOYING;
             }
+            return JobDeploymentStatus.MISSING;
+        }
+        return JobDeploymentStatus.READY;
+    }
+
+    public DeleteControl shutdownAndDelete(
```
Aitozi, line 107: Can we document here to clarify that what is to be deleted ? Like HA data and cluster resource ?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkUtils.java
gyfora: You are absolutely correct, this only works for kubernetes HA (using configmaps). 

It's a good idea to add comments to the code and highlight in the README.

In the future we could support other HA providers but that is out of scope in this ticket.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/JobDeploymentStatus.java
gyfora: Sounds good, I was trying to preserve the current behaviour :)

Aitozi: I have create the [ticket](https://issues.apache.org/jira/browse/FLINK-26399) for this

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkUtils.java
Aitozi: Make sense to me, I'm OK to merge the current shape with some more comments for clarification.  

But I have sense that we may have hard work to do the nice clean up work for other HA providers, It's a bit of out of scope of the operator responsibility or ability, maybe we should extend at the Flink to support `deleteAndCleanUpHA` ?. Do you have some inputs for this cc @wangyang0918

gyfora:
I discovered that we need to wait for cluster shutdown before deleting the configmaps otherwise they might be recreated. Adding this now

gyfora:
I have addressed the comments @wangyang0918 @tweise , let me know if there is anything else.

I had to improve the waitForClusterShutdown logic to be able to create a safe point where we can delete the configmaps without the jobmanager recreating them.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/JobDeploymentStatus.java
```diff
@@ -0,0 +1,58 @@
...
+
+import io.javaoperatorsdk.operator.api.reconciler.UpdateControl;
+
+import java.util.concurrent.TimeUnit;
+
+import static org.apache.flink.kubernetes.operator.reconciler.BaseReconciler.PORT_READY_DELAY_SECONDS;
+import static org.apache.flink.kubernetes.operator.reconciler.BaseReconciler.REFRESH_SECONDS;
+
+/** Status of the Flink job deployment. */
+public enum JobDeploymentStatus {
```
tweise, line 30: nit: if this applies to session mode, then the `Job` part would be misleading. But this can be dealt with later as we will need to iterate on the states more. It's already great to have a starting point!

tweise:
@gyfora I'm going to run through some of the scenarios manually to verify. Hopefully I can add the missing tests in the next days.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkUtils.java
wangyang0918: The most appropriate way to the HA data clean up is `HighAvailabilityServices#closeAndCleanupAllData()`. It should work both for ZooKeeper and K8s HA. But I agree we could comment the limitation here and do the improvement later.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkUtils.java
```diff
@@ -108,21 +116,109 @@ private static void mergeInto(JsonNode toNode, JsonNode fromNode) {
...
+                PodList jmPodList =
+                        kubernetesClient
+                                .pods()
+                                .inNamespace(namespace)
+                                .withLabel(
+                                        Constants.LABEL_TYPE_KEY, Constants.LABEL_TYPE_NATIVE_TYPE)
+                                .withLabel(
+                                        Constants.LABEL_COMPONENT_KEY,
+                                        Constants.LABEL_COMPONENT_JOB_MANAGER)
+                                .withLabel(Constants.LABEL_APP_KEY, clusterId)
```
wangyang0918, line 183: Could be simplified by `KubernetesUtils.getJobManagerSelectors(clusterId)`

wangyang0918 (approved):
+1 after the minor comments are addressed.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkUtils.java
```diff
@@ -108,21 +116,109 @@ private static void mergeInto(JsonNode toNode, JsonNode fromNode) {
...
+                break;
+            }
+            LOG.info("Waiting for cluster shutdown... ({})", i);
+            try {
+                Thread.sleep(1000);
+            } catch (InterruptedException e) {
+                throw new RuntimeException(e);
+            }
+        }
+    }
```
tweise, line 215: It would be good to log that shutdown was completed, otherwise the last thing we see is:
```2022-02-28 21:12:45,381 o.a.f.k.o.u.FlinkUtils         [INFO ] [default.basic-example] Waiting for cluster shutdown... (5)
2022-02-28 21:12:46,395 o.a.f.k.o.u.FlinkUtils         [INFO ] [default.basic-example] Waiting for cluster shutdown... (6)
```

tweise (approved):
Verified reconciliation scenarios.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkUtils.java
```diff
@@ -108,21 +116,109 @@ private static void mergeInto(JsonNode toNode, JsonNode fromNode) {
...
+                    .configMaps()
+                    .inNamespace(namespace)
+                    .withLabels(
+                            KubernetesUtils.getConfigMapLabels(
+                                    clusterId, LABEL_CONFIGMAP_TYPE_HIGH_AVAILABILITY))
+                    .delete();
+        }
+    }
+
+    /** We need this due to the buggy flink kube cluster client behaviour for now. */
```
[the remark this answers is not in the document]
gyfora: what I meant originally by this, is that the deployment client wont let you submit a new flink job as long as there is still a service around even if marked for deletion. Maybe this is not even a bug, but in any case now the comment is irrelevant with the canged behaviour, will remove it.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/JobDeploymentStatus.java
gyfora: I could rename this to cluster deployment status, then it wouldn't be confusing in case of session

gyfora: or better yet JobManager as you initially named :)

PR #29, opened 2022-03-01
[FLINK-26405] Add validation check of num of JM replica

bgeng777:
- Add JM replicas check in the validator to make sure when HA is enabled, the num of JM should be 1.
- Add the relevant test.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultDeploymentValidator.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultDeploymentValidator.java
@@ -18,6 +18,7 @@
 package org.apache.flink.kubernetes.operator.validation;
 
 import org.apache.flink.configuration.Configuration;
+import org.apache.flink.configuration.HighAvailabilityOptions;
 import org.apache.flink.configuration.MemorySize;
 import org.apache.flink.kubernetes.operator.crd.FlinkDeployment;
 import org.apache.flink.kubernetes.operator.crd.spec.FlinkDeploymentSpec;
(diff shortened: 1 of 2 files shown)


bgeng777:
cc @wangyang0918 @gyfora

gyfora (approved):
Looks good! 👍

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultDeploymentValidator.java
```diff
@@ -86,12 +87,26 @@
...
 
-        return validateResources("JobManager", jmSpec.getResource());
+        return firstPresent(
+                validateResources("JobManager", jmSpec.getResource()),
+                validateJmReplicas("JobManager", jmSpec.getReplicas(), confMap));
+    }
+
+    private Optional<String> validateJmReplicas(
+            String component, int replicas, Map<String, String> confMap) {
+        if (!confMap.containsKey(HighAvailabilityOptions.HA_MODE.key()) && replicas != 1) {
```
wangyang0918, line 102: We should use `HighAvailabilityMode.isHighAvailabilityModeActivated(conf)` instead here.

The following is the pre-check in Flink.
```
    public int getReplicas() {
        final int replicas =
                flinkConfig.get(KubernetesConfigOptions.KUBERNETES_JOBMANAGER_REPLICAS);
        if (replicas < 1) {
            throw new IllegalConfigurationException(
                    String.format(
                            "'%s' should not be configured less than one.",
                            KubernetesConfigOptions.KUBERNETES_JOBMANAGER_REPLICAS.key()));
        } else if (replicas > 1
                && !HighAvailabilityMode.isHighAvailabilityModeActivated(flinkConfig)) {
            throw new IllegalConfigurationException(
                    "High availability should be enabled when starting standby JobManagers.");
        }
        return replicas;
    }
```

gyfora: you are right @wangyang0918 that would be a bit cleaner, we could push a small fixup for this or include this in some followup work. 

Sorry for merging this quickly

bgeng777: My bad for using wrong judgement method. I have followed Yang's advice to fix it in this new #30.

PR #31, opened 2022-03-01
[FLINK-26399] Make some option of operator configurable

Aitozi:
As described in issue, this pr is meant to make some option configurable, with some minor code snippet improvements.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/OperatorConfigOptions.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/OperatorConfigOptions.java
@@ -1,0 +1,41 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.kubernetes.operator.config;
+
+import org.apache.flink.configuration.ConfigOption;
+import org.apache.flink.configuration.ConfigOptions;
+
+/** This class holds configuration constants used by flink operator. */
+public class OperatorConfigOptions {
+
+    public static final ConfigOption<Integer> OPERATOR_RECONCILER_RESCHEDULE_INTERVAL_IN_SEC =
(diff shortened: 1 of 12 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/BaseReconciler.java
```diff
@@ -41,17 +43,32 @@
 
     private static final Logger LOG = LoggerFactory.getLogger(BaseReconciler.class);
 
-    public static final int REFRESH_SECONDS = 60;
-    public static final int PORT_READY_DELAY_SECONDS = 10;
+    public final int reconcileDelayInSec;
+    public final int portCheckIntervalInSec;
```
gyfora, line 47: Maybe these could be captured into it's own configclass for easy access and that way it's going to be much easier to extend in the future

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/JobManagerDeploymentStatus.java
```diff
@@ -41,14 +38,15 @@
     /** JobManager deployment not found, probably not started or killed by user. */
     MISSING;
 
-    public UpdateControl<FlinkDeployment> toUpdateControl(FlinkDeployment flinkDeployment) {
+    public UpdateControl<FlinkDeployment> toUpdateControl(
+            FlinkDeployment flinkDeployment, int reconcileDelayInSec, int portCheckIntervalInSec) {
```
gyfora, line 42: If we have a dedicated configclass for all these values, then it's enough to pass that and can be changed/extended later

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/OperatorConfigOptions.java
```diff
@@ -0,0 +1,41 @@
...
+
+    public static final ConfigOption<Integer> OPERATOR_RECONCILER_RESCHEDULE_DELAY_IN_SEC =
+            ConfigOptions.key("operator.reconciler.reschedule.delay.sec")
+                    .intType()
+                    .defaultValue(60)
+                    .withDescription(
+                            "The delay in second for the controller to reschedule the reconcile process");
+
+    public static final ConfigOption<Integer> OPERATOR_RECONCILER_PORT_CHECK_INTERVAL_IN_SEC =
+            ConfigOptions.key("operator.reconciler.port-check.interval.sec")
```
gyfora, line 35: I think they both should be called interval for consistency. Could we change this in all places to make it consistent please?

gyfora (changes requested):
Added one "minor" comment, otherwise looks really good :)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/OperatorConfigOptions.java
gyfora: also since the checking logic will be moved to the observer, let's call it `operator.observer....`

gyfora:
Also since the refactor commit has been merged now, please rebase on that. Sorry for the trouble

Aitozi:
No trouble :)   I have addressed your comments, please take a look again @gyfora

gyfora (approved):
Looks good, merging. We will probably have to introduce new config options and reconsider the names at one point because now they are used for too many things.

PR #33, opened 2022-03-02
[FLINK-26257] Enable metrics configuration for Prometheus

morhidi:
- added documentation on how to enable Prometheus metrics reporter on the operator
- add option for exposing a dedicated port for metrics on the operator container
- minor fix in the operator metric group name that works for Prometheus

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,3 +84,62 @@ Considering the cost of running the builds, the stability, and the maintainabili
 All the unit tests, integration tests, and the end-to-end tests will be triggered for each PR.
 
 Note: Please make sure the CI passed before merging.
+
+## Operator Metrics
+
+The operator extends the [Flink Metric System](https://nightlies.apache.org/flink/flink-docs-master/docs/ops/metrics/) that allows gathering and exposing metrics to centralized monitoring solutions. The well known [Metric Reporters](https://nightlies.apache.org/flink/flink-docs-master/docs/deployment/metric_reporters) are shipped in the operator image and are ready to use.
+
+### Slf4j
+The default metrics reporter in the operator is Slf4j. It does not require any external monitoring systems, and it is enabled in the operator [Helm chart](helm/flink-operator/templates/flink-operator.yaml) by default, mainly for demonstrating purposes.
+```properties
+metrics.reporter.slf4j.factory.class: org.apache.flink.metrics.slf4j.Slf4jReporterFactory
+metrics.reporter.slf4j.interval: 1 MINUTE
+```
+To use a more robust production grade monitoring solution the configuration needs to be changed.
+
+### Prometheus
+The following example shows how to enable the Prometheus metric reporter:
+```properties
+metrics.reporter.prom.class: org.apache.flink.metrics.prometheus.PrometheusReporter
+metrics.reporter.prom.port: 9999
+```
+Some metric reporters, including the Prometheus, needs a port to be exposed on the container. This can be achieved be defining a value for the otherwise empty `metrics.port` variable.
+Either in the [values.yaml](helm/flink-operator/values.yaml) file:
+```yaml
+metrics:
+  port: 9999
+```
+or using the option `--set metrics.port=9999` in the command line.
+
+The Prometheus Operator among other options provides an elegant, declarative way to specify how group of pods should be monitored using custom resources.
+
+To install the Prometheus operator via Helm run:
+
+```shell
+helm repo add prometheus-community https://prometheus-community.github.io/helm-charts
+helm install prometheus prometheus-community/kube-prometheus-stack
+```
+The Grafana dashboard can be accessed through port-forwarding:
+```shell
+kubectl port-forward deployment/prometheus-grafana 3000
+```
+To enable the operator metrics in Prometheus create a `pod-monitor.yaml` file with the following content:
+```yaml
+apiVersion: monitoring.coreos.com/v1
+kind: PodMonitor
+metadata:
+  name: flink-operator
+  labels:
+    release: prometheus
+spec:
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: flink-operator
+  podMetricsEndpoints:
+      - port: metrics
+```
+and apply it on your Kubernetes environment:
+```shell
+kubectl create -f pod-monitor.yaml
+```
+Once the custom resource is created in the Kubernetes environment the operator metrics are ready to explore [http://localhost:3000/explore](http://localhost:3000/explore).
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/metrics/KubernetesOperatorMetricGroup.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/metrics/KubernetesOperatorMetricGroup.java
@@ -30,7 +30,7 @@ import java.util.Map;
 public class KubernetesOperatorMetricGroup
         extends AbstractMetricGroup<KubernetesOperatorMetricGroup> {
 
-    private static final String GROUP_NAME = "kubernetes-operator";
+    private static final String GROUP_NAME = "k8soperator";
     private String namespace;
     private String name;
     private String hostname;
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/metrics/KubernetesOperatorMetricOptions.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/metrics/KubernetesOperatorMetricOptions.java
@@ -23,9 +23,8 @@ import org.apache.flink.configuration.ConfigOptions;
 /** Configuration options for metrics. */
 public class KubernetesOperatorMetricOptions {
     public static final ConfigOption<String> SCOPE_NAMING_KUBERNETES_OPERATOR =
-            ConfigOptions.key("metrics.scope.kubernetes-operator")
-                    .stringType()
-                    .defaultValue("<host>.kubernetes-operator.<namespace>.<name>")
+            ConfigOptions.key("metrics.scope.k8soperator")
+                    .defaultValue("<host>.k8soperator.<namespace>.<name>")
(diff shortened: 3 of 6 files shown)


morhidi:
cc @wangyang0918 @Aitozi @gyfora

README.md
```diff
@@ -83,3 +83,62 @@ Considering the cost of running the builds, the stability, and the maintainabili
...
+## Operator Metrics
+
+The operator extends the [Flink Metric System](https://nightlies.apache.org/flink/flink-docs-master/docs/ops/metrics/) that allows gathering and exposing metrics to centralized monitoring solutions. The well known [Metric Reporters](https://nightlies.apache.org/flink/flink-docs-master/docs/deployment/metric_reporters) are shipped in the operator image and are ready to use. 
+
+### Slf4j
+The default metrics reporter in the operator is Slf4j. It does not require any external monitoring systems, and it is enabled in the operator Helm chart by default, mainly for demonstrating purposes.
+```properties
+metrics.reporter.slf4j.factory.class: org.apache.flink.metrics.slf4j.Slf4jReporterFactory
+metrics.reporter.slf4j.interval: 1 MINUTE
+```
```
SteNicholas, line 96: It's better to add the template reference link(https://github.com/apache/flink-kubernetes-operator/blob/main/helm/flink-operator/templates/flink-operator.yaml) at the end of this introduction.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/metrics/KubernetesOperatorMetricGroup.java
```diff
@@ -30,7 +30,7 @@
 public class KubernetesOperatorMetricGroup
         extends AbstractMetricGroup<KubernetesOperatorMetricGroup> {
 
-    private static final String GROUP_NAME = "kubernetes-operator";
+    private static final String GROUP_NAME = "kubernetes_operator";
```
SteNicholas, line 33: Please sync `kubernetes_operator` with [KubernetesOperatorMetricGroupTest#70](https://github.com/apache/flink-kubernetes-operator/blob/main/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/metrics/KubernetesOperatorMetricGroupTest.java#L70)

SteNicholas (commented):
@morhidi, thanks for your contribution. I left the minor comments for this pull request. Please take a look.

morhidi:
> @morhidi, thanks for your contribution. I left the minor comments for this pull request. Please take a look.

will do, thanks!

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/metrics/KubernetesOperatorMetricGroup.java
morhidi: fixed

README.md
morhidi: Added a link when we first mention it

morhidi:
@SteNicholas added the requested changes, PTAL

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/metrics/KubernetesOperatorMetricOptions.java
```diff
@@ -23,8 +23,8 @@
 /** Configuration options for metrics. */
 public class KubernetesOperatorMetricOptions {
     public static final ConfigOption<String> SCOPE_NAMING_KUBERNETES_OPERATOR =
-            ConfigOptions.key("metrics.scope.kubernetes-operator")
-                    .defaultValue("<host>.kubernetes-operator.<namespace>.<name>")
+            ConfigOptions.key("metrics.scope.kubernetes_operator")
```
Aitozi, line 26: nit: the underline format for flink option is not a common style.

helm/flink-operator/templates/flink-operator.yaml
```diff
@@ -46,6 +46,10 @@ spec:
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           command: ["/docker-entrypoint.sh", "operator"]
+          ports:
```
Aitozi, line 49: Do we have to expose port by default ?  It's could be decided by a variable like `--set metric.port=9999` ? If not present, we can skip to declare, Right?

Aitozi (commented):
@morhidi LGTM otherwise some little comments, please take a look.

README.md
SteNicholas: @morhidi, LGTM.

helm/flink-operator/templates/flink-operator.yaml
```diff
@@ -136,7 +140,6 @@ data:
   flink-conf.yaml: |+
     metrics.reporter.slf4j.factory.class: org.apache.flink.metrics.slf4j.Slf4jReporterFactory
     metrics.reporter.slf4j.interval: 1 MINUTE
-    # metrics.scope.kubernetes-operator: <host>.kubernetes-operator.<namespace>.<name>
```
SteNicholas, line 139: Why do you remove this line?

SteNicholas (commented):
@morhidi, thanks for your updates. The `kubernetes_operator` naming format doesn't follow the naming pattern in Flink config option. IMO, you could rollback the changes about `kubernetes_operator`.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/metrics/KubernetesOperatorMetricOptions.java
morhidi: Thanks, I'll revert this change, it was requested in the PR

helm/flink-operator/templates/flink-operator.yaml
morhidi: Good idea, I'll make it optional

helm/flink-operator/templates/flink-operator.yaml
morhidi: I'll put it back

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/metrics/KubernetesOperatorMetricOptions.java
morhidi: Reverted

helm/flink-operator/templates/flink-operator.yaml
morhidi: Fixed

Aitozi:
@morhidi Thanks for the updating. Regarding to the group name, I check the code in `org.apache.flink.metrics.prometheus.AbstractPrometheusReporter#replaceInvalidChars` It seems already use the `_` to replace all the illegal character. Do you encounter the problem when you test with `PrometheusReporter` ?

morhidi:
> @morhidi Thanks for the updating. Regarding to the group name, I check the code in `org.apache.flink.metrics.prometheus.AbstractPrometheusReporter#replaceInvalidChars` It seems already use the `_` to replace all the illegal character. Do you encounter the problem when you test with `PrometheusReporter` ?

Yes it complained about it.

morhidi:
> > @Aitozi @SteNicholas how about calling the group `kubernetesoperator`. To align with `jobmanager`, `taskmanager` used in core flink. It seems to be more safe then using either `kubernetes_operator` or `kubernetes-operator`
> 
> @morhidi, the group name `kubernetesoperator` is too long.

since the `operator` is reserved I'm open for any other suggestions

SteNicholas:
@morhidi, what about naming to `k8soperator`?

gyfora:
I dont mind the `k8soperator` if the number doesnt cause a problem 😄

Aitozi:
I'm ok with `k8soperator` :) . But I think the `PrometheusReporter` should also handle the illegal metric name. I will take a look then.

morhidi:
`k8soperator ` it is then :) @gyfora @Aitozi @SteNicholas , PTAL

SteNicholas (approved):
LGTM. cc @wangyang0918 @gyfora

README.md
```diff
@@ -84,3 +84,62 @@ Considering the cost of running the builds, the stability, and the maintainabili
...
+To use a more robust production grade monitoring solution the configuration needs to be changed.
+
+### Prometheus
+The following example shows how to enable the Prometheus metric reporter:
+```properties
+metrics.reporter.prom.class: org.apache.flink.metrics.prometheus.PrometheusReporter
+metrics.reporter.prom.port: 9999
+```
+Some metric reporters, including the Prometheus, needs a port to be exposed on the container. This can be achieved be defining a value for the otherwise empty`metrics.port` variable.
```
wangyang0918, line 106:
```suggestion
Some metric reporters, including the Prometheus, needs a port to be exposed on the container. This can be achieved be defining a value for the otherwise empty `metrics.port` variable.
```

wangyang0918 (approved):
LGTM and manual verification succeeds.

Aitozi (approved):
LGTM +1

Aitozi:
> I'm ok with `k8soperator` :) . But I think the `PrometheusReporter` should already handle the illegal metric name. I will take a closer look then.

@morhidi Sorry for bother you again, I have reproduced the error with group name of `kubernetes-operator`. It is caused by not apply the `CharacterFilter` from the reporter after I applied it to `KubernetesOperatorMetricGroup` It will work with `PrometheusReporter`

```
    @Override
    protected final String getGroupName(CharacterFilter filter) {
        return GROUP_NAME;
    }
```

to 

```
    @Override
    protected final String getGroupName(CharacterFilter filter) {
        return filter.filterCharacters(GROUP_NAME);
    }
```
It's ok with the current solution, just meant to make it clear :)

PR #35, opened 2022-03-02
[FLINK-26308] Limit RBAC to namespaces watched

mbalassi:
Provides the option for the user to switch off creating the cluster scoped RBAC if the `watchNamespaces` value is enabled in the Helm chart. By default it retains the currrent behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ The webhook can be disabled during helm install by passing the `--set webhook.cr
 ### Watching only specific namespaces
 
 The operator supports watching a specific list of namespaces for FlinkDeployment resources. You can enable it by setting the `--set watchNamespaces={flink-test}` parameter.
+When this is enabled role-based access control is only created specifically for these namespaces for the operator and the jobmanagers, otherwise it defaults to cluster scope.
 
 ## User Guide
 ### Create a new Flink deployment
--- a/helm/flink-operator/templates/rbac.yaml
+++ b/helm/flink-operator/templates/rbac.yaml
@@ -16,16 +16,10 @@
 # limitations under the License.
 ################################################################################
 
----
-{{- if .Values.rbac.create }}
----
-apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRole
-metadata:
-  name: flink-operator
-  namespace: {{ .Values.operatorNamespace.name }}
-  labels:
-    {{- include "flink-operator.labels" . | nindent 4 }}
+{{/*
+RBAC rules used to create the (cluster)role based on the operator scope
+*/}}
+{{- define "flink-operator.rbacRules" }}
 rules:
   - apiGroups:
       - flink-operator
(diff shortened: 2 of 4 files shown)


mbalassi:
cc @wangyang0918 @gyfora

gyfora (approved):
Looks good! 👍

wangyang0918:
@mbalassi Sorry for the late response. I am afraid after this change, the k8s operator could not work well if the flink-kubernetes-operator runs in `default` namespace while the watchNamespaces is `flink-test`.

gyfora:
@wangyang0918 do you mean that we have to create an additional role for the operator itself?

mbalassi:
Hi @wangyang0918!

Could you specify what the issue you encountered is? I have tested cases similar to the one you mention and did not run into problems yet. Happy to modify if it is insufficient.

mbalassi:
Hi @wangyang0918!

Could you specify what the issue you encountered is? I have tested cases similar to the one you mention and did not run into problems yet. Happy to modify if it is insufficient.

wangyang0918:
I think we might have following issues what cause the k8s operator could not work.
* FLINK-26436, we are watching the deployment in all namespaces
* The default Flink rest service exposed type is `LoadBalancer`. So k8s operator need the permissions to list nodes. It has 
 been changed to `ClusterIP` in FLINK-24503.
```
2022-03-03 13:35:36,774 ERROR io.javaoperatorsdk.operator.processing.event.ReconciliationDispatcher [] - Error during event processing ExecutionScope{ resource id: CustomResourceID{name='flink-example-statemachine', namespace='flink-test'}, version: 2490360} failed.
org.apache.flink.kubernetes.operator.exception.ReconciliationException: io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: GET at: https://10.96.0.1/api/v1/nodes. Message: Forbidden!Configured service account doesn't have access. Service account may have been revoked. nodes is forbidden: User "system:serviceaccount:default:flink-operator" cannot list resource "nodes" in API group "" at the cluster scope.
        at org.apache.flink.kubernetes.operator.controller.FlinkDeploymentController.reconcile(FlinkDeploymentController.java:137) ~[flink-kubernetes-operator-1.0-SNAPSHOT-shaded.jar:1.0-SNAPSHOT]
        at org.apache.flink.kubernetes.operator.controller.FlinkDeploymentController.reconcile(FlinkDeploymentController.java:57) ~[flink-kubernetes-operator-1.0-SNAPSHOT-shaded.jar:1.0-SNAPSHOT]
        at io.javaoperatorsdk.operator.processing.Controller$2.execute(Controller.java:101) ~[flink-kubernetes-operator-1.0-SNAPSHOT-shaded.jar:1.0-SNAPSHOT]
        at io.javaoperatorsdk.operator.processing.Controller$2.execute(Controller.java:76) ~[flink-kubernetes-operator-1.0-SNAPSHOT-shaded.jar:1.0-SNAPSHOT]
        at io.javaoperatorsdk.operator.api.monitoring.Metrics.timeControllerExecution(Metrics.java:34) ~[flink-kubernetes-operator-1.0-SNAPSHOT-shaded.jar:1.0-SNAPSHOT]
        at io.javaoperatorsdk.operator.processing.Controller.reconcile(Controller.java:75) ~[flink-kubernetes-operator-1.0-SNAPSHOT-shaded.jar:1.0-SNAPSHOT]
        at io.javaoperatorsdk.operator.processing.event.ReconciliationDispatcher.reconcileExecution(ReconciliationDispatcher.java:151) ~[flink-kubernetes-operator-1.0-SNAPSHOT-shaded.jar:1.0-SNAPSHOT]
        at io.javaoperatorsdk.operator.processing.event.ReconciliationDispatcher.handleReconcile(ReconciliationDispatcher.java:117) ~[flink-kubernetes-operator-1.0-SNAPSHOT-shaded.jar:1.0-SNAPSHOT]
        at io.javaoperatorsdk.operator.processing.event.ReconciliationDispatcher.handleDispatch(ReconciliationDispatcher.java:82) ~[flink-kubernetes-operator-1.0-SNAPSHOT-shaded.jar:1.0-SNAPSHOT]
        at io.javaoperatorsdk.operator.processing.event.ReconciliationDispatcher.handleExecution(ReconciliationDispatcher.java:51) [flink-kubernetes-operator-1.0-SNAPSHOT-shaded.jar:1.0-SNAPSHOT]
        at io.javaoperatorsdk.operator.processing.event.EventProcessor$ControllerExecution.run(EventProcessor.java:385) [flink-kubernetes-operator-1.0-SNAPSHOT-shaded.jar:1.0-SNAPSHOT]
        at java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source) [?:?]
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source) [?:?]
        at java.lang.Thread.run(Unknown Source) [?:?]
```
* We might do not need to create the service account in watchNamespaces unless we want the `flink-operator` service account be used by flink jobmanager/taskmanager.


After more consideration, I think we might need to separate the RBAC for flink-kubernetes-operator and flink jobmanager/taskmanager. The later should have less permissions, like this https://github.com/wangyang0918/flink-native-k8s-operator/blob/master/deploy/flink-rbac.yaml.

wangyang0918:
> Hi @wangyang0918!
> 
> Could you specify what the issue you encountered is? I have tested cases similar to the one you mention and did not run into problems yet. Happy to modify if it is insufficient.

If you check the logs of flink-kubernetes-operator, I think you could find some exceptions and the `status` of FlinkDeployment could not be updated correctly.

mbalassi:
Thanks @wangyang0918!

I missed the loadbalancer issue, because I was testing with a job the had the following in the flinkConfig: 😄 
```
kubernetes.rest-service.exposed.type: ClusterIP
```
How do you suggest we should solve this? This means we already should work with flink 1.15 out of the box, but will fail for older flink versions in this specific case (fortunately the default behavior is still the clusterrole, so it works for users just starting out). 

I agree with separating the service accounts, makes my change much cleaner. Will do this under the same Jira.

wangyang0918:
For this configuration, maybe we could set the rest service exposed type to `ClusterIP` in `FlinkUtils#loadDefaultConfig()` if it is not configured explicitly in the ConfigMap `flink-default-config`. Of cause, it could also take effect if the user override it via FlinkDeployment `.spec.flinkConfiguration`.


Separating the service accounts is a great effort. Thanks for doing this.

PR #37, opened 2022-03-03
[FLINK-26447] Clean up webhook jar and dependency management

SteNicholas:
Currently the webhook module builds it's own shaded jar which includes the operator shaded jar contents as well that is unnecessary and simply adds to the size of the flink operator image. Operator dependencies should be configured in the `provided` scope and the operator shaded jar could be put on the classpath when the webhook starts.

**The brief change log**
- Updates the scope of the dependencies `flink-kubernetes-operator`to `provided` and remove the `flink-core` dependencies in the `pom.xml` of the `flink-kubernetes-webhook` module.
- Adds the classpath option with the path of `OPERATOR_JAR` for webhook starting in `docker-entrypoint.sh`.

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -31,7 +31,8 @@ elif [ "$1" = "operator" ]; then
 elif [ "$1" = "webhook" ]; then
     echo "Starting Webhook"
 
-    exec java -jar $LOG_CONFIG /$WEBHOOK_JAR
+    # Adds the operator shaded jar on the classpath when the webhook starts
+    exec java -cp /$OPERATOR_JAR:/$WEBHOOK_JAR $LOG_CONFIG org.apache.flink.kubernetes.operator.admission.FlinkOperatorWebhook
 fi
 
 args=("${args[@]}")
--- a/flink-kubernetes-webhook/pom.xml
+++ b/flink-kubernetes-webhook/pom.xml
@@ -36,66 +36,6 @@ under the License.
             <groupId>org.apache.flink</groupId>
             <artifactId>flink-kubernetes-operator</artifactId>
             <version>${project.version}</version>
-            <exclusions>
-                <exclusion>
-                    <groupId>org.apache.flink</groupId>
-                    <artifactId>flink-core</artifactId>
-                </exclusion>
-                <exclusion>
-                    <groupId>org.apache.flink</groupId>
-                    <artifactId>flink-shaded-netty</artifactId>
-                </exclusion>
-            </exclusions>
-        </dependency>
-
-        <dependency>
-            <groupId>org.apache.flink</groupId>
-            <artifactId>flink-core</artifactId>
-            <version>${flink.version}</version>
-        </dependency>
-
-        <dependency>
-            <groupId>org.apache.flink</groupId>
-            <artifactId>flink-shaded-netty</artifactId>
-            <version>4.1.70.Final-${flink.shaded.version}</version>
         </dependency>
     </dependencies>
-
-    <build>
-        <plugins>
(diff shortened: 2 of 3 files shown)


flink-kubernetes-webhook/pom.xml
```diff
@@ -61,41 +62,4 @@ under the License.
         </dependency>
     </dependencies>
 
-    <build>
-        <plugins>
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-shade-plugin</artifactId>
```
gyfora, line 68: I think we probably need to keep the shade plugin to include the netty classes in our jar. 

But we can definitely remove the `flink-core` dependency and the exclusions for the `flink-kubernetes-operator`

gyfora (changes requested):

docker-entrypoint.sh
```diff
@@ -31,7 +31,8 @@ elif [ "$1" = "operator" ]; then
 elif [ "$1" = "webhook" ]; then
     echo "Starting Webhook"
 
-    exec java -jar $LOG_CONFIG /$WEBHOOK_JAR
+    # Adds the operator shaded jar on the classpath when the webhook starts
+    exec java -classpath $LOG_CONFIG /$OPERATOR_JAR -jar $LOG_CONFIG /$WEBHOOK_JAR
```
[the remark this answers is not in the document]
SteNicholas: @gyfora, sorry for the mistake. I have removed the unused first `$LOG_CONFIG`.

docker-entrypoint.sh
```diff
@@ -31,7 +31,8 @@ elif [ "$1" = "operator" ]; then
 elif [ "$1" = "webhook" ]; then
     echo "Starting Webhook"
 
-    exec java -jar $LOG_CONFIG /$WEBHOOK_JAR
+    # Adds the operator shaded jar on the classpath when the webhook starts
+    exec java -Xbootclasspath/a:/$OPERATOR_JAR -jar $LOG_CONFIG /$WEBHOOK_JAR
```
gyfora, line 35: did you try:
```
exec java $LOG_CONFIG -classpath /$OPERATOR_JAR:/$WEBHOOK_JAR org.apache.flink.kubernetes.operator.admission.FlinkOperatorWebhook
```

gyfora: you should be able to run the webhook on minikube and test there

SteNicholas:
@gyfora, sorry for the mistake to validate the changes on kubeclient using directly executing the java command.

gyfora:
I will test and merge this later today! thank you :)

docker-entrypoint.sh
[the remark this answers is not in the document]
SteNicholas: @gyfora, I have tried both -classpath and -jar together, but meet with `NoFoundClassDefError`.

gyfora:
Please check this commit: https://github.com/gyfora/flink-kubernetes-operator/commit/80f020b3915aee2fdafdc1a38f09c14764d6150e 

Seems to work for me

gyfora:
> Please check this commit: [gyfora@80f020b](https://github.com/gyfora/flink-kubernetes-operator/commit/80f020b3915aee2fdafdc1a38f09c14764d6150e)
> 
> Seems to work for me

We could actually remove the whole shading because the operator jar includes every dependency that we need :)

SteNicholas:
@gyfora, I have verified the above commit which is validated successly and previously I didn't remove all flink dependencies in `flink-kubernetes-webhook` module, which is my fault to miss these are all in `flink-kubernetes-operator` module.

PR #38, opened 2022-03-03
[FLINK-26181] Support manual savepoint triggering in the operator

morhidi:


morhidi:
cc @wangyang0918 @gyfora @tweise

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
```diff
@@ -168,4 +169,18 @@ public void stopSessionCluster(
...
     }
+
+    public void triggerSavepoint(FlinkDeployment deployment, Configuration conf) throws Exception {
+        LOG.info("Triggering savepoint on " + deployment.getMetadata().getName());
+        try (ClusterClient<String> clusterClient = getClusterClient(conf)) {
+            CompletableFuture<String> result =
+                    clusterClient.triggerSavepoint(
+                            JobID.fromHexString(deployment.getStatus().getJobStatus().getJobId()),
+                            null);
+            deployment.getStatus().getJobStatus().setSavepointLocation(result.get());
```
SteNicholas, line 180: Does this need to check the result is valid?

morhidi: fair point, I'll look into this

morhidi: for the time being, the job gets into an error state if it fails to create the checkpoint:

```reconciliationStatus:
    error: 'java.util.concurrent.ExecutionException: org.apache.flink.runtime.checkpoint.CheckpointException:
      Checkpoint triggering task Source: Custom Source (1/2) of job 00000000000000000000000000000000
      is not being executed at the moment. Aborting checkpoint. Failure reason: Not
      all required tasks are currently running.'
```

morhidi: fair point though we could add a more sophisticated error handling

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/SavepointUtils.java
```diff
@@ -0,0 +1,43 @@
...
+/** Savepoint utilities. */
+public class SavepointUtils {
+
+    public static final String SAVEPOINT_ANNOTATION_KEY =
+            "flinkdeployments.flink.apache.org/savepoint";
+    public static final String SAVEPOINT_ANNOTATION_VALUE = "trigger";
+
+    public static boolean checkSavepointAnnotation(FlinkDeployment flinkDeployment) {
+        return flinkDeployment.getMetadata().getAnnotations() != null
+                && SAVEPOINT_ANNOTATION_VALUE.equals(
```
SteNicholas, line 31: Could the value of the `SAVEPOINT_ANNOTATION_VALUE` be a boolean value which is user-friendly and no need to be a specified value？

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
SteNicholas: @morhidi, it's better to add the check of the result and log the failure information for checkpoint failure.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/SavepointUtils.java
morhidi: To my best knowledge k8s annotations are key-value pairs.

SteNicholas: @morhidi, I mean that the `SAVEPOINT_ANNOTATION_KEY` could be `flinkdeployments.flink.apache.org/trigger-savepoint` and `SAVEPOINT_ANNOTATION_VALUE` could be `true`.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
morhidi: @SteNicholas what kind of check do you think is relevant here on the result returned? We could wrap the call in a try-catch, but in this case the user won't get any feedback on the failed savepoint operation. @gyfora @wangyang0918 what do you think? Shall we introduce a new Savepoint object instead of the current savepointLocation field to track the result, error, timestamp etc?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/SavepointUtils.java
morhidi: We can do that, yeah.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
gyfora: Yes I think we could have a Savepoint/SavepointInfo object in the `lastSavepoint` field of the jobstatus to add metadata like creation time that could be valuable

gyfora: If you want you can do that as part of the ticket, or as a follow up work I dont mind either way :)

morhidi: Let's do this in a separate ticket. It affects other parts of the code too cancel with savepoint, restore from savepoint. etc.

gyfora:
@morhidi what happens when the user simultaneously changes the deployment spec and annotates for savepoint? 

Seems like in that case savepointing will be executed immediately after redeploy which will certainly fail, causing the whole update to fail from the operator perspective.

We should clearly define what is the expectation in these cases and add a test case to the controllertest

morhidi:
> @morhidi what happens when the user simultaneously changes the deployment spec and annotates for savepoint?
> 
> Seems like in that case savepointing will be executed immediately after redeploy which will certainly fail, causing the whole update to fail from the operator perspective.
> 
> We should clearly define what is the expectation in these cases and add a test case to the controllertest

At first I though by adding an extra guard `if (!specChanged && SavepointUtils.checkSavepointAnnotation(flinkApp))` would postpone the savepoint creation until we have a running job. But it turned out it is not enough, the tasks should run as well. We can think about this tomorrow.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
tweise: Yes, we should capture the last successful savepoint (or maybe n) in the status. I think we should also capture the error as event on the CR,  not just the log.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/JobReconciler.java
```diff
@@ -109,6 +110,14 @@ public JobReconciler(
             }
         }
 
+        if (SavepointUtils.checkSavepointAnnotation(flinkApp)) {
+            triggerSavePoint(flinkApp, effectiveConfig);
+            SavepointUtils.removeSavepointAnnotation(flinkApp);
```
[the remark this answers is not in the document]
morhidi: Most probably you can enforce such a thing. We can introduce a counter based triggering mechanism later too, but that's less convenient for use IMO.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
morhidi: If we count savepoint failures as reconcile errors, it will be stored there. It's a nice user experience, once you remove the savepoint annotation the reconcile loop starts again.

tweise: Is there a reason not to do that? Since the user expects the savepoint to be taken I think we should treat it as error if it fails.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/JobReconciler.java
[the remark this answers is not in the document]
wangyang0918: I agree with removing the annotation after savepoint is done successfully. Then we will have a semantic that the savepoint is performing if the annotation exists.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/JobReconciler.java
```diff
@@ -109,6 +110,14 @@ public JobReconciler(
             }
         }
 
+        if (SavepointUtils.checkSavepointAnnotation(flinkApp)) {
+            triggerSavePoint(flinkApp, effectiveConfig);
```
wangyang0918, line 114: I think we need to an asynchronous way to do the savepoint since it will take little long time and block the operator thread.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
wangyang0918: I also agree with counting savepoint failures as reconcile errors.

wangyang0918 (commented):
Thanks for working on this useful feature. The changes look good to me. My biggest concern is about the synchronous savepoint execution since it will block the operator thread.

README.md
```diff
@@ -83,3 +83,29 @@ Considering the cost of running the builds, the stability, and the maintainabili
...
+
+
+## Managing Savepoints
+### Manual savepoint triggering
+The operator enables manual savepoint triggering for users. Savepoints can be taken by attaching the `flinkdeployments.flink.apache.org/savepoint=trigger` annotation to the FlinkDeployment's metadata:
+```yaml
+kind: FlinkDeployment
+metadata:
+  annotations:
+    flinkdeployments.flink.apache.org/savepoint: trigger
```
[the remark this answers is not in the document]
SteNicholas: @wangyang0918, IMO, @morhidi addressed my comments for this and forget to update the document.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/JobReconciler.java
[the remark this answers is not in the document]
tweise: My question was related to whether modifying the CR may fail because the operator is normally only expected to modify the status field: https://kubernetes.io/docs/concepts/overview/working-with-objects/kubernetes-objects/#object-spec-and-status

I don't know what the convention for modifying metadata is though.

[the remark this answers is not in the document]
tweise: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status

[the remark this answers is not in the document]
wangyang0918: I am not aware of any convention which prevents we could not remove the annotation. Correct me if I am wrong.

README.md
```diff
@@ -83,3 +83,29 @@ Considering the cost of running the builds, the stability, and the maintainabili
 All the unit tests, integration tests, and the end-to-end tests will be triggered for each PR.
 
 Note: Please make sure the CI passed before merging.
+
+
+## Managing Savepoints
+### Manual savepoint triggering
+The operator enables manual savepoint triggering for users. Savepoints can be taken by attaching the `flinkdeployments.flink.apache.org/savepoint=trigger` annotation to the FlinkDeployment's metadata:
```
SteNicholas, line 90:
```suggestion
The operator enables manual savepoint triggering for users. Savepoints can be taken by attaching the `flinkdeployments.flink.apache.org/trigger-savepoint=true` annotation to the FlinkDeployment's metadata:
```

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/JobReconciler.java
gyfora: I think we can do the Asynch way pretty nicely, as the rest api allows async trigger/polling for the savepoint status. So the flow could be:

 1. If savepoint annotation is set to `trigger`
    - Send savepoint trigger to flink rest api
    - Set savepoint annotation to `in-progress/pending`
    - Store pending savepoint information (trigger id) to job status
 2. If savepoint annotation is set to `in-progress/pending` Observer should poll for savepoint trigger id
   - If finished, remove annotation, update last savepoint
   - If still pending, reschedule

wangyang0918: This is really a great idea. Big +1 for this.

One more thing, if the savepoint is in `in-progress/pending`, we should not schedule a new one. This could be done via `validator`.

morhidi: Ok, I'll work in these ideas today

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/JobReconciler.java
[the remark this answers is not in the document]
tweise: @morhidi we should perhaps reconsider this approach of modifying the CR as it has further reaching implications. Essentially it makes the CR update non-idempotent, which can lead to operational problems. The same CR posted multiple times is not guaranteed to have the same outcome. And this is an issue as the source of the CR would normally be managed in an external system (git repo for example) and applied through a CD pipeline.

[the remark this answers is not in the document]
morhidi: @tweise what would be the alternative solution?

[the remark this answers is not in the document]
Aitozi: I think the concern make sense, since the operator's semantic is try to reconcile to the desired state, and the desired state changed by the operator itself is not common way.
Can we attach the annotation of  `flinkdeployments.flink.apache.org/trigger-savepoint-timestamp=1646411987`
and stored the last triggered timestamp in status. If the status timestamp < the spec timestamp , It will need to trigger again and update the status.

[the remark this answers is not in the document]
morhidi: How about just not removing the annotation?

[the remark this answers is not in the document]
tweise: How about an explicit field `triggerSavepointNonce` that can be modified whenever a trigger is desired. That would work naturally with reconcile semantics.

morhidi:
I still owe the PR with a few tests, but wanted to share how far I got @gyfora @tweise @wangyang0918 @Aitozi @SteNicholas PTAL

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/OperatorConfigOptions.java
```diff
@@ -38,4 +38,11 @@
                     .withDescription(
                             "The interval in second for the controller to reschedule the reconcile process to "
                                     + "wait for deployment to be ready");
+
+    public static final ConfigOption<Integer>
+            OPERATOR_OBSERVER_SAVEPOINT_TRIGGER_GRACE_PERIOD_IN_SEC =
+                    ConfigOptions.key("operator.observer.savepoint.trigger.grace-period.sec")
+                            .intType()
+                            .defaultValue(5)
+                            .withDescription("TODO");
```
gyfora, line 47: Just marking TODO here as well ^^

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/status/SavepointInfo.java
```diff
@@ -0,0 +1,34 @@
...
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+/** Stores savepoint related information. */
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+@Builder
+public class SavepointInfo {
```
gyfora, line 30: Should be `@Value` without the others

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/Observer.java
```diff
@@ -129,13 +135,55 @@ private boolean observeFlinkJobStatus(FlinkDeployment flinkApp, Configuration ef
         }
     }
 
+    public boolean observeSavepointStatus(FlinkDeployment flinkApp, Configuration effectiveConfig) {
+        SavepointInfo savepointInfo = flinkApp.getStatus().getJobStatus().getSavepointInfo();
+        if (savepointInfo.getTriggerId() == null) {
+            LOG.info("Checkpointing not in progress");
```
gyfora, line 141: should be DEBUG

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/Observer.java
```diff
@@ -129,13 +135,55 @@ private boolean observeFlinkJobStatus(FlinkDeployment flinkApp, Configuration ef
...
+        } catch (Exception e) {
+            LOG.error("Exception while fetching savepoint info", e);
+            return false;
+        }
+
+        if (!savepointFetchResult.isTriggered()) {
+            String error = savepointFetchResult.getError();
+            if (error != null
+                    || !SavepointUtils.withinGracePeriod(operatorConfiguration, savepointInfo)) {
+                LOG.info("Savepoint not triggered, setting reconciliation error");
```
gyfora, line 156: should be ERROR

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/Observer.java
```diff
@@ -129,13 +135,55 @@ private boolean observeFlinkJobStatus(FlinkDeployment flinkApp, Configuration ef
...
+                savepointInfo.setTriggerId(null);
+                ReconciliationStatus reconciliationStatus =
+                        flinkApp.getStatus().getReconciliationStatus();
+                reconciliationStatus.setSuccess(false);
+                reconciliationStatus.setError(error != null ? error : "Savepoint status unknown");
```
gyfora, line 161: We should move the FlinkDeploymentController#updateForReconciliationError into a static utility and reuse here

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/BaseReconciler.java
```diff
@@ -66,4 +69,30 @@ private DeleteControl shutdownAndDelete(
...
+        reconciliationStatus.setError(null);
+
+        if (reconciliationStatus.getLastReconciledSpec() != null) {
+            int oldSavepointGeneration =
+                    reconciliationStatus.getLastReconciledSpec().getJob().getSavepointGeneration();
+            FlinkDeploymentSpec cloned = CloneUtils.clone(flinkApp.getSpec());
+            cloned.getJob().setSavepointGeneration(oldSavepointGeneration);
+            reconciliationStatus.setLastReconciledSpec(cloned);
+        } else {
+            reconciliationStatus.setLastReconciledSpec(flinkApp.getSpec());
```
gyfora, line 85: We could clone/copy `flinkApp.getSpec()` before set it in both branches. They should both be copies for consistency

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/BaseReconciler.java
```diff
@@ -66,4 +69,30 @@ private DeleteControl shutdownAndDelete(
...
+
+    protected void updateSavepointReconciliationSuccess(FlinkDeployment flinkApp) {
+        ReconciliationStatus reconciliationStatus = flinkApp.getStatus().getReconciliationStatus();
+        reconciliationStatus.setSuccess(true);
+        reconciliationStatus.setError(null);
+        reconciliationStatus
+                .getLastReconciledSpec()
+                .getJob()
+                .setSavepointGeneration(flinkApp.getSpec().getJob().getSavepointGeneration());
+    }
```
gyfora, line 97: Maybe these 2 methods could move into a utility class together with `updateForReconciliationError`

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/JobReconciler.java
```diff
@@ -189,4 +208,9 @@ protected void shutdown(FlinkDeployment flinkApp, Configuration effectiveConfig)
 
         FlinkUtils.deleteCluster(flinkApp, kubernetesClient, true);
     }
+
+    protected void triggerSavePoint(FlinkDeployment deployment, Configuration effectiveConfig)
```
gyfora, line 212: could be private

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
```diff
@@ -168,4 +181,78 @@ public void stopSessionCluster(
...
+                    JobID.fromHexString(deployment.getStatus().getJobStatus().getJobId()));
+
+            TriggerResponse response =
+                    clusterClient
+                            .sendRequest(
+                                    savepointTriggerHeaders,
+                                    savepointTriggerMessageParameters,
+                                    new SavepointTriggerRequestBody(null, false))
+                            .get();
+            LOG.info("Savepoint triggered: " + response.getTriggerId().toHexString());
```
gyfora, line 202: should be debug

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/JobReconciler.java
```diff
@@ -101,12 +104,22 @@ public JobReconciler(
...
+                                    .getLastSavepoint()
+                                    .getLocation();
                     // Upgrade mode changes from savepoint -> last-state
                     deployFlinkJob(
                             flinkApp, effectiveConfig, Optional.ofNullable(savepointLocation));
                 }
             }
+            updateForSpecReconciliationSuccess(flinkApp);
+        } else if (SavepointUtils.shouldTriggerSavepoint(flinkApp)) {
+            if (flinkApp.getStatus().getJobStatus().getSavepointInfo().getTriggerId() == null) {
```
gyfora, line 119: This check could also be part of `shouldTriggerSavepoint` to make it cleaner

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/JobReconciler.java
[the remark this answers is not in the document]
morhidi: I introduced a new field savepointGeneration in the CR @tweise that serves as a counter

gyfora (changes requested):
Thank you @morhidi , looks really good. I have added a few smaller comments.

It would be also good to point out some of the changes that were required and why (moving reconciliation status update inside the reconciler etc). It makes sense to me, but some of these things are pretty implicit in the PR now.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/status/SavepointInfo.java
morhidi: You sure? That's an immutable version of the `@Data` without setters for example.

gyfora: You are right sorry

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
morhidi: I would prefer to keep this as info

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/OperatorConfigOptions.java
morhidi: yup, fixed

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/Observer.java
morhidi: fixed

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/Observer.java
morhidi: fixed

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/Observer.java
morhidi: done

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/BaseReconciler.java
morhidi: done

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/BaseReconciler.java
morhidi: ack

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/JobReconciler.java
morhidi: fixed

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/JobReconciler.java
morhidi: makes more sense, yeah

morhidi:
> Thank you @morhidi , looks really good. I have added a few smaller comments.
> 
> It would be also good to point out some of the changes that were required and why (moving reconciliation status update inside the reconciler etc). It makes sense to me, but some of these things are pretty implicit in the PR now.

Yeah, my intention was with the update is to allow the interested parties to goof around with it. I'll add same explanation tomorrow along with the tests and docs.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/CloneUtils.java
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.flink.kubernetes.operator.utils;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+/** Utility class for cloning objects. */
+public class CloneUtils {
```
Aitozi, line 25: this is duplicate with `org.apache.flink.kubernetes.operator.admission.admissioncontroller.clone.ObjectMapperCloner` Can extract and use the common one

Aitozi (commented):
Thanks for the update @morhidi , I like the current design of `savepointGeneration`. I left some minor comments please take a look.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/Observer.java
```diff
@@ -129,13 +135,53 @@ private boolean observeFlinkJobStatus(FlinkDeployment flinkApp, Configuration ef
         }
     }
 
+    public boolean observeSavepointStatus(FlinkDeployment flinkApp, Configuration effectiveConfig) {
```
Aitozi, line 138: this can be `private`

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/Observer.java
```diff
@@ -129,13 +135,53 @@ private boolean observeFlinkJobStatus(FlinkDeployment flinkApp, Configuration ef
...
+        } catch (Exception e) {
+            LOG.error("Exception while fetching savepoint info", e);
+            return false;
+        }
+
+        if (!savepointFetchResult.isTriggered()) {
+            String error = savepointFetchResult.getError();
+            if (error != null
+                    || !SavepointUtils.withinGracePeriod(operatorConfiguration, savepointInfo)) {
+                LOG.error("Savepoint not triggered, setting reconciliation error");
```
Aitozi, line 156: Also log the error if not null

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/JobReconciler.java
```diff
@@ -101,12 +104,20 @@ public JobReconciler(
...
+                                    .getSavepointInfo()
+                                    .getLastSavepoint()
+                                    .getLocation();
                     // Upgrade mode changes from savepoint -> last-state
                     deployFlinkJob(
                             flinkApp, effectiveConfig, Optional.ofNullable(savepointLocation));
                 }
             }
+            ReconciliationUtils.updateForSpecReconciliationSuccess(flinkApp);
+        } else if (SavepointUtils.shouldTriggerSavepoint(flinkApp)) {
```
Aitozi, line 118: IMO, This should be one of the sub-condition of `specChanged`. Please correct me if I'm wrong.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/status/SavepointInfo.java
```diff
@@ -0,0 +1,34 @@
...
+    private Savepoint lastSavepoint;
+    private String triggerId;
+    private long triggerTimestamp;
```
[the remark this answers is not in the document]
gyfora: savepointGeneration is part of the status in the lastReconciledSpec field already

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/CloneUtils.java
```diff
@@ -0,0 +1,41 @@
...
+            return null;
+        }
+        try {
+            return (T)
+                    objectMapper.readValue(
+                            objectMapper.writeValueAsString(object), object.getClass());
+        } catch (JsonProcessingException e) {
+            throw new IllegalStateException(e);
+        }
+    }
```
gyfora, line 40: We should move this under `ReconciliationUtils` to avoid creating too many util classes

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
```diff
@@ -0,0 +1,58 @@
...
+
+import org.apache.flink.kubernetes.operator.crd.FlinkDeployment;
+import org.apache.flink.kubernetes.operator.crd.spec.FlinkDeploymentSpec;
+import org.apache.flink.kubernetes.operator.crd.status.ReconciliationStatus;
+import org.apache.flink.kubernetes.operator.utils.CloneUtils;
+
+/** Reconsiliation utilities. */
+public class ReconciliationUtils {
+
+    public static void updateForSpecReconciliationSuccess(FlinkDeployment flinkApp) {
```
gyfora, line 28: I think we should come up with a better name for this method, not sure what 😄

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkOperatorConfiguration.java
```diff
@@ -20,25 +20,17 @@
...
-    }
-
-    public int getReconcileIntervalInSec() {
-        return reconcileIntervalInSec;
-    }
-
-    public int getPortCheckIntervalInSec() {
-        return portCheckIntervalInSec;
-    }
+    private final int savepointTriggerGracePeriodInSec;
```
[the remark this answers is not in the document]
morhidi: indeed, good catch

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/Observer.java
morhidi: yup, thx

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/Observer.java
morhidi: That is a failure case, I'll adjust the code to make it more clear

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/JobReconciler.java
morhidi: We cannot take a savepoint while reconciling the cluster, hence it is separated to another reconcile loop.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/CloneUtils.java
morhidi: done, thanks

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
morhidi: I'll leave this for the creative mindset then :)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/CloneUtils.java
morhidi: I'm hesitating adding the webhook module as a dependency for now. We can think of a shared core module later. Nice catch tho.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/Observer.java
Aitozi: The webhook is depend on the operator module now, we can let the webhook to use the util in the operator

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/CloneUtils.java
Aitozi: The webhook is depend on the operator module now, we can let the webhook to use the util in the operator

morhidi: Yep, we can do this, preferably in a separate ticket.

gyfora: I agree with @Aitozi , lets reuse the clone utility method in the webhook that we just created. Should be a super simple change, let's include it in this one

wangyang0918 (commented):
Sorry for the late response. It seems that the savepoint trigger approach is going to a different direction. I do not insist on the `annotation` way and I am also not against introducing a new field `savepointGeneration` in the CR.

Just one question, should `savepointGeneration` have to increase positively? If it should, then it is not very easy to trigger a new savepoint.

gyfora:
@wangyang0918 We reconsidered the annotatiton approach based on feedback from @tweise and other problems with the busy reconcile loops due to status updates

As for for the `savepointGeneration` semantics, indeed the increasing logic makes triggering slightly more difficult. I am not against making the check `!=` instead. @morhidi wdyt?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/Observer.java
```diff
@@ -129,13 +135,54 @@ private boolean observeFlinkJobStatus(FlinkDeployment flinkApp, Configuration ef
...
+        } catch (Exception e) {
+            LOG.error("Exception while fetching savepoint info", e);
+            return false;
+        }
+
+        if (!savepointFetchResult.isTriggered()) {
+            String error = savepointFetchResult.getError();
+            if (error != null
+                    || SavepointUtils.gracePeriodEnded(operatorConfiguration, savepointInfo)) {
+                LOG.error("Savepoint trigger failure, setting reconciliation error");
```
SteNicholas, line 157: Could this add the above error in the error log?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/Observer.java
```diff
@@ -129,13 +135,54 @@ private boolean observeFlinkJobStatus(FlinkDeployment flinkApp, Configuration ef
...
+            }
+            LOG.info("Savepoint operation not running, waiting within grace period");
+        }
+        if (savepointFetchResult.getSavepoint() == null) {
+            LOG.info("Savepoint not completed yet");
+            return false;
+        }
+
+        savepointInfo.setLastSavepoint(savepointFetchResult.getSavepoint());
+        savepointInfo.setTriggerId(null);
```
SteNicholas, line 171: Why does this set the trigger id of the savepoint info to null?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/Observer.java
```diff
@@ -129,13 +135,54 @@ private boolean observeFlinkJobStatus(FlinkDeployment flinkApp, Configuration ef
...
+                LOG.error("Savepoint trigger failure, setting reconciliation error");
+                savepointInfo.setTriggerId(null);
+                ReconciliationUtils.updateForReconciliationError(
+                        flinkApp, error != null ? error : "Savepoint status unknown");
+                return false;
+            }
+            LOG.info("Savepoint operation not running, waiting within grace period");
+        }
+        if (savepointFetchResult.getSavepoint() == null) {
+            LOG.info("Savepoint not completed yet");
```
SteNicholas, line 166: Does this need to set the trigger id of the savepoint info to null?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/Observer.java
```diff
@@ -129,13 +135,54 @@ private boolean observeFlinkJobStatus(FlinkDeployment flinkApp, Configuration ef
...
+    }
+
     private boolean isReadyToReconcile(FlinkDeployment flinkApp, Configuration effectiveConfig) {
         JobManagerDeploymentStatus jmDeploymentStatus =
                 flinkApp.getStatus().getJobManagerDeploymentStatus();
 
         switch (jmDeploymentStatus) {
             case READY:
-                return observeFlinkJobStatus(flinkApp, effectiveConfig);
+                boolean clusterStable = observeFlinkJobStatus(flinkApp, effectiveConfig);
```
SteNicholas, line 181: Does this be simplified as follows:
```
return observeFlinkJobStatus(flinkApp, effectiveConfig) && observeSavepointStatus(flinkApp, effectiveConfig);
```

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/Observer.java
gyfora: no, this means the savepoint is not completed yet, we will need to observe again

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/JobReconciler.java
```diff
@@ -189,4 +206,9 @@ protected void shutdown(FlinkDeployment flinkApp, Configuration effectiveConfig)
 
         FlinkUtils.deleteCluster(flinkApp, kubernetesClient, true);
     }
+
+    private void triggerSavePoint(FlinkDeployment deployment, Configuration effectiveConfig)
```
SteNicholas, line 210:
```suggestion
    private void triggerSavepoint(FlinkDeployment deployment, Configuration effectiveConfig)
```

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/Observer.java
gyfora: Let's keep it in 2 lines to be more readable.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/Observer.java
gyfora: the trigger id is not null if a savepoint is in progress, this means that it has cmpleted

SteNicholas (changes requested):
@morhidi , sorry for late feedback. I have left the smaller comments on your updates.

wangyang0918:
> @wangyang0918 We reconsidered the annotatiton approach based on feedback from @tweise and other problems with the busy reconcile loops due to status updates
As for for the savepointGeneration semantics, indeed the increasing logic makes triggering slightly more difficult. I am not against making the check != instead. @morhidi wdyt?

Just like @tweise has said, maybe `savepointGeneration` could be renamed to `triggerSavepointNonce`, the value is the `trigger-id`.

gyfora:
if we are going with the "nonce" word I think it should be `savepointTriggerNonce` not the other way around :)

wangyang0918:
> if we are going with the "nonce" word I think it should be `savepointTriggerNonce` not the other way around :)

Make sense. I am not an expert of English grammar :)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/JobReconciler.java
morhidi: thanks

README.md
```diff
@@ -143,3 +143,14 @@ and apply it on your Kubernetes environment:
...
+
+# Savepoints
+Savepoints can be triggered manually by flipping the value of the boolean variable `savepointTriggerNonce` in the job specification:
+```yaml
+ job:
+    jarURI: local:///opt/flink/examples/streaming/StateMachineExample.jar
+    parallelism: 2
+    upgradeMode: savepoint
+    state: running
+    savepointTriggerNonce: false
```
gyfora, line 155: `savepointTriggerNonce` should be an integer type, otherwise it defeats the purpose

gyfora (commented):
`savepointTriggerNonce` must be an integer and not boolean

README.md
morhidi: ok I misunderstood it, i'll fix it then

morhidi: pushed it, @gyfora thx, makes more sense now

PR #39, opened 2022-03-04
The operator restart ingress is lost

leinenglian:


wangyang0918:
@leinenglian Please create a ticket [here](https://issues.apache.org/jira/projects/FLINK/issues) if you believe it is a bug.
When an agreement is reached under the ticket, you will get assigned and then start the working.

PR #41, opened 2022-03-04
[FLINK-26377] Extract the Reconciler interface

Aitozi:
This PR is meant to extract the reconciler interface and create the instance of reconciler on the fly

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/FlinkReconciler.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/FlinkReconciler.java
@@ -1,0 +1,40 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.kubernetes.operator.reconciler;
+
+import org.apache.flink.configuration.Configuration;
+import org.apache.flink.kubernetes.operator.crd.FlinkDeployment;
+
+import io.javaoperatorsdk.operator.api.reconciler.Context;
+import io.javaoperatorsdk.operator.api.reconciler.DeleteControl;
+import io.javaoperatorsdk.operator.api.reconciler.UpdateControl;
+
+/** The internal interface of reconciler. */
+public interface FlinkReconciler {
+
+    UpdateControl<FlinkDeployment> reconcile(
+            String operatorNamespace,
+            FlinkDeployment flinkApp,
+            Context context,
+            Configuration effectiveConfig)
+            throws Exception;
+
+    DeleteControl cleanup(
+            String operatorNamespace, FlinkDeployment flinkApp, Configuration effectiveConfig);
+}
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/FlinkReconcilerFactory.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/FlinkReconcilerFactory.java
@@ -1,0 +1,74 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.kubernetes.operator.reconciler;
+
+import org.apache.flink.kubernetes.operator.config.FlinkOperatorConfiguration;
+import org.apache.flink.kubernetes.operator.crd.FlinkDeployment;
+import org.apache.flink.kubernetes.operator.service.FlinkService;
+
+import io.fabric8.kubernetes.client.KubernetesClient;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+/** The factory to create reconciler based on app mode. */
+public class FlinkReconcilerFactory {
+
+    private final KubernetesClient kubernetesClient;
+    private final FlinkService flinkService;
+    private final FlinkOperatorConfiguration operatorConfiguration;
+    private final Map<Mode, FlinkReconciler> reconcilerMap;
+
+    public FlinkReconcilerFactory(
+            KubernetesClient kubernetesClient,
+            FlinkService flinkService,
+            FlinkOperatorConfiguration operatorConfiguration) {
+        this.kubernetesClient = kubernetesClient;
+        this.flinkService = flinkService;
+        this.operatorConfiguration = operatorConfiguration;
+        this.reconcilerMap = new ConcurrentHashMap<>();
+    }
+
+    public FlinkReconciler getOrCreate(FlinkDeployment flinkApp) {
+        return reconcilerMap.computeIfAbsent(
+                getMode(flinkApp),
+                mode -> {
+                    switch (mode) {
+                        case SESSION:
+                            return new SessionReconciler(
+                                    kubernetesClient, flinkService, operatorConfiguration);
+                        case APPLICATION:
+                            return new JobReconciler(
+                                    kubernetesClient, flinkService, operatorConfiguration);
+                        default:
+                            throw new UnsupportedOperationException(
+                                    String.format("Unsupported running mode: %s", mode));
+                    }
+                });
+    }
+
+    private Mode getMode(FlinkDeployment flinkApp) {
+        return flinkApp.getSpec().getJob() != null ? Mode.APPLICATION : Mode.SESSION;
+    }
+
+    enum Mode {
+        APPLICATION,
+        SESSION
+    }
+}
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/FlinkOperator.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/FlinkOperator.java
@@ -23,8 +23,7 @@ import org.apache.flink.kubernetes.operator.controller.FlinkControllerConfig;
 import org.apache.flink.kubernetes.operator.controller.FlinkDeploymentController;
 import org.apache.flink.kubernetes.operator.metrics.OperatorMetricUtils;
 import org.apache.flink.kubernetes.operator.observer.Observer;
-import org.apache.flink.kubernetes.operator.reconciler.JobReconciler;
-import org.apache.flink.kubernetes.operator.reconciler.SessionReconciler;
+import org.apache.flink.kubernetes.operator.reconciler.FlinkReconcilerFactory;
 import org.apache.flink.kubernetes.operator.service.FlinkService;
 import org.apache.flink.kubernetes.operator.utils.FlinkUtils;
 import org.apache.flink.kubernetes.operator.validation.DefaultDeploymentValidator;
@@ -60,12 +59,10 @@ public class FlinkOperator {
                 FlinkOperatorConfiguration.fromConfiguration(defaultConfig.getOperatorConfig());
 
         Observer observer = new Observer(flinkService);
-        JobReconciler jobReconciler =
-                new JobReconciler(client, flinkService, operatorConfiguration);
-        SessionReconciler sessionReconciler =
-                new SessionReconciler(client, flinkService, operatorConfiguration);
 
         FlinkDeploymentValidator validator = new DefaultDeploymentValidator();
+        FlinkReconcilerFactory factory =
+                new FlinkReconcilerFactory(client, flinkService, operatorConfiguration);
 
         FlinkDeploymentController controller =
                 new FlinkDeploymentController(
@@ -75,8 +72,7 @@ public class FlinkOperator {
                         namespace,
                         validator,
                         observer,
-                        jobReconciler,
-                        sessionReconciler);
+                        factory);
 
         FlinkControllerConfig controllerConfig = new FlinkControllerConfig(controller);
         controller.setControllerConfig(controllerConfig);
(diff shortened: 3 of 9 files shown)


Aitozi:
I introduced the enum `Mode` to distinguish the different run mode in this pr, It is for internal use now, If we have the consensus for the operator supported mode later, we can make this a field at `FlinkDeployment` then.

cc @gyfora @wangyang0918 @tweise PTAL

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/CompositeReconciler.java
```diff
@@ -0,0 +1,86 @@
...
+        return getOrCreateReconciler(mode).cleanup(operatorNamespace, flinkApp, effectiveConfig);
+    }
+
+    private FlinkReconciler getOrCreateReconciler(Mode mode) {
+        return reconcilerMap.computeIfAbsent(
+                mode,
+                m -> factory.create(kubernetesClient, flinkService, operatorConfiguration, mode));
+    }
+
+    private Mode inferMode(FlinkDeployment flinkApp) {
```
gyfora, line 79: I don't really see the value of having a Mode enum, it just adds unnecessary complexity, this simple if branch could be part of the factory logic, then we wouldnt be restricted to hardcoded set of modes

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/FlinkReconciler.java
```diff
@@ -0,0 +1,45 @@
...
+public interface FlinkReconciler {
+
+    UpdateControl<FlinkDeployment> reconcile(
+            String operatorNamespace,
+            FlinkDeployment flinkApp,
+            Context context,
+            Configuration effectiveConfig)
+            throws Exception;
+
+    default DeleteControl cleanup(
```
gyfora, line 41: Why does this have a default basically empty implementaiton? I think we should keep it abstract (required)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/CompositeReconciler.java
```diff
@@ -0,0 +1,86 @@
...
+import io.fabric8.kubernetes.client.KubernetesClient;
+import io.javaoperatorsdk.operator.api.reconciler.Context;
+import io.javaoperatorsdk.operator.api.reconciler.DeleteControl;
+import io.javaoperatorsdk.operator.api.reconciler.UpdateControl;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+/** The composite reconciler which will use the target reconciler based on the app mode. */
+public class CompositeReconciler implements FlinkReconciler {
```
gyfora, line 35: I think if we get rid of the mode we dont even need the CompositeReconciler and use the ReconcilerFactory in place of it.

gyfora (changes requested):
Thanks for the work!

I think we need to make this simpler and avoid having too many layers. I think the Mode enum, and the CompositeReconciler is just added complexity and extra layers. 

I would prefer to use the reconcilerfactory directly from the controller to create the reconciler for the current app. And the factory can return the type of reconciler based on it's logic.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/FlinkReconciler.java
Aitozi: This aligns to the `io.javaoperatorsdk.operator.api.reconciler.Reconciler`, It provides the `default` cleanup. I will remove it.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/CompositeReconciler.java
Aitozi: It acts the lookup key in the reconcileMap, By this we can choose the target reconciler respect to the `FlinkDeploymentApp` mode. My thought is not have to create `Reconciler` each loop, So i use a KeyMap to store it.

gyfora: the factory can still distinghuish the reconcilers and cache them, but this should be an implementation detail of the factory at this point

gyfora: Basically your CompositeReconciler + factory logic should be The Factory :)

Aitozi: Get it, thanks, I will change the implementation

Aitozi:
Thanks for your suggestion, I have addressed your comments. Please take a look again @gyfora

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/FlinkReconcilerFactory.java
```diff
@@ -0,0 +1,74 @@
...
+            KubernetesClient kubernetesClient,
+            FlinkService flinkService,
+            FlinkOperatorConfiguration operatorConfiguration) {
+        this.kubernetesClient = kubernetesClient;
+        this.flinkService = flinkService;
+        this.operatorConfiguration = operatorConfiguration;
+        this.reconcilerMap = new ConcurrentHashMap<>();
+    }
+
+    public FlinkReconciler getOrCreate(FlinkDeployment flinkApp) {
```
gyfora, line 48: Since we have only 2 reconcilers and no resource implications, I would prefer to instantiate and populate the map in the constructor.

This way we can avoid any errors after the reconciler has started

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/FlinkReconciler.java
```diff
@@ -0,0 +1,43 @@
...
+import org.apache.flink.configuration.Configuration;
+import org.apache.flink.kubernetes.operator.crd.FlinkDeployment;
+
+import io.javaoperatorsdk.operator.api.reconciler.Context;
+import io.javaoperatorsdk.operator.api.reconciler.DeleteControl;
+import io.javaoperatorsdk.operator.api.reconciler.UpdateControl;
+
+/**
+ * The internal interface of reconciler, It aligns to the {@link
+ * io.javaoperatorsdk.operator.api.reconciler.Reconciler}.
```
gyfora, line 30: This is a bit different from the operator SDK reconciler as in our case we have validator, observer and reconciler so it is only a subset of the responsitiblities.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/FlinkReconcilerFactory.java
```diff
@@ -0,0 +1,74 @@
...
+                    }
+                });
+    }
+
+    private Mode getMode(FlinkDeployment flinkApp) {
+        return flinkApp.getSpec().getJob() != null ? Mode.JOB : Mode.SESSION;
+    }
+
+    enum Mode {
+        JOB,
```
gyfora, line 71: Instead of JOB we could call it APPLICATION to better represent the deployment mode

gyfora (commented):
Added some minor comments but I think this looks nice and simple now. We can see how this evolves with new deployment modes like standalone etc.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/FlinkReconcilerFactory.java
Aitozi: renamed

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/FlinkReconciler.java
Aitozi: removed the extra comments.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/FlinkReconcilerFactory.java
Aitozi: I lean to keep lazy instantiate behavior, If we instantiate and populate the map in the constructor it will lose the meaning of the `ReconcilerFactory`, It will act like the previous with two pre-instantiated reconciler.

gyfora: Ok we can keep it like this now :)

gyfora (approved):

gyfora:
LGTM, let's give some time for others to chime in with any comments before merging :)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/FlinkReconcilerFactory.java
```diff
@@ -0,0 +1,74 @@
...
+                getMode(flinkApp),
+                mode -> {
+                    switch (mode) {
+                        case SESSION:
+                            return new SessionReconciler(
+                                    kubernetesClient, flinkService, operatorConfiguration);
+                        case APPLICATION:
+                            return new JobReconciler(
+                                    kubernetesClient, flinkService, operatorConfiguration);
+                        default:
```
SteNicholas, line 59: The default branch could be removed because the return value of the `getMode` method only has `APPLICATION` or `SESSION` mode, no other mode here.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/FlinkReconcilerFactory.java
```diff
@@ -0,0 +1,74 @@
...
+        this.reconcilerMap = new ConcurrentHashMap<>();
+    }
+
+    public FlinkReconciler getOrCreate(FlinkDeployment flinkApp) {
+        return reconcilerMap.computeIfAbsent(
+                getMode(flinkApp),
+                mode -> {
+                    switch (mode) {
+                        case SESSION:
+                            return new SessionReconciler(
```
SteNicholas, line 54: IMO, the `BaseReconciler` could add a abstract method like `getMode`, and register the mapping to the `FlinkReconcilerFactory`. Thus, the `getOrCreate` could return the reconciler by the mode.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/FlinkReconciler.java
```diff
@@ -0,0 +1,40 @@
...
+
+import org.apache.flink.configuration.Configuration;
+import org.apache.flink.kubernetes.operator.crd.FlinkDeployment;
+
+import io.javaoperatorsdk.operator.api.reconciler.Context;
+import io.javaoperatorsdk.operator.api.reconciler.DeleteControl;
+import io.javaoperatorsdk.operator.api.reconciler.UpdateControl;
+
+/** The internal interface of reconciler. */
+public interface FlinkReconciler {
```
SteNicholas, line 29: Is it better to naming the interface to `Reconciler`? Because the `BaseReconciler` is the implementation of this interface.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/FlinkReconciler.java
```diff
@@ -0,0 +1,40 @@
...
+
+    UpdateControl<FlinkDeployment> reconcile(
```
SteNicholas, line 31: Please add the comment of the interface.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/FlinkReconciler.java
```diff
@@ -0,0 +1,40 @@
...
+            String operatorNamespace,
+            FlinkDeployment flinkApp,
+            Context context,
+            Configuration effectiveConfig)
+            throws Exception;
+
+    DeleteControl cleanup(
```
SteNicholas, line 38: Ditto.

SteNicholas (commented):
@Aitozi, I left the minor comments for the `Reconciler` interface extraction. cc @gyfora

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/FlinkReconciler.java
tweise: +1, I think `Flink` is redundant

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/FlinkReconcilerFactory.java
tweise, line 48: Ditto, `Flink` in the name doesn't resonate. I would say either just `ReconcilerFactory` or `FlinkDeploymentReconcilerFactory`

tweise:
> I introduced the enum Mode to distinguish the different run mode in this pr, It is for internal use now, If we have the consensus for the operator supported mode later, we can make this a field at FlinkDeployment then.

In its present form I would prefer to not add it to the CRD. It is redundant, presence or absence of job spec provides the same information. I think future standalone mode is orthogonal.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/FlinkReconcilerFactory.java
Aitozi: Yes, But the `default` branch is forced. So I throw an exception there.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/FlinkReconciler.java
Aitozi: will rename it

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/FlinkReconcilerFactory.java
Aitozi: I have considered this way, But I do not quite see the usage of the `getMode`. So I lean to keep the current way, we just have to maintain the instantiation in the `ReconcilerFactory`, and instantiate at need.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/FlinkReconcilerFactory.java
Aitozi: renamed to `ReconcilerFactory`

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/FlinkReconciler.java
Aitozi: updated

Aitozi:
> > I introduced the enum Mode to distinguish the different run mode in this pr, It is for internal use now, If we have the consensus for the operator supported mode later, we can make this a field at FlinkDeployment then.
> 
> In its present form I would prefer to not add it to the CRD. It is redundant, presence or absence of job spec provides the same information. I think future standalone mode is orthogonal.

Agree, I have not add it to the CRD, It's created for the internal use now.

Aitozi:
I have addressed your comments @tweise @SteNicholas , please take a look again.

PR #42, opened 2022-03-04
[FLINK-26308][FLINK-26471] Separate Flink job role from operator role

mbalassi:
This separates the Flink job role from the operator role, which fixes two issues:
1. That when the `operatorNamespace` is the same as one of the `watchNamespaces` the install fails due to attempted duplicated role creation which I introduced in #35.
2. That deleting the operator role affected jobs described in FLINK-26471.

@wangyang0918 has raised another issue with #35 [here](https://github.com/apache/flink-kubernetes-operator/pull/35#issuecomment-1058750206), that prior to 1.15 when using `watchNamespaces` unless the user explicitly sets 
```
kubernetes.rest-service.exposed.type: ClusterIP
```
I would like to avoid overwriting the default Flink behavior in the operator, my suggestion is to present a warning to the user in the admission control validation and we might a line to the Readme / docs anyway.

Also while testing this change I have discovered that the webhook fails when we have multiple entries in `watchNamespaces`. I will fix these two issues in a separate PR early next week.

### DIFF
--- a/examples/basic-checkpoint-ha.yaml
+++ b/examples/basic-checkpoint-ha.yaml
@@ -40,7 +40,7 @@ spec:
       cpu: 1
   podTemplate:
     spec:
-      serviceAccount: flink-operator
+      serviceAccount: flink
       containers:
         - name: flink-main-container
           volumeMounts:
--- a/examples/basic.yaml
+++ b/examples/basic.yaml
@@ -26,7 +26,7 @@ spec:
   flinkVersion: 1.14.3
   flinkConfiguration:
     taskmanager.numberOfTaskSlots: "2"
-  serviceAccount: flink-operator
+  serviceAccount: flink
   jobManager:
     replicas: 1
     resource:
--- a/helm/flink-operator/templates/rbac.yaml
+++ b/helm/flink-operator/templates/rbac.yaml
@@ -17,7 +17,7 @@
 ################################################################################
 
 {{/*
-RBAC rules used to create the (cluster)role based on the operator scope
+RBAC rules used to create the operator (cluster)role based on the scope
 */}}
 {{- define "flink-operator.rbacRules" }}
 rules:
@@ -68,6 +68,27 @@ rules:
     verbs:
       - "*"
 {{- end }}
+
+{{/*
+RBAC rules used to create the job (cluster)role based on the scope
+*/}}
+{{- define "flink-operator.jobRbacRules" }}
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - pods
+      - configmaps
+    verbs:
+      - '*'
+  - apiGroups:
+      - apps
+    resources:
+      - deployments
+    verbs:
+      - '*'
+{{- end }}
+
 ---
 {{- if .Values.rbac.create }}
 ---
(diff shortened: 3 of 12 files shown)


helm/flink-operator/values.yaml
```diff
@@ -36,11 +36,16 @@ rbac:
...
 
-serviceAccount:
+operatorServiceAccount:
   create: true
   annotations: {}
   name: "flink-operator"
 
+jobServiceAccount:
+  create: true
+  annotations: {"helm.sh/resource-policy": keep}
```
gyfora, line 46: could this simply be?
```
annotations: 
    "helm.sh/resource-policy" : keep
```

examples/basic.yaml
```diff
@@ -19,14 +19,14 @@
 apiVersion: flink.apache.org/v1alpha1
 kind: FlinkDeployment
 metadata:
-  namespace: default
+  namespace: flink-user
```
gyfora, line 22: I think we should not change this in the example

gyfora (commented):
Added some minor comments, I will test this manually tomorrow

helm/flink-operator/templates/rbac.yaml
```diff
@@ -132,7 +190,25 @@ roleRef:
   apiGroup: rbac.authorization.k8s.io
 subjects:
   - kind: ServiceAccount
-    name: {{ template "flink-operator.serviceAccountName" . }}
+    name: {{ include "flink-operator.serviceAccountName" . }}
+    namespace: {{ .Values.operatorNamespace.name }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
```
wangyang0918, line 196: Given that watchNamespaces is not configured, I am thinking whether we really need to create the `ClusterRole` and `ClusterRoleBinding` for flink job. Maybe `Role` and `RoleBinding` is enough. If users want to create a FlinkDeployment in a different namespace with flink-k8s-operator, then they should create the `serviceAccount` `Role` `RoleBinding` manually.

examples/basic-checkpoint-ha.yaml
```diff
@@ -40,7 +40,7 @@ spec:
       cpu: 1
   podTemplate:
     spec:
-      serviceAccount: flink-operator
+      serviceAccount: flink
```
wangyang0918, line 43: We could also update the serviceAccount in `e2e-tests/data/cr.yaml` to `flink`.

wangyang0918 (commented):
Manually verified with/without `watchNamespaces` configured. It works well.
I just have left two comments.

helm/flink-operator/templates/rbac.yaml
mbalassi: Fair point, given that we are trying to limit the footprint of the roles created I will create only role/rolebing in the operator namespace.

examples/basic-checkpoint-ha.yaml
mbalassi: Yes, great catch.

mbalassi:
Thanks @wangyang0918 and @gyfora. I have addressed your comments.

mbalassi:
I added an admittedly ugly fix for the webhook not working for multiple namespaces. @wangyang0918 maybe you have a nicer approach for this. The summary is that when we pass the `watchNamespaces` array through the template the output looks as follows:
```
namespaceSelector:
    matchExpressions:
      - key: kubernetes.io/metadata.name
        operator: In
        values: [flink-user default]
```
Which creates a selector that cannot be matched, the correct value would be `["flink-user", "default"]`.

PR #43, opened 2022-03-05
[hotfix] Make flink-kubernetes-shaded dep provided

gyfora:
Solves a problem where we had a compile dependency on a fatjar which does not really work in many cases.


gyfora:
cc @wangyang0918 @tweise

wangyang0918 (approved):
Thanks for the fix. Verified via `mvn clean test` locally.

PR #44, opened 2022-03-07
[FLINK-26370] Use configurable ThreadPool for Flink cluster communication

kelemensanyi:
This PR configures the operator sdk to use an unbounded thread pool for its control loop.

Blocking threads in the reconciler loop for a resource will not hold back the reconciliation of other resources anymore.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/FlinkOperator.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/FlinkOperator.java
@@ -17,6 +17,7 @@
 
 package org.apache.flink.kubernetes.operator;
 
+import org.apache.flink.annotation.VisibleForTesting;
 import org.apache.flink.kubernetes.operator.config.DefaultConfig;
 import org.apache.flink.kubernetes.operator.config.FlinkOperatorConfiguration;
 import org.apache.flink.kubernetes.operator.controller.FlinkControllerConfig;
@@ -30,6 +31,7 @@ import org.apache.flink.kubernetes.operator.validation.DefaultDeploymentValidato
 import org.apache.flink.kubernetes.operator.validation.FlinkDeploymentValidator;
 
 import io.fabric8.kubernetes.client.DefaultKubernetesClient;
+import io.fabric8.kubernetes.client.KubernetesClient;
 import io.javaoperatorsdk.operator.Operator;
 import io.javaoperatorsdk.operator.api.config.ConfigurationService;
 import io.javaoperatorsdk.operator.api.config.ConfigurationServiceOverrider;
@@ -37,31 +39,37 @@ import io.javaoperatorsdk.operator.config.runtime.DefaultConfigurationService;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.concurrent.Executors;
+
 /** Main Class for Flink native k8s operator. */
 public class FlinkOperator {
     private static final Logger LOG = LoggerFactory.getLogger(FlinkOperator.class);
 
-    public static void main(String... args) {
+    private final Operator operator;
+    private final FlinkDeploymentController controller;
+    private final FlinkControllerConfig controllerConfig;
+
+    public FlinkOperator() {
+        this(FlinkUtils.loadDefaultConfig());
+    }
+
+    public FlinkOperator(DefaultConfig defaultConfig) {
 
         LOG.info("Starting Flink Kubernetes Operator");
-        DefaultConfig defaultConfig = FlinkUtils.loadDefaultConfig();
         OperatorMetricUtils.initOperatorMetrics(defaultConfig.getOperatorConfig());
 
-        DefaultKubernetesClient client = new DefaultKubernetesClient();
+        KubernetesClient client = new DefaultKubernetesClient();
         String namespace = client.getNamespace();
         if (namespace == null) {
             namespace = "default";
         }
 
-        ConfigurationService configurationService =
-                new ConfigurationServiceOverrider(DefaultConfigurationService.instance())
-                        .checkingCRDAndValidateLocalModel(false)
-                        .build();
-
-        Operator operator = new Operator(client, configurationService);
-
         FlinkOperatorConfiguration operatorConfiguration =
                 FlinkOperatorConfiguration.fromConfiguration(defaultConfig.getOperatorConfig());
+
+        ConfigurationService configurationService = getConfigurationService(operatorConfiguration);
+
+        operator = new Operator(client, configurationService);
 
         FlinkService flinkService = new FlinkService(client, operatorConfiguration);
 
@@ -70,7 +78,7 @@ public class FlinkOperator {
                 new ReconcilerFactory(client, flinkService, operatorConfiguration);
         ObserverFactory observerFactory = new ObserverFactory(flinkService, operatorConfiguration);
 
-        FlinkDeploymentController controller =
+        controller =
                 new FlinkDeploymentController(
                         defaultConfig,
                         operatorConfiguration,
@@ -80,13 +88,40 @@ public class FlinkOperator {
                         reconcilerFactory,
                         observerFactory);
 
-        FlinkControllerConfig controllerConfig =
+        controllerConfig =
                 new FlinkControllerConfig(controller, operatorConfiguration.getWatchedNamespaces());
         controller.setControllerConfig(controllerConfig);
         controllerConfig.setConfigurationService(configurationService);
+    }
 
+    private ConfigurationService getConfigurationService(
+            FlinkOperatorConfiguration operatorConfiguration) {
+
+        ConfigurationServiceOverrider configOverrider =
+                new ConfigurationServiceOverrider(DefaultConfigurationService.instance())
+                        .checkingCRDAndValidateLocalModel(false);
+
+        int parallelism = operatorConfiguration.getReconcilerMaxParallelism();
+        if (parallelism == -1) {
+            configOverrider = configOverrider.withExecutorService(Executors.newCachedThreadPool());
+        } else {
+            configOverrider = configOverrider.withConcurrentReconciliationThreads(parallelism);
+        }
+        return configOverrider.build();
+    }
+
+    public void run() {
         operator.register(controller, controllerConfig);
         operator.installShutdownHook();
         operator.start();
+    }
+
+    @VisibleForTesting
+    protected Operator getOperator() {
+        return operator;
+    }
+
+    public static void main(String... args) {
+        new FlinkOperator().run();
     }
 }
--- a/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/FlinkOperatorTest.java
+++ b/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/FlinkOperatorTest.java
@@ -1,0 +1,71 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.kubernetes.operator;
+
+import org.apache.flink.configuration.Configuration;
+import org.apache.flink.kubernetes.operator.config.DefaultConfig;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import java.util.Optional;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.ThreadPoolExecutor;
+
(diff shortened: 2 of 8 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/FlinkOperator.java
```diff
@@ -27,33 +27,43 @@
...
 
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
 /** Main Class for Flink native k8s operator. */
 public class FlinkOperator {
     private static final Logger LOG = LoggerFactory.getLogger(FlinkOperator.class);
 
-    public static void main(String... args) {
+    final Operator operator;
```
gyfora, line 49: let's keep this private final and add a protected getter with the @VisibleForTesting annotation

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/FlinkOperator.java
```diff
@@ -78,12 +88,27 @@ public static void main(String... args) {
...
         operator.register(controller, controllerConfig);
         operator.installShutdownHook();
         operator.start();
     }
+
+    protected ConfigurationService getConfigurationService() {
+        return new ConfigurationServiceDecorator(DefaultConfigurationService.instance()) {
+            @Override
+            public ExecutorService getExecutorService() {
+                return Executors.newCachedThreadPool();
```
gyfora, line 106: Should this always be unlimited size? Maybe we should make this configurable, what do you think

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/FlinkOperatorTest.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.IntStream;
+
+/** @link FlinkOperator unit tests. */
+public class FlinkOperatorTest {
+
+    @Test
+    public void testExecutorServiceCanRun100ThreadsParallel() throws Exception {
```
gyfora, line 36: Seems like the test is testing that cached executor service works that it should. This feels pretty unnecessary, it would be enough I think to assert that the executor service that we get back from the operator is the expected type/config.

Even better would be to actually start the operator and try to submit multiple different deployment (different names) in parallel and see that it handles it.

gyfora (commented):
Thank you for the contribution! I have added a few comments

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/FlinkOperator.java
kelemensanyi: The only case I can see when I would set a limit is when I would be getting out of memory. In such a case by setting a limit we could keep the operator running.

I've added a new config option for that: `RECONCILIATION_MAX_PARALLELISM`

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/FlinkOperator.java
kelemensanyi: Fixed now.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/FlinkOperatorTest.java
kelemensanyi: I originally opted for the functional test, because
- the `ExecutorService` interface does not allow to discover its configuration and
- the Controller delegated the construction of the executor service to the standard library, which might return another type of class in future releases.

Functional testing is also not an ideal solution, since we can not test if the pool is really unbounded and as you suggested we might do a bit more testing than it is strictly necessary.

I don't have a strong opinion here, so why not do the simplest then: I've changed the tests as you've suggested. 👍

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/FlinkOperator.java
```diff
@@ -74,12 +88,43 @@ public static void main(String... args) {
...
+                return new ThreadPoolExecutor(
+                        0, maxPoolSize, 60L, TimeUnit.SECONDS, new SynchronousQueue<>());
+            }
+
+            private int getReconciliationMaxPoolSize() {
+                int value =
+                        EnvUtils.getPositiveOrMinusOneInt(
+                                EnvUtils.ENV_RECONCILIATION_MAX_PARALLELISM,
+                                Integer.MAX_VALUE,
+                                "Specify a positive number or -1 for infinite parallelism.");
```
gyfora, line 116: I think it would be a bit nicer and (a lot simpler) to add this to `OperatorConfigOptions` and expose it through the `FlinkOperatorConfiguration` class instead of the env variable. 

We have some env variable based configs still in a few places but I think for things like this the operator config is much better.

gyfora (changes requested):
The changes look good, I only added one comment regarding moving configs from env -> operatorconfiguration. Sorry for the trouble :) 

Also please rebase your commits on top of the master. And squash them together if possible.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/FlinkOperator.java
kelemensanyi: No worries. :) Fixed now, please have another look.

gyfora:
Looks good @kelemensanyi  thank you! I will test it locally and merge afterwards :)

helm/flink-operator/values.yaml
```diff
@@ -61,6 +61,7 @@ operatorConfiguration:
     metrics.reporter.slf4j.interval: 5 MINUTE
 
     operator.reconciler.reschedule.interval.sec: 15
+    operator.reconciliation.max.parallelism: -1
```
gyfora, line 64: we do not need to add this here, it's already in the defaults

gyfora (changes requested):
I have tested this, works well in the unbounded case but unfortunately compeltely fails if we set the number of threads.

The current behaviour with the SynchronizedQueue will reject submissions in case the current thread is in use.

The rejection exception completely breaks the operator and it will stop processing new events afterwards:

```
flink-operator Exception in thread "pool-3-thread-2" java.util.concurrent.RejectedExecutionException: Task flinkdeploymentcontroller -> ExecutionScope{ resource id: CustomResourceID{name='basic-example-2', namespace='default'}, version: 19080} rejected from java.util.concurrent.ThreadPoolExecutor@336f3b56[Running, pool size = 1, active threads = 1, queued tasks = 0, completed tasks = 3]
flink-operator     at java.base/java.util.concurrent.ThreadPoolExecutor$AbortPolicy.rejectedExecution(Unknown Source)
flink-operator     at java.base/java.util.concurrent.ThreadPoolExecutor.reject(Unknown Source)
flink-operator     at java.base/java.util.concurrent.ThreadPoolExecutor.execute(Unknown Source)
flink-operator     at io.javaoperatorsdk.operator.api.config.ExecutorServiceManager$InstrumentedExecutorService.execute(ExecutorServiceManager.java:158)
flink-operator     at io.javaoperatorsdk.operator.processing.event.EventProcessor.submitReconciliationExecution(EventProcessor.java:153)
flink-operator     at io.javaoperatorsdk.operator.processing.event.EventProcessor.eventProcessingFinished(EventProcessor.java:213)
flink-operator     at io.javaoperatorsdk.operator.processing.event.EventProcessor$ControllerExecution.run(EventProcessor.java:386)
flink-operator     at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
flink-operator     at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
flink-operator     at java.base/java.lang.Thread.run(Unknown Source)
```

We should probably use a LinkedBlockingQueue for these cases and set the corePoolSize also to the same as maxPoolsize if I understand correctly

gyfora:
I looked into this a bit more closely. I think the correct solution here is the following:

Remove ConfigurationServiceDecorator and use `ConfigurationServiceOverrider` that does already this, you can see here (https://github.com/apache/flink-kubernetes-operator/pull/91) and here https://github.com/java-operator-sdk/java-operator-sdk/blob/v2.1.1/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationServiceOverrider.java

If the user specifies unbounded threadpool we override the executor service directly like you did now. If the user specifies a bounded number of threads, we only override the number of threads with withConcurrentReconciliationThreads. That will default to `Executors.newFixedThreadPool(concurrentReconciliationThreads());` which seems to work well already. 

Also the default number of threads currently is 5.

helm/flink-operator/values.yaml
kelemensanyi: That's removed now.

kelemensanyi:
Thanks @gyfora for finding out the details about the configuration. It sounds good what you suggested.

I think I implemented that as well, but while testing locally I see the operator restarting sometimes and sometimes it hangs as well. I'm not sure if that's only my local environment, but I have the impression that somehow exceptions don't always get into the logs.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/FlinkOperator.java
```diff
@@ -30,47 +31,55 @@
...
-        Operator operator = new Operator(client, configurationService);
-
         FlinkOperatorConfiguration operatorConfiguration =
                 FlinkOperatorConfiguration.fromConfiguration(defaultConfig.getOperatorConfig());
 
+        ConfigurationServiceOverrider configOverrider =
+                new ConfigurationServiceOverrider(DefaultConfigurationService.instance())
+                        .checkingCRDAndValidateLocalModel(false);
+        configOverrider = setupExecutorService(configOverrider, operatorConfiguration);
+        ConfigurationService configurationService = configOverrider.build();
```
gyfora, line 71: this is getting long we can move it into a dedicated method or one of the util classes and probably merge with `setupExecutorService`

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/FlinkOperatorTest.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.ThreadPoolExecutor;
+
+/** @link FlinkOperator unit tests. */
+public class FlinkOperatorTest {
+
+    @Test
+    public void testExecutorServiceUsesReconciliationMaxParallelismFromConfig() {
+        final var defaultConfig = FlinkUtils.loadDefaultConfig();
+        final var operatorConfig =
```
gyfora, line 35: We can remove the `final` keyword from everywhere in this method it's not really necessary and just adds more code

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/FlinkOperatorTest.java
```diff
@@ -0,0 +1,50 @@
...
+                FlinkOperatorConfiguration.fromConfiguration(defaultConfig.getOperatorConfig());
+        final int maxParallelism = operatorConfig.getReconcilerMaxParallelism();
+        final int threadCount = maxParallelism == -1 ? Integer.MAX_VALUE : maxParallelism;
+        System.out.println(threadCount);
```
gyfora, line 39: pls remove print

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/FlinkOperatorTest.java
```diff
@@ -0,0 +1,50 @@
...
+
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.ThreadPoolExecutor;
+
+/** @link FlinkOperator unit tests. */
+public class FlinkOperatorTest {
+
+    @Test
+    public void testExecutorServiceUsesReconciliationMaxParallelismFromConfig() {
+        final var defaultConfig = FlinkUtils.loadDefaultConfig();
```
gyfora, line 34: you should probably create the flink config directly by instantiating a `new Configuration` that will avoid env dependent test behavior and you can easily test the 2 expected behaviour modes here

gyfora (changes requested):
Added some cosmetic comments, will test this later :)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/FlinkOperator.java
kelemensanyi: Fixed.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/FlinkOperatorTest.java
kelemensanyi: Fixed.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/FlinkOperatorTest.java
kelemensanyi: Fixed.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/FlinkOperatorTest.java
kelemensanyi: Good point. Fixed now.

kelemensanyi:
I still couldn't make it work in a stable manner locally. Would be great if you could give it a try @gyfora .

PR #45, opened 2022-03-07
[FLINK-26508] Webhook should only validate on /validate endpoint end log errors for others

SteNicholas:
The current webhook implementations accept requests on all paths and execute validations, which should restrict this to only /validate and log errors on other rest paths.

**The brief change log**
- `AdmissionHandler` adds the check whether the request path is equal to `/validate` and the log of the error on other rest paths.

### DIFF
--- a/flink-kubernetes-webhook/src/main/java/org/apache/flink/kubernetes/operator/admission/AdmissionHandler.java
+++ b/flink-kubernetes-webhook/src/main/java/org/apache/flink/kubernetes/operator/admission/AdmissionHandler.java
@@ -37,11 +37,14 @@ import org.apache.flink.shaded.netty4.io.netty.handler.codec.http.HttpResponse;
 import org.apache.flink.shaded.netty4.io.netty.handler.codec.http.HttpResponseStatus;
 import org.apache.flink.shaded.netty4.io.netty.handler.codec.http.HttpVersion;
 import org.apache.flink.shaded.netty4.io.netty.handler.codec.http.LastHttpContent;
+import org.apache.flink.shaded.netty4.io.netty.handler.codec.http.QueryStringDecoder;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.fabric8.kubernetes.api.model.GenericKubernetesResource;
 import io.fabric8.kubernetes.api.model.admission.v1.AdmissionReview;
 import org.apache.commons.lang3.exception.ExceptionUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nonnull;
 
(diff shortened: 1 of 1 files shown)


flink-kubernetes-webhook/src/main/java/org/apache/flink/kubernetes/operator/admission/AdmissionHandler.java
```diff
@@ -66,15 +70,26 @@ public AdmissionHandler(Validator<GenericKubernetesResource> validator) {
...
+                review = objectMapper.readValue(in, AdmissionReview.class);
+                AdmissionReview response = validatingController.handle(review);
+                sendResponse(ctx, objectMapper.writeValueAsString(response));
+            } catch (Exception e) {
+                sendError(ctx, ExceptionUtils.getStackTrace(e));
+            }
+        } else {
+            String error =
+                    String.format(
+                            "Webhook accepts the illegal request path '%s', only accepts the request path '/validate'.",
```
gyfora, line 89: Could we please change this to: 
```
Illegal path requested: '%s'. Only /validate is accepted.
```

gyfora (changes requested):
Looks good, I had a small comment to this. 

@SteNicholas would you have time adding a simple test for this ? That would also cover https://issues.apache.org/jira/browse/FLINK-26137 . Two birds, one stone :)

SteNicholas:
@gyfora ,  I would like to create webhook REST api test in this pull request later.

flink-kubernetes-webhook/src/main/java/org/apache/flink/kubernetes/operator/admission/AdmissionHandler.java
SteNicholas: @gyfora , of course yes. The above error message looks good to me.

flink-kubernetes-webhook/src/main/java/org/apache/flink/kubernetes/operator/admission/AdmissionHandler.java
```diff
@@ -66,15 +70,25 @@ public AdmissionHandler(Validator<GenericKubernetesResource> validator) {
...
-        try {
-            InputStream in = new ByteBufInputStream(msgContent);
-            review = objectMapper.readValue(in, AdmissionReview.class);
-            AdmissionReview response = validatingController.handle(review);
-            sendResponse(ctx, objectMapper.writeValueAsString(response));
-        } catch (Exception e) {
-            sendError(ctx, ExceptionUtils.getStackTrace(e));
+        QueryStringDecoder decoder = new QueryStringDecoder(httpRequest.uri());
+        String path = decoder.path();
+        if ("/validate".equals(path)) {
```
Aitozi, line 75: nit: could make `/validate` static and reuse it in the error msg :)

Aitozi, line 75: nit: could make `/validate` static and reuse it in the error msg :)

SteNicholas:
@gyfora , I will push a commit for the Rest test for `AdmissionReview`. Does I need to push another pull request for the test?

PR #46, opened 2022-03-08
[FLINK-26261] Introduce status and create event for terminal deployment error

tweise:
The PR is a sketch for handling a deployment that results in a permanent error state (invalid service account) that would require a CR update to resolve. I do not particularly like the exception based flow, but this scenario requires access to the k8s client.

The event logged looks like this (kubectl describe):
```
ReplicaFailure  FailedCreate  4h48m  JobManagerDeployment  pods "basic-example-d5c476cb6-" is forbidden: error looking up service account default/invalid-flink-operator: serviceaccount "invalid-flink-operator" not found
```


tweise:
@gyfora fyi we should probably rename `JobManagerDeploymentStatus` to something more general like `DeploymentStatus` as that should cover all relevant states, whether they are related to JobManager or not. If that is something agreeable, I'm going to add a ticket.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
```diff
@@ -113,29 +118,70 @@ public DeleteControl cleanup(FlinkDeployment flinkApp, Context context) {
...
-            updateForReconciliationError(flinkApp, ide.getMessage());
+        } catch (DeploymentFailedException dfe) {
+            updateForDeploymentFailed(flinkApp, dfe);
             return UpdateControl.updateStatus(flinkApp);
         } catch (Exception e) {
             throw new ReconciliationException(e);
         }
     }
 
+    private void updateForDeploymentFailed(
```
gyfora, line 143: I think the method name should say something about the event being fired. This is not really just a status update as the others.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobManagerDeploymentStatus.java
```diff
@@ -54,8 +58,9 @@
                                 operatorConfiguration.getPortCheckIntervalInSec(),
                                 TimeUnit.SECONDS);
             case MISSING:
+            case ERROR:
             default:
-                return null;
+                return UpdateControl.noUpdate();
```
gyfora, line 63: An error will trigger a status update, is `noUpdate` correct here?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
```diff
@@ -113,29 +118,70 @@ public DeleteControl cleanup(FlinkDeployment flinkApp, Context context) {
...
+                        .withFirstTimestamp(dfe.deployCondition.getLastTransitionTime())
+                        .withLastTimestamp(dfe.deployCondition.getLastUpdateTime())
+                        .withMessage(dfe.getMessage())
+                        .withNewMetadata()
+                        .withGenerateName(flinkApp.getMetadata().getName())
+                        .withNamespace(flinkApp.getMetadata().getNamespace())
+                        .endMetadata()
+                        .withNewSource()
+                        .withComponent(dfe.component)
+                        .endSource();
```
gyfora, line 177: This should be extracted to a utility method that returns `Event`

gyfora (commented):
Looks really good and simple @tweise . I added a few smaller comments. It would be nice to have 1-2 new cases in the Observer/Controller tests to cover the new functionality.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
```diff
@@ -92,8 +94,11 @@ public DeleteControl cleanup(FlinkDeployment flinkApp, Context context) {
         LOG.info("Stopping cluster {}", flinkApp.getMetadata().getName());
         Configuration effectiveConfig =
                 FlinkUtils.getEffectiveConfig(flinkApp, defaultConfig.getFlinkConfig());
-
-        observer.observe(flinkApp, context, effectiveConfig);
+        try {
+            observer.observe(flinkApp, context, effectiveConfig);
+        } catch (DeploymentFailedException dfe) {
```
SteNicholas, line 99: Does this need to log the exception for the failure of deployment?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
```diff
@@ -113,29 +118,70 @@ public DeleteControl cleanup(FlinkDeployment flinkApp, Context context) {
...
                     reconcilerFactory
                             .getOrCreate(flinkApp)
                             .reconcile(operatorNamespace, flinkApp, context, effectiveConfig);
             updateForReconciliationSuccess(flinkApp);
             return updateControl;
-        } catch (InvalidDeploymentException ide) {
-            LOG.error("Reconciliation failed", ide);
-            updateForReconciliationError(flinkApp, ide.getMessage());
+        } catch (DeploymentFailedException dfe) {
+            updateForDeploymentFailed(flinkApp, dfe);
```
SteNicholas, line 136: Why does this remove the error log?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
```diff
@@ -113,29 +118,70 @@ public DeleteControl cleanup(FlinkDeployment flinkApp, Context context) {
...
+                flinkApp.getMetadata().getNamespace(),
+                flinkApp.getMetadata().getName(),
+                dfe.getMessage());
+        flinkApp.getStatus().setJobManagerDeploymentStatus(JobManagerDeploymentStatus.ERROR);
+        updateForReconciliationError(flinkApp, dfe.getMessage());
+
+        // TODO: avoid repeated event
+        EventBuilder evtb =
+                new EventBuilder()
+                        .withApiVersion("v1")
```
SteNicholas, line 156: Coud this use the ApiVersion of the FlinkDeployment?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobManagerDeploymentStatus.java
```diff
@@ -37,7 +37,11 @@
     DEPLOYING,
 
     /** JobManager deployment not found, probably not started or killed by user. */
-    MISSING;
+    // TODO: currently a mix of SUSPENDED and ERROR, needs cleanup
```
SteNicholas, line 40: Does this need to add the `SUSPENDED` status?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobManagerDeploymentStatus.java
```diff
@@ -37,7 +37,11 @@
...
+    MISSING,
+
+    /** Deployment in terminal error, requires spec change for reconciliation to continue. */
+    ERROR;
```
SteNicholas, line 44: Should the `Observer#isReadyToReconcile` add the `ERROR` case branch?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/Observer.java
```diff
@@ -94,6 +96,14 @@ private void observeJmDeployment(
...
                     status);
 
+            List<DeploymentCondition> conditions = status.getConditions();
+            for (DeploymentCondition dc : conditions) {
+                if ("FailedCreate".equals(dc.getReason())
+                        && "ReplicaFailure".equals(dc.getType())) {
+                    throw new DeploymentFailedException(
+                            DeploymentFailedException.COMPONENT_JOBMANAGER, dc);
+                }
+            }
```
gyfora, line 106: We should not check for new Errors if we are already in ERROR deploymant status, otherwise it wont be possible to submit a new spec to fix the job.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobManagerDeploymentStatus.java
tweise: I think yes, but outside of this PR. We should get rid of MISSING also.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
tweise: No, that version is `flink.apache.org/v1alpha1`

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
tweise: Relevant information is already logged, for example:
```
[ERROR] [default.basic-example] Deployment default/basic-example failed with pods "basic-example-d5c476cb6-" is forbidden: error looking up service account default/invalid-flink-operator: serviceaccount "invalid-flink-operator" not found
```

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/Observer.java
tweise: Done, exception is now only raised when not already in ERROR

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobManagerDeploymentStatus.java
tweise: Good catch, done!

tweise:
@gyfora PTAL

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobManagerDeploymentStatus.java
tweise: seems to work as expected

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
tweise: renamed

gyfora (approved):
Looks great @tweise I will play with this locally a little before merging.

SteNicholas (approved):
LGTM.

PR #50, opened 2022-03-10
[FLINK-26594] Helm - Replace operatorNamespace with Release Namespace

haoxins:
It will be more natural for the end users to customize the namespace with Helm's built-in objects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ A Kubernetes operator for Apache Flink, implemented in Java. See [FLIP-212](http
 The operator is managed helm chart. To install run:
 ```
  cd helm/flink-operator
- helm install flink-operator .
+ helm install flink-operator . --namespace flink-operator --create-namespace
 ```
 
 ### Validating webhook
--- a/helm/flink-operator/templates/serviceaccount.yaml
+++ b/helm/flink-operator/templates/serviceaccount.yaml
@@ -22,7 +22,7 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: {{ include "flink-operator.serviceAccountName" . }}
-  namespace: {{ Release.Namespace }}
+  namespace: {{ .Release.Namespace }}
   labels:
     {{- include "flink-operator.labels" . | nindent 4 }}
   {{- with .Values.operatorServiceAccount.annotations }}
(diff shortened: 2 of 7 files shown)


gyfora:
Hi @haoxins , I think the change makes sense but could you please open a JIRA ticket first?

helm/flink-operator/values.yaml
```diff
@@ -18,9 +18,6 @@
 
 ---
 
-operatorNamespace:
-  name: default
```
gyfora, line 22: We should document this in the README so that it's clear how to select the namespace :) at least one example would be good

haoxins: > I think the change makes sense but could you please open a JIRA ticket first?

Sure, will do~

haoxins: https://issues.apache.org/jira/browse/FLINK-26594

gyfora: perfect, thank you :)

gyfora:
cc @mbalassi

README.md
```diff
@@ -7,7 +7,7 @@ A Kubernetes operator for Apache Flink, implemented in Java. See [FLIP-212](http
 The operator is managed helm chart. To install run:
 ```
  cd helm/flink-operator
- helm install flink-operator .
+ helm install flink-operator . --namespace flink-operator --create-namespace
```
tweise, line 10: Is it the intention to make these parameters mandatory?

mbalassi (changes requested):
Hi @haoxins,

I agree that connecting the operator namespace to the helm release namespace makes a lot of sense. Do you think making it so that if the user does not specify a namespace it just defaults to the `default` namespace is a decent approach?

helm/flink-operator/templates/serviceaccount.yaml
```diff
@@ -22,7 +22,7 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: {{ include "flink-operator.serviceAccountName" . }}
-  namespace: {{ .Values.operatorNamespace.name }}
+  namespace: {{ Release.Namespace }}
```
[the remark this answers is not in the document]
haoxins: yeah, thanks~

haoxins:
> Hi @haoxins,
> 
> I agree that connecting the operator namespace to the helm release namespace makes a lot of sense. Do you think making it so that if the user does not specify a namespace it just defaults to the `default` namespace is a decent approach?

I think if the user does not specify the `--namespace` param, the value will be the default specified in the user's local Kubernetes configuration file (the user's current K8s `ns`).

mbalassi (approved):
Thanks - makes sens, will give it another round of testing and merge tomorrow.

PR #51, opened 2022-03-10
[FLINK-26572] Improve reconcile reschedule configs and defaults

gyfora:
Make config names cleaner and reduce reschedule delays while the job is deploying.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkOperatorConfiguration.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkOperatorConfiguration.java
@@ -26,26 +26,33 @@ import lombok.Value;
 @Value
 public class FlinkOperatorConfiguration {
 
-    int reconcileIntervalInSec;
-
-    int portCheckIntervalInSec;
-
-    int savepointTriggerGracePeriodInSec;
+    int reconcileIntervalSeconds;
+    int progressCheckIntervalSeconds;
+    int restApiReadyDelaySeconds;
+    int savepointTriggerGracePeriodSeconds;
 
     public static FlinkOperatorConfiguration fromConfiguration(Configuration operatorConfig) {
-        int reconcileIntervalInSec =
+        int reconcileIntervalSeconds =
                 operatorConfig.getInteger(
                         OperatorConfigOptions.OPERATOR_RECONCILER_RESCHEDULE_INTERVAL_IN_SEC);
-        int portCheckIntervalInSec =
-                operatorConfig.getInteger(
-                        OperatorConfigOptions.OPERATOR_OBSERVER_PORT_CHECK_INTERVAL_IN_SEC);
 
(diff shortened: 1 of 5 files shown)


gyfora:
cc @tweise @wangyang0918

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkOperatorConfiguration.java
```diff
@@ -26,26 +26,33 @@
...
+    int savepointTriggerGracePeriodSeconds;
 
     public static FlinkOperatorConfiguration fromConfiguration(Configuration operatorConfig) {
-        int reconcileIntervalInSec =
+        int reconcileIntervalSeconds =
                 operatorConfig.getInteger(
                         OperatorConfigOptions.OPERATOR_RECONCILER_RESCHEDULE_INTERVAL_IN_SEC);
-        int portCheckIntervalInSec =
+
+        int restApiReadyDelaySeconds =
```
wangyang0918, line 39: Out the scope of this PR.

To be honest, I do not like the `restApiReadyDelaySeconds` configuration. This make the `JobManagerDeploymentStatus.READY` not really ready for accepting REST API calls. This will be problem if we are running session cluster.

wangyang0918 (approved):
+1

I just left an unrelated comment.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkOperatorConfiguration.java
tweise: Why would this affect session cluster? There is a delay between the port becoming available and the actual rest server being ready which absent any other reliable technique to check we have the (now configurable!) delay for.

tweise (approved):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkOperatorConfiguration.java
wangyang0918: For example, if something is wrong or very slow with the leader election, `JobManagerDeploymentStatus` is `READY ` but the session cluster is not ready for accepting job submission. What I mean is to update the `JobManagerDeploymentStatus` the `READY` only when the flink cluster is actually working.

gyfora: I agree, the current check is not very robust and works only in a very specific scenario. Let's open a ticket to track this effort and see if we can come up with something better :)

wangyang0918: I have created a ticket FLINK-26605 to track this effort.

PR #52, opened 2022-03-10
[FLINK-26574] Expose config files in helm chart

gyfora:
Main changes:
 - Move the default config and logging properties to files from the helm templates
 - Allow users to easily override operator and flink default configs from helm chart values
 - Allow users to use easily use their own configmaps instead of ones created by helm
 
This PR should be merged only after https://github.com/apache/flink-kubernetes-operator/pull/51 as the config keys have already been adapted to that PR.

### DIFF
--- a/helm/flink-operator/values.yaml
+++ b/helm/flink-operator/values.yaml
@@ -55,10 +55,32 @@ webhook:
   #   name: jks-password-secret
   #   key: password-key
 
-imagePullSecrets: []
-nameOverride: ""
-fullnameOverride: ""
+operatorConfiguration:
+  create: true
+  # Set append to false to replace configuration files
+  # append: true
+  flink-conf.yaml: |+
+    metrics.reporter.slf4j.factory.class: org.apache.flink.metrics.slf4j.Slf4jReporterFactory
+    metrics.reporter.slf4j.interval: 5 MINUTE
+
+    operator.reconciler.reschedule.interval.sec: 15
+    operator.observer.progress-check.interval.sec: 5
+  log4j2.properties: |+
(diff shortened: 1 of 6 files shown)


gyfora:
cc @morhidi @mbalassi

morhidi:
Thank @gyfora, I'll play with this today.

morhidi:
We can also consider putting all the default config values in the values.yaml as is:
https://github.com/elastic/helm-charts/blob/d915d8d53721d7a0535b0956509c6a6e8756f3ec/elasticsearch/values.yaml#L31

gyfora:
@morhidi based on the feedback I made the following improvements:
Allow users to append-to or replace all config files (yaml and logging props) from values.

Default behaviour is append (to the default configs defined in the conf directories)

helm/flink-operator/values.yaml
```diff
@@ -55,10 +55,32 @@ webhook:
...
+  flink-conf.yaml: |+
+    metrics.reporter.slf4j.factory.class: org.apache.flink.metrics.slf4j.Slf4jReporterFactory
+    metrics.reporter.slf4j.interval: 5 MINUTE
+
+    operator.reconciler.reschedule.interval.sec: 15
+    operator.observer.progress-check.interval.sec: 5
+  log4j2.properties: |+
+    rootLogger.level = DEBUG
+  # Set append to false to replace configuration files
+  # append: true
```
morhidi, line 69: Looks better :) I suggest to move this property under `create: true`. I missed it at first sight.

morhidi (changes requested):

PR #54, opened 2022-03-11
Use localhost for REST API when running outside k8s

tweise:
A small improvement to have a good default when running the operator outside of k8s. (I find myself switching back and forth and don't want to rely on local files.) As mentioned in the comment, this can become an operator config if we agree that this is useful.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
@@ -32,6 +32,7 @@ import org.apache.flink.configuration.Configuration;
 import org.apache.flink.configuration.RestOptions;
 import org.apache.flink.kubernetes.configuration.KubernetesConfigOptions;
 import org.apache.flink.kubernetes.kubeclient.decorators.ExternalServiceDecorator;
+import org.apache.flink.kubernetes.operator.config.FlinkOperatorConfiguration;
 import org.apache.flink.kubernetes.operator.crd.FlinkDeployment;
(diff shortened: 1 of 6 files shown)


morhidi:
I ended up adding the namespaced rest service names (basic-example-rest.default) into the `/etc/hosts` file when running the operator locally. This fix doesn't harm though. LGTM

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
```diff
@@ -140,8 +141,13 @@ public boolean isJobManagerPortReady(Configuration config) {
         final String clusterId = config.get(KubernetesConfigOptions.CLUSTER_ID);
         final String namespace = config.get(KubernetesConfigOptions.NAMESPACE);
         final int port = config.getInteger(RestOptions.PORT);
-        final String host =
+        String host =
                 ExternalServiceDecorator.getNamespacedExternalServiceName(clusterId, namespace);
+        if (EnvUtils.get("KUBERNETES_SERVICE_HOST") == null) {
+            // not running in k8s, simplify local development
+            // TODO: make this an operator config and pass that config through
```
gyfora, line 148: If we are planning to make this configurable I suggest we do it straight away. Let's not leave TODOs for such minor additions.

tweise: Haha yes, that is actually what I'm planning to do if the change as such is acceptable. Will have to pass `DefaultConfig` for that, not just the effective config for Flink. But that is probably a change worthwhile doing anyways as we are likely to add more settings that govern the operator behavior going forward.

tweise:
@gyfora PTAL

PR #55, opened 2022-03-13
[FLINK-26546] Extract Observer Interface

Aitozi:
This PR is mean to extract the `Observer` interface to simplify the observe logic. The main logic is aligned to the `Reconciler` interface

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/BaseObserver.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/BaseObserver.java
@@ -1,0 +1,117 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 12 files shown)


Aitozi:
cc @morhidi @gyfora @wangyang0918

gyfora:
Thanks @Aitozi, looks good at a first glance, I will review this in detail tomorrow morning :)

wangyang0918:
Just a personal taste, compared to inheritance, I prefer composition. The problem of inheritance is that we do not know which methods in `BaseReconciler` and `BaseObserver` will be overridden by a subclass. Currently, it is not a major problem since we have very few methods in the interface.

Note: I do not mean we have to change in this PR.

Aitozi:
@wangyang0918 Thanks for sharing your insights. IMO, one of the benefits of inherited `BaseXX` class is that it can provide the `template` ability and leave it implemented by the subclass. The composition seems can not achieve the same effect. 
For example, in the `BaseReconciler#cleanup` provide the common template part, and the both subclass only have to handle how to `shutdown` the corresponding component.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/BaseObserver.java
```diff
@@ -0,0 +1,117 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.List;
+import java.util.Optional;
+
+/** The base observer. */
+public abstract class BaseObserver implements Observer {
+
+    protected final Logger logger = LoggerFactory.getLogger(this.getClass());
```
bgeng777, line 42: nit: the logger can be static to reduce extra memory footprint.

Aitozi: It's intended for the sharing of `logger` in subclass

gyfora:
I also like the BaseXX approach, I think now we have a nice clean separation of observers, and reconcilers :)

PR #56, opened 2022-03-14
[FLINK-26473] Check for deployment errors when listJobs fails

tweise:
Adds handling for listJobs errors and with that ongoing error checking for the underlying deployment.

Need to add test coverage, this is for early comments.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/exception/DeploymentFailedException.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/exception/DeploymentFailedException.java
@@ -19,6 +19,7 @@ package org.apache.flink.kubernetes.operator.exception;
 
 import org.apache.flink.kubernetes.operator.crd.FlinkDeployment;
 
+import io.fabric8.kubernetes.api.model.ContainerStateWaiting;
 import io.fabric8.kubernetes.api.model.Event;
 import io.fabric8.kubernetes.api.model.EventBuilder;
 import io.fabric8.kubernetes.api.model.apps.DeploymentCondition;
@@ -26,15 +27,33 @@ import io.fabric8.kubernetes.api.model.apps.DeploymentCondition;
 /** Exception to signal terminal deployment failure. */
 public class DeploymentFailedException extends RuntimeException {
     public static final String COMPONENT_JOBMANAGER = "JobManagerDeployment";
+    public static final String REASON_CRASH_LOOP_BACKOFF = "CrashLoopBackOff";
+
     private static final long serialVersionUID = -1070179896083579221L;
 
     public final String component;
-    public final DeploymentCondition deployCondition;
+    public final String type;
+    public final String reason;
+    public final String lastTransitionTime;
+    public final String lastUpdateTime;
 
     public DeploymentFailedException(String component, DeploymentCondition deployCondition) {
         super(deployCondition.getMessage());
         this.component = component;
-        this.deployCondition = deployCondition;
+        this.type = deployCondition.getType();
+        this.reason = deployCondition.getReason();
+        this.lastTransitionTime = deployCondition.getLastTransitionTime();
+        this.lastUpdateTime = deployCondition.getLastUpdateTime();
+    }
+
+    public DeploymentFailedException(
+            String component, String type, ContainerStateWaiting stateWaiting) {
+        super(stateWaiting.getMessage());
+        this.component = component;
+        this.type = type;
+        this.reason = stateWaiting.getReason();
+        this.lastTransitionTime = null;
+        this.lastUpdateTime = null;
     }
 
     public static Event asEvent(DeploymentFailedException dfe, FlinkDeployment flinkApp) {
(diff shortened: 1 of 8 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/Observer.java
```diff
@@ -75,12 +80,15 @@ private boolean isApplicationClusterReady(FlinkDeployment dep) {
...
+            FlinkDeployment flinkApp,
+            Context context,
+            Configuration effectiveConfig,
+            boolean checkOnReady) {
         FlinkDeploymentStatus deploymentStatus = flinkApp.getStatus();
         JobManagerDeploymentStatus previousJmStatus =
                 deploymentStatus.getJobManagerDeploymentStatus();
 
-        if (JobManagerDeploymentStatus.READY == previousJmStatus) {
+        if (JobManagerDeploymentStatus.READY == previousJmStatus && !checkOnReady) {
```
tweise, line 91: Rather than adding the flag, this can be moved into observe.

SteNicholas: @tweise +1, the branch that the REST server to be ready is only covered in `observe` method. The exception like `TimeoutException` doesn't need to call `isJobManagerPortReady` to check.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/Observer.java
```diff
@@ -127,14 +136,26 @@ private void observeJmDeployment(
                     return;
                 }
             }
+            // checking the pod is expensive; we only do it when the deployment isn't ready
+            PodList jmPods = flinkService.getJmPodList(flinkApp, effectiveConfig);
+            for (Pod pod : jmPods.getItems()) {
+                for (ContainerStatus cs : pod.getStatus().getContainerStatuses()) {
+                    ContainerStateWaiting csw = cs.getState().getWaiting();
+                    if ("CrashLoopBackOff".equals(csw.getReason())) {
```
SteNicholas, line 144: Could this check be `!StringUtils.isEmpty(csw.getReason())  && csw.getReason().toLowerCase().contains("CrashLoopBackOff".toLowerCase())`?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/Observer.java
```diff
@@ -96,23 +104,24 @@ private void observeJmDeployment(
             if (status != null
                     && status.getAvailableReplicas() != null
                     && spec.getReplicas().intValue() == status.getReplicas()
-                    && spec.getReplicas().intValue() == status.getAvailableReplicas()
-                    && flinkService.isJobManagerPortReady(effectiveConfig)) {
-
-                // typically it takes a few seconds for the REST server to be ready
+                    && spec.getReplicas().intValue() == status.getAvailableReplicas()) {
```
gyfora, line 107: Could we please move these 4 conditions into a `deploymentExists` or something similar method to simplify the flow?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/Observer.java
```diff
@@ -96,23 +104,24 @@ private void observeJmDeployment(
...
+                if (flinkService.isJobManagerPortReady(effectiveConfig)) {
```
gyfora, line 108: I don't really see the point in moving this into a nested if branch with fall-through logic. I think the simple if-else before was correct.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/Observer.java
```diff
@@ -96,23 +104,24 @@ private void observeJmDeployment(
...
+                    // typically it takes a few seconds for the REST server to be ready
+                    LOG.info(
+                            "JobManager deployment {} in namespace {} port ready, waiting for the REST API...",
+                            flinkApp.getMetadata().getName(),
+                            flinkApp.getMetadata().getNamespace());
+                    deploymentStatus.setJobManagerDeploymentStatus(
+                            JobManagerDeploymentStatus.DEPLOYED_NOT_READY);
+                    return;
+                }
+            } else {
```
gyfora, line 118: We could also remove the else branch here (as we return early in other cases) to flatten the logic a bit to make it more readable.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/Observer.java
```diff
@@ -127,14 +136,26 @@ private void observeJmDeployment(
...
                 }
             }
+            // checking the pod is expensive; we only do it when the deployment isn't ready
+            PodList jmPods = flinkService.getJmPodList(flinkApp, effectiveConfig);
+            for (Pod pod : jmPods.getItems()) {
+                for (ContainerStatus cs : pod.getStatus().getContainerStatuses()) {
+                    ContainerStateWaiting csw = cs.getState().getWaiting();
+                    if ("CrashLoopBackOff".equals(csw.getReason())) {
+                        LOG.warn("JobManager pod fails: {} {}", csw.getReason(), csw.getMessage());
+                        return;
```
gyfora, line 146: Could we move this check to a separate method?

gyfora (changes requested):
Thank you @tweise! I will wait for the tests before doing a second pass on this. I have added some comments regarding the branching in the observe logic.

As we are planning to add more things to this parts in the coming tickets, we should try to streamline this and move checks to different methods when possible.

tweise:
@gyfora PTAL

gyfora (approved):
Looks good!

PR #58, opened 2022-03-14
[FLINK-26528] Trigger the updateControl only when the FlinkDeployment…

Aitozi:
… have changed

This PR is meant to eliminate the duplicate status updating for `FlinkDeployment`. To achieve this, we compare the object with the original copy before reconcile. If status changed, the `UpdateControl` will apply, otherwise, It will produce no update.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/exception/DeploymentFailedException.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/exception/DeploymentFailedException.java
@@ -26,6 +26,7 @@ import io.fabric8.kubernetes.api.model.apps.DeploymentCondition;
 /** Exception to signal terminal deployment failure. */
 public class DeploymentFailedException extends RuntimeException {
     public static final String COMPONENT_JOBMANAGER = "JobManagerDeployment";
+    private static final long serialVersionUID = -1070179896083579221L;
 
     public final String component;
     public final DeploymentCondition deployCondition;
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobManagerDeploymentStatus.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobManagerDeploymentStatus.java
@@ -19,6 +19,7 @@ package org.apache.flink.kubernetes.operator.observer;
 
 import org.apache.flink.kubernetes.operator.config.FlinkOperatorConfiguration;
 import org.apache.flink.kubernetes.operator.crd.FlinkDeployment;
+import org.apache.flink.kubernetes.operator.reconciler.ReconciliationUtils;
 import org.apache.flink.kubernetes.operator.utils.SavepointUtils;
 
 import io.javaoperatorsdk.operator.api.reconciler.UpdateControl;
@@ -45,7 +46,9 @@ public enum JobManagerDeploymentStatus {
     ERROR;
 
     public UpdateControl<FlinkDeployment> toUpdateControl(
-            FlinkDeployment flinkDeployment, FlinkOperatorConfiguration operatorConfiguration) {
+            FlinkDeployment originalCopy,
+            FlinkDeployment flinkDeployment,
+            FlinkOperatorConfiguration operatorConfiguration) {
         int rescheduleAfterSec;
         switch (this) {
             case DEPLOYING:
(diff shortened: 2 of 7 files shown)


Aitozi:
cc @wangyang0918 @gyfora

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/exception/DeploymentFailedException.java
```diff
@@ -26,6 +26,7 @@
 /** Exception to signal terminal deployment failure. */
 public class DeploymentFailedException extends RuntimeException {
     public static final String COMPONENT_JOBMANAGER = "JobManagerDeployment";
+    private static final long serialVersionUID = -1070179896083579221L;
```
gyfora, line 29: Why did you add the serialVersionUIDs?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobManagerDeploymentStatus.java
```diff
@@ -45,7 +46,9 @@
     ERROR;
 
     public UpdateControl<FlinkDeployment> toUpdateControl(
-            FlinkDeployment flinkDeployment, FlinkOperatorConfiguration operatorConfiguration) {
+            FlinkDeployment originalCopy,
```
gyfora, line 49: Maybe instead of adding `originalCopy` here we could simply change this method to return the reschedule delay only. I think that would fit better in the current flow. Then we can use the `ReconciliationUtils.toUpdateControl` in the controller

What do you think?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
```diff
@@ -74,4 +77,22 @@ public static void updateForReconciliationError(FlinkDeployment flinkApp, String
             throw new IllegalStateException(e);
         }
     }
+
+    public static UpdateControl<FlinkDeployment> toUpdateControl(
+            FlinkDeployment originalCopy, FlinkDeployment current) {
+        UpdateControl<FlinkDeployment> updateControl;
+        if (!Objects.equals(originalCopy.getSpec(), current.getSpec())) {
+            throw new UnsupportedOperationException(
+                    "The spec changed during reconcile is not supported.");
```
gyfora, line 86: I think this is a valid safeguard but I think the error should be something like:
```
Detected spec change after reconcile, this probably indicates a bug.
```

gyfora (changes requested):
Added some minor comments :)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/exception/DeploymentFailedException.java
Aitozi: The IDE complains this

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobManagerDeploymentStatus.java
Aitozi: Agree, I will improve it

Aitozi:
Updated according to your comments, PTAL @gyfora

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobManagerDeploymentStatus.java
```diff
@@ -44,7 +41,7 @@
     /** Deployment in terminal error, requires spec change for reconciliation to continue. */
     ERROR;
 
-    public UpdateControl<FlinkDeployment> toUpdateControl(
+    public Time rescheduleAfter(
```
gyfora, line 44: I think we should use the built in java Duration class instead of the Flink Time

gyfora:
Added one last minor comment :)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobManagerDeploymentStatus.java
Aitozi: Fixed

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/exception/DeploymentFailedException.java
wangyang0918: nit: we could always set the `serialVersionUIDs` to `1L` when introduce a serializable class. BTW, do we really need to serialize `DeploymentFailedException`?

https://flink.apache.org/contributing/code-style-and-quality-java.html

Aitozi: All the `Exception` class are `Serializable`, So we should add `serialVersionUIDs` for them. I will keep in mind to set the `serialVersionUIDs` to `1L` when introduce serializable class next time.

PR #59, opened 2022-03-14
[FLINK-26620] Mark CRD classes experimental

SteNicholas:
We should mark the CRD classes experimental for the first preview release. This is the only API of the operator that we plan to release.

**The brief change log**

- Use `@Experimental` annotation to mark the classes of the `org.apache.flink.kubernetes.operator.crd` package experimental.


gyfora:
Should we also mark UpgradeMode and JobState?

SteNicholas:
@gyfora, IMO, the enums don't need to add the `@Experimental` annotation.

gyfora:
I think we need to annotate the enums also, they affect the schema of the generated crd and what can or cannot be submitted. So 100% part of the api :)

PR #60, opened 2022-03-14
[FLINK-26552] Try to use @EnableKubernetesMockClient(crud = true) in controller test

SteNicholas:
The controller test currently uses the `KubernetesMockserver` directly. We could try to use `@EnableKubernetesMockClient(crud = true)` in the `FlinkDeploymentControllerTest` test.

**The brief change log**

- Adds the `@EnableKubernetesMockClient(crud = true)` annotation to `FlinkDeploymentControllerTest` and removes the `KubernetesMockserver` and `KubernetesClient` start and stop.


gyfora:
I think with this change you could also remove the mockserver related logic from the beginning of the `testUpgradeNotReadyCluster` method.

gyfora (approved):

PR #62, opened 2022-03-15
[FLINK-26605] Check if JM can serve rest api calls every time before reconcile

bgeng777:
- Add `isJobManagerServing()` method to use the rest call(i.e. "/jobs/overview") to check if JM can actually serve requests after JM is deployed.
- Add the previous check in the `BaseObserver` logic to be executed every time before reconcile when JM state is `READY`.
- Enhance `SessionObserverTest` and `FlinkDeploymentControllerTest`  to include the case when JM port is ready but the rest service is not available.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,7 +52,7 @@ jobs:
           export SHELL=/bin/bash
           export DOCKER_BUILDKIT=1
           eval $(minikube -p minikube docker-env)
-          docker build -f ./Dockerfile -t flink-kubernetes-operator:ci-latest .
+          docker build --progress=plain --no-cache -f ./Dockerfile -t flink-kubernetes-operator:ci-latest .
           docker images
       - name: Start the operator
         run: |
@@ -107,7 +107,7 @@ jobs:
           export SHELL=/bin/bash
           export DOCKER_BUILDKIT=1
           eval $(minikube -p minikube docker-env)
-          docker build -f ./Dockerfile -t flink-kubernetes-operator:ci-latest .
+          docker build --progress=plain --no-cache -f ./Dockerfile -t flink-kubernetes-operator:ci-latest --progress plain .
           docker images
       - name: Start the operator
         run: |
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkOperatorConfiguration.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkOperatorConfiguration.java
@@ -35,6 +35,7 @@ public class FlinkOperatorConfiguration {
     Duration progressCheckInterval;
     Duration restApiReadyDelay;
     Duration savepointTriggerGracePeriod;
+    Duration flinkClientTimeout;
     String flinkServiceHostOverride;
     Set<String> watchedNamespaces;
 
@@ -52,6 +53,9 @@ public class FlinkOperatorConfiguration {
                 operatorConfig.get(
                         OperatorConfigOptions.OPERATOR_OBSERVER_SAVEPOINT_TRIGGER_GRACE_PERIOD);
 
+        Duration flinkClientTimeout =
+                operatorConfig.get(OperatorConfigOptions.OPERATOR_OBSERVER_FLINK_CLIENT_TIMEOUT);
+
         String flinkServiceHostOverride = null;
         if (EnvUtils.get("KUBERNETES_SERVICE_HOST") == null) {
             // not running in k8s, simplify local development
@@ -65,6 +69,7 @@ public class FlinkOperatorConfiguration {
                 progressCheckInterval,
                 restApiReadyDelay,
                 savepointTriggerGracePeriod,
+                flinkClientTimeout,
                 flinkServiceHostOverride,
                 watchedNamespaces);
     }
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/SessionObserver.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/SessionObserver.java
@@ -25,6 +25,8 @@ import org.apache.flink.kubernetes.operator.service.FlinkService;
 
 import io.javaoperatorsdk.operator.api.reconciler.Context;
 
+import java.util.concurrent.TimeoutException;
+
 /** The observer of the {@link org.apache.flink.kubernetes.operator.config.Mode#SESSION} cluster. */
 public class SessionObserver extends BaseObserver {
 
@@ -35,6 +37,19 @@ public class SessionObserver extends BaseObserver {
 
     @Override
     public void observe(FlinkDeployment flinkApp, Context context, Configuration effectiveConfig) {
-        observeJmDeployment(flinkApp, context, effectiveConfig);
+        if (isClusterReady(flinkApp)) {
+            // Check if session cluster can serve rest calls following our practice in JobObserver
+            try {
+                flinkService.listJobs(effectiveConfig);
+            } catch (Exception e) {
+                logger.error("REST service in session cluster is bad now", e);
+                if (e instanceof TimeoutException) {
+                    // check for problems with the underlying deployment
+                    observeJmDeployment(flinkApp, context, effectiveConfig);
+                }
+            }
+        } else {
+            observeJmDeployment(flinkApp, context, effectiveConfig);
+        }
     }
 }
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
@@ -138,7 +138,11 @@ public class FlinkService {
 
     public Collection<JobStatusMessage> listJobs(Configuration conf) throws Exception {
         try (ClusterClient<String> clusterClient = getClusterClient(conf)) {
-            return clusterClient.listJobs().get(10, TimeUnit.SECONDS);
+            return clusterClient
+                    .listJobs()
+                    .get(
+                            operatorConfiguration.getFlinkClientTimeout().getSeconds(),
+                            TimeUnit.SECONDS);
         }
     }
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,7 +35,7 @@ COPY $WEBHOOK_DIR/src ./$WEBHOOK_DIR/src
 
 COPY tools ./tools
 
-RUN --mount=type=cache,target=/root/.m2 mvn clean install
+RUN --mount=type=cache,target=/root/.m2 mvn -ntp clean install
 
 # stage
 FROM openjdk:11-jre
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/OperatorConfigOptions.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/OperatorConfigOptions.java
@@ -53,4 +53,11 @@ public class OperatorConfigOptions {
                     .defaultValue(Duration.ofSeconds(10))
                     .withDescription(
                             "The interval before a savepoint trigger attempt is marked as unsuccessful");
+
+    public static final ConfigOption<Duration> OPERATOR_OBSERVER_FLINK_CLIENT_TIMEOUT =
+            ConfigOptions.key("operator.observer.flink.client.timeout.sec")
+                    .durationType()
+                    .defaultValue(Duration.ofSeconds(10))
+                    .withDescription(
+                            "The timeout for the observer to wait the flink rest client to return.");
 }
--- a/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/TestingFlinkService.java
+++ b/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/TestingFlinkService.java
@@ -37,6 +37,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.TimeoutException;
 import java.util.stream.Collectors;
 
 /** Flink service mock for tests. */
@@ -79,8 +80,11 @@ public class TestingFlinkService extends FlinkService {
 
     @Override
     public List<JobStatusMessage> listJobs(Configuration conf) throws Exception {
-        if (jobs.isEmpty()) {
+        if (jobs.isEmpty() && !sessions.isEmpty()) {
             throw new Exception("Trying to list a job without submitting it");
+        }
+        if (!isPortReady) {
+            throw new TimeoutException("JM port is unavailable");
         }
         return jobs.stream().map(t -> t.f1).collect(Collectors.toList());
     }
--- a/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentControllerTest.java
+++ b/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentControllerTest.java
@@ -78,6 +78,7 @@ public class FlinkDeploymentControllerTest {
                     Duration.ofSeconds(2),
                     Duration.ofSeconds(3),
                     Duration.ofSeconds(4),
+                    Duration.ofSeconds(5),
                     null,
                     Collections.emptySet());
 
--- a/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/observer/JobObserverTest.java
+++ b/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/observer/JobObserverTest.java
@@ -99,13 +99,28 @@ public class JobObserverTest {
                 deployment.getStatus().getJobStatus().getJobName());
         assertTrue(
                 Long.valueOf(deployment.getStatus().getJobStatus().getUpdateTime())
-                                .compareTo(
-                                        Long.valueOf(
-                                                deployment
-                                                        .getStatus()
-                                                        .getJobStatus()
-                                                        .getStartTime()))
+                        .compareTo(
+                                Long.valueOf(
+                                        deployment
+                                                .getStatus()
+                                                .getJobStatus()
+                                                .getStartTime()))
                         >= 0);
+        // Test job manager is unavailable suddenly
+        flinkService.setPortReady(false);
+        observer.observe(deployment, readyContext, conf);
+        assertEquals(
+                JobManagerDeploymentStatus.DEPLOYING,
+                deployment.getStatus().getJobManagerDeploymentStatus());
+        flinkService.setPortReady(true);
+        observer.observe(deployment, readyContext, conf);
+        assertEquals(
+                JobManagerDeploymentStatus.DEPLOYED_NOT_READY,
+                deployment.getStatus().getJobManagerDeploymentStatus());
+        observer.observe(deployment, readyContext, conf);
+        assertEquals(
+                JobManagerDeploymentStatus.READY,
+                deployment.getStatus().getJobManagerDeploymentStatus());
 
         // Test listing failure
         flinkService.clear();
--- a/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/observer/JobObserverTest.java
+++ b/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/observer/JobObserverTest.java
@@ -99,12 +99,12 @@ public class JobObserverTest {
                 deployment.getStatus().getJobStatus().getJobName());
         assertTrue(
                 Long.valueOf(deployment.getStatus().getJobStatus().getUpdateTime())
-                        .compareTo(
-                                Long.valueOf(
-                                        deployment
-                                                .getStatus()
-                                                .getJobStatus()
-                                                .getStartTime()))
+                                .compareTo(
+                                        Long.valueOf(
+                                                deployment
+                                                        .getStatus()
+                                                        .getJobStatus()
+                                                        .getStartTime()))
                         >= 0);
         // Test job manager is unavailable suddenly
         flinkService.setPortReady(false);
@@ -112,6 +112,7 @@ public class JobObserverTest {
         assertEquals(
                 JobManagerDeploymentStatus.DEPLOYING,
                 deployment.getStatus().getJobManagerDeploymentStatus());
+        // Job manager recovers
         flinkService.setPortReady(true);
         observer.observe(deployment, readyContext, conf);
         assertEquals(
--- a/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/observer/SessionObserverTest.java
+++ b/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/observer/SessionObserverTest.java
@@ -48,40 +48,50 @@ public class SessionObserverTest {
 
     @Test
     public void observeSessionCluster() {
-        FlinkService flinkService = new TestingFlinkService();
+        TestingFlinkService flinkService = new TestingFlinkService();
         SessionObserver observer =
                 new SessionObserver(
                         flinkService,
                         FlinkOperatorConfiguration.fromConfiguration(new Configuration()));
         FlinkDeployment deployment = TestUtils.buildSessionCluster();
+        Configuration conf = FlinkUtils.getEffectiveConfig(deployment, new Configuration());
         deployment
                 .getStatus()
                 .getReconciliationStatus()
                 .setLastReconciledSpec(deployment.getSpec());
 
-        observer.observe(
-                deployment,
-                readyContext,
-                FlinkUtils.getEffectiveConfig(deployment, new Configuration()));
+        observer.observe(deployment, readyContext, conf);
 
         assertEquals(
                 JobManagerDeploymentStatus.DEPLOYED_NOT_READY,
                 deployment.getStatus().getJobManagerDeploymentStatus());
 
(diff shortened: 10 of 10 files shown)


bgeng777:
cc @gyfora @wangyang0918

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
```diff
@@ -129,6 +132,26 @@ public boolean isJobManagerPortReady(Configuration config) {
...
+                    new URL(
+                            clusterClient.getWebInterfaceURL()
+                                    + JobsOverviewHeaders.getInstance().getTargetRestEndpointURL());
+            // TODO: add support for https if necessary
+            HttpURLConnection connection = (HttpURLConnection) url.openConnection();
+            connection.setRequestMethod("GET");
+            connection.connect();
+            if (connection.getResponseCode() == HttpURLConnection.HTTP_OK) {
+                return true;
+            }
```
gyfora, line 148: Why did you chose to call the rest api this way instead of through the clusterclient directly?

bgeng777: Thanks for this catch. I neglect that the `clusterClient.listJobs()` will actually send reqesut the same with this raw http blocking call. Have updated to use the clusterClient.

wangyang0918:
I think we do not need `OPERATOR_OBSERVER_REST_READY_DELAY_IN_SEC` now. Right?

bgeng777:
> I think we do not need `OPERATOR_OBSERVER_REST_READY_DELAY_IN_SEC` now. Right?

That's right. IIUC, the reason behind this pr is to ensure the JM can serve rest call before taking other actions. From this point,  `OPERATOR_OBSERVER_REST_READY_DELAY_IN_SEC` is redundant as we will do the check every time before the reconcile. But currently, `OPERATOR_OBSERVER_REST_READY_DELAY_IN_SEC` is used to set `rescheduleAfterSecfor` when the JM status is `DEPLOYED_NOT_READY`. If we remove `OPERATOR_OBSERVER_REST_READY_DELAY_IN_SEC`, do you mean we set a fix value(e.g. 10s) for reschedule interval?

BTW, another thought during the impl: it seems that the 2 states(`JM is launched but port is not ready` and `JM is launched, port is ready but rest service is not ready`) can be merged as what we really care is if the JM can server instead of if the JM port is ready.

gyfora:
@bgeng777 I think if we remove OPERATOR_OBSERVER_REST_READY_DELAY_IN_SEC we can simply use the `progressCheckIntervalSeconds` which would be the same interval as for savepoints/deploying state.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkOperatorConfiguration.java
```diff
@@ -38,7 +38,7 @@ public static FlinkOperatorConfiguration fromConfiguration(Configuration operato
 
         int restApiReadyDelaySeconds =
                 operatorConfig.getInteger(
-                        OperatorConfigOptions.OPERATOR_OBSERVER_REST_READY_DELAY_IN_SEC);
+                        OperatorConfigOptions.OPERATOR_OBSERVER_PROGRESS_CHECK_INTERVAL_IN_SEC);
```
gyfora, line 41: we can simply use `progressCheckIntervalSeconds` and remove `restApiReadyDelaySeconds` from here

gyfora (commented):
Let's give @tweise a chance to review this also, he has been working on a related PR: https://github.com/apache/flink-kubernetes-operator/pull/56

We should make sure these 2 work well together.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkOperatorConfiguration.java
bgeng777: I totally agree with your point. I believe `restApiReadyDelaySeconds` field is not necessary any more. I am currently working on fixing the CI issue and will take your advice asap.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
```diff
@@ -129,6 +129,16 @@ public boolean isJobManagerPortReady(Configuration config) {
         return false;
     }
 
+    public boolean isJobManagerServing(Configuration config) {
+        try (ClusterClient<String> clusterClient = getClusterClient(config)) {
+            clusterClient.listJobs().get(10, TimeUnit.SECONDS);
+            return true;
+        } catch (Exception ignored) {
+        }
```
SteNicholas, line 137: Does this need to log the exception for listing the jobs of the cluster client?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
```diff
@@ -129,6 +129,16 @@ public boolean isJobManagerPortReady(Configuration config) {
         return false;
     }
 
+    public boolean isJobManagerServing(Configuration config) {
+        try (ClusterClient<String> clusterClient = getClusterClient(config)) {
+            clusterClient.listJobs().get(10, TimeUnit.SECONDS);
+            return true;
```
SteNicholas, line 135: Does this have another way to check the JobManager is serving? The exception way doesn't look good to me.

bgeng777: I do agree the try catch solution is not elegant but this pr aims at verifying the REST service of JM.
For network issue, to my best knowledge, try catch is the most widely used choice. Do you have any suggestion?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
bgeng777: The `observeJmDeployment()` method will output log WARN msg when the rest service is not working. 
Adding it here seems to be redundant?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
SteNicholas: @bgeng777, what about using the `clusterClient#getWebInterfaceURL` to check the serving?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkOperatorConfiguration.java
bgeng777: `restApiReadyDelaySeconds` removed.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
bgeng777: `getWebInterfaceURL` is neat but one drawback of `clusterClient#getWebInterfaceURL` is that it only returns the current web url(i.e. address of the leader of jm) but does not guarantee the JM can actually work. In `isJobManagerPortReady`, we currently have already used `getWebInterfaceURL` and will 'telnet' the port in the url. In the jira, we think it may not be enough and that's why I finally choose the `listJob` call.

wangyang0918: The `getWebInterfaceURL` might not work in the current situation because we are using `StandaloneClientHAServices` when creating the rest cluster client.

.github/workflows/ci.yml
```diff
@@ -52,7 +52,7 @@ jobs:
           export SHELL=/bin/bash
           export DOCKER_BUILDKIT=1
           eval $(minikube -p minikube docker-env)
-          docker build -f ./Dockerfile -t flink-kubernetes-operator:ci-latest .
+          docker build --progress=plain --no-cache -f ./Dockerfile -t flink-kubernetes-operator:ci-latest .
```
tweise, line 55: Are these changes unrelated?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
tweise: Please move the timeout into the config, the previous delay was also configurable. Did you check how the timeout behaves? If REST API is not available when the call is made and then becomes available after 5s, will the list call still fail after waiting for 10s? I believe that's what I had seen when testing this.

bgeng777: hi @tweise, I check the flink code and find the underlying impl of `listJob` is [retriable](https://github.com/apache/flink/blob/46bb03848ea67fa7b8952f757d57a2cda6ab16aa/flink-clients/src/main/java/org/apache/flink/client/program/rest/RestClusterClient.java#L833). The retry delay is controlled by the flink config `rest.retry.delay` (default val is 3s). For your case, if the initial call falis, it should retry each 3s until achieving the max retry attempts(default is 20). So do you mean you meet unexpected failure of this list call? Any steps that I can follow to repeat it? 

For the configurable timeout, in fact, I reuse the `listJobs()` in our `FlinkService`. I want to know if you are meaning that we should make both of them wait a configurable timeout. thanks.

.github/workflows/ci.yml
bgeng777: You are right, they are unrelated. You can check the commit history. It is a hotfix used for helping debug the github CI problem. Yang and I think it may be useful for our further development so I leave it here. Do you have any suggestion?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
bgeng777: Hi @tweise , I tried following steps to test the actual behavior of `clusterClient.listJobs()`:
1. Apply the `basic-session-example` yaml and wait for everything to be fine
2. Run ` kubectl scale deployment basic-session-example --replicas=0` so JM port and rest call are both unavailable
3. Rescale the replicas to 1 quickly when the `isJobManagerServing` is waiting.
I keep monitoring the operator log (I added some local dummy debug info)
image

From the log, I find that the rest call is sent at 13:56:05(In the k8s log, it is 05:56:05 due to time lag) and keeps waiting. Then when JM is started at 13:56:07 and everything becomes ok again around 13:56:13, the rest call succeeds. So I believe the answer to "will the list call still fail after waiting for 10s" is that it will not fail due to the retry.

gyfora:
Looks good, I think if you rebase it is good to go :)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/BaseObserver.java
```diff
@@ -62,8 +62,22 @@ protected void observeJmDeployment(
...
                 deploymentStatus.getJobManagerDeploymentStatus();
 
+        if (JobManagerDeploymentStatus.READY == previousJmStatus) {
+            if (!flinkService.isJobManagerServing(effectiveConfig)) {
+                deploymentStatus.setJobManagerDeploymentStatus(
+                        JobManagerDeploymentStatus.DEPLOYED_NOT_READY);
+                logger.warn("The job manager is currently not ready for serving.");
+            }
+            return;
+        }
```
gyfora, line 72: This is not really your fault be now the JobObserver won't call observe if the job is already in a ready state (while the sessionobserver always does). I think we need to remove the if logic in the jobObserver and just have your change

gyfora (changes requested):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/BaseObserver.java
gyfora: you can probably reproduce the problem by adding a test case to the JobObserverTest

bgeng777: Thanks for pointing this out. I remove the `observeJmDeployment` call in `observeFlinkJobStatus`. Let me know if that is our expected change.

gyfora: don't remove that :) I am talking about:
https://github.com/apache/flink-kubernetes-operator/blob/main/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobObserver.java#L50

and please add a testcase similar to what you added for the session observer so we can guard against future regressions here

bgeng777: Got it. Will add it now

wangyang0918: +1 for adding a test to cover this.

bgeng777: Add tests for `observeApplicationCluster()` to cover the case that JM is ready and later becomes unavailable and finally recovers. Let me know if it is enough.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/BaseObserver.java
```diff
@@ -62,8 +62,22 @@ protected void observeJmDeployment(
...
+
         if (JobManagerDeploymentStatus.DEPLOYED_NOT_READY == previousJmStatus) {
-            deploymentStatus.setJobManagerDeploymentStatus(JobManagerDeploymentStatus.READY);
+            // check if the JM is ready for accepting job submission
+            if (flinkService.isJobManagerServing(effectiveConfig)) {
+                deploymentStatus.setJobManagerDeploymentStatus(JobManagerDeploymentStatus.READY);
+            } else {
+                logger.info("The job manager has not been ready for serving yet...");
+            }
```
gyfora, line 80: Thank you @bgeng777 for adding the extra tests. Before merging this I would like to clear up the check flow a little because what you and what @tweise has done conflicts a little.

I think we should not `return` from the observe method if `flinkService.isJobManagerServing(effectiveConfig)` is false. In those cases we should always check the deployment too which happen later down in this method.

This is the whole idea with @tweise 's recent change here: https://github.com/apache/flink-kubernetes-operator/blob/main/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobObserver.java#L74

With your change @bgeng777 we can probably get rid of the observe call there but we still need to validate the deployment if the lest api does not respond.

gyfora (changes requested):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/BaseObserver.java
gyfora: Also we should use the same **configurable** timeout in both `flinkservice.listJobs` and `flinkservice. isJobManagerServing ` (probably the isJobManagerServing could call listJobs internally) because if the hardcoded 10 seconds fail for any reason the whole logic breaks

gyfora: I think we are pretty close to getting this right, let's cover these few corner cases and we are going to be in a very good shape :)

bgeng777: Thanks a lot for the detailed explanation! @gyfora 
1. After checking [FLINK-26473](https://issues.apache.org/jira/browse/FLINK-26473) and again, I agree with cancelling the `return` to make it fall back to the deployment. When the rest service is unavailabe, it is reasonable to check the deployment to find if anything bad happens. I will update the logic here.

2. Adding configurable timeout as @tweise and you suggest makes sense to me. As we achieve consensus on setting the same config for both methods, reusing the same logic is absolutely better.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
```diff
@@ -136,9 +136,20 @@ public boolean isJobManagerPortReady(Configuration config) {
...
+            return false;
+        }
+    }
+
     public Collection<JobStatusMessage> listJobs(Configuration conf) throws Exception {
         try (ClusterClient<String> clusterClient = getClusterClient(conf)) {
-            return clusterClient.listJobs().get(10, TimeUnit.SECONDS);
+            return clusterClient
+                    .listJobs()
+                    .get(operatorConfiguration.getProgressCheckIntervalSeconds(), TimeUnit.SECONDS);
```
gyfora, line 152: I think we need a new config in the OperatorConfiguraiton for this. We could call it `flinkClientTimeoutSeconds` -> `flink.client.timeout.seconds`

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
tweise: I don't think this is correct and we need to add test coverage. We need to cover this: https://github.com/apache/flink-kubernetes-operator/pull/56/files#diff-1d1ff01af8cbc1561adcc4e62e4f8f4280c7aa8c7b3f0624eb5f27f09aa9e96fR73

tweise: It might be better to leave this logic in the observer.

gyfora: @tweise I dont really understand your comment, it seems to refer to changes you introduced in your PR.

gyfora: It would be very nice of @bgeng777 to add test coverage to guard the current behaviour but it's not easy to understand all these corner cases that should have been covered already by tests. Let's try to test and review this carefully and we should then take the time to add the missing test coverage for the older features.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/BaseObserver.java
```diff
@@ -62,9 +62,22 @@ protected void observeJmDeployment(
         JobManagerDeploymentStatus previousJmStatus =
                 deploymentStatus.getJobManagerDeploymentStatus();
 
+        if (JobManagerDeploymentStatus.READY == previousJmStatus) {
+            if (flinkService.isJobManagerServing(effectiveConfig)) {
```
tweise, line 66: Since we are already listing jobs when the deployment is ready, do we really need this? I would prefer we avoid redundant calls to the REST API. The previous delay was there to not run into listJobs error.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
tweise: I will try to cover the listJobs error scenario in a separate PR soon.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/BaseObserver.java
bgeng777: For JobObserver, I do agree, it is somehow redundant when deployment is ready. For Job, we first `observeJmDeployment`(1 call of `listJobs()`) and then `observeFlinkJobStatus()`(1 or 2 call of `listJobs()`). We could reduce at least 1 extra rest call if we refactor some codes.
But I think we may still need this check here for SessionObserver. For Session, we will only `observeJmDeployment`. Without this check, we will fall into the deployment check always and in current logic of `main` branch, we would change the JobManagerDeploymentStatus to `DEPLOYED_NOT_READY` as well which is not desired.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
bgeng777: updated.

wangyang0918:
@bgeng777 I think we might need to step back and figure out what we are trying to fix in this PR. After #56, the original problems described in the FLINK-26605 might not be true.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/BaseObserver.java
gyfora: What we could do @bgeng777 is to pass a flag to the JM observer whether to check Rest API for ready jobs. As you explained this should be false for the jobobserver as that checks the rest api anyways. I think for deployed_not_ready state we need to always check to have a good way of telling when it becomes ready.

gyfora: Or we could just keep the if check before observing in the job observer like it was before :)

gyfora:
I agree @wangyang0918 , the case of failing applications/jobs is well covered now with the recent changes. 

This PR with some modifications could improve the session case and the general logic of moving from deployed_not_ready to ready

bgeng777:
Hi @wangyang0918 @gyfora, thanks a lot for your reply.
It's my bad to not pay enough attention to @tweise's recent [PR](https://github.com/apache/flink-kubernetes-operator/pull/56) for [FLINK-26473](https://issues.apache.org/jira/browse/FLINK-26473) as these 2 PR were developed at the same time. As Yang said, the goal of [FLINK-26605](https://issues.apache.org/jira/browse/FLINK-26605) is to solve 2 problems:
1. Check if JobManager is really ready for serving when the state is READY.
2. Solve the case that JobManager crashed backoff.
I agree that #56 has already solved above problems.

This PR is trying to solve above 2 problems in a somehow different way and leads to confusion. As Gyula's said, this PR should swicth the target to improve the session case. In fact, the session observer will keep switching between READY and DEPLOYED_NOT_READY due to logic in main branch which should be a bug I believe.

So I would like to clarify the goal of this PR will include after above discussion:
1. Refine the control flow in the `observeJmDeployment` based on current main branch to improve logic of moving from DEPLOYED_NOT_READY to READY to fix the bug of session observer.
2. Rename `operator.observer.rest-ready.delay.sec` to `operator.observer.flink.client.timeout.sec` to control the waiting time of `listJobs`. IMO, it can be used as previous config for waiting the rest api to be fine as well.
3. Modify the Dockerfile to output more meaningful message in the CI.

What I would do to rollback extra changes:
1. Remove `isJobManagerServing` method. Instead, I will keep using `isJobManagerPortReady` as well to detect if JM is ready to work in Session case. The reason is that the timeout solution is expensive and IMO, as I said in [FLINK-26655](https://issues.apache.org/jira/browse/FLINK-26655), `isJobManagerServing` and `isJobManagerPortReady` should do the same thing(i.e. check if JM can work correctly) actually. If we find a good approach for checking the readiness of JM later, we can update it then.

I will finish above changes today. Please let me know if there are unwanted changes or I ignore some cases.

gyfora:
@bgeng777 I agree with a simplified approach here, one comment is to please check https://github.com/apache/flink-kubernetes-operator/pull/80/files and use a simialr approach for the new config (no need for `.sec` if we use a duration type :))

bgeng777:
sorry for bother you guys again: @gyfora @wangyang0918 @tweise 
Main changes are only in `SessionObserver`: if the cluster state is READY, then check the rest service via `listJobs` using the same idea with `JobObserver` to make sure it can serve correctly (not only port is ready). if it can not serve or state is not READY, go to check the JM deployment.

gyfora (approved):
Thanks @bgeng777 . I think this has been simplified a lot, and thanks for adding a test for the JobObserver to check if the cluster suddenly becomes unavailable. 

If you could add the same test to the sessionobservertest also it is +1 from me

bgeng777:
@gyfora Paying much attention for the bug fix before. Now same tests added for SessionObserver :)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/SessionObserver.java
```diff
@@ -35,6 +37,19 @@ public SessionObserver(
 
     @Override
     public void observe(FlinkDeployment flinkApp, Context context, Configuration effectiveConfig) {
```
wangyang0918, line 39: Sorry for the late response. It seems that the current implementation could be optimized by a similar logic with `JobObserver#observe()`. Because the `JobManagerDeploymentStatus` could be set to `READY` in `observeJmDeployment`.

```
if (!isClusterReady(flinkApp)) {
      observeJmDeployment(flinkApp, context, effectiveConfig);
}

if (isClusterReady(flinkApp)) {
        // Check if session cluster can serve rest calls following our practice in JobObserver
        try {
            flinkService.listJobs(effectiveConfig);
        } catch (Exception e) {
            logger.error("REST service in session cluster is bad now", e);
            if (e instanceof TimeoutException) {
                // check for problems with the underlying deployment
                observeJmDeployment(flinkApp, context, effectiveConfig);
            }
        }
}
```

PR #64, opened 2022-03-15
[FLINK-26646][hotfix] Make append explicit for config overrides

gyfora:


gyfora:
cc @wangyang0918

helm/flink-operator/values.yaml
```diff
@@ -58,7 +58,7 @@ webhook:
 operatorConfiguration:
   create: true
   # Set append to false to replace configuration files
-  # append: true
+  append: true
```
wangyang0918, line 61: So now the default behavior is append, not replace. Right?

gyfora: Before this change, append was the default even if not specified. That was the source of the helm template complications :D 

Now it is explicitly set to append=true in the default values.yaml  but if you omit it , it will be false.

gyfora: I think this is more predictable behaviour now

wangyang0918: Make sense.

wangyang0918 (approved):
+1

PR #65, opened 2022-03-15
[FLINK-26649] Add startTime in JobStatus

SteNicholas:
The current `JobStatus.updateTime` actually means the start time, not the last update time. We could introduce a new field `startTime` and make `updateTime` reflect its original intention.

**The brief change log**
- Adds the field `startTime` in `JobStatus` to represent the start time.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobObserver.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobObserver.java
@@ -87,8 +87,8 @@ public class JobObserver extends BaseObserver {
         status.setState(newJob.getJobState().name());
         status.setJobName(newJob.getJobName());
         status.setJobId(newJob.getJobId().toHexString());
-        // track the start time, changing timestamp would cause busy reconciliation
-        status.setUpdateTime(String.valueOf(newJob.getStartTime()));
+        status.setStartTime(String.valueOf(newJob.getStartTime()));
+        status.setUpdateTime(String.valueOf(System.currentTimeMillis()));
     }
 
     private void observeSavepointStatus(FlinkDeployment flinkApp, Configuration effectiveConfig) {
--- a/docs/content/docs/custom-resource/reference.md
+++ b/docs/content/docs/custom-resource/reference.md
@@ -135,7 +135,8 @@ under the License.
 | jobName | java.lang.String | Name of the job. |
 | jobId | java.lang.String | Flink JobId of the Job. |
 | state | java.lang.String | Last observed state of the job. |
-| updateTime | java.lang.String | Start time of the job. |
+| startTime | java.lang.String | Start time of the job. |
+| updateTime | java.lang.String | Update time of the job. |
 | savepointInfo | org.apache.flink.kubernetes.operator.crd.status.SavepointInfo | Information about pending and last savepoint for the job. |
 
 ### ReconciliationStatus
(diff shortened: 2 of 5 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobObserver.java
```diff
@@ -87,8 +87,8 @@ private void updateJobStatus(JobStatus status, List<JobStatusMessage> clusterJob
         status.setState(newJob.getJobState().name());
         status.setJobName(newJob.getJobName());
         status.setJobId(newJob.getJobId().toHexString());
-        // track the start time, changing timestamp would cause busy reconciliation
-        status.setUpdateTime(String.valueOf(newJob.getStartTime()));
+        status.setStartTime(String.valueOf(newJob.getStartTime()));
+        status.setUpdateTime(String.valueOf(System.currentTimeMillis()));
```
tweise, line 91: Have you confirmed that this does not cause a busy reconcile loop? I actually think we should just rename the existing field.

wangyang0918: The current javaoperatorsdk behavior is a little strange. When `generationAwareEventProcessing` is configured to `false`, not `true`, then the reconciliation will not be triggered if `.spec` is not changed.

[1]. https://javaoperatorsdk.io/docs/features

gyfora: @wangyang0918 exactly, we only want to trigger new reconcile automatically when the Spec itself is changed by the user. This does not affect reconciliations that we schedule ourselves but avoids immediate reconciliation triggers when the status changes.

gyfora: GenerationAware here means that it is aware that we should only reconcile if the generation (spec) changes. So it might be a bit confusing but it seems to work :)

wangyang0918: @gyfora So do you mean to need to set `generationAwareEventProcessing` to `false` to enable the filtering? I am afraid the default value is `true`.

gyfora: we want it to be aware of the generation so we want it to be `true` which is the default. True enables filtering, false triggers reconcile for every spec/status change

wangyang0918: The description in the java doc[1] is not consistent with the documentation.

[1]. https://github.com/java-operator-sdk/java-operator-sdk/blob/main/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/ControllerConfiguration.java#L27

image

gyfora: Yea it should say, `To turn off this feature...`

wangyang0918: I have to verify whether it really works.

wangyang0918: FYI: it is indeed a typo in the documentation. They have fixed this.

I have verified the status changes will not cause busy reconciliation.

wangyang0918 (approved):
LGTM.

@gyfora @tweise do you have other concerns?

gyfora (approved):

wangyang0918:
Merging this PR.

PR #67, opened 2022-03-15
[FLINK-26545] Ingress rules in managed namespace

morhidi:
- moved the ingress(es) into the managed namespaces
- the host names in the rules are created as` <clusterid>.<namespace>.<domain>` now


morhidi:
cc @wangyang0918 @gyfora

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/BaseReconciler.java
```diff
@@ -60,7 +59,11 @@ private DeleteControl shutdownAndDelete(
             FlinkUtils.deleteCluster(flinkApp, kubernetesClient, true);
         }
         IngressUtils.updateIngressRules(
-                flinkApp, effectiveConfig, operatorNamespace, kubernetesClient, true);
+                flinkApp,
+                effectiveConfig,
+                flinkApp.getMetadata().getNamespace(),
```
Aitozi, line 64: this do not have pass the `flinkApp.getMetadata().getNamespace()`, it can get from the `flinkApp`

Aitozi:
LGTM otherwise a minor comment

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/BaseReconciler.java
morhidi: yup, I missed that. Thanks!

helm/flink-operator/templates/ingress.yaml
```diff
@@ -17,6 +17,23 @@
 ################################################################################
 ---
 {{- if .Values.ingress.create }}
+{{- if .Values.watchNamespaces }}
```
wangyang0918, line 20: How could we create the ingress when `watchNamespaces` is not configured? By default, the operator is watching all the namespaces, I am afraid it still could not update the ingress rules when running in a different namespace with operator.

All in all, I am suggesting to create the ingress in the java code, not the helm chart.

morhidi: You're right the all namespace scenario is not covered properly. We discussed this yesterday with @gyfora and @mbalassi and haven't had the chance to update the PR yet.

My main motivation behind using the Helm chart for creating the Ingresses was to use a more predictable/static manner. To my best knowledge Ingresses require at least an LB address, and a wildcard DNS entry with the current implementation to work properly, sometimes firewall/NW changes. 

So my suggestion for the ingress logic is the following:

- pre-create Ingresses for all watchNamespaces when the Operator is started (Let's use Java then)
- create Ingresses on the fly whenever a CR is created in a new namespace
- Set the operator deployment as the owner of all Ingresses and never delete them to avoid dynamic LB generation.

Let me know what you think @wangyang0918

wangyang0918: Thanks for sharing the progress. For the owner reference, I am afraid we could not set the owner of all the ingress in different namespaces to the operator deployment. Refer to here[1] for more information.

I am just thinking in a different way. Could we only create one ingress for the operator and let the operator do the HTTP requests redirecting? Benefit from this, we could easily integrate a unified authentication if necessary. I believe hadoop YARN[2] is running in a similar way to proxy the application master UI. 

And maybe we could also consider to provide a webUI for the operator itself in the future, which could greatly lower the bar to use our k8s operator.


[1]. https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#OwnerReference
[2]. https://hadoop.apache.org/docs/stable/hadoop-yarn/hadoop-yarn-site/WebApplicationProxy.html#:~:text=The%20Web%20Application%20Proxy%20is,web%20based%20attacks%20through%20YARN.

morhidi: This should be doable with an ingress controller too, then. I'll dig into this.

morhidi: @wangyang0918 So it seems my assumption about the one LB address per Ingress is not correct:

> With the NGINX Ingress controller you can also have multiple ingress objects for multiple environments or namespaces with the same network load balancer; with the ALB, each ingress object requires a new load balancer.



https://aws.amazon.com/blogs/opensource/network-load-balancer-nginx-ingress-controller-eks/

It seems we can create one ingress per deployment, and it get's merged by nginx. I'll test this and see if it works. Having a single ingress per deplyment would solve the ownership issue as well.

wangyang0918: It makes sense to me if the implementation does not rely on a specific cloud provision.

wangyang0918: I will get to the PR soon and verify it on the ACK(alibaba container kubernetes).

wangyang0918 (approved):
LGTM.

I have verified on the ACK(alibaba container kubernetes). It also works as expected. All the ingresses could share a same LoadBalancer.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/IngressUtils.java
```diff
@@ -29,7 +33,8 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.Optional;
+import java.util.Collections;
+import java.util.List;
 
 /** Ingress utilities. */
 public class IngressUtils {
```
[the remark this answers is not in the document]
morhidi: Removed

morhidi:
> LGTM.
> 
> I have verified on the ACK(alibaba container kubernetes). It also works as expected. All the ingresses could share a same LoadBalancer.

This is good news! Thank you for your assistance @wangyang0918

morhidi:
For the record I filed a new Jira https://issues.apache.org/jira/browse/FLINK-26706 with some improvements on ingress front

PR #70, opened 2022-03-16
[FLINK-26660] Specify eventsource name when watching multiple namespaces

gyfora:


mbalassi (approved):
LGTM, thanks. Tested in the original `minikube` environment, where I have reported the issue and it solved the problem.

PR #74, opened 2022-03-16
[FLINK-26676] Make ClusterIP the default rest service type

gyfora:
This is intended to solve some problems encountered when watching specific / multiple namespaces.

cc @mbalassi @wangyang0918


wangyang0918 (approved):
Thanks for creating this PR.

+1 for merging.

PR #75, opened 2022-03-16
[FLINK-26640] Make FlinkVersion enum and required

gyfora:
This PR changes the FlinkVersion from string to enum.

I simply copied the FlinkVersion enum from flink master so that we can adapt this directly once we bump the flink dep to at least 1.15.

### DIFF
--- a/helm/flink-operator/crds/flinkdeployments.flink.apache.org-v1.yml
+++ b/helm/flink-operator/crds/flinkdeployments.flink.apache.org-v1.yml
@@ -26,6 +26,22 @@ spec:
               serviceAccount:
                 type: string
               flinkVersion:
+                enum:
+                - v1_3
+                - v1_4
+                - v1_5
+                - v1_6
+                - v1_7
+                - v1_8
+                - v1_9
+                - v1_10
+                - v1_11
+                - v1_12
+                - v1_13
+                - v1_14
+                - v1_15
+                - v1_16
+                - versionStr
                 type: string
               ingressDomain:
                 type: string
@@ -9114,6 +9130,22 @@ spec:
                       serviceAccount:
                         type: string
                       flinkVersion:
+                        enum:
+                        - v1_3
+                        - v1_4
+                        - v1_5
+                        - v1_6
+                        - v1_7
+                        - v1_8
+                        - v1_9
+                        - v1_10
+                        - v1_11
+                        - v1_12
+                        - v1_13
+                        - v1_14
+                        - v1_15
+                        - v1_16
+                        - versionStr
                         type: string
                       ingressDomain:
                         type: string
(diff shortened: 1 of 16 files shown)


gyfora:
cc @tweise

helm/flink-operator/crds/flinkdeployments.flink.apache.org-v1.yml
```diff
@@ -26,6 +26,22 @@ spec:
               serviceAccount:
                 type: string
               flinkVersion:
+                enum:
+                - v1_3
```
tweise, line 30: Do we really want those old (irrelevant) Flink versions here? It might be better to only list the versions that are actually supported. That probably means that we need to have our own enum. And that should be fine, since the validation can still check it is one of `FlinkVersion`. That would also avoid having to copy the entire `FlinkVersion` class, we can add that check when 1.15 becomes the new minimum version.

gyfora: I don't have a problem with having our own enum either (that way we would have immediate validation for the accepted versions). @wangyang0918 , @morhidi what do you think?

morhidi: makes perfect sense to me

gyfora:
Updated the PR @tweise please take a look

helm/flink-operator/crds/flinkdeployments.flink.apache.org-v1.yml
wangyang0918: Having our own enum of FlinkVersion also makes sense to me.

wangyang0918:
> I am wondering if we should require the user to explicitly set the flink version or provide a default value like I did now

It is not very easy to predict the flinkVersion if a user customized image is used.

wangyang0918 (approved):

gyfora:
> > I am wondering if we should require the user to explicitly set the flink version or provide a default value like I did now
> 
> It is not very easy to predict the flinkVersion if a user customized image is used.

@wangyang0918 are you suggesting that it should be explicitly required always in the yaml?

wangyang0918:
> are you suggesting that it should be explicitly required always in the yaml?

For now, I would say yes.

gyfora:
Sounds good, we can relax this later. I will update the PR (+examples, docs)

tweise:
@gyfora I just realized that my comment wasn't posted..  I'm actually not sure if we want to restrict the upper bound of the Flink version. I imagine that in the long run the underlying Flink dependencies will converge towards a good compatibility story and then only the minimum version matters. When no version is specified, we assume the minimum version. When a higher version is specified, features of that higher version can be used.

tweise:
I also think it would be nice to have a way to infer the version when the official docker image is used!

PR #77, opened 2022-03-17
[FLINK-26696] Reuse the method of judgment

zuston:
Minor optimization.


wangyang0918 (approved):
+1. Thanks for the fixing.

PR #78, opened 2022-03-17
[FLINK-26697] Detailed log with flink app name

zuston:


gyfora:
Hi @zuston! Thanks for the effort but this goes the complete opposite way of what we are trying to do here: https://issues.apache.org/jira/browse/FLINK-26454

The operator exposes all this info as part of the MDC so it is enough to configure the logging properties to see them. So we should remove the name, namespace etc from the actual logging and use this instead.

gyfora:
@SteNicholas hasn't started working on this so either you or I can pick that ticket up instead. I agree that we should do this ASAP because the logging is quite messing at the moment

zuston:
Sorry i missed this [ticket](https://issues.apache.org/jira/browse/FLINK-26454). I will close this PR and discuss with @SteNicholas offline. Maybe i can take over.

PR #80, opened 2022-03-18
[FLINK-26714] Remove .sec from OperatorConfigOptions and use Duration type config instead

SteNicholas:
The `OperatorConfigOptions` options mostly have .sec in the names to denote seconds intervals. The Duration type configs should be used so that users can specify them in a more user friendly way and remove .sec from the name.

**The brief change log**
- Modifies the type of configs to `Duration` and removes `.sec` from `OperatorConfigOptions`.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkOperatorConfiguration.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkOperatorConfiguration.java
@@ -24,36 +24,33 @@ import org.apache.flink.kubernetes.operator.utils.OperatorUtils;
 
 import lombok.Value;
 
+import java.time.Duration;
 import java.util.Set;
 
 /** Configuration class for operator. */
 @Value
 public class FlinkOperatorConfiguration {
 
-    int reconcileIntervalSeconds;
-    int progressCheckIntervalSeconds;
-    int restApiReadyDelaySeconds;
-    int savepointTriggerGracePeriodSeconds;
+    Duration reconcileInterval;
+    Duration progressCheckInterval;
+    Duration restApiReadyDelay;
+    Duration savepointTriggerGracePeriod;
     String flinkServiceHostOverride;
     Set<String> watchedNamespaces;
 
     public static FlinkOperatorConfiguration fromConfiguration(Configuration operatorConfig) {
-        int reconcileIntervalSeconds =
-                operatorConfig.getInteger(
-                        OperatorConfigOptions.OPERATOR_RECONCILER_RESCHEDULE_INTERVAL_IN_SEC);
+        Duration reconcileInterval =
+                operatorConfig.get(OperatorConfigOptions.OPERATOR_RECONCILER_RESCHEDULE_INTERVAL);
 
-        int restApiReadyDelaySeconds =
-                operatorConfig.getInteger(
-                        OperatorConfigOptions.OPERATOR_OBSERVER_REST_READY_DELAY_IN_SEC);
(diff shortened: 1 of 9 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkOperatorConfiguration.java
```diff
@@ -30,30 +30,33 @@
 @Value
 public class FlinkOperatorConfiguration {
 
-    int reconcileIntervalSeconds;
-    int progressCheckIntervalSeconds;
-    int restApiReadyDelaySeconds;
-    int savepointTriggerGracePeriodSeconds;
+    long reconcileIntervalSeconds;
```
Aitozi, line 33: We can keep the `Duration` here, let it to be converted when used.

Aitozi: By this we can also save the `seconds` in variable name.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkOperatorConfiguration.java
```diff
@@ -24,36 +24,33 @@
...
 
 /** Configuration class for operator. */
 @Value
 public class FlinkOperatorConfiguration {
 
-    int reconcileIntervalSeconds;
-    int progressCheckIntervalSeconds;
-    int restApiReadyDelaySeconds;
-    int savepointTriggerGracePeriodSeconds;
+    Duration reconcileIntervalSeconds;
```
mbalassi, line 34: If these are `Duration` types now please remove the `Seconds` suffix from their name too.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkOperatorConfiguration.java
```diff
@@ -24,36 +24,33 @@
...
+    Duration restApiReadyDelaySeconds;
+    Duration savepointTriggerGracePeriodSeconds;
     String flinkServiceHostOverride;
     Set<String> watchedNamespaces;
 
     public static FlinkOperatorConfiguration fromConfiguration(Configuration operatorConfig) {
-        int reconcileIntervalSeconds =
-                operatorConfig.getInteger(
-                        OperatorConfigOptions.OPERATOR_RECONCILER_RESCHEDULE_INTERVAL_IN_SEC);
+        Duration reconcileIntervalSeconds =
```
mbalassi, line 42: please remove the Seconds suffix

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobManagerDeploymentStatus.java
```diff
@@ -44,7 +44,7 @@
 
     public Duration rescheduleAfter(
             FlinkDeployment flinkDeployment, FlinkOperatorConfiguration operatorConfiguration) {
-        int rescheduleAfterSec;
+        Duration rescheduleAfterSec;
```
mbalassi, line 47: please remove the Seconds suffix

mbalassi (changes requested):
Hi @SteNicholas!

It makes sense to take this approach, please adjust the variable names too to reflect your change.

mbalassi (approved):
Thanks, LGTM. Let us make sure the CI runs successfuly.

PR #83, opened 2022-03-19
[FLINK-26707] Rework README

gyfora:


gyfora:
Is there anything else you think we should include here? cc @wangyang0918 @mbalassi @morhidi

gyfora:
Would be nice to have some minimal graphics here, something with Flink and Kubernetes logos 😄

gyfora:
I would like to create a flow chart like this here: https://golangexample.com/a-kubernetes-operator-that-manages-flink-applications-on-kubernetes/

I will work on that part next week and we can add it to the readme then

mbalassi (approved):
Thanks, @gyfora. LGTM, I agree that a simply flow diagram would be great here.
it might still make sense to have a dedicated page for the webhook that you delete here, but we can add back later.

docs/content/docs/operations/helm.md
```diff
@@ -25,3 +25,32 @@ under the License.
...
+```
+
+Alternatively to install the operator (and also the helm chart) to a specific namespace:
+
+```
+helm install flink-operator helm/flink-operator --namespace flink-operator --create-namespace
+```
+
+Note that in this case you will need to update the namespace in the examples accordingly or the `default`
+namespace to the [watched namespaces](#watching-only-specific-namespaces).
```
Aitozi, line 42: nit: we'd better also add an uninstall guide here

wangyang0918:
A straightforward diagram will help the user to know how is flink-kubernetes-operator working at first glance.

PR #85, opened 2022-03-19
[FLINK-26538] Add ability to restart flink deployment

Aitozi:
This PR is meant to add ability to restart deployment with one-shot command. 

**Brief change**

- Add `restartNonce` in the `FlinkDeployment#spec`
- JobReconciler will react to the `restartNonce` change by restarting the flink deployment.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/JobReconciler.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/JobReconciler.java
@@ -90,16 +90,14 @@ public class JobReconciler extends BaseReconciler {
             }
             JobState currentJobState = lastReconciledSpec.getJob().getState();
             JobState desiredJobState = jobSpec.getState();
-
             UpgradeMode upgradeMode = jobSpec.getUpgradeMode();
             JobState stateAfterReconcile = currentJobState;
             if (currentJobState == JobState.RUNNING) {
                 if (desiredJobState == JobState.RUNNING) {
-                    LOG.info("Upgrading running job, suspending first...");
+                    LOG.info("Upgrading/Restarting running job, suspending first...");
                 }
                 printCancelLogs(upgradeMode);
-                suspendJob(flinkApp, upgradeMode, effectiveConfig);
-                stateAfterReconcile = JobState.SUSPENDED;
+                stateAfterReconcile = suspendJob(flinkApp, upgradeMode, effectiveConfig);
             }
             if (currentJobState == JobState.SUSPENDED && desiredJobState == JobState.RUNNING) {
                 if (upgradeMode == UpgradeMode.STATELESS) {
@@ -202,17 +200,19 @@ public class JobReconciler extends BaseReconciler {
                 effectiveConfig);
     }
 
-    private void suspendJob(
+    private JobState suspendJob(
             FlinkDeployment flinkApp, UpgradeMode upgradeMode, Configuration effectiveConfig)
             throws Exception {
         final Optional<String> savepointOpt =
                 internalSuspendJob(flinkApp, upgradeMode, effectiveConfig);
 
         JobStatus jobStatus = flinkApp.getStatus().getJobStatus();
-        jobStatus.setState(JobState.SUSPENDED.name());
+        JobState stateAfterReconcile = JobState.SUSPENDED;
+        jobStatus.setState(stateAfterReconcile.name());
         flinkApp.getStatus().setJobManagerDeploymentStatus(JobManagerDeploymentStatus.MISSING);
         savepointOpt.ifPresent(
                 location -> jobStatus.getSavepointInfo().setLastSavepoint(Savepoint.of(location)));
+        return stateAfterReconcile;
     }
 
     @Override
--- a/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/reconciler/JobReconcilerTest.java
+++ b/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/reconciler/JobReconcilerTest.java
@@ -312,6 +312,54 @@ public class JobReconcilerTest {
         assertNull(flinkService.listJobs().get(0).f0);
     }
 
+    @Test
+    public void triggerRestart() throws Exception {
+        Context context = TestUtils.createContextWithReadyJobManagerDeployment();
+        TestingFlinkService flinkService = new TestingFlinkService();
+
+        JobReconciler reconciler = new JobReconciler(null, flinkService, operatorConfiguration);
+        FlinkDeployment deployment = TestUtils.buildApplicationCluster();
+        Configuration config = FlinkUtils.getEffectiveConfig(deployment, new Configuration());
+
+        reconciler.reconcile(deployment, context, config);
+        List<Tuple2<String, JobStatusMessage>> runningJobs = flinkService.listJobs();
+        verifyAndSetRunningJobsToStatus(deployment, runningJobs);
+
+        // Test restart job
+        FlinkDeployment restartJob = ReconciliationUtils.clone(deployment);
+        restartJob.getSpec().setRestartNonce(1L);
+        reconciler.reconcile(restartJob, context, config);
+        assertEquals(
+                JobState.SUSPENDED,
+                restartJob
+                        .getStatus()
+                        .getReconciliationStatus()
+                        .getLastReconciledSpec()
+                        .getJob()
+                        .getState());
+        runningJobs = flinkService.listJobs();
+        assertEquals(0, runningJobs.size());
+
+        reconciler.reconcile(restartJob, context, config);
+        assertEquals(
+                JobState.RUNNING,
+                restartJob
+                        .getStatus()
+                        .getReconciliationStatus()
+                        .getLastReconciledSpec()
+                        .getJob()
+                        .getState());
+        runningJobs = flinkService.listJobs();
+        assertEquals(1, runningJobs.size());
+        assertEquals(
+                1L,
+                restartJob
+                        .getStatus()
+                        .getReconciliationStatus()
+                        .getLastReconciledSpec()
+                        .getRestartNonce());
+    }
+
     private void verifyAndSetRunningJobsToStatus(
             FlinkDeployment deployment, List<Tuple2<String, JobStatusMessage>> runningJobs) {
         assertEquals(1, runningJobs.size());
--- a/docs/content/docs/custom-resource/reference.md
+++ b/docs/content/docs/custom-resource/reference.md
@@ -55,6 +55,7 @@ This page serves as a full reference for FlinkDeployment custom resource definit
 | jobManager | org.apache.flink.kubernetes.operator.crd.spec.JobManagerSpec | JobManager specs. |
 | taskManager | org.apache.flink.kubernetes.operator.crd.spec.TaskManagerSpec | TaskManager specs. |
 | job | org.apache.flink.kubernetes.operator.crd.spec.JobSpec | Job specification for application deployments. Null for session clusters. |
+| restartNonce | java.lang.Long | Nonce used to manually trigger restart for the cluster. In order to trigger restart, change  the number to anything other than the current value. |
 | logConfiguration | java.util.Map<java.lang.String,java.lang.String> | Log configuration overrides for the Flink deployment. Format logConfigFileName ->  configContent. |
 
 ### FlinkSessionJobSpec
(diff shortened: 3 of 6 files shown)


Aitozi:
cc @wangyang0918 @gyfora @tweise PTAL.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/reconciler/JobReconcilerTest.java
```diff
@@ -199,6 +200,38 @@ public void triggerSavepoint() throws Exception {
...
+
+        // Test restart job
+        FlinkDeployment restartJob = ReconciliationUtils.clone(deployment);
+        restartJob.getSpec().setRestartNonce(1L);
+        reconciler.reconcile(restartJob, context, config);
+
+        runningJobs = flinkService.listJobs();
+        assertEquals(1, runningJobs.size());
+        long newJobStartTime = runningJobs.get(0).f1.getStartTime();
+        Assertions.assertTrue(newJobStartTime > jobStartTime);
```
bgeng777, line 225: nit: it looks that in the file, we have already imported `org.junit.jupiter.api.Assertions.assertEquals` and `org.junit.jupiter.api.Assertions.assertTrue`. Maybe we can use these asserts without `Assertion.` prefix.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/JobReconciler.java
```diff
@@ -114,6 +119,14 @@ public void reconcile(FlinkDeployment flinkApp, Context context, Configuration e
         }
     }
 
+    private void restart(FlinkDeployment flinkApp, Configuration effectiveConfig) throws Exception {
+        LOG.info("Restart application cluster now.");
+        String jobIdString = flinkApp.getStatus().getJobStatus().getJobId();
+        JobID jobID = jobIdString != null ? JobID.fromHexString(jobIdString) : null;
+        flinkService.cancelJob(jobID, UpgradeMode.LAST_STATE, effectiveConfig);
```
bgeng777, line 126: IIUC, reason for using the `flinkService.cancelJob` with `LAST_STATE` mode is to delete the cluster directly?
But in `LAST_STATE` mode, we will not delete HA config. Is that expected?

Aitozi: Yes. IMO, the restart semantic will let the job keep running from the latest checkpoint. So the HA config is retained.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/reconciler/JobReconcilerTest.java
Aitozi: Good catch, removed.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/JobReconciler.java
bgeng777: I recheck the JIRA, the restart should be a shortcut of `suspend -> delete current deployment -> restart job from latest state in a new deployment", right? If that is the case, I think your code works fine.

Aitozi: Yes, It's exactly as you said

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
```diff
@@ -65,6 +65,22 @@ public static void updateForReconciliationError(FlinkDeployment flinkApp, String
         reconciliationStatus.setError(err);
     }
 
+    public static boolean triggerRestart(FlinkDeployment flinkApp) {
+        Long restartNonce = flinkApp.getSpec().getRestartNonce();
+        boolean everReconcileSucceed =
```
wangyang0918, line 70: IIRC, the `triggerRestart` is never called with `lastReconciledSpec` is null since `JobReconciler#reconcile()` will directly return in such case.

wangyang0918 (changes requested):
Thanks for creating this PR. I have two concerns about the current changes.
* What the expected behavior when `restartNonce` happens with other fields changes(e.g. `.job.upgradeMode`, `.image`)?
* If the HA is not enabled, we will lost state after `restartNonce`.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/JobReconciler.java
```diff
@@ -114,6 +119,14 @@ public void reconcile(FlinkDeployment flinkApp, Context context, Configuration e
         }
     }
 
+    private void restart(FlinkDeployment flinkApp, Configuration effectiveConfig) throws Exception {
+        LOG.info("Restart application cluster now.");
+        String jobIdString = flinkApp.getStatus().getJobStatus().getJobId();
+        JobID jobID = jobIdString != null ? JobID.fromHexString(jobIdString) : null;
```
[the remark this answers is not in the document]
Aitozi: not necessary, will replace it with `null`.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
Aitozi: oh, yes. Thanks for point out.

Aitozi:
@wangyang0918 Thanks for your review. I will think over your two concerns a bit more.

wangyang0918:
I agree that `restartNonce` should respect the `UpgradeMode` and try to reuse the current spec change behavior.

The HA is not always necessary for `restartNonce`. We have already done the sanity check in webhook when using `last-state`. However, we might still have state loss in some corner scenarios. This will be covered by FLINK-26577.

gyfora:
I would suggest we use mechanism that I implemented in https://github.com/apache/flink-kubernetes-operator/pull/103 to simply transition to SUSPENDED when restarting and then the next reconciliation loop will take care of the redployment

Aitozi:
> #103

OK, I will take a look.

Aitozi:
I'm working on this now.

Aitozi:
After this change, this PR only introduce the `restartNonce` spec field, when this changed, it reuse the logic of upgrade to restart the job. cc @wangyang0918 @gyfora PTAL again.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/JobReconciler.java
```diff
@@ -202,17 +200,19 @@ private void printCancelLogs(UpgradeMode upgradeMode) {
...
 
-    private void suspendJob(
+    private JobState suspendJob(
             FlinkDeployment flinkApp, UpgradeMode upgradeMode, Configuration effectiveConfig)
             throws Exception {
         final Optional<String> savepointOpt =
                 internalSuspendJob(flinkApp, upgradeMode, effectiveConfig);
 
         JobStatus jobStatus = flinkApp.getStatus().getJobStatus();
-        jobStatus.setState(JobState.SUSPENDED.name());
```
Aitozi, line 212: I wonder do we need to set the `JobStatus#state` to `SUSPEND` here? One is the observed state, one is the desired state. Maybe we could just clear the state and let the next reconcile to sync the state by observer?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultDeploymentValidator.java
```diff
@@ -245,6 +241,19 @@
...
+                && ReconciliationUtils.triggerRestart(deployment)) {
+            if (newJob.getState() != JobState.RUNNING) {
+                return Optional.of(
+                        String.format("Cannot restart job to %s state", newJob.getState()));
+            }
+            if (oldJob.getState() != JobState.RUNNING) {
+                return Optional.of(
+                        String.format("Cannot restart job in %s state", oldJob.getState()));
+            }
+        }
```
gyfora, line 255: I think we can remove these checks together with the `ReconciliationUtils.triggerRestart(deployment))` I dont think we are achieving anything and we actually restrict cases when the user want to restart a failing job

gyfora (changes requested):
I left a small comment to remove validation. We already have some necessary validation in place for upgrade operations that will also cover this case.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultDeploymentValidator.java
Aitozi: This validation is try to guard that job should target `RUNNING` state and also in `RUNNING` state when user try to use the restartNonce. The `RUNNING` state is the desire state of the job not the observed state here. Otherwise, what will we do if the job is in `SUSPENDED` state and change the `restartNonce` ?

gyfora: nothing will happen, but why would anything happen anyways :) without the validaiton the restartNonce will behave like any other spec change

gyfora: Instead of adding this specific validation for the `restartNonce` field we could discuss whether we should allow spec changes in suspended state. I think it is fine

Aitozi: > but why would anything happen anyways

I think the `restartNonce` have the semantic of bring the job to a `start/running` status somehow? If not go with this check it will work like ask job to restart once, but it still suspend :).

> we could discuss whether we should allow spec changes in suspended state

If JobState changes from suspended to suspended, with other spec changes, It will not take effect, I think it will not bring bad impact to allow the change. The problem of the `restartNonce` field change in suspended state is only the problem of the semantic, as I said first.

gyfora: I think the way you implemented it the restartNonce will not change the desired state. And I think it's good like this. This is just a field to trigger restart without having to change the actual spec.

Aitozi: About the removal of the validation, I want to keep it to Monday to wait for more inputs. Not insist on it, Just want to hear more opinions on it :)

gyfora: All I am trying to say, is that the way it is implemented, changing the `restartNonce` field works as any other spec change and it triggers a simple upgrade of the deplyoment. (this is perfectly good)

Given this, I don't see the point in introducing restrictions on this field as we also dont have restrictions on when an upgrade can be triggered. I just want to keep it conistent and simple.

We can of course wait for others to chime in here as well:
@wangyang0918 @morhidi @tweise

wangyang0918: I lean to make `restartNonce` working as a normal fields and do not have special validation.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/JobReconciler.java
wangyang0918: I prefer to update the status as early as possible.

wangyang0918: I prefer to update the status as early as possible since we are pretty sure the job has already been cancelled.

Aitozi: Get it

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultDeploymentValidator.java
Aitozi: @gyfora @wangyang0918 Thanks for your suggestion, I have removed the validation accordingly

PR #93, opened 2022-03-22
[FLINK-26706] Introduce Ingress URL templating

morhidi:
- users can customise the endpoint address of the Flink UI
- eliminates the need for a wildcard DNS entry `*.example.com`
- enables path and domain based routing with ingress

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/IngressUtils.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/IngressUtils.java
@@ -20,6 +20,8 @@ package org.apache.flink.kubernetes.operator.utils;
 import org.apache.flink.configuration.Configuration;
 import org.apache.flink.configuration.RestOptions;
 import org.apache.flink.kubernetes.operator.crd.FlinkDeployment;
+import org.apache.flink.kubernetes.operator.exception.ReconciliationException;
+import org.apache.flink.util.Preconditions;
 
 import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.fabric8.kubernetes.api.model.OwnerReference;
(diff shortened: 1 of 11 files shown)


morhidi:
cc @wangyang0918 @gyfora I'm sharing with you guys while working on the validation. Let me know what you think.

wangyang0918:
Great work. I will have a look on this PR today.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/IngressUtils.java
```diff
@@ -122,4 +154,27 @@ public static void setOwnerReference(HasMetadata owner, List<HasMetadata> resour
...
+        } catch (MalformedURLException e) {
+            LOG.error(e.getMessage());
+            throw new ReconciliationException(
+                    String.format(
+                            "Unable to process the Ingress template(%s). Error: %s",
+                            ingressTemplate, e.getMessage()));
+        }
+    }
+
+    private static String addProtocol(String url) {
```
wangyang0918, line 173: It seems we do not use the protocol when creating the ingress. Why do we need this?

wangyang0918 (commented):
This PR looks really good to me.

When I am testing this PR in my minikube, I find we must add a trailing slash in the URL. For example, `http://192.168.64.5/default/advanced-ingress/`. Otherwise, the Flink webUI could not show up. Could we find some solution for this?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/IngressUtils.java
morhidi: I used the `URL` class as a parser basically to get the domain/path parts from the template (instead of writing our own parser) A valid URL needs the protocol though, so it's a hack basically.

morhidi:
> This PR looks really good to me.
> 
> When I am testing this PR in my minikube, I find we must add a trailing slash in the URL. For example, `http://192.168.64.5/default/advanced-ingress/`. Otherwise, the Flink webUI could not show up. Could we find some solution for this?

I couldn't find a solution for this yet. Maybe we need another rule to handle the root context.

morhidi:
> > This PR looks really good to me.
> > When I am testing this PR in my minikube, I find we must add a trailing slash in the URL. For example, `http://192.168.64.5/default/advanced-ingress/`. Otherwise, the Flink webUI could not show up. Could we find some solution for this?
> 
> I couldn't find a solution for this yet. Maybe we need another rule to handle the root context.

See https://github.com/kubernetes/ingress-nginx/issues/4266 for the background. So something like this would solve the missing trailing slash. (I don't personally like the solution tho)
```
ingress:
    template: "/{{namespace}}/{{name}}(/|$)(.*)"
    className: "nginx"
    annotations:
      nginx.ingress.kubernetes.io/rewrite-target: "/$2"
      nginx.ingress.kubernetes.io/enable-rewrite-log: "true"
      nginx.ingress.kubernetes.io/configuration-snippet: |
        if ($uri = "/default/advanced-ingress") {rewrite .* $1/default/advanced-ingress/ permanent;}
```

wangyang0918 (approved):
+1 for merging.

robsonpeixoto:
This solution is amazing! Thanks a lot =D
But add this feature in annotation os very useful, for example:

```
ingress:
    template: "/{{namespace}}/{{name}}(/|$)(.*)"
    className: "nginx"
    annotations:
      nginx.ingress.kubernetes.io/rewrite-target: "/$2"
      nginx.ingress.kubernetes.io/enable-rewrite-log: "true"
      nginx.ingress.kubernetes.io/configuration-snippet: |
        if ($uri = "/{{namespace}}/{{name}}") {rewrite .* $1/{{namespace}}/{{name}}/ permanent;}
```

WDYT?

PR #96, opened 2022-03-22
[FLINK-26804] Improve e2e test status check stability

mbalassi:
I went with a simple extra wait-time approach as the tools already existed for this.

The meaningful alternative is to add expected whitelisted states, like `DEPLOYED_NOT_READY` instead of `READY` where we do wait and fail otherwise. I felt that this was unnecessarily complex. 

I will retrigger the e2e tests 3 times before merging this to establish confidence that it does solve the issue.


mbalassi:
I run the test altogether 5 times successfully, it seems stable now.

wangyang0918 (approved):
Thanks for improving the e2e test stability.

+1 for merging.

PR #101, opened 2022-03-23
[FLINK-26768] Clear errors when observed running job successfully

wangyang0918:
I change the `JobReconciler#reconcile()` a bit in this PR. Even though, the spec does not change, we still have a chance to update the reconciliationStatus.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/BaseObserver.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/BaseObserver.java
@@ -22,6 +22,7 @@ import org.apache.flink.configuration.Configuration;
 import org.apache.flink.kubernetes.operator.config.FlinkOperatorConfiguration;
 import org.apache.flink.kubernetes.operator.crd.FlinkDeployment;
 import org.apache.flink.kubernetes.operator.crd.status.FlinkDeploymentStatus;
+import org.apache.flink.kubernetes.operator.crd.status.ReconciliationStatus;
 import org.apache.flink.kubernetes.operator.exception.DeploymentFailedException;
 import org.apache.flink.kubernetes.operator.service.FlinkService;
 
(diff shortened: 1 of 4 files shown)


gyfora:
@wangyang0918 I wonder if the whole logic could be simplified as follows:

if after the reconcile step, we did not throw an exception and there is still an error, clear the error. This way this could be simple 2 lines in the controller logic after reconcile.

wangyang0918:
So you are suggesting that we could simply clear the error no matter we have observed running job or not. Right?

gyfora:
You are right we should actually not clear errors if the JobDeployment is in ERROR state. So maybe the Observer already could clear errors

gyfora:
I am not completely sure what I am suggesting but it feels strange to have this logic inside the reconciler.

wangyang0918:
Of cause, the observer could also clear the error. The reason why I put this logic in the reconciler is that observer should not update the `reconciliationStatus`.

gyfora:
At the moment we use the reconciliationStatus/error to store both errors related to the reconciliation and when we dedect errors with the deployment.

So it is a bit overused field actually which causes some complications now :)

wangyang0918:
> At the moment we use the reconciliationStatus/error to store both errors related to the reconciliation and when we dedect errors with the deployment.
> 
> So it is a bit overused field actually which causes some complications now :)

I agree with you about this. So do you think we need to separate the two different errors? One is related to the reconciliation and another is about detecting the JobManager deployment.

gyfora:
I would not rush with separating this because it's not always clear what kind of error it is. In most cases the deployment error is actually related to the reconciliation (we submitted an invalid image or service account config or something).

I think the best thing we could do is make sure errors are cleared in a consistent way and later we can introduce new fields if we see the need.

What do you think about this flow:
1. Observer observes deployment
2. If we are not in JMDeployment.ERROR state clear the error information
3. The reconciler will reset error if necessery

This might cause a back and forth setting of error states in cases where the job is running but there is an invalid reconcile request but this happens within the controller loop so the user will not see this.

wangyang0918:
> I would not rush with separating this because it's not always clear what kind of error it is. In most cases the deployment error is actually related to the reconciliation (we submitted an invalid image or service account config or something).
> 
> I think the best thing we could do is make sure errors are cleared in a consistent way and later we can introduce new fields if we see the need.
> 
> What do you think about this flow:
> 
> 1. Observer observes deployment
> 2. If we are not in JMDeployment.ERROR state clear the error information
> 3. The reconciler will set the error again if necessery (that is how it already works)
> 
> This might cause a back and forth setting of error states in cases where the job is running but there is an invalid reconcile request but this happens within the controller loop so the user will not see this.

Nice suggestion and comments have been addressed accordingly.

gyfora:
Looks good, I will test this a bit later :)

PR #102, opened 2022-03-23
[FLINK-26815] Tests in FlinkConfigBuilderTest are skipped

morhidi:
- hot fix for the NPE caused by [FLINK-26706](https://issues.apache.org/jira/browse/FLINK-26706) Introduce Ingress URL templating
- bumped surfire version see https://issues.apache.org/jira/browse/SUREFIRE-1741


morhidi:
cc @wangyang0918 @gyfora

PR #103, opened 2022-03-23
[FLINK-26830] Transition to SUSPEND before redeploying on job upgrades

gyfora:
Break redeployments into two steps to avoid getting into an inconsistent state if job submission fails.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
@@ -17,14 +17,17 @@
 
 package org.apache.flink.kubernetes.operator.reconciler;
 
+import org.apache.flink.kubernetes.operator.config.FlinkOperatorConfiguration;
 import org.apache.flink.kubernetes.operator.crd.FlinkDeployment;
 import org.apache.flink.kubernetes.operator.crd.spec.FlinkDeploymentSpec;
+import org.apache.flink.kubernetes.operator.crd.spec.JobState;
 import org.apache.flink.kubernetes.operator.crd.status.ReconciliationStatus;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.javaoperatorsdk.operator.api.reconciler.UpdateControl;
 
+import java.time.Duration;
 import java.util.Objects;
 
 /** Reconciliation utilities. */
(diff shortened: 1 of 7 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
```diff
@@ -93,6 +101,26 @@ public static void updateForReconciliationError(FlinkDeployment flinkApp, String
...
+        }
+
+        ReconciliationStatus reconciliationStatus = current.getStatus().getReconciliationStatus();
+        if (current.getSpec().getJob() != null
+                && current.getSpec().getJob().getState() == JobState.RUNNING
+                && reconciliationStatus != null
+                && reconciliationStatus.isSuccess()
+                && reconciliationStatus.getLastReconciledSpec().getJob().getState()
+                        == JobState.SUSPENDED) {
+            return updateControl.rescheduleAfter(0);
```
wangyang0918, line 116: Even though, we schedule a ZERO delay reconciliation, we still have a chance other CR changes sneak in and then applied together with next reconciliation. Maybe this is not a problem since we are always orienting toward the final state.

wangyang0918 (commented):
Separating the upgrade process into two step(suspending, redeploying) makes lots of sense to me.



When I am testing this PR, I find another issue when change upgrade mode from `stateless` to `savepoint` with `examples/basic.yaml`. The reconciliation failed because of savepoint path is not configured. Combined with FLINK-26577, it makes me think more about necessity of allowing upgrade mode change during the running state.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
gyfora: Maybe there is a better way provided by the SDK  (to move to the beginning of the queue) but you could actually argue that it's better to get the new updates before finishing the upgrade, that way you wouldnt have to redeploy twice

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentControllerTest.java
```diff
@@ -403,8 +425,18 @@ public void testUpgradeNotReadyCluster(FlinkDeployment appCluster, boolean allow
 
             flinkService.setPortReady(true);
             testController.reconcile(appCluster, context);
-            testController.reconcile(appCluster, context);
-
+            if (appCluster.getSpec().getJob() != null
```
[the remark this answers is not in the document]
gyfora: good point

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
wangyang0918: Yes. I think we do not need to move the redeploy event to the beginning of the queue. As a side effect, we save an unnecessary restart.

wangyang0918 (approved):
+1

PR #107, opened 2022-03-24
[hotfix][FLINK-26832] Some minor logging / observe fixes

gyfora:


gyfora:
cc @bgeng777

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/BaseObserver.java
```diff
@@ -103,10 +111,12 @@ protected void observeJmDeployment(
                 return;
             }
 
+            logger.info("JobManager is deployed but not ready yet", status);
```
bgeng777, line 114: The old code seems more clear:
```
logger.info("JobManager deployment exists but not ready, status {}", status);
```
The new description `deployed but not ready yet` is not wrong but is too close to the enum `DEPLOYED_NOT_READY`, while we actully set the status to `DEPLOYING`.

gyfora: At this point it is always deploying status so no point in printing the current status. I could change it to 
```
JobManager is being deployed
```

gyfora: fixed

bgeng777:
Thanks for improving the logging! LGTM.

The optimization of skipping observation of JM when the job is in `suspended` state can reduce unnecessary checks, as we will not care JM when there is no job running. 
I am not perfectly sure if we need to make some changes for incoming session job. But I think it is out of the scope of this PR.

PR #109, opened 2022-03-24
[FLINK-26577] Always trigger a savepoint when upgrade mode changed to last-state

wangyang0918:
> At the moment there are several corner cases which can lead to accidental state loss (or at least weird behaviour) when switching to last-state upgrade mode from other modes.
>
> 2 cases that immediately come to mind:
>
> savepoint to last-state: 
When the new upgrade mode is last-state, the job deployment will simply be deleted. If HA was not enabled previously, the last savepoint might be very far back in time.
>
> stateless to last-state:
If checkpointing and HA is not enabled, the deployment will simply be killed like previously and we might start a job from empty state. Maybe taking a savepoint would be the right approach in this case and continue from there.

After this PR, a savepoint will always be triggered when users change upgrade mode from stateless/savepoint to last-state. This is a safe guard so that we will never lose state.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/JobReconciler.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/JobReconciler.java
@@ -34,6 +34,7 @@ import org.apache.flink.kubernetes.operator.utils.FlinkUtils;
 import org.apache.flink.kubernetes.operator.utils.IngressUtils;
 import org.apache.flink.kubernetes.operator.utils.SavepointUtils;
 import org.apache.flink.runtime.jobgraph.SavepointConfigOptions;
+import org.apache.flink.util.Preconditions;
 
 import io.fabric8.kubernetes.client.KubernetesClient;
 import io.javaoperatorsdk.operator.api.reconciler.Context;
@@ -82,7 +83,11 @@ public class JobReconciler extends BaseReconciler {
         }
 
         boolean specChanged = !flinkApp.getSpec().equals(lastReconciledSpec);
-        if (specChanged && readyForSpecChanges(flinkApp)) {
+        if (specChanged) {
+            if (!inUpgradeableState(flinkApp)) {
+                LOG.info("Waiting for upgradeable state");
+                return;
+            }
             JobState currentJobState = lastReconciledSpec.getJob().getState();
             JobState desiredJobState = jobSpec.getState();
 
@@ -92,7 +97,7 @@ public class JobReconciler extends BaseReconciler {
                 if (desiredJobState == JobState.RUNNING) {
                     LOG.info("Upgrading running job, suspending first...");
                 }
-                printCancelLogs(upgradeMode, flinkApp.getMetadata().getName());
+                printCancelLogs(upgradeMode);
                 suspendJob(flinkApp, upgradeMode, effectiveConfig);
                 stateAfterReconcile = JobState.SUSPENDED;
             }
@@ -113,9 +118,12 @@ public class JobReconciler extends BaseReconciler {
         }
     }
 
-    private boolean readyForSpecChanges(FlinkDeployment deployment) {
-        if (deployment.getSpec().getJob().getUpgradeMode() != UpgradeMode.SAVEPOINT) {
-            // Only savepoint upgrade mode needs a running job
+    private boolean inUpgradeableState(FlinkDeployment deployment) {
+        if (deployment.getSpec().getJob().getUpgradeMode() != UpgradeMode.SAVEPOINT
+                && !ReconciliationUtils.isUpgradeModeChangedToLastStateAndHADisabledPreviously(
+                        deployment)) {
+            // Only savepoint upgrade mode or changed from stateless/savepoint to last-state while
+            // HA disabled previously need a running job
             return true;
         }
         return deployment.getStatus().getJobManagerDeploymentStatus()
@@ -155,7 +163,7 @@ public class JobReconciler extends BaseReconciler {
         deployFlinkJob(flinkApp, effectiveConfig, savepointOpt);
     }
 
-    private void printCancelLogs(UpgradeMode upgradeMode, String name) {
+    private void printCancelLogs(UpgradeMode upgradeMode) {
         switch (upgradeMode) {
             case STATELESS:
                 LOG.info("Cancelling job");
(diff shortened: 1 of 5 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/JobReconciler.java
```diff
@@ -82,7 +83,11 @@ public void reconcile(FlinkDeployment flinkApp, Context context, Configuration e
         }
 
         boolean specChanged = !flinkApp.getSpec().equals(lastReconciledSpec);
-        if (specChanged && readyForSpecChanges(flinkApp)) {
+        if (specChanged) {
+            if (!readyForSpecChanges(flinkApp)) {
```
gyfora, line 87: Could we please rename this method and changed the logged line? `readyForSpecChanges` is a little confusing.

We could call it `inUpgradeableState` that would describe the logic better

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/JobReconciler.java
```diff
@@ -114,6 +119,9 @@ public void reconcile(FlinkDeployment flinkApp, Context context, Configuration e
     }
 
     private boolean readyForSpecChanges(FlinkDeployment deployment) {
+        if (isUpgradeModeChangedToLastState(deployment)) {
+            return isJobRunning(deployment);
```
gyfora, line 123: I think we could allow upgrading to last-state even SUSPENDED.  So we could simply change the IF branch at  line 125 to:

```
if (deployment.getSpec().getJob().getUpgradeMode() != UpgradeMode.SAVEPOINT && !isUpgradeModeChangedToLastState(deployment))
```

This way if previous job was with savepoint we would pick up the last savepoint  otherwise we either have state or not depending on the HA setting. Which would be fair for a suspended job.

gyfora (changes requested):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/JobReconciler.java
wangyang0918: I think you are right. We should allow upgrade to last-state even SUSPENDED.

wangyang0918:
Pushed a `fixup` commit to address the comments.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/reconciler/JobReconcilerTest.java
```diff
@@ -143,6 +144,10 @@ public void testUpgradeModeChangeFromSavepointToLastState() throws Exception {
         deployment.getSpec().getJob().setState(JobState.RUNNING);
         deployment.getSpec().setImage("new-image-2");
 
+        // Wait ready for spec changes
+        deployment.getStatus().setJobManagerDeploymentStatus(JobManagerDeploymentStatus.READY);
+        deployment.getStatus().getJobStatus().setState(JobState.RUNNING.name());
+
```
gyfora, line 150: This should not be necessary, in this test we simulate the SUSPENDED -> RUNNING

wangyang0918: good point.

gyfora:
Looks good! I am going to test this locally

wangyang0918:
Maybe we also need the sanity check for savepoint config options when changing from stateless/savepoint to last-state. I will do it tomorrow or in a follow-up ticket.

wangyang0918:
> Maybe the only thing I would improve now that I am playing around with it is that if we have HA enabled for a job and the job is already running , we probably should not trigger a savepoint when switching to last-state.
> 
> Do you see any corner cases with this logic?

Actually, I also have this improvement in my mind at the very begging and not sure about whether to introduce this. Because we could not guarantee that there is a successful checkpoint in the ConfigMap. You could argue that the state will also be lost when JobManager failed over. But changing the upgrade mode is not a frequent operation, I think triggering a savepoint is reasonable. WDYT?

gyfora:
I see the problems but I think this is a critical behaviour in certain production situations. Let's take this scenario:

You have a job with SAVEPOINT upgrade mode (+checkpointing/HA) running. The job starts having some memory (or other) problems and goes into a failure loop. The only way to solve it is to redeploy a new version but you cannot take savepoints anymore.

At this point you NEED to use the last-state strategy and if that forces a savepoint, you are completely screwed.

wangyang0918:
> I see the problems but I think this is a critical behaviour in certain production situations. Let's take this scenario:
> 
> You have a job with SAVEPOINT upgrade mode (+checkpointing/HA) running. The job starts having some memory (or other) problems and goes into a failure loop. The only way to solve it is to redeploy a new version but you cannot take savepoints anymore.
> 
> At this point you NEED to use the last-state strategy and if that forces a savepoint, you are completely screwed.

Fair point. I will improve the current implementation.

gyfora:
@wangyang0918 do you think there is a simple way to check the configmaps and that they have any checkpoint info in them?

wangyang0918:
> do you think there is a simple way to check the configmaps and that they have any checkpoint info in them?
> 
> Maybe we could rely on some pre-existing flink utility

Even though we have a successful checkpoint in the ConfigMap, it still could be an old one since the checkpoint interval is too big. So I am suggesting only check the HA is enabled previously.

gyfora:
Sounds good we do not need to solve all the hard problems right now :)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
```diff
@@ -118,6 +122,29 @@ public static void updateForReconciliationError(FlinkDeployment flinkApp, String
         return updateControl.rescheduleAfter(rescheduleAfter.toMillis());
     }
 
+    public static boolean isUpgradeModeChangedToLastStateAndHADisabledPreviously(
+            FlinkDeployment flinkApp) {
+        final FlinkDeploymentSpec lastReconciledSpec =
+                Preconditions.checkNotNull(
+                        flinkApp.getStatus().getReconciliationStatus().getLastReconciledSpec());
+        if (lastReconciledSpec.getJob() == null || flinkApp.getSpec().getJob() == null) {
```
gyfora, line 130: These will never be null for jobs so I think we can remove the check.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
```diff
@@ -118,6 +122,29 @@ public static void updateForReconciliationError(FlinkDeployment flinkApp, String
...
+            return false;
+        }
+        if (lastReconciledSpec.getFlinkConfiguration() == null
+                || lastReconciledSpec.getFlinkConfiguration().isEmpty()) {
+            return false;
```
gyfora, line 135: If upgrade mode is last state these will never be null due to validation, so we can remove the check

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultDeploymentValidator.java
```diff
@@ -241,6 +243,19 @@
             return Optional.of("Cannot switch from job to session cluster");
         }
 
+        if (StringUtils.isNullOrWhitespaceOnly(
+                        newSpec.getFlinkConfiguration()
+                                .get(CheckpointingOptions.SAVEPOINT_DIRECTORY.key()))
+                && deployment.getStatus().getJobManagerDeploymentStatus()
```
gyfora, line 249: While it might be correct wouldn’t it be clearer to check the last reconciled job state == RUNNING?

gyfora (commented):
Added some minor simplification comments, will test this in the afternoon

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultDeploymentValidator.java
wangyang0918: Imagine the current `JobManagerDeploymentStatus` is `ERROR`, and `isUpgradeModeChangedToLastStateAndHADisabledPreviously()` is true, the savepoint path also needs to be configured. Right? Because it might turn into `READY` in the following reconciliation.

PR #110, opened 2022-03-24
[FLINK-26836] Add sanity check for state.savepoints.dir when using savepoint upgrade mode

SteNicholas:
Add sanity check for `state.savepoints.dir` when using savepoint upgrade mode and manually triggering endpoint for the running job.

**The brief change log**
- `DefaultDeploymentValidator` adds the check for `state.savepoints.dir` when the upgrade mode is `SAVEPOINT` or `savepointTriggerNonce` isn't null.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultDeploymentValidator.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultDeploymentValidator.java
@@ -17,6 +17,7 @@
 
 package org.apache.flink.kubernetes.operator.validation;
 
+import org.apache.flink.configuration.CheckpointingOptions;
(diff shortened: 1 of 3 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultDeploymentValidator.java
```diff
@@ -146,6 +148,22 @@
             return Optional.of("Job could not be upgraded with last-state while HA disabled");
         }
 
+        if (StringUtils.isNullOrWhitespaceOnly(
+                Configuration.fromMap(confMap)
```
gyfora, line 152: We could move `Configuration.fromMap(confMap)` to a variable and reuse in this method.

gyfora (approved):
Added a minor comment, otherwise looks good!

PR #118, opened 2022-03-28
[FLINK-26214] Publish Docker images to GitHub Registry

mbalassi:
I managed to make this work with simply using `flink-operator` for the docker image name, after reconsideration I prefer this for brevity.
We only publish images for the `main` branch after this PR, I will proceed with working on the release branch/tag setup.
We should improve this ideally where we only tag the image with the `latest` tag if the CI finishes successfully.

### DIFF
--- a/.github/workflows/docker_push.yml
+++ b/.github/workflows/docker_push.yml
@@ -16,15 +16,21 @@
 # limitations under the License.
 ################################################################################
 
-name: "Push to Docker Registry"
+name: "Build Docker Image"
 on:
   workflow_dispatch:
   push:
     branches:
       - main
+      - 'release-*'
+    tags:
+      - 'v*'
+  pull_request:
+    branches:
+      - main
+      - 'release-*'
 jobs:
-  github_registry_push:
-    name: github_registry_push
+  build_image:
     runs-on: ubuntu-latest
     permissions:
       packages: write
@@ -34,7 +40,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Log in to the Container registry
-        uses: docker/login-action@f054a8b539a109f9f41c372932f1ae047eff08c9
+        uses: docker/login-action@v1
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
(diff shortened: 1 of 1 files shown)


mbalassi:
I tested via [mbalassi/packages](https://github.com/mbalassi/flink-kubernetes-operator/pkgs/container/flink-operator), and could already manually push to [apache/packages](https://github.com/orgs/apache/packages/container/package/flink-operator).

.github/workflows/docker_push.yml
```diff
@@ -0,0 +1,59 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      packages: write
+      contents: read
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v2
+
+      - name: Log in to the Container registry
+        uses: docker/login-action@f054a8b539a109f9f41c372932f1ae047eff08c9
```
tisonkun, line 37:
```suggestion
        uses: docker/login-action@v1
```

.github/workflows/docker_push.yml
```diff
@@ -0,0 +1,59 @@
...
+            type=raw,value=latest,enable=${{ github.ref == format('refs/heads/{0}', github.event.repository.default_branch) }}
+            type=sha,prefix=,format=short
+
+      - name: Build and push Docker images
+        uses: docker/build-push-action@ad44023a93711e3deb337508980b4b5e9bcdc5dc
+        with:
+          context: .
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
```
tisonkun, line 59:
```suggestion
          labels: ${{ steps.meta.outputs.labels }}

```

.github/workflows/docker_push.yml
```diff
@@ -0,0 +1,59 @@
...
+            ghcr.io/${{ github.repository_owner }}/flink-operator
+          tags: |
+            type=raw,value=latest,enable=${{ github.ref == format('refs/heads/{0}', github.event.repository.default_branch) }}
+            type=sha,prefix=,format=short
+
+      - name: Build and push Docker images
+        uses: docker/build-push-action@ad44023a93711e3deb337508980b4b5e9bcdc5dc
+        with:
+          context: .
+          push: true
```
tisonkun, line 57: You may trigger the docker packaging logic on `pull_request` also while skip the push phase so that we can check before merging a pull request. You can refer to [this workflow](https://github.com/engula/engula/blob/93cd8b3be0cd8ffc6a3a7f331e6bde2ca9d15284/.github/workflows/docker.yml) as an example.

tisonkun (commented):
Generally looks good! Comments inline.

.github/workflows/docker_push.yml
mbalassi: Great!

.github/workflows/docker_push.yml
mbalassi: Good call, I cleaned up references like this

mbalassi:
Thanks, @tisonkun I addressed your comments. I went with an approach based on the [Semantic versioning](https://github.com/docker/metadata-action#semver) example of the docker metadata action.

.github/workflows/docker_push.yml
```diff
@@ -0,0 +1,67 @@
...
+
+name: "Build Docker Image"
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+      - 'release-*'
+    tags:
+      - 'v*'
```
gyfora, line 27: this should be `release-*

mbalassi: Good catch

.github/workflows/docker_push.yml
```diff
@@ -0,0 +1,67 @@
...
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Extract metadata (tags, labels) for Docker
+        id: meta
+        uses: docker/metadata-action@v3
+        with:
+          images: |
+            ghcr.io/${{ github.repository_owner }}/flink-operator
+          tags: |
+            type=raw,value=main,enable=${{ github.ref == 'refs/heads/main') }}
```
gyfora, line 56: remove extra `)` please

PR #119, opened 2022-03-28
[examples] chore - remove Flink patch version number

haoxins:


mbalassi (approved):
Thanks, it aligns better with the `flinkVersion` field this way.

PR #120, opened 2022-03-28
[FLINK-26881] Pick up the savepoint directory in the flink config when trigger savepoint

wangyang0918:
This PR allows to pick the savepoint directory in the flink config instead of `null` when trigger savepoint. After this PR, upgrading FlinkDeployment from stateless to savepoint could actually work.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
@@ -28,6 +28,7 @@ import org.apache.flink.client.deployment.application.ApplicationConfiguration;
 import org.apache.flink.client.deployment.application.cli.ApplicationClusterDeployer;
 import org.apache.flink.client.program.ClusterClient;
 import org.apache.flink.client.program.rest.RestClusterClient;
+import org.apache.flink.configuration.CheckpointingOptions;
 import org.apache.flink.configuration.Configuration;
 import org.apache.flink.configuration.RestOptions;
 import org.apache.flink.kubernetes.configuration.KubernetesConfigOptions;
@@ -51,6 +52,7 @@ import org.apache.flink.runtime.rest.messages.job.savepoints.SavepointStatusMess
 import org.apache.flink.runtime.rest.messages.job.savepoints.SavepointTriggerHeaders;
 import org.apache.flink.runtime.rest.messages.job.savepoints.SavepointTriggerMessageParameters;
 import org.apache.flink.runtime.rest.messages.job.savepoints.SavepointTriggerRequestBody;
+import org.apache.flink.util.Preconditions;
 
 import io.fabric8.kubernetes.api.model.PodList;
 import io.fabric8.kubernetes.client.KubernetesClient;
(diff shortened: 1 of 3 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
```diff
@@ -172,10 +174,14 @@ public boolean isJobManagerPortReady(Configuration config) {
                     clusterClient.cancel(jobID).get(1, TimeUnit.MINUTES);
                     break;
                 case SAVEPOINT:
+                    final String savepointDirectory =
+                            Preconditions.checkNotNull(
+                                    conf.get(CheckpointingOptions.SAVEPOINT_DIRECTORY));
+                    final long timeout = operatorConfiguration.getFlinkClientTimeout().getSeconds();
```
gyfora, line 180: The flink client timeout that we have in the operator config is mostly calibrated for listing actions and has 10 sec default. I think that might cause problems here

wangyang0918: I will keep the timeout as before and introduce a new config option if necessary in the future.

gyfora: How about taking the checkpoint timeout from the flink config directly here? (And use the default for that from flink)

wangyang0918: I am afraid we could not get it from the flink config and it could also be overridden by user.

```
env.getCheckpointConfig().setCheckpointTimeout(60000);
```

gyfora: It should somehow be configurable I agree on that otherwise this will fail for larger states :/

gyfora: Worst case we could introduce a savepoint timeout config for now

bgeng777: Is the [execution.checkpointing.timeout](https://nightlies.apache.org/flink/flink-docs-master/zh/docs/deployment/config/#execution-checkpointing-timeout) what we want?

wangyang0918: The user configured timeout has high priority. Right?

gyfora: We could ignore that in this case and make a note of it

gyfora: I agree with using the execution.checkpointing.timeout

wangyang0918: I will verify whether it also take effect for savepoint, not only checkpoint.

wangyang0918:
Pushed a `fixup` to address the comment.

PR #121, opened 2022-03-28
[hotfix] Disable flink-kubernetes-operator shade step on apache-release profile

gyfora:
This fix disables the shade plugin for the flink-kubernetes-operator module when using the apache-release in order to not deploy the fatjar as part of the release process.


mbalassi (approved):
LGTM, we should land this to `main` too.

PR #123, opened 2022-03-28
[FLINK-26214] Helm chart uses GitHub packages for docker image

mbalassi:
Adds the new package to the helm chart and removes the now unnecessary steps from the quickstart.

### DIFF
--- a/docs/content/docs/try-flink-kubernetes-operator/quick-start.md
+++ b/docs/content/docs/try-flink-kubernetes-operator/quick-start.md
@@ -63,20 +63,13 @@ minikube start --kubernetes-version=v1.21.5
 
 We also recommend [k9s](https://k9scli.io/) as GUI for kubernetes, but it is optional for this quickstart guide.
 
-## Building the operator image from source
+## Get the source code
+
+You can get the operator source code from GitHub:
 
 ```shell
 git clone https://github.com/apache/flink-kubernetes-operator.git
 cd flink-kubernetes-operator
-```
-
-You can build the docker image of the operator via the following command from the project root folder:
-```bash
-docker build . -t flink-operator
-```
-When using minikube you need to make this image available within the running minikube container:
-```bash
-minikube image load flink-operator
 ```
 
 ## Deploying the operator
--- a/helm/flink-operator/values.yaml
+++ b/helm/flink-operator/values.yaml
@@ -23,9 +23,9 @@
 # watchNamespaces: ["flink"]
 
 image:
-  repository: flink-operator
+  repository: ghcr.io/apache/flink-operator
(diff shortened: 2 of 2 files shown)


helm/flink-operator/values.yaml
```diff
@@ -23,9 +23,9 @@
 # watchNamespaces: ["flink"]
 
 image:
-  repository: flink-operator
+  repository: ghcr.io/apache/flink-operator
   pullPolicy: IfNotPresent
-  tag: latest
+  tag: main
```
tisonkun, line 28: I may prefer use latest in the CD workflow which is the default manner in the Docker world. Could you explain a bit why do we use `main` instead of `latest` in the CD workflow? IIRC `docker pull <image>` is default to `latest`.

docs/content/docs/try-flink-kubernetes-operator/quick-start.md
```diff
@@ -63,22 +63,15 @@ minikube start --kubernetes-version=v1.21.5
 
 We also recommend [k9s](https://k9scli.io/) as GUI for kubernetes, but it is optional for this quickstart guide.
 
-## Building the operator image from source
+## Get the source code
```
bgeng777, line 66: Should we use `## Getting the source code` to follow our naming pattern of each section?

docs/content/docs/try-flink-kubernetes-operator/quick-start.md
```diff
@@ -63,22 +63,15 @@ minikube start --kubernetes-version=v1.21.5
...
+
+You can get the operator source code from GitHub:
 
 ```shell
 git clone https://github.com/apache/flink-kubernetes-operator.git
 cd flink-kubernetes-operator
 ```
 
-You can build the docker image of the operator via the following command from the project root folder:
```
gyfora, line 73: Why would you want to get the source code if you are not going to build the operator?

helm/flink-operator/values.yaml
wangyang0918: Given that this is the snapshot docker image of latest main branch, I also think `latest` tag might be better here.

wangyang0918:
Out the scope of this PR, I am thinking how we could install the flink-kubernetes-operator like airflow. Then we do not force the users to download the source code.

```
helm repo add apache-airflow https://airflow.apache.org
helm upgrade --install airflow apache-airflow/airflow --namespace airflow --create-namespace
```

https://airflow.apache.org/docs/helm-chart/stable/index.html

mbalassi:
Thanks or the feedback, team. I am closing this as per #126 we are probably taking a better approach to this.

PR #125, opened 2022-03-29
[FLINK-26903] Add correct NOTICE file to flink-kubernetes-operator

wangyang0918:
Given that `flink-kubernetes-shaded` already contains a NOTICE file, and `flink-kubernetes-webhook` does not bundle any dependencies. This PR is trying to add a correct NOTICE file only for the `flink-kubernetes-operator` module.

I have reviewed all the bundled dependencies in the uber jar to make sure they are listed in the NOTICE file with correct license.

cc @gyfora @tweise correct me if I miss anything.

### DIFF
--- a/flink-kubernetes-operator/pom.xml
+++ b/flink-kubernetes-operator/pom.xml
@@ -61,6 +61,12 @@ under the License.
             <groupId>org.apache.flink</groupId>
             <artifactId>flink-clients_${scala.version}</artifactId>
             <version>${flink.version}</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.apache.flink</groupId>
+                    <artifactId>flink-scala_${scala.version}</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
 
         <dependency>
--- a/flink-kubernetes-operator/src/main/resources/META-INF/NOTICE
+++ b/flink-kubernetes-operator/src/main/resources/META-INF/NOTICE
@@ -1,0 +1,113 @@
+flink-kubernetes-operator
+Copyright 2014-2021 The Apache Software Foundation
+
+This project includes software developed at
+The Apache Software Foundation (http://www.apache.org/).
+
+This project bundles the following dependencies under the Apache Software License 2.0 (http://www.apache.org/licenses/LICENSE-2.0.txt)
+
+- com.fasterxml.jackson.core:jackson-annotations:2.13.1
+- com.fasterxml.jackson.core:jackson-core:2.13.1
+- com.fasterxml.jackson.core:jackson-databind:2.13.1
+- com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:2.13.1
+- com.fasterxml.jackson.datatype:jackson-datatype-jsr310:2.13.1
+- com.github.mifmif:generex:1.0.2
+- com.google.auto.service:auto-service-annotations:1.0.1
+- com.google.auto.service:auto-service:1.0.1
+- com.google.auto:auto-common:1.2
+- com.google.code.findbugs:jsr305:1.3.9
+- com.google.errorprone:error_prone_annotations:2.7.1
+- com.google.guava:failureaccess:1.0.1
+- com.google.guava:guava:31.0.1-jre
+- com.google.guava:listenablefuture:9999.0-empty-to-avoid-conflict-with-guava
+- com.google.j2objc:j2objc-annotations:1.3
+- com.squareup.okhttp3:logging-interceptor:3.12.12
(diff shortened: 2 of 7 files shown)


flink-kubernetes-operator/pom.xml
```diff
@@ -61,6 +61,12 @@ under the License.
             <groupId>org.apache.flink</groupId>
             <artifactId>flink-clients_${scala.version}</artifactId>
             <version>${flink.version}</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.apache.flink</groupId>
+                    <artifactId>flink-scala_${scala.version}</artifactId>
+                </exclusion>
+            </exclusions>
```
gyfora, line 69: What is the reason for excluding this?

flink-kubernetes-operator/src/main/resources/META-INF/NOTICE
```diff
@@ -0,0 +1,113 @@
...
+- org.apache.flink:flink-rpc-akka-loader:1.14.4
+- org.apache.flink:flink-rpc-core:1.14.4
+- org.apache.flink:flink-runtime:1.14.4
+- org.apache.flink:flink-shaded-asm-7:7.1-14.0
+- org.apache.flink:flink-shaded-force-shading:14.0
+- org.apache.flink:flink-shaded-guava:30.1.1-jre-14.0
+- org.apache.flink:flink-shaded-jackson:2.12.4-14.0
+- org.apache.flink:flink-shaded-netty:4.1.65.Final-14.0
+- org.apache.flink:flink-shaded-zookeeper-3:3.4.14-14.0
+- org.apache.flink:flink-streaming-java_2.12:1.14.4
```
gyfora, line 91: Do we need to list Flink dependencies here?

flink-kubernetes-operator/pom.xml
wangyang0918: All the interfaces we are using is scala free, so do we really need to bundle scala in the uber jar?

flink-kubernetes-operator/src/main/resources/META-INF/NOTICE
wangyang0918: Since `flink-kubernetes-operator` is a separate project and we bundle Flink jars, that's why I also list the Flink dependencies here. Maybe you are right. But I am not sure.

flink-kubernetes-operator/pom.xml
gyfora: why would it be in the client if it's not using it?

flink-kubernetes-operator/src/main/resources/META-INF/NOTICE
gyfora: we can leave it there for now, no harm here. But we can revisit this later and remove if not necessary

flink-kubernetes-operator/pom.xml
gyfora: I dont actually know if we are going to hit any issues here, just playing the devil's advocate :)

gyfora: safest would be for now is to keep it included

wangyang0918: The scala dependency was introduced by `flink-streaming-java_2.12`. I remember the root cause is about akka.

gyfora (approved):
Let's merge this if you feel comfortoble with the scala exclusion. I will create an RC2 afterwards.

Please also merge this change to `release-0.1` branch

flink-kubernetes-operator/pom.xml
wangyang0918: I am almost sure we are not using scala classes in the `flink-kubernetes-operator`. That is the reason I excluded it to reduce the jar size. But I could add it back if you are concerned about this.

gyfora: I am not concerned should be ok :)

flink-kubernetes-operator/src/main/resources/META-INF/NOTICE
zentol: All Flink artifacts come with a nice notice file that if properly merged should remove the need to explicitly mention it here.

flink-kubernetes-operator/pom.xml
zentol: streaming-java had a transitive dependency on flink-scala in 1.14-, which was removed in 1.15.0. To keep the dependency graph identical we added an explicit dependency on flink-scala.
If you don't run any application within the operator itself then you shouldn't need it.

That said you also shouldn't need a bunch of other stuff, like the QS client, rpc-akka-loader, asm, flink-optimizer etc etc.

PR #127, opened 2022-03-29
[BP-0.1] [FLINK-26903] Add correct NOTICE file to flink-kubernetes-operator

wangyang0918:
Backport #125 to release-0.1 branch

### DIFF
--- a/flink-kubernetes-operator/pom.xml
+++ b/flink-kubernetes-operator/pom.xml
@@ -61,6 +61,12 @@ under the License.
             <groupId>org.apache.flink</groupId>
             <artifactId>flink-clients_${scala.version}</artifactId>
             <version>${flink.version}</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.apache.flink</groupId>
+                    <artifactId>flink-scala_${scala.version}</artifactId>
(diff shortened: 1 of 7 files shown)


wangyang0918:
Will merge this PR after CI passed.

zentol:
Having the jars contain the NOTICE files is imo insufficient as discoverability within the image is basically 0. You should extract & merge the NOTICE files for all jars, similarly to what we do for flink-dist.

gyfora:
@zentol we could include this in the notice for the operator as well, but since the jars contain the notices themselves we don't really need to do this. Anyone checking the contents of the container should be able to discover these notices. Thet are not part of any bundle that we publist to dist.apache.org.

gyfora:
We can open a jira for this to make it nicer for 1.0.0 together with other NOTICE improvements

wangyang0918:
I have created a ticket FLINK-26914 to track the NOTICE file improvements.

mbalassi:
@zentol thanks for looking into this. I agree that the current setup could be improved, and support @wangyang0918's suggestion to solve this for 1.0.0, and live with it for 0.1.0.

PR #130, opened 2022-03-29
[FLINK-26214] Support cross platform image builds

mbalassi:
This enables us to publish docker build for both `amd64` and `arm64`. You can verify this via:

```shell
% docker manifest inspect ghcr.io/mbalassi/flink-operator:149986a
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
   "manifests": [
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 3055,
         "digest": "sha256:017787b0c5ccf724f676ed73f26c2928a2d9d58c4464e4ba7fb00ca3f2470158",
         "platform": {
            "architecture": "amd64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 3055,
         "digest": "sha256:c37a552f048c0431672dc190e735f4a618bd0c42030de4fea9cdc02dbab88fee",
         "platform": {
            "architecture": "arm64",
            "os": "linux"
         }
      }
   ]
}
```

While working on this I noticed instability of the maven tests in the docker builds due to growing latency, so I decided to cut down on the number of times we run the tests for a single commit/PR. The docker build does not run the tests after this change by default, but the CI does.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,6 +17,7 @@
 ################################################################################
 # Build
 FROM maven:3.8.4-openjdk-11 AS build
+ARG SKIP_TESTS=true
 
 WORKDIR /app
 ENV SHADED_DIR=flink-kubernetes-shaded
@@ -35,7 +36,7 @@ COPY $WEBHOOK_DIR/src ./$WEBHOOK_DIR/src
 
 COPY tools ./tools
 
-RUN --mount=type=cache,target=/root/.m2 mvn -ntp clean install
+RUN --mount=type=cache,target=/root/.m2 mvn -ntp clean install -DskipTests=$SKIP_TESTS
 
 # stage
 FROM openjdk:11-jre
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,7 +56,7 @@ jobs:
           export SHELL=/bin/bash
           export DOCKER_BUILDKIT=1
           eval $(minikube -p minikube docker-env)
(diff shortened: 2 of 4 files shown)


Dockerfile
```diff
@@ -35,7 +36,7 @@ COPY $WEBHOOK_DIR/src ./$WEBHOOK_DIR/src
 
 COPY tools ./tools
 
-RUN --mount=type=cache,target=/root/.m2 mvn -ntp clean install
+RUN --mount=type=cache,target=/root/.m2 mvn -ntp clean install -DskipTests=$SKIP_TESTS
```
Aitozi, line 39: I have also think about this, can we just mvn build when building docker image, Because the CI already run the test. (Make the `Build with Maven` not skipTests as now)

Aitozi: Get it.

tisonkun:
@mbalassi it seems that we're now publishing images with sha tags only and no more `main`?

mbalassi:
@tisonkun On the `main` branch we do tag them with `main`, but we also publish the the images on the release branches (currently `release-0.1`), those are the ones you see appearing without the `main` tag. We should tag these with the branch name too, my here with the `semver` tag seems bugged.

Screenshot 2022-03-30 at 12 44 33

These 2 images are my latest push on the 2 actively built branches.

PR #131, opened 2022-03-29
[FLINK-26892] Observe current status before validating CR changes

bgeng777:
- Adjust current logic in `FlinkDeploymentController` from `validate with new config -> observe with new config -> reconcile with new config `to `observe with latest validated config -> validate new config -> reconcile with new config` to get rid of the case that once a wrong config is applied without webhook, the controller will get stuck on the validation error and lose control of the previous running deployment.
- Refactor `FlinkConfigBuilder` to add support of building config with metadata and deployment spec.
- Add corresponding test.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/Mode.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/Mode.java
@@ -19,13 +19,33 @@
 package org.apache.flink.kubernetes.operator.config;
 
 import org.apache.flink.kubernetes.operator.crd.FlinkDeployment;
+import org.apache.flink.kubernetes.operator.crd.spec.FlinkDeploymentSpec;
 
 /** The mode of {@link FlinkDeployment}. */
 public enum Mode {
     APPLICATION,
     SESSION;
 
+    /**
+     * Return the mode of the given FlinkDeployment for Observer and Reconciler. Note, switching
+     * mode for an existing deployment is not allowed.
+     *
+     * @param flinkApp given FlinkDeployment
+     * @return Mode
+     */
     public static Mode getMode(FlinkDeployment flinkApp) {
-        return flinkApp.getSpec().getJob() != null ? APPLICATION : SESSION;
+        // Try to use lastReconciledSpec if it exists.
+        // The mode derived from last-reconciled spec or current spec should be same.
+        // If they are different, observation phase will use last-reconciled spec and validation
+        // phase will fail.
+        FlinkDeploymentSpec lastReconciledSpec =
+                flinkApp.getStatus().getReconciliationStatus().getLastReconciledSpec();
+        return lastReconciledSpec == null
+                ? getMode(flinkApp.getSpec())
+                : getMode(lastReconciledSpec);
+    }
+
+    private static Mode getMode(FlinkDeploymentSpec spec) {
+        return spec.getJob() != null ? APPLICATION : SESSION;
     }
 }
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
@@ -104,22 +104,23 @@ public class FlinkDeploymentController
 
     @Override
     public UpdateControl<FlinkDeployment> reconcile(FlinkDeployment flinkApp, Context context) {
-        FlinkDeployment originalCopy = ReconciliationUtils.clone(flinkApp);
         LOG.info("Starting reconciliation");
-
-        Optional<String> validationError = validator.validate(flinkApp);
-        if (validationError.isPresent()) {
-            LOG.error("Validation failed: " + validationError.get());
-            ReconciliationUtils.updateForReconciliationError(flinkApp, validationError.get());
-            return ReconciliationUtils.toUpdateControl(
-                    operatorConfiguration, originalCopy, flinkApp, false);
-        }
-
-        Configuration effectiveConfig =
-                FlinkUtils.getEffectiveConfig(flinkApp, defaultConfig.getFlinkConfig());
-
+        FlinkDeployment originalCopy = ReconciliationUtils.clone(flinkApp);
         try {
-            observerFactory.getOrCreate(flinkApp).observe(flinkApp, context, effectiveConfig);
+            observerFactory
+                    .getOrCreate(flinkApp)
+                    .observe(flinkApp, context, defaultConfig.getFlinkConfig());
+
+            Configuration effectiveConfig =
+                    FlinkUtils.getEffectiveConfig(flinkApp, defaultConfig.getFlinkConfig());
+            Optional<String> validationError = validator.validate(flinkApp);
+            if (validationError.isPresent()) {
+                LOG.error("Validation failed: " + validationError.get());
+                ReconciliationUtils.updateForReconciliationError(flinkApp, validationError.get());
+                return ReconciliationUtils.toUpdateControl(
+                        operatorConfiguration, originalCopy, flinkApp, false);
+            }
+
             reconcilerFactory.getOrCreate(flinkApp).reconcile(flinkApp, context, effectiveConfig);
         } catch (DeploymentFailedException dfe) {
             handleDeploymentFailed(flinkApp, dfe);
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
@@ -92,13 +92,13 @@ public class FlinkDeploymentController
     @Override
     public DeleteControl cleanup(FlinkDeployment flinkApp, Context context) {
         LOG.info("Deleting FlinkDeployment");
-        Configuration effectiveConfig =
-                FlinkUtils.getEffectiveConfig(flinkApp, defaultConfig.getFlinkConfig());
         try {
-            observerFactory.getOrCreate(flinkApp).observe(flinkApp, context, effectiveConfig);
+            observerFactory.getOrCreate(flinkApp).observe(flinkApp, context);
         } catch (DeploymentFailedException dfe) {
             // ignore during cleanup
         }
+        Configuration effectiveConfig =
+                FlinkUtils.getEffectiveConfig(flinkApp, defaultConfig.getFlinkConfig());
         return reconcilerFactory.getOrCreate(flinkApp).cleanup(flinkApp, effectiveConfig);
     }
 
@@ -107,12 +107,7 @@ public class FlinkDeploymentController
         LOG.info("Starting reconciliation");
         FlinkDeployment originalCopy = ReconciliationUtils.clone(flinkApp);
         try {
-            observerFactory
-                    .getOrCreate(flinkApp)
-                    .observe(flinkApp, context, defaultConfig.getFlinkConfig());
-
-            Configuration effectiveConfig =
-                    FlinkUtils.getEffectiveConfig(flinkApp, defaultConfig.getFlinkConfig());
+            observerFactory.getOrCreate(flinkApp).observe(flinkApp, context);
             Optional<String> validationError = validator.validate(flinkApp);
             if (validationError.isPresent()) {
                 LOG.error("Validation failed: " + validationError.get());
@@ -120,7 +115,8 @@ public class FlinkDeploymentController
                 return ReconciliationUtils.toUpdateControl(
                         operatorConfiguration, originalCopy, flinkApp, false);
             }
-
+            Configuration effectiveConfig =
+                    FlinkUtils.getEffectiveConfig(flinkApp, defaultConfig.getFlinkConfig());
             reconcilerFactory.getOrCreate(flinkApp).reconcile(flinkApp, context, effectiveConfig);
         } catch (DeploymentFailedException dfe) {
             handleDeploymentFailed(flinkApp, dfe);
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobObserver.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobObserver.java
@@ -21,13 +21,11 @@ package org.apache.flink.kubernetes.operator.observer;
 import org.apache.flink.configuration.Configuration;
 import org.apache.flink.kubernetes.operator.config.FlinkOperatorConfiguration;
 import org.apache.flink.kubernetes.operator.crd.FlinkDeployment;
-import org.apache.flink.kubernetes.operator.crd.spec.FlinkDeploymentSpec;
 import org.apache.flink.kubernetes.operator.crd.status.FlinkDeploymentStatus;
 import org.apache.flink.kubernetes.operator.crd.status.JobStatus;
 import org.apache.flink.kubernetes.operator.crd.status.SavepointInfo;
 import org.apache.flink.kubernetes.operator.reconciler.ReconciliationUtils;
 import org.apache.flink.kubernetes.operator.service.FlinkService;
-import org.apache.flink.kubernetes.operator.utils.FlinkUtils;
 import org.apache.flink.kubernetes.operator.utils.SavepointUtils;
 import org.apache.flink.runtime.client.JobStatusMessage;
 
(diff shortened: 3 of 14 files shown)


bgeng777:
cc @gyfora @wangyang0918 @Aitozi

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
```diff
@@ -104,9 +105,30 @@ public DeleteControl cleanup(FlinkDeployment flinkApp, Context context) {
...
+                            flinkApp.getMetadata(),
+                            lastReconciledSpec,
+                            defaultConfig.getFlinkConfig());
+            try {
+                observerFactory
+                        .getOrCreate(flinkApp)
+                        .observe(flinkApp, context, latestValidatedConfig);
+            } catch (DeploymentFailedException dfe) {
+                handleDeploymentFailed(flinkApp, dfe);
+                LOG.info("Reconciliation successfully completed");
```
Aitozi, line 124: This log is not right

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
```diff
@@ -104,9 +105,30 @@ public DeleteControl cleanup(FlinkDeployment flinkApp, Context context) {
 
     @Override
     public UpdateControl<FlinkDeployment> reconcile(FlinkDeployment flinkApp, Context context) {
-        FlinkDeployment originalCopy = ReconciliationUtils.clone(flinkApp);
         LOG.info("Starting reconciliation");
+        FlinkDeployment originalCopy = ReconciliationUtils.clone(flinkApp);
+        FlinkDeploymentSpec lastReconciledSpec =
+                flinkApp.getStatus().getReconciliationStatus().getLastReconciledSpec();
+        if (lastReconciledSpec != null) {
+            Configuration latestValidatedConfig =
```
Aitozi, line 113: nit: maybe called `lastValidatedConfig` better, because it from the `lastReconciledSpec`

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkConfigBuilder.java
```diff
@@ -50,13 +51,20 @@
...
 
     public FlinkConfigBuilder(FlinkDeployment deploy, Configuration flinkConfig) {
-        this.deploy = deploy;
-        this.spec = this.deploy.getSpec();
+        this.meta = deploy.getMetadata();
+        this.spec = deploy.getSpec();
+        this.effectiveConfig = new Configuration(flinkConfig);
+    }
+
+    public FlinkConfigBuilder(
```
Aitozi, line 64: Can we directly pass the namespace and clusterId here? It seems we only use these two fields

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkConfigBuilder.java
```diff
@@ -50,13 +51,20 @@
...
 public class FlinkConfigBuilder {
-    private final FlinkDeployment deploy;
+    private final ObjectMeta meta;
     private final FlinkDeploymentSpec spec;
     private final Configuration effectiveConfig;
 
     public FlinkConfigBuilder(FlinkDeployment deploy, Configuration flinkConfig) {
-        this.deploy = deploy;
-        this.spec = this.deploy.getSpec();
+        this.meta = deploy.getMetadata();
```
Aitozi, line 59: this(deploy.getMetadata(), deploy.getSpec(), flinkConfig)

Aitozi (commented):
@bgeng777 Overall looks good to me, I left some minor comments, please take a look

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
bgeng777: IIUC, when catching a DeploymentFailedException, our current code would log `"Reconciliation successfully completed"`.
I agree we can improve this log msg.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkConfigBuilder.java
bgeng777: Nice catch. Fixed.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
bgeng777: Agree. Fixed.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkConfigBuilder.java
bgeng777: I considered this solution as well. I am not sure if we will use other fields of the meta in the future. One example could be `annotations`: I notice that in our flink's `KubernetesConfigOptions`, there are option like `JOB_MANAGER_ANNOTATIONS` and `TASK_MANAGER_ANNOTATIONS`. If we have plan to support such fields in our cr, we may need to add back the `ObjectMeta`. WDYT?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
Aitozi: Oh, I ignored it

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkConfigBuilder.java
Aitozi: Get it

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
```diff
@@ -104,9 +105,31 @@ public DeleteControl cleanup(FlinkDeployment flinkApp, Context context) {
...
         LOG.info("Starting reconciliation");
+        FlinkDeployment originalCopy = ReconciliationUtils.clone(flinkApp);
+        FlinkDeploymentSpec lastReconciledSpec =
+                flinkApp.getStatus().getReconciliationStatus().getLastReconciledSpec();
+        if (lastReconciledSpec != null) {
+            Configuration lastValidatedConfig =
+                    FlinkUtils.getEffectiveConfig(
+                            flinkApp.getMetadata(),
+                            lastReconciledSpec,
+                            defaultConfig.getFlinkConfig());
```
gyfora, line 117: Maybe this whole logic could be embedded inside the observer. 
That would change it to simply `.observe(flinkApp, context)`

It would also simplify the controller loop which is very important

gyfora (changes requested):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
gyfora: I think it also fits better into the Observer because using the correct config is actually part of the observation logic

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
```diff
@@ -104,9 +105,31 @@ public DeleteControl cleanup(FlinkDeployment flinkApp, Context context) {
...
+            try {
+                observerFactory
+                        .getOrCreate(flinkApp)
+                        .observe(flinkApp, context, lastValidatedConfig);
+            } catch (DeploymentFailedException dfe) {
```
gyfora, line 122: We could also keep everything inside the original try-catch block and just do the validation there also. Maybe that will make it look nicer :)

bgeng777:
@gyfora when following your excellent suggestion, I notice that there are many duplicated lines in different `observe()` methods in our `JobObserver` and `SessionObserver`. The only difference is what to observe when the cluster is ready. It looks like we can abstract some common methods in `BaseObserver` like:
```java
public void observe(FlinkDeployment flinkApp, Context context, Configuration defaultConfig) {
    FlinkDeploymentSpec lastReconciledSpec = ...
    // Nothing has been launched so skip observing
    if (lastReconciledSpec == null) {
        return;
    }
  
    Configuration lastValidatedConfig = ....
    if (!isClusterReady(flinkApp)) {
        observeJmDeployment(flinkApp, context, lastValidatedConfig);
    }
    if (isClusterReady(flinkApp)) {
        observeWhenClusterReady();
    }
    clearErrorsIfJobManagerDeploymentNotInErrorStatus(flinkApp);
  }
```
And then implement customized `observeWhenClusterReady()` in `JobObserver` and `SessionObserver`. WDYT?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobObserver.java
```diff
@@ -46,14 +48,24 @@ public JobObserver(
     }
 
     @Override
-    public void observe(FlinkDeployment flinkApp, Context context, Configuration effectiveConfig) {
+    public void observe(FlinkDeployment flinkApp, Context context, Configuration defaultConfig) {
```
gyfora, line 51: I suggest we not pass the defaultConfig to the `observe` method but pass the `DefaultConfig` to the Observer constructor directly. This way we can avoid accidentally using the wrong config.

gyfora:
@bgeng777 I agree with the suggestion to move the shared logic to the BaseObserver. As long as we don't change the current behaviour I think this is a good time to simplify it as much as we can by factoring out the common bits to the BaseObserver

bgeng777:
@gyfora thanks for your advice. PTAK at the latest commit, which does 2 things:
1. Add a new parameter in the constructor of `BaseObserver` and remove the `effectiveConfig` parameter in the `Observer#observe()` interface. I did not directly use `DefaultConfig` class as it would introduce extra construction of `FlinkOperatorConfiguration`. Besides, `operatorConfiguration` and `flinkConfig` are for different purposes and such modification can increase the readability.
2. Follow above discusstion to factoring out common codes to the `BaseObserver` and adjust tests.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentControllerTest.java
```diff
@@ -397,7 +397,45 @@ public void testUpgradeNotReadyCluster() {
         testUpgradeNotReadyCluster(appCluster, false);
     }
 
-    public void testUpgradeNotReadyCluster(FlinkDeployment appCluster, boolean allowUpgrade) {
+    @Test
+    public void verifyReconcileWithBadConfig() throws Exception {
```
wangyang0918, line 401: This test does not cover that observer should use previous validated config, not the current invalid one. Right?

gyfora (changes requested):

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentControllerTest.java
bgeng777: You are right. Current test only verifies the order of `observe()` and `validate()`.
I enhanced the test to verify the previous validated config is really used by modifying the rest port.

bgeng777:
@wangyang0918 @gyfora Would you mind taking another look?

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentControllerTest.java
```diff
@@ -397,7 +402,59 @@ public void testUpgradeNotReadyCluster() {
...
+                () -> {
+                    testController.reconcile(appCluster, context);
+                });
+        assertEquals(
+                JobManagerDeploymentStatus.READY,
+                appCluster.getStatus().getJobManagerDeploymentStatus());
+        // Verify the saved rest port in lastReconciledSpec is actually used in observe().
+        // Otherwise, the NaN rest port
+        // would lead to exception
+        appCluster.getSpec().getFlinkConfiguration().put(RestOptions.PORT.key(), "NaN");
```
wangyang0918, line 444: I am afraid we still not verify the observer is using last reconciled config. I suggest to do it in the `TestingFlinkService#listJobs`. To achieve that, we could set a consumer of `TestingFlinkService`, just like `TestingClusterClient`.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentControllerTest.java
```diff
@@ -397,7 +402,59 @@ public void testUpgradeNotReadyCluster() {
...
+                appCluster.getStatus().getReconciliationStatus().getError());
+        assertTrue(updateControl.isUpdateStatus());
+        assertEquals(
+                JobManagerDeploymentStatus.DEPLOYED_NOT_READY,
+                appCluster.getStatus().getJobManagerDeploymentStatus());
+
+        // Check the exception
+        appCluster.getSpec().getJobManager().setReplicas(1);
+        appCluster.getSpec().getJob().setJarURI(null);
+        assertDoesNotThrow(
```
wangyang0918, line 434: What's the purpose of this assert?

bgeng777: It is for @gyfora's latest [comment](https://github.com/apache/flink-kubernetes-operator/pull/131#discussion_r839643933) to make sure  we first do validation and then get the effective config.

wangyang0918: I think the test will also fail without `assertDoesNotThrow`. Right?

wangyang0918 (approved):
+1 I am giving the pass now.

gyfora (changes requested):
Looks good, one minor thing we still need to change is in the:

`org.apache.flink.kubernetes.operator.config.Mode` logic.

We need to make sure to get the mode based on the last-reconciled spec if there is one. We do not allow mode switching but this is only enforced in the validation phase.

bgeng777:
> Looks good, one minor thing we still need to change is in the:
> 
> `org.apache.flink.kubernetes.operator.config.Mode` logic.
> 
> We need to make sure to get the mode based on the last-reconciled spec if there is one. We do not allow mode switching but this is only enforced in the validation phase.

@gyfora My bad for ignoring the `Mode` when fetching an `Observer` instance. I have updated the `getMode(flinkApp)` to make sure it returns correct mode when last-reconciled spec exists.
Also add a dedicated test to verify the `Mode` is selected correctly for observation when operator mode is changed wrongly.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/Mode.java
```diff
@@ -19,13 +19,33 @@
...
-        return flinkApp.getSpec().getJob() != null ? APPLICATION : SESSION;
+        // Try to use lastReconciledSpec if it exists.
+        // The mode derived from last-reconciled spec or current spec should be same.
+        // If they are different, observation phase will use last-reconciled spec and validation
+        // phase will fail.
+        FlinkDeploymentSpec lastReconciledSpec =
+                flinkApp.getStatus().getReconciliationStatus().getLastReconciledSpec();
+        return lastReconciledSpec == null
+                ? getMode(flinkApp.getSpec())
+                : getMode(flinkApp.getSpec());
```
gyfora, line 45: should this be `getMode(lastReconciledSpec)` here ? Maybe something is wrong with the test you added :)

gyfora: no worries, we have code review for a reason :)

PR #133, opened 2022-03-30
[FLINK-26916] Remove job graph from K8s HA ConfigMaps when shutting down cluster

wangyang0918:
After shutting down the cluster with HA ConfigMap retained, we could simply delete the JobGraph information in the ConfigMap. When the FlinkDeployment is resumed, the EmbeddedExecutor will generate a new JobGraph, which picks up the latest job config changes, and then submit the job. It could also recover from the latest checkpoint.

This is a short-term but not temporary hack and it could work both for 1.14 and 1.15.


wangyang0918:
cc @gyfora

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
```diff
@@ -202,14 +206,18 @@ public boolean isJobManagerPortReady(Configuration config) {
...
                     throw new RuntimeException("Unsupported upgrade mode " + upgradeMode);
             }
         }
-        FlinkUtils.waitForClusterShutdown(kubernetesClient, conf);
+        FlinkUtils.waitForClusterShutdown(kubernetesClient, namespace, clusterId);
+        if (HighAvailabilityMode.isHighAvailabilityModeActivated(conf)) {
+            // Delete the job graph in the HA ConfigMaps so that the newly changed job config(e.g.
+            // parallelism) could take effect
+            FlinkUtils.deleteJobGraphInKubernetesHA(clusterId, namespace, kubernetesClient);
+        }
```
gyfora, line 220: Wouldn't it be more safe to execute this logic before deployment? Technically we could have an error after shutting down the job but before deleting the jobgraph.

wangyang0918: I think your suggestion makes sense. I will update the PR.

wangyang0918:
Pushed a `fixup` to address the comments.

e2e-tests/test_last_state_upgrade.sh
```diff
@@ -44,6 +44,17 @@ function wait_for_jobmanager_running() {
     wait_for_logs $jm_pod_name "Rest endpoint listening at" ${TIMEOUT} || exit 1
 }
 
+function assert_available_slots() {
+  expected=$1
+  ip=$(minikube ip)
+  actual=$(curl http://$ip/default/${CLUSTER_ID}/overview 2>/dev/null | grep -E -o '"slots-available":[0-9]+' | awk -F':' '{print $2}')
```
gyfora, line 50: wouldnt slots available depend on TM config only? that used to work even before I think

wangyang0918: I do not think so. If the parallelism is configured 2, we will get 0 available slot after the job is running.

gyfora: ah okay, I misunderstood the check. good point

gyfora (approved):
Thank you, this is a good fix. Please merge this to main and backpot to the release branch :)

e2e-tests/test_last_state_upgrade.sh
wangyang0918: I think you mean the "slots-total" :)

gyfora: yep

PR #138, opened 2022-03-30
[FLINK-26924] Rename docker image and helm chart to flink-kubernetes-operator

mbalassi:
I aimed to update all externally released mentions of `flink-operator` to `flink-kubernetes-operator` 

I left references to it within the helm chart itself, for example the k8s role used by the operator is still called `flink-operator` by default, templates are still prefixed with `flink-operator` etc. 
Another area where I did leave `flink-operator` is in our JUnit tests, for example for test namespaces.

Let me know if you prefer deeper with this effort, in these cases I felt that these mentions are either explicitly internal or only exists within the k8s cluster, explicitly linked to the k8s operator.


.github/workflows/docker_push.yml
```diff
@@ -54,7 +54,7 @@ jobs:
         uses: docker/metadata-action@v3
         with:
           images: |
-            ghcr.io/${{ github.repository_owner }}/flink-operator
+            ghcr.io/${{ github.repository }}
```
gyfora, line 57: why was /flink-operator removed here?

gyfora (approved):
Seems to be good. Had one minor question, and I will test this locally

.github/workflows/docker_push.yml
mbalassi: `github.repository_owner` == `apache`
`github.repository` == `apache/flink-kubernetes-operator`

In short now that the image name equals the repo name I can write this shorter.

PR #139, opened 2022-03-30
[FLINK-26893] Validate checkpoint config with last-state upgrade mode

SteNicholas:
When last-state upgrade mode is selected, the validation whether checkpointing is turned on (in addition to the HA settings) should be considered.

**The brief change log**
- `DefaultDeploymentValidator` adds the validation whether checkpointing is turned on when upgrade mode is `LAST_STATE`.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkConfigBuilder.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkConfigBuilder.java
@@ -30,6 +30,8 @@ import org.apache.flink.kubernetes.configuration.KubernetesDeploymentTarget;
 import org.apache.flink.kubernetes.operator.crd.FlinkDeployment;
 import org.apache.flink.kubernetes.operator.crd.spec.FlinkDeploymentSpec;
 import org.apache.flink.kubernetes.operator.crd.spec.Resource;
+import org.apache.flink.kubernetes.operator.crd.spec.UpgradeMode;
+import org.apache.flink.streaming.api.environment.ExecutionCheckpointingOptions;
 import org.apache.flink.util.StringUtils;
 
 import io.fabric8.kubernetes.api.model.ObjectMeta;
@@ -41,6 +43,7 @@ import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.nio.file.Files;
+import java.time.Duration;
 import java.util.Collections;
 
 import static org.apache.flink.configuration.DeploymentOptionsInternal.CONF_DIR;
@@ -93,6 +96,17 @@ public class FlinkConfigBuilder {
             effectiveConfig.set(
                     REST_SERVICE_EXPOSED_TYPE,
                     KubernetesConfigOptions.ServiceExposedType.ClusterIP);
+        }
+
+        // With last-state upgrade mode, set the default value of 'execution.checkpointing.interval'
+        // to 5 minutes when HA is enabled.
+        if (spec.getJob() != null
+                && spec.getJob().getUpgradeMode() == UpgradeMode.LAST_STATE
+                && FlinkUtils.isKubernetesHAActivated(effectiveConfig)
+                && !effectiveConfig.contains(
+                        ExecutionCheckpointingOptions.CHECKPOINTING_INTERVAL)) {
+            effectiveConfig.set(
+                    ExecutionCheckpointingOptions.CHECKPOINTING_INTERVAL, Duration.ofMinutes(5));
         }
         return this;
     }
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkConfigBuilder.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkConfigBuilder.java
@@ -102,7 +102,6 @@ public class FlinkConfigBuilder {
         // to 5 minutes when HA is enabled.
         if (spec.getJob() != null
                 && spec.getJob().getUpgradeMode() == UpgradeMode.LAST_STATE
-                && FlinkUtils.isKubernetesHAActivated(effectiveConfig)
                 && !effectiveConfig.contains(
                         ExecutionCheckpointingOptions.CHECKPOINTING_INTERVAL)) {
             effectiveConfig.set(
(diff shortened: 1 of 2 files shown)


SteNicholas:
@wangyang0918 @gyfora, could you help to review?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultDeploymentValidator.java
```diff
@@ -145,10 +146,29 @@
...
+                return Optional.of(
+                        "Job could not be upgraded with last-state while Kubernetes HA disabled");
+            } else if (!configuration.containsKey(
+                            ExecutionCheckpointingOptions.CHECKPOINTING_INTERVAL.key())
+                    || configuration
+                                    .get(ExecutionCheckpointingOptions.CHECKPOINTING_INTERVAL)
+                                    .compareTo(
+                                            configuration.get(
+                                                    ExecutionCheckpointingOptions
+                                                            .MIN_PAUSE_BETWEEN_CHECKPOINTS))
```
gyfora, line 160: I dont really understant the point of this check. Is this necessary?

wangyang0918:
I believe many jobs are enabling the checkpoint via `StreamExecutionEnvironment.enableCheckpointing()`. This strict check will not work for them.

gyfora:
@wangyang0918 that's true unfortunately

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultDeploymentValidator.java
SteNicholas: @gyfora, this check refers to `CheckpointConfig#setCheckpointInterval` and is necessary. @wangyang0918 WDYT?

wangyang0918:
> for the jobs that enable the checkpoint via StreamExecutionEnvironment.enableCheckpointing(), we could guide the user to use the configuration method to open checkpoint.

This will not work for session clusters since each job has different checkpoint interval.


>  in Flink, when calling the StreamExecutionEnvironment.enableCheckpointing(), the execution.checkpointing.interval should be set with value of the interval for enableCheckpointing().

This is impossible to set the value to flink config in the CR.


What I am thinking is to add a default value for `execution.checkpointing.interval` when HA enabled. It could be overridden by user specified CR or `StreamExecutionEnvironment.enableCheckpointing()`. Then we do not force users should set it explicitly in the CR.

@SteNicholas @gyfora WDYT?

SteNicholas:
> What I am thinking is to add a default value for `execution.checkpointing.interval` when HA enabled. It could be overridden by user specified CR or StreamExecutionEnvironment.enableCheckpointing(). Then we do not force users should set it explicitly in the CR.

@wangyang0918, this makes sense to me. I will update the implementation to add a default value for `execution.checkpointing.interval` when HA enabled. But what's the recommend default value for `execution.checkpointing.interval`?

wangyang0918:
@gyfora Does it make sense to you that adding a default value(maybe 120s) for `execution.checkpointing.interval` when last-state upgrade mode is used. It could be overridden by users via explicit config options or programmatically.

gyfora:
@wangyang0918 I would probably go for a more conservative 5 or 10 minute interval maybe

wangyang0918:
> I would probably go for a more conservative 5 or 10 minute interval maybe

Make sense. @SteNicholas Would you like to update this PR accordingly?

SteNicholas:
> > I would probably go for a more conservative 5 or 10 minute interval maybe
> 
> Make sense. @SteNicholas Would you like to update this PR accordingly?

@wangyang0918, I'm updating this PR according to above suggestion.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
```diff
@@ -114,6 +115,10 @@ public DeleteControl cleanup(FlinkDeployment flinkApp, Context context) {
                 ReconciliationUtils.updateForReconciliationError(flinkApp, validationError.get());
                 return ReconciliationUtils.toUpdateControl(
                         operatorConfiguration, originalCopy, flinkApp, false);
+            } else if (flinkApp.getSpec()
+                    .getFlinkConfiguration()
+                    .containsKey(ExecutionCheckpointingOptions.CHECKPOINTING_INTERVAL.key())) {
+                FlinkUtils.setCheckpointingInterval(originalCopy.getSpec().getFlinkConfiguration());
```
gyfora, line 121: We should simply add this logic to the FlinkConfigBuilder instead of modifying the spec

SteNicholas:
@gyfora, thanks for your review. I have addressed your comments for `FlinkConfigBuilder`. Please help to take a look.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkConfigBuilder.java
```diff
@@ -94,6 +97,17 @@ public FlinkConfigBuilder applyFlinkConfiguration() {
                     REST_SERVICE_EXPOSED_TYPE,
                     KubernetesConfigOptions.ServiceExposedType.ClusterIP);
         }
+
+        // With last-state upgrade mode, set the default value of 'execution.checkpointing.interval'
+        // to 5 minutes when HA is enabled.
+        if (spec.getJob() != null
+                && spec.getJob().getUpgradeMode() == UpgradeMode.LAST_STATE
+                && FlinkUtils.isKubernetesHAActivated(effectiveConfig)
```
gyfora, line 105: The HA config is already validated for last-state mode by the DeploymentValidator, no need to check again here

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkConfigBuilder.java
```diff
@@ -94,6 +97,16 @@ public FlinkConfigBuilder applyFlinkConfiguration() {
...
         }
+
+        // With last-state upgrade mode, set the default value of 'execution.checkpointing.interval'
+        // to 5 minutes when HA is enabled.
+        if (spec.getJob() != null
+                && spec.getJob().getUpgradeMode() == UpgradeMode.LAST_STATE
+                && !effectiveConfig.contains(
+                        ExecutionCheckpointingOptions.CHECKPOINTING_INTERVAL)) {
+            effectiveConfig.set(
+                    ExecutionCheckpointingOptions.CHECKPOINTING_INTERVAL, Duration.ofMinutes(5));
```
gyfora, line 108: Let's move the 5 minutes into a constant in FlinkConfigBuilder and I think we are good to go :)

gyfora (changes requested):
Left one last small comment, otherwise looks good

PR #143, opened 2022-04-01
[FLINK-26959] Simplify quickstart

gyfora:
This should be merged after the 0.1.0 release

### DIFF
--- a/docs/content/docs/try-flink-kubernetes-operator/quick-start.md
+++ b/docs/content/docs/try-flink-kubernetes-operator/quick-start.md
@@ -26,8 +26,8 @@ under the License.
 
 # Quick Start
 
-This document provides a quick introduction to using the Flink Kubernetes Operator. Readers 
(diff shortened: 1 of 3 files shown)


docs/content/docs/try-flink-kubernetes-operator/quick-start.md
```diff
@@ -123,4 +118,10 @@ To expose the Flink Dashboard you may add a port-forward rule or start a tunnel:
 minikube tunnel
```
wangyang0918, line 118: `minikube tunnel` could not work now because we are using `ClusterIP` for exposing the rest service. I am suggesting to use `kubectl port-forward svc/basic-example-rest 8081`.
Also we could refer to the ingress[1] solution here.

[1]. https://nightlies.apache.org/flink/flink-kubernetes-operator-docs-main/docs/operations/ingress/

gyfora: good catch, fixed

PR #145, opened 2022-04-01
[FLINK-26996] Break the reconcile after first create session cluster

Aitozi:
When I test session cluster, I found that it will always start twice for the session cluster. I think this fix should be included in the release-0.1

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/SessionReconciler.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/SessionReconciler.java
@@ -57,13 +57,14 @@ public class SessionReconciler extends BaseReconciler {
             flinkApp.getStatus()
                     .setJobManagerDeploymentStatus(JobManagerDeploymentStatus.DEPLOYING);
             IngressUtils.updateIngressRules(flinkApp, effectiveConfig, kubernetesClient);
+        } else {
+            boolean specChanged = !flinkApp.getSpec().equals(lastReconciledSpec);
+            if (specChanged) {
+                upgradeSessionCluster(flinkApp, effectiveConfig);
+                IngressUtils.updateIngressRules(flinkApp, effectiveConfig, kubernetesClient);
+            }
         }
 
-        boolean specChanged = !flinkApp.getSpec().equals(lastReconciledSpec);
-        if (specChanged) {
-            upgradeSessionCluster(flinkApp, effectiveConfig);
-            IngressUtils.updateIngressRules(flinkApp, effectiveConfig, kubernetesClient);
-        }
         ReconciliationUtils.updateForSpecReconciliationSuccess(flinkApp, null);
     }
 
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/SessionReconciler.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/SessionReconciler.java
@@ -57,12 +57,14 @@ public class SessionReconciler extends BaseReconciler {
             flinkApp.getStatus()
                     .setJobManagerDeploymentStatus(JobManagerDeploymentStatus.DEPLOYING);
             IngressUtils.updateIngressRules(flinkApp, effectiveConfig, kubernetesClient);
-        } else {
-            boolean specChanged = !flinkApp.getSpec().equals(lastReconciledSpec);
-            if (specChanged) {
-                upgradeSessionCluster(flinkApp, effectiveConfig);
-                IngressUtils.updateIngressRules(flinkApp, effectiveConfig, kubernetesClient);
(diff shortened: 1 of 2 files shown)


Aitozi:
pls take a look cc @gyfora

gyfora:
I will check this when I get home. What happens when it is submitted the second time? Will we get an error or 2 clusters?

Aitozi:
> I will check this when I get home. What happens when it is submitted the second time? Will we get an error or 2 clusters?

It will trigger an upgade after creating session cluster.Because the current spec always not equal to lastReconcileSpec which is null

gyfora:
Thanks I understand now. It's an unfortunate bug that delays session deployments, I would not consider it a release blocker at this stage. @wangyang0918 what do you think? We can still backport this to the release-0.1 branch and if we ever have a 0.1.1 release, included it there.

Could you also please add a test case to guard this behavior?

Aitozi:
A test is added to guard this and also create a ticket for it. 
Since in release-0.1 the functional of session cluster is not complete, I'm OK not consider it as a release blocker.

wangyang0918:
@Aitozi @gyfora I also just forgot to verify the session cluster. I lean to not make this as the blocker of release-0.1. Because the session cluster could eventually start up and v0.1 is a preview release. I just want to release the next major version 0.2 asap, which will include a fully functional session cluster support. Maybe also include the metrics improvements, and some feedback from the users.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/SessionReconciler.java
```diff
@@ -57,13 +57,14 @@ public void reconcile(FlinkDeployment flinkApp, Context context, Configuration e
             flinkApp.getStatus()
                     .setJobManagerDeploymentStatus(JobManagerDeploymentStatus.DEPLOYING);
             IngressUtils.updateIngressRules(flinkApp, effectiveConfig, kubernetesClient);
+        } else {
```
wangyang0918, line 60: Compared with else, I prefer a `return` at the end if `if...`

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/reconciler/SessionReconcilerTest.java
```diff
@@ -0,0 +1,59 @@
...
+/** Tests for {@link SessionReconciler}. */
+public class SessionReconcilerTest {
+
+    private final FlinkOperatorConfiguration operatorConfiguration =
+            FlinkOperatorConfiguration.fromConfiguration(new Configuration());
+
+    @Test
+    public void testStartSession() throws Exception {
+        Context context = TestUtils.createEmptyContext();
+        var count = new AtomicInteger();
```
wangyang0918, line 42: We could set the initial value to `0` explicitly.

```
AtomicInteger count = new AtomicInteger(0);
```

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/reconciler/SessionReconcilerTest.java
```diff
@@ -0,0 +1,59 @@
...
+                        super.submitSessionCluster(deployment, conf);
+                        count.addAndGet(1);
+                    }
+                };
+
+        SessionReconciler reconciler =
+                new SessionReconciler(null, flinkService, operatorConfiguration);
+        FlinkDeployment deployment = TestUtils.buildSessionCluster();
+        reconciler.reconcile(deployment, context, new Configuration());
+        assertEquals(count.get(), 1);
```
wangyang0918, line 57: The first argument of `assertEquals` is expected value. So this should be `assertEquals(1, count.get());`

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/reconciler/SessionReconcilerTest.java
Aitozi: Fixed

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/reconciler/SessionReconcilerTest.java
Aitozi: thanks, fixed

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/SessionReconciler.java
Aitozi: Fixed

wangyang0918 (approved):
+1

gyfora:
Merging this @Aitozi please prepare the backport pr for `release-1.0`

PR #149, opened 2022-04-02
[FLINK-27012] Caching maven dependencies to speed up workflows in github actions

wangyang0918:
Using [cache](https://github.com/marketplace/actions/cache) to speed up workflows.


mbalassi (approved):
Thanks, LGTM

PR #150, opened 2022-04-02
[FLINK-26611] Document operator config options

bgeng777:
- Add `Deployment` -> `Configuration` tab in the doc website. This page can be used to document other configs as well. (e.g. https://issues.apache.org/jira/browse/FLINK-27008)
- Add document of operator config options.

### DIFF
--- a/docs/content/docs/operations/configuration.md
+++ b/docs/content/docs/operations/configuration.md
@@ -1,0 +1,38 @@
+---
+title: "Configuration"
+weight: 2
+type: docs
+aliases:
+- /operations/configuration.html
+---
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+# Configuration
+
+## Operator
+
+| Key  | Default | Type | Description |
+| ---- | ------- | ---- | ----------- |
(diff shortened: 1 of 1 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/OperatorConfigOptions.java
```diff
@@ -33,7 +33,7 @@
                     .durationType()
                     .defaultValue(Duration.ofSeconds(60))
                     .withDescription(
-                            "The interval for the controller to reschedule the reconcile process");
+                            "The interval for the controller to reschedule the reconcile process.");
```
SteNicholas, line 36: I have updated this in #141.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/OperatorConfigOptions.java
```diff
@@ -61,7 +61,7 @@
                     .durationType()
                     .defaultValue(Duration.ofSeconds(10))
                     .withDescription(
-                            "The interval before a savepoint trigger attempt is marked as unsuccessful");
+                            "The interval before a savepoint trigger attempt is marked as unsuccessful.");
```
SteNicholas, line 64: Ditto.

docs/content/docs/deployment/configuration.md
```diff
@@ -0,0 +1,38 @@
+---
```
SteNicholas, line 1: Could the document of configuration be generated by `OperatorConfigOptions`?

gyfora: Not very easy to do unfortunately. flink-docs contains the annotaiton processing logic for flink but it's not flexible

gyfora (changes requested):
The changes look good but I would prefer to have them in the `operations` section for now.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/OperatorConfigOptions.java
bgeng777: Thanks for the information. As we do not automatically generate documents for now, I will roll back this change and only keep it in our doc website.

bgeng777:
> The changes look good but I would prefer to have them in the `operations` section for now.

I am ok to both. Previous design follows the pattern of flink doc. Now, only a `configuration` page is added in `operations` sec.

gyfora:
Merging this, @bgeng777 please open a backport PR for the `release-0.1` branch

PR #151, opened 2022-04-02
[BP][FLINK-26611] Document operator config options

bgeng777:
- Backport #150 for [FLINK-26611] Document operator config options


bgeng777:
cc @gyfora

PR #152, opened 2022-04-02
[BP][FLINK-26996] Break the reconcile after first create session cluster

Aitozi:
Bp for https://github.com/apache/flink-kubernetes-operator/pull/145


Aitozi:
cc @gyfora

PR #153, opened 2022-04-03
[FLINK-27000] Support to set JVM args for operator

Aitozi:
Provide a way to set jvm args and also print the start up options using the flink tools

---

By this we can also easily remote debug in the minikube environment 

```
helm install flink-operator helm/flink-kubernetes-operator  --set image.repository=aitozi/flink-java-operator --set image.tag=latest --set jvmArgs.operator='-agentlib:jdwp=transport=dt_socket\,server=y\,suspend=y\,address=*:5005'

kubectl port-forward {operator-pod} 5005:5005
```

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -27,12 +27,12 @@ if [ "$1" = "help" ]; then
 elif [ "$1" = "operator" ]; then
     echo "Starting Operator"
 
-    exec java -cp /$FLINK_KUBERNETES_SHADED_JAR:/$OPERATOR_JAR $LOG_CONFIG org.apache.flink.kubernetes.operator.FlinkOperator
+    exec java $JVM_ARGS -cp /$FLINK_KUBERNETES_SHADED_JAR:/$OPERATOR_JAR $LOG_CONFIG org.apache.flink.kubernetes.operator.FlinkOperator
 elif [ "$1" = "webhook" ]; then
     echo "Starting Webhook"
 
     # Adds the operator shaded jar on the classpath when the webhook starts
-    exec java -cp /$FLINK_KUBERNETES_SHADED_JAR:/$OPERATOR_JAR:/$WEBHOOK_JAR $LOG_CONFIG org.apache.flink.kubernetes.operator.admission.FlinkOperatorWebhook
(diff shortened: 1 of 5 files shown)


docker-entrypoint.sh
```diff
@@ -27,12 +27,12 @@ if [ "$1" = "help" ]; then
 elif [ "$1" = "operator" ]; then
     echo "Starting Operator"
 
-    exec java -cp /$FLINK_KUBERNETES_SHADED_JAR:/$OPERATOR_JAR $LOG_CONFIG org.apache.flink.kubernetes.operator.FlinkOperator
+    exec java $JVM_ARGS -cp /$FLINK_KUBERNETES_SHADED_JAR:/$OPERATOR_JAR $LOG_CONFIG org.apache.flink.kubernetes.operator.FlinkOperator
```
gyfora, line 30: I think we should put the JVM_ARGS after the log config, to give the user an opportunity of overwriting every setting

helm/flink-kubernetes-operator/values.yaml
```diff
@@ -78,3 +78,8 @@ metrics:
 imagePullSecrets: []
 nameOverride: ""
 fullnameOverride: ""
+
+# Set the jvm start up options for webhook and operator
+jvmArgs:
+  webhook: ""
+  operator: ""
```
gyfora, line 85: This looks pretty awkward. We should start structuring the helm options a bit more. 
I will probably open a ticket for this.

gyfora (changes requested):

docker-entrypoint.sh
Aitozi: Good point, fixed

helm/flink-kubernetes-operator/values.yaml
Aitozi: I also feel the helm option is a little mess now 😄, do you have some suggestion for this ? Or let it be improved in your ticket?

gyfora: I will try to work on this in another ticket, we need to clean up many things . It’s ok as it is

PR #161, opened 2022-04-08
[FLINK-27141] Improve FlinkService#waitForClusterShutdown logic

SteNicholas:
The current `waitForClusterShutdown` logic has a hardcoded 60 second timeout and prints a log message after every second, which should be improved to make the max timeout configurable and only log a message once every 5-10 seconds to make this more reasonable.

**The brief change log**

- `waitForClusterShutdown` adds the max timeout configuration `operator.reconciler.flink.shutdown.cluster.timeout` and logs a message once every 5 seconds.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/OperatorConfigOptions.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/OperatorConfigOptions.java
@@ -70,9 +70,17 @@ public class OperatorConfigOptions {
                     .withDescription(
                             "The timeout for the observer to wait the flink rest client to return.");
 
-    public static final ConfigOption<Duration> OPERATOR_CANCEL_JOB_TIMEOUT =
+    public static final ConfigOption<Duration> OPERATOR_RECONCILER_FLINK_CANCEL_JOB_TIMEOUT =
             ConfigOptions.key("operator.reconciler.flink.cancel.job.timeout")
                     .durationType()
                     .defaultValue(Duration.ofMinutes(1))
-                    .withDescription("The timeout for the operator to cancel job.");
+                    .withDescription(
+                            "The timeout for the reconciler to wait for flink to cancel job.");
+
+    public static final ConfigOption<Duration> OPERATOR_RECONCILER_FLINK_SHUTDOWN_CLUSTER_TIMEOUT =
+            ConfigOptions.key("operator.reconciler.flink.shutdown.cluster.timeout")
+                    .durationType()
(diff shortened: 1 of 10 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/OperatorConfigOptions.java
```diff
@@ -70,9 +70,17 @@
...
+    public static final ConfigOption<Duration> OPERATOR_RECONCILER_FLINK_CANCEL_JOB_TIMEOUT =
             ConfigOptions.key("operator.reconciler.flink.cancel.job.timeout")
                     .durationType()
                     .defaultValue(Duration.ofMinutes(1))
-                    .withDescription("The timeout for the operator to cancel job.");
+                    .withDescription(
+                            "The timeout for the reconciler to wait for flink to cancel job.");
+
+    public static final ConfigOption<Duration> OPERATOR_RECONCILER_FLINK_SHUTDOWN_CLUSTER_TIMEOUT =
+            ConfigOptions.key("operator.reconciler.flink.shutdown.cluster.timeout")
```
gyfora, line 81: should be `operator.reconciler.flink.cluster.shutdown.timeout` instead

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkUtils.java
```diff
@@ -207,7 +213,10 @@ public static void waitForClusterShutdown(
             if (!jobManagerRunning && !serviceRunning) {
                 break;
             }
-            LOG.info("Waiting for cluster shutdown... ({})", i);
+            // log a message waiting to shutdown Flink cluster every 5 seconds.
+            if (i % 4 == 0) {
```
gyfora, line 217: Wouldn't this actually print it every 4 seconds?

SteNicholas: @gyfora , i start from 0, therefore print the log every 5 seconds.

gyfora: I think that is actually incorrect, it will print:
1, 5, 9, 13...

SteNicholas: @gyfora, sorry for this mistake. I will update this as soon as possible.

PR #163, opened 2022-04-09
[FLINK-27154] Disable web.cancel.enable for application clusters

SteNicholas:
`web.cancel.enable` should be set to false for application deployments to avoid users accidentally cancelling jobs.

**The brief change log**

- `FlinkConfigBuilder` set `web.cancel.enable` to false for application deployments.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkConfigBuilder.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkConfigBuilder.java
@@ -102,9 +102,11 @@ public class FlinkConfigBuilder {
         }
 
         if (spec.getJob() != null) {
-            // Set 'web.cancel.enable' to false for application deployments to avoid users
-            // accidentally cancelling jobs.
-            effectiveConfig.set(CANCEL_ENABLE, false);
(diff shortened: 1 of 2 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkConfigBuilder.java
```diff
@@ -100,16 +101,22 @@ public FlinkConfigBuilder applyFlinkConfiguration() {
...
-                && spec.getJob().getUpgradeMode() == UpgradeMode.LAST_STATE
-                && !effectiveConfig.contains(
-                        ExecutionCheckpointingOptions.CHECKPOINTING_INTERVAL)) {
-            effectiveConfig.set(
-                    ExecutionCheckpointingOptions.CHECKPOINTING_INTERVAL,
-                    DEFAULT_CHECKPOINTING_INTERVAL);
+        if (spec.getJob() != null) {
+            // Set 'web.cancel.enable' to false for application deployments to avoid users
+            // accidentally cancelling jobs.
+            effectiveConfig.set(CANCEL_ENABLE, false);
```
gyfora, line 107: We should not set this if the user explicitly set it to true

PR #164, opened 2022-04-11
[FLINK-26871] Handle session job spec change

Aitozi:
This PR is meant to support the upgrade/trigger savepoint for the session job. The brief change as below:

- Complete the `FlinkSessionJobReconciler` to handle spec change
- Complete the `SessionJobObserver` to observe the job status and savepoint result
- Add the `FlinkSessionJobReconcilerTest` and `SessionJobObserverTest` to verify the change.

I have done the basic verification locally to verify the upgrade with `sateless` and `savepoint` and trigger savepoint. But as I test, the `cancel` operation will always cleanup the HA data. So I have not figured out how to support `last_state` mode in session job. So I disable the `last_state` mode now.

An e2e test will be introduce in a following PR to guard the session job functional.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/sessionjob/SessionJobObserver.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/sessionjob/SessionJobObserver.java
@@ -17,13 +17,119 @@
 
 package org.apache.flink.kubernetes.operator.observer.sessionjob;
 
+import org.apache.flink.configuration.Configuration;
+import org.apache.flink.kubernetes.operator.config.FlinkOperatorConfiguration;
 import org.apache.flink.kubernetes.operator.crd.FlinkSessionJob;
+import org.apache.flink.kubernetes.operator.crd.status.JobStatus;
+import org.apache.flink.kubernetes.operator.observer.JobStatusObserver;
 import org.apache.flink.kubernetes.operator.observer.Observer;
+import org.apache.flink.kubernetes.operator.observer.SavepointObserver;
+import org.apache.flink.kubernetes.operator.observer.context.VoidObserverContext;
+import org.apache.flink.kubernetes.operator.reconciler.ReconciliationUtils;
+import org.apache.flink.kubernetes.operator.reconciler.sessionjob.SessionJobHelper;
+import org.apache.flink.kubernetes.operator.service.FlinkService;
+import org.apache.flink.kubernetes.operator.utils.FlinkUtils;
+import org.apache.flink.kubernetes.operator.utils.OperatorUtils;
+import org.apache.flink.runtime.client.JobStatusMessage;
+import org.apache.flink.util.Preconditions;
 
 import io.javaoperatorsdk.operator.api.reconciler.Context;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.List;
+import java.util.Optional;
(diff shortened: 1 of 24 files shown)


Aitozi:
cc @gyfora @wangyang0918 Please take a look when you are free.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/sessionjob/SessionJobObserver.java
```diff
@@ -17,13 +17,175 @@
...
+                    LOG.info(
+                            "Job status successfully updated from {} to {}",
+                            previousJobStatus,
+                            targetJobStatus);
+                }
+                return true;
+            } else {
+                LOG.info("No job found for JobID: {}", jobId);
+                sessionJob.getStatus().getJobStatus().setState(JOB_STATE_UNKNOWN);
+            }
```
gyfora, line 120: should we return false here?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/sessionjob/SessionJobObserver.java
```diff
@@ -17,13 +17,175 @@
...
+
+        } catch (Exception e) {
+            LOG.error("Exception while listing jobs", e);
+            sessionJob.getStatus().getJobStatus().setState(JOB_STATE_UNKNOWN);
+            return false;
+        }
+        return true;
+    }
+
+    private void observeSavepointStatus(
```
gyfora, line 130: Seems like there is a lot of code duplication here with the ApplicationObserver, I wonder if we can factor this out somehow

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/sessionjob/SessionJobObserver.java
```diff
@@ -17,13 +17,175 @@
...
+        }
+        LOG.info("Savepoint status updated with latest completed savepoint info");
+        savepointInfo.updateLastSavepoint(savepointFetchResult.getSavepoint());
+    }
+
+    /**
+     * Update previous job status based on the job list from the cluster and return the target
+     * status.
+     */
+    private String updateJobStatus(JobStatus status, List<JobStatusMessage> clusterJobStatuses) {
```
gyfora, line 179: this also looks basically duplicated from the ApplicationObserver

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/sessionjob/SessionJobObserver.java
Aitozi: Yes, I have the same feeling, I will try to improve this

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/sessionjob/SessionJobObserver.java
Aitozi: Yes

Aitozi:
@gyfora I introduced the `SavepointObserver` and `JobStatusObserver` to solve the code duplication. Please take a look again

Aitozi:
Force pushed to resolve conflicts

PR #165, opened 2022-04-11
[FLINK-26140] Support rollback strategies

gyfora:
This PR is designed to lay the groundwork for upgrade rollback strategies. It contains the following changes, improvements and new features on a high level:

**CRD**
 - Remove `error` & `success` fields from `ReconciliationStatus` and move `error` to the top level `FlinkDeploymentStatus/FlinkSessionJobStatus` class (the success flag has been removed completely)
 - Move configuration management from Controller to Reconciler
 - Introduce `reconciliationTimestamp`, `lastStableSpec` and `state` fields in `ReconciliationStatus`
 
**Configuration**
```
kubernetes.operator.upgrade.rollback.enabled: true
kubernetes.operator.upgrade.readiness.timeout: 60s
```
 
**Stable deployments and roll-back logic:**
The rollback logic relies on tracking the last stable / running deployment spec (in addition to the lastReconciledSpec). The observer decides (in the future based on the cunfigured settings) when a job is stable and updates the lastStableSpec accordingly.

If rollback is enabled, the readiness timeout defines the maximum time a new deployment can take to get into a stable / running state before it is rolled back.

**Future improvements that are not part of this PR:**
 - Allow custom stability conditions such as successful checkpoints etc.
 - Roll back before end of readiness timeout on fatal errors
 - Support rolling back sessionjob deployments

### DIFF
--- a/docs/content/docs/custom-resource/reference.md
+++ b/docs/content/docs/custom-resource/reference.md
@@ -170,6 +170,7 @@ This page serves as a full reference for FlinkDeployment custom resource definit
 | jobStatus | org.apache.flink.kubernetes.operator.crd.status.JobStatus | Last observed status of the Flink job on Application deployments. |
 | jobManagerDeploymentStatus | org.apache.flink.kubernetes.operator.crd.status.JobManagerDeploymentStatus | Last observed status of the JobManager deployment. |
 | reconciliationStatus | org.apache.flink.kubernetes.operator.crd.status.ReconciliationStatus | Status of the last reconcile operation. |
+| error | java.lang.String | Error information about the Flink deployment. |
 
 ### FlinkSessionJobReconciliationStatus
 **Class**: org.apache.flink.kubernetes.operator.crd.status.FlinkSessionJobReconciliationStatus
@@ -178,8 +179,6 @@ This page serves as a full reference for FlinkDeployment custom resource definit
 
 | Parameter | Type | Docs |
 | ----------| ---- | ---- |
-| success | boolean | True if last reconciliation step was successful. |
-| error | java.lang.String | If success == false, error information about the reconciliation failure. |
 | lastReconciledSpec | java.lang.String | Last reconciled job spec. Used to decide whether further reconciliation steps are necessary. |
 
 ### FlinkSessionJobStatus
@@ -191,6 +190,7 @@ This page serves as a full reference for FlinkDeployment custom resource definit
 | ----------| ---- | ---- |
 | jobStatus | org.apache.flink.kubernetes.operator.crd.status.JobStatus | Last observed status of the job. |
 | reconciliationStatus | org.apache.flink.kubernetes.operator.crd.status.FlinkSessionJobReconciliationStatus | Status of the last reconcile operation. |
+| error | java.lang.String | Error information about the session job. |
 
 ### JobManagerDeploymentStatus
 **Class**: org.apache.flink.kubernetes.operator.crd.status.JobManagerDeploymentStatus
@@ -219,6 +219,17 @@ This page serves as a full reference for FlinkDeployment custom resource definit
 | updateTime | java.lang.String | Update time of the job. |
 | savepointInfo | org.apache.flink.kubernetes.operator.crd.status.SavepointInfo | Information about pending and last savepoint for the job. |
 
+### ReconciliationState
+**Class**: org.apache.flink.kubernetes.operator.crd.status.ReconciliationState
+
+**Description**: Current state of the reconciliation.
+
+| Value | Docs |
+| ----- | ---- |
+| DEPLOYED | The last reconciledSpec is currently deployed. |
+| ROLLING_BACK | In the process of rolling back to the lastStableSpec. |
+| ROLLED_BACK | Rolled back to the lastStableSpec. |
+
 ### ReconciliationStatus
 **Class**: org.apache.flink.kubernetes.operator.crd.status.ReconciliationStatus
 
@@ -226,9 +237,10 @@ This page serves as a full reference for FlinkDeployment custom resource definit
 
 | Parameter | Type | Docs |
 | ----------| ---- | ---- |
-| success | boolean | True if last reconciliation step was successful. |
-| error | java.lang.String | If success == false, error information about the reconciliation failure. |
+| reconciliationTimestamp | long | Epoch timestamp of the last successful reconcile operation. |
 | lastReconciledSpec | java.lang.String | Last reconciled deployment spec. Used to decide whether further reconciliation steps are  necessary. |
+| lastStableSpec | java.lang.String | Last stable deployment spec according to the specified stability condition. If a rollback  strategy is defined this will be the target to roll back to. |
+| state | org.apache.flink.kubernetes.operator.crd.status.ReconciliationState | Deployment state of the last reconciled spec. |
 
 ### Savepoint
 **Class**: org.apache.flink.kubernetes.operator.crd.status.Savepoint
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/AbstractDeploymentObserver.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/AbstractDeploymentObserver.java
@@ -25,11 +25,12 @@ import org.apache.flink.kubernetes.operator.crd.spec.JobSpec;
 import org.apache.flink.kubernetes.operator.crd.spec.JobState;
 import org.apache.flink.kubernetes.operator.crd.status.FlinkDeploymentStatus;
 import org.apache.flink.kubernetes.operator.crd.status.JobManagerDeploymentStatus;
+import org.apache.flink.kubernetes.operator.crd.status.ReconciliationState;
 import org.apache.flink.kubernetes.operator.crd.status.ReconciliationStatus;
 import org.apache.flink.kubernetes.operator.exception.DeploymentFailedException;
 import org.apache.flink.kubernetes.operator.observer.Observer;
+import org.apache.flink.kubernetes.operator.reconciler.ReconciliationUtils;
 import org.apache.flink.kubernetes.operator.service.FlinkService;
-import org.apache.flink.kubernetes.operator.utils.FlinkUtils;
 
 import io.fabric8.kubernetes.api.model.ContainerStateWaiting;
 import io.fabric8.kubernetes.api.model.ContainerStatus;
@@ -64,6 +65,32 @@ public abstract class AbstractDeploymentObserver implements Observer<FlinkDeploy
         this.flinkService = flinkService;
         this.operatorConfiguration = operatorConfiguration;
         this.flinkConfig = flinkConfig;
+    }
+
+    @Override
+    public void observe(FlinkDeployment flinkApp, Context context) {
+        ReconciliationStatus reconciliationStatus = flinkApp.getStatus().getReconciliationStatus();
+        FlinkDeploymentSpec lastReconciledSpec =
+                reconciliationStatus.deserializeLastReconciledSpec();
+
+        // Nothing has been launched so skip observing
+        if (lastReconciledSpec == null
+                || reconciliationStatus.getState() == ReconciliationState.ROLLING_BACK) {
+            return;
+        }
+
+        Configuration observeConfig = ReconciliationUtils.getDeployedConfig(flinkApp, flinkConfig);
+        if (!isJmDeploymentReady(flinkApp)) {
+            observeJmDeployment(flinkApp, context, observeConfig);
+        }
+        if (isJmDeploymentReady(flinkApp)) {
+            if (observeFlinkCluster(flinkApp, context, observeConfig)) {
+                if (reconciliationStatus.getState() != ReconciliationState.ROLLED_BACK) {
+                    reconciliationStatus.markReconciledSpecAsStable();
+                }
+            }
+        }
+        clearErrorsIfDeploymentIsHealthy(flinkApp);
     }
 
     protected void observeJmDeployment(
@@ -159,16 +186,16 @@ public abstract class AbstractDeploymentObserver implements Observer<FlinkDeploy
         }
     }
 
-    protected boolean isClusterReady(FlinkDeployment dep) {
+    protected boolean isJmDeploymentReady(FlinkDeployment dep) {
         return dep.getStatus().getJobManagerDeploymentStatus() == JobManagerDeploymentStatus.READY;
     }
 
-    protected void clearErrorsIfJobManagerDeploymentNotInErrorStatus(FlinkDeployment dep) {
-        if (dep.getStatus().getJobManagerDeploymentStatus() != JobManagerDeploymentStatus.ERROR) {
-            final ReconciliationStatus reconciliationStatus =
-                    dep.getStatus().getReconciliationStatus();
-            reconciliationStatus.setSuccess(true);
-            reconciliationStatus.setError(null);
+    protected void clearErrorsIfDeploymentIsHealthy(FlinkDeployment dep) {
+        FlinkDeploymentStatus status = dep.getStatus();
+        ReconciliationStatus reconciliationStatus = status.getReconciliationStatus();
+        if (status.getJobManagerDeploymentStatus() != JobManagerDeploymentStatus.ERROR
+                && reconciliationStatus.isLastReconciledSpecStable()) {
+            status.setError(null);
         }
     }
 
@@ -189,34 +216,15 @@ public abstract class AbstractDeploymentObserver implements Observer<FlinkDeploy
                 && lastReconciledSpec.getJob().getState() == JobState.SUSPENDED;
     }
 
-    public void observe(FlinkDeployment flinkApp, Context context) {
-        FlinkDeploymentSpec lastReconciledSpec =
-                flinkApp.getStatus().getReconciliationStatus().deserializeLastReconciledSpec();
-        // Nothing has been launched so skip observing
-        if (lastReconciledSpec == null) {
-            return;
-        }
-
-        Configuration lastValidatedConfig =
-                FlinkUtils.getEffectiveConfig(
-                        flinkApp.getMetadata(), lastReconciledSpec, this.flinkConfig);
-        if (!isClusterReady(flinkApp)) {
-            observeJmDeployment(flinkApp, context, lastValidatedConfig);
-        }
-        if (isClusterReady(flinkApp)) {
-            observeIfClusterReady(flinkApp, context, lastValidatedConfig);
-        }
-        clearErrorsIfJobManagerDeploymentNotInErrorStatus(flinkApp);
-    }
-
     /**
      * Observe the flinkApp status when the cluster is ready. It will be implemented by child class
      * to reflect the changed status on the flinkApp resource.
      *
      * @param flinkApp the target flinkDeployment resource
      * @param context the context with which the operation is executed
-     * @param lastValidatedConfig the last validated config
+     * @param deployedConfig config that is deployed on the Flink cluster
+     * @return true if cluster state is stable
      */
-    public abstract void observeIfClusterReady(
-            FlinkDeployment flinkApp, Context context, Configuration lastValidatedConfig);
+    protected abstract boolean observeFlinkCluster(
+            FlinkDeployment flinkApp, Context context, Configuration deployedConfig);
 }
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/ApplicationObserver.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/ApplicationObserver.java
@@ -49,7 +49,7 @@ public class ApplicationObserver extends AbstractDeploymentObserver {
                 new JobStatusObserver<>(flinkService) {
                     @Override
                     public void onTimeout(ApplicationObserverContext ctx) {
-                        observeJmDeployment(ctx.flinkApp, ctx.context, ctx.lastValidatedConfig);
+                        observeJmDeployment(ctx.flinkApp, ctx.context, ctx.deployedConfig);
                     }
 
                     @Override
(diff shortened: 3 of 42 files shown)


gyfora:
Opening this for review/discussion, I will work on adding tests for the core logic.

cc @wangyang0918 @tweise @Aitozi

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
```diff
@@ -111,12 +114,34 @@ public void reconcile(FlinkDeployment flinkApp, Context context, Configuration e
...
+        deployFlinkJob(
+                flinkApp,
+                rollbackConfig,
+                Optional.ofNullable(
+                                flinkApp.getStatus()
+                                        .getJobStatus()
+                                        .getSavepointInfo()
+                                        .getLastSavepoint())
+                        .map(Savepoint::getLocation));
+        flinkApp.getStatus().getReconciliationStatus().setRolledBack(true);
```
Aitozi, line 142: Do we have to update the `lastReconciledSpec` to `lastStableSpec` after rollback?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
```diff
@@ -111,12 +114,34 @@ public void reconcile(FlinkDeployment flinkApp, Context context, Configuration e
...
 
+    private void rollbackApplication(FlinkDeployment flinkApp) throws Exception {
+        LOG.warn("Rolling back deployment to last stable spec");
+        Configuration rollbackConfig =
+                FlinkUtils.getEffectiveConfig(
+                        flinkApp.getMetadata(),
+                        flinkApp.getStatus().getReconciliationStatus().deserializeLastStableSpec(),
+                        defaultConfig.getFlinkConfig());
+        suspendJob(flinkApp, UpgradeMode.LAST_STATE, rollbackConfig);
+        deployFlinkJob(
```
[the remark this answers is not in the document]
gyfora: I have thought about that but I decided against it for 2 main reasons:
 1. It is pretty difficult to track the rollback process in 2 steps as we do not want to touch the lastReconciledSpec etc at this point which we use to perform the 2 step upgrade.
 2. We don't need to persist savepoint information in status
 
It would be good to catch rollback errors and put the job in SUSPENDED state when the rollback deployment fails though.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
```diff
@@ -62,40 +64,42 @@ public static void updateForSpecReconciliationSuccess(
...
 
     public static void updateForReconciliationError(FlinkSessionJob flinkSessionJob, String err) {
-        FlinkSessionJobReconciliationStatus reconciliationStatus =
-                flinkSessionJob.getStatus().getReconciliationStatus();
-        reconciliationStatus.setSuccess(false);
-        reconciliationStatus.setError(err);
+        flinkSessionJob.getStatus().setError(err);
+    }
+
+    public static void updateLastStableSpec(FlinkDeployment deployment) {
```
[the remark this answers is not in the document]
gyfora: I agree, I will move some if these simple utilities there

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/SessionObserver.java
```diff
@@ -50,4 +52,9 @@ public void observeIfClusterReady(
             }
         }
     }
+
+    @Override
+    protected void updateStableSpec(FlinkDeployment deployment) {
```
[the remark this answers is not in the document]
gyfora: There is no condition there at the moment, we will have to come up with something. It's probably enough to be able to list jobs.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
Aitozi: Oh, I get it now. It will leave the `lastReconciledSpec = currentSpec`, but mark the `lastReconciledSpec` have be rollbacked

gyfora: Yes, this way it is clear from the status what happened.

morhidi:
Thank you @gyfora for working on such a great feature. I'll have closer look later today. For the first read I was wondering if we could call the property `kubernetes.operator.upgrade.readiness.timeout` as `kubernetes.operator.deployment.readiness.timeout`. We could probably use this property for new deployments too and simply bring it to a terminal error state if the job fails to come up properly. WDYT?

Aitozi:
> Based on some discussion with @morhidi we decided to have the rollback related configs simply as part of the `flinkConfiguration` in order to avoid polluting the CRD

Actually, I also have the same sense before. I think we are lack of an explicit rule to guide that which should be in the CRD and which should be in the flink config.

gyfora:
@Aitozi @wangyang0918 I think at this point most new configs/features should probably go straight to `flinkConfiguration`
We should aim to keep the CRD slim and only keep the most important settings there

gyfora:
@Aitozi I actually also introduced a two step rollback logic here, first marking with `ROLLING_BACK` in order to allow dealing with errors during the rollback process. 

I am still working through some scenarios and will add tests to cover these

gyfora:
@Aitozi @morhidi @wangyang0918 @tweise 

Please take a look, I have added some test coverage for the rollback logic.

Sorry for the massive change but I had to make some cleanup/minor changes to the existing utils and reconciler otherwise the whole thing would simply be too complex and more messy :)

gyfora:
I will rebase on @Aitozi 's changes tomorrow

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/AbstractDeploymentObserver.java
```diff
@@ -159,16 +185,18 @@ private void checkCrashLoopBackoff(FlinkDeployment flinkApp, Configuration effec
...
-            final ReconciliationStatus reconciliationStatus =
-                    dep.getStatus().getReconciliationStatus();
-            reconciliationStatus.setSuccess(true);
-            reconciliationStatus.setError(null);
+        FlinkDeploymentStatus status = dep.getStatus();
+        ReconciliationStatus reconciliationStatus = status.getReconciliationStatus();
+        if (status.getJobManagerDeploymentStatus() != JobManagerDeploymentStatus.ERROR
+                && reconciliationStatus
+                        .getLastReconciledSpec()
+                        .equals(reconciliationStatus.getLastStableSpec())) {
```
bgeng777, line 198: nit: looks like we can directly use `reconciliationStatus.isLastReconciledSpecStable`.

Besides, I want to make sure I under the new condition correctly: previously, we will try to clear errors and once the JM deployment is not in ERROR status, we do the cleaning. After introducing the `isLastReconciledSpecStable` check, we only `markReconciledSpecAsStable` when `isJmDeploymentReady` && `observeFlinkCluster` && `reconciliationStatus.getState() != ReconciliationStatus.State.ROLLED_BACK`. As a result, if JM is in states like `DEPLOYED_NOT_READY`, the error in status will not be cleared.  
It seems that the later clear will not influence the correctness of the current logic as whether error info is null is only checked in `isJobUpgradeInProgress` but may make the method name `clearErrorsIfJobManagerDeploymentNotInErrorStatus` misleading. WDYT?

gyfora: The intention here was to avoid losing error information that caused a rollback operation thats why I introduced the additional isStable check. I have now updated the method name to: `clearErrorsIfDeploymentIsHealthy` that is more appropriate

gyfora:
I have rebased to the latest changes @Aitozi

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultValidator.java
```diff
@@ -261,7 +261,7 @@ private Optional<String> validateSpecChange(FlinkDeployment deployment) {
                     && deployment.getStatus().getJobManagerDeploymentStatus()
                             != JobManagerDeploymentStatus.MISSING
                     && ReconciliationUtils.isUpgradeModeChangedToLastStateAndHADisabledPreviously(
-                            deployment)) {
+                            deployment, new Configuration())) {
```
bgeng777, line 264: A concern for the refactor:
I do agree that introducing the defaultConfig in this validation as it is possible that some users may modify the default filnk-conf.yaml(e.g. adding HA config). But I am a little concerned that should we pass en empty config or using `FlinkUtils.loadConfiguration(EnvUtils.get(EnvUtils.ENV_FLINK_CONF_DIR))` ? So the Validator can work correctly if users do not specify HA config in Spec but add them in flink-conf.yaml.

docs/content/docs/custom-resource/reference.md
```diff
@@ -226,9 +226,32 @@ This page serves as a full reference for FlinkDeployment custom resource definit
 
 | Parameter | Type | Docs |
 | ----------| ---- | ---- |
-| success | boolean | True if last reconciliation step was successful. |
-| error | java.lang.String | If success == false, error information about the reconciliation failure. |
+
+### State
```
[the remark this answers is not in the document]
gyfora: This is generated based on the code. Maybe naming the enum State could be improved

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
```diff
@@ -102,36 +111,84 @@ public void reconcile(FlinkDeployment flinkApp, Context context, Configuration e
...
+
+        LOG.warn("Executing roll-back operation");
+
+        FlinkDeploymentSpec rollbackSpec = reconciliationStatus.deserializeLastStableSpec();
+        Configuration rollbackConfig =
+                FlinkUtils.getEffectiveConfig(flinkApp.getMetadata(), rollbackSpec, defaultConfig);
+
+        UpgradeMode upgradeMode = flinkApp.getSpec().getJob().getUpgradeMode();
+
+        suspendJob(
```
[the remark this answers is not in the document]
gyfora: I am afraid it is not that simple and it is by design like this. Let me try to explain.

I think we earlier decided against making changes to the spec itself as it breaks the idempotency of the operator logic. If we simply change the spec then an external process might simply reapply what was previous rolled back causing a loop.

This is also the reason why we do not upgrade the `lastReconciledSpec`. That field shows the last spec that the user have sent in that was reconciled by the controller. Even if it was rolled back later that does not change.

The ReconciliationStatus.State shows what actually happened to the lastReconciledSpec: DEPLOYED means it is running on the cluster, ROLLING_BACK/ROLLED_BACK actually signal that it was reverted after it was DEPLOYED.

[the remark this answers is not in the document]
gyfora: Of course I am happy to iterate over the design and change anything, just wanted to explain the current design in detail.

With the current flow, it is always clear what spec the user sent in, what was deployed and what has been rolled back.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultValidator.java
gyfora: Good catch, The only reason I have this here because I am waiting for this PR to be completed by @SteNicholas : https://github.com/apache/flink-kubernetes-operator/pull/162

I didnt want to double implement the logic

gyfora: But I will add the proper config there anyways, as you suggested :)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
```diff
@@ -102,36 +111,84 @@ public void reconcile(FlinkDeployment flinkApp, Context context, Configuration e
...
+                flinkApp,
+                upgradeMode == UpgradeMode.STATELESS
+                        ? UpgradeMode.STATELESS
+                        : UpgradeMode.LAST_STATE,
+                rollbackConfig);
+        deployFlinkJob(
+                rollbackSpec.getJob(),
+                flinkApp.getStatus(),
+                rollbackConfig,
+                Optional.ofNullable(
```
Aitozi, line 166: If it is in `STATELESS` mode, do we need to start with the last savepoint?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/AbstractDeploymentObserver.java
```diff
@@ -159,16 +185,16 @@ private void checkCrashLoopBackoff(FlinkDeployment flinkApp, Configuration effec
...
         return dep.getStatus().getJobManagerDeploymentStatus() == JobManagerDeploymentStatus.READY;
     }
 
-    protected void clearErrorsIfJobManagerDeploymentNotInErrorStatus(FlinkDeployment dep) {
-        if (dep.getStatus().getJobManagerDeploymentStatus() != JobManagerDeploymentStatus.ERROR) {
-            final ReconciliationStatus reconciliationStatus =
-                    dep.getStatus().getReconciliationStatus();
-            reconciliationStatus.setSuccess(true);
-            reconciliationStatus.setError(null);
+    protected void clearErrorsIfDeploymentIsHealthy(FlinkDeployment dep) {
```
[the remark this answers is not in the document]
gyfora: Well it's both stable and no errors thats why I went for healthy instead :)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
gyfora: good catch, in that case we should always pass empty

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/RollbackTest.java
```diff
@@ -0,0 +1,214 @@
...
+        // Validate stable job
+        var jobStatus = deployment.getStatus().getJobStatus();
+        assertTrue(reconciliationStatus.isLastReconciledSpecStable());
+
+        triggerRollback.run();
+
+        assertFalse(reconciliationStatus.isLastReconciledSpecStable());
+        assertEquals(ReconciliationStatus.State.ROLLING_BACK, reconciliationStatus.getState());
+        assertEquals(
+                "Deployment is not ready within the configured timeout, rolling-back.",
```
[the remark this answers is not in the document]
gyfora: I think that would be an overkill :)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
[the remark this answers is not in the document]
wangyang0918: After more consideration, I believe the current approach is more appropriate.

gyfora:
Thanks for testing this @wangyang0918 , I will leave this open for a little more to allow for more inputs given the complexity of the change.

wangyang0918:
One more comment, I forgot to note that the session job rollback has not been supported. I assume it will be done in a separate ticket.

gyfora:
@wangyang0918 I just pushed a commit before your comment :)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/ApplicationObserver.java
```diff
@@ -70,23 +70,28 @@ protected Optional<String> updateJobStatus(
...
                     .ifPresent(
                             error ->
                                     ReconciliationUtils.updateForReconciliationError(
                                             flinkApp, error));
         }
+        return isJobReady(jobStatus);
+    }
+
+    private boolean isJobReady(JobStatus jobStatus) {
+        return org.apache.flink.api.common.JobStatus.RUNNING.name().equals(jobStatus.getState());
```
tweise, line 95: Can we add a TODO here? RUNNING doesn't mean that the job is executing as expected, even a job that flip flops may intermittently have RUNNING status.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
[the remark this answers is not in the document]
tweise: Agreed that we should not change the spec, which in some case may also not be possible due to access control. The reason not to change `lastReconciledSpec` is more related to the implementation as it pertains to the status, nevertheless I think the current approach makes sense.

tweise (approved):

wangyang0918 (changes requested):
@gyfora I find a problem when having more tests. The `ROLLED_BACK` application could not be suspended correctly.

This issue could be easily reproduced by updating `.spec.job.state` to suspended for a `ROLLED_BACK` application. I am afraid we need to fix this before merging.

BTW, would you like to create a ticket for supporting rollback for session job?

gyfora:
@wangyang0918 thanks, I will work on fixing this and covering with tests. Also I have a few follow up Jiras in mind including the sessionjob support 

I am traveling today so I might need a day or so

wangyang0918:
@gyfora No hurry, take your time.

gyfora:
@wangyang0918 I think for the suspend to work, I need to fix 2 things basically:
 1. If the user suspends a job, the spec should be automatically marked as `lastStableSpec`
 2. If an upgrade after a suspend fails, we should roll back to the suspended state (simply stop the job)
 
 Will work on this over the weekend

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/RollbackTest.java
```diff
@@ -265,5 +265,28 @@ public void testRollback(
         assertTrue(reconciliationStatus.isLastReconciledSpecStable());
         assertEquals(ReconciliationState.DEPLOYED, reconciliationStatus.getState());
         assertNull(deployment.getStatus().getError());
+
+        if (deployment.getSpec().getJob() != null) {
```
wangyang0918, line 269: It seems that the tests do not cover the scenario which is suspending from `ROLLED_BACK` state.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/RollbackTest.java
```diff
@@ -265,5 +265,28 @@ public void testRollback(
...
+            deployment.getSpec().getJob().setState(JobState.SUSPENDED);
+            testController.reconcile(deployment, context);
+            Thread.sleep(500);
+            testController.reconcile(deployment, TestUtils.createEmptyContext());
+            testController.reconcile(deployment, context);
+            testController.reconcile(deployment, context);
```
wangyang0918, line 275: I think all the tests could still pass while removing these lines.

wangyang0918 (commented):
Now this PR could work while suspending a `ROLLED_BACK` FlinkDeployment.

wangyang0918:
> @wangyang0918 I think for the suspend to work, I need to fix 2 things basically:
> 
> 1. If the user suspends a job, the spec should be automatically marked as `lastStableSpec`
> 2. If an upgrade after a suspend fails, we should roll back to the suspended state (simply stop the job)
> 
> Will work on this over the weekend

It seems that the current implementation is not consistent with fix 2. But I lean to the current behavior that we should not roll back to suspended state. This will help users to quickly figure out the incorrect fields updates.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/ApplicationObserver.java
gyfora: yes I will add a Jira for this before merging

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/RollbackTest.java
gyfora: thanks, will add now

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/RollbackTest.java
gyfora: good point this should not be necessary

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/RollbackTest.java
gyfora: fixed

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/RollbackTest.java
gyfora: fixed

gyfora:
> > @wangyang0918 I think for the suspend to work, I need to fix 2 things basically:
> > 
> > 1. If the user suspends a job, the spec should be automatically marked as `lastStableSpec`
> > 2. If an upgrade after a suspend fails, we should roll back to the suspended state (simply stop the job)
> > 
> > Will work on this over the weekend
> 
> It seems that the current implementation is not consistent with fix 2. But I lean to the current behavior that we should not roll back to suspended state. This will help users to quickly figure out the incorrect fields updates.

Yes, I changed my mind along the logic that you also highlighted. If the previous state was suspended, it is better to not roll back to allow users to find errors isntead

wangyang0918 (approved):
+1 for merging.

gyfora:
I have opened the following jiras for follow-up work:
https://issues.apache.org/jira/browse/FLINK-27281
https://issues.apache.org/jira/browse/FLINK-27279
https://issues.apache.org/jira/browse/FLINK-27280

cc @tweise @Aitozi

PR #168, opened 2022-04-16
[FLINK-27161] Support fetch user jar from different source

Aitozi:
This pr is meant to support fetch user jar from different source. 

**The brief change log:**

1. Introduced `ArtifactFetcher` interface, with two implementation to support `http` and `filesystem` based resource
2. Copy the oss fs jar to the plugins dir as the default/example jar. Users can support other filesystem by copy other fs jars to the plugins as mentioned in [flink](https://nightlies.apache.org/flink/flink-docs-release-1.14/docs/deployment/resource-providers/standalone/docker/#advanced-customization) .

It is verified by unit test to verify the local filesystem and http. Manually verify the oss resource


Aitozi:
Please help take a look when you are free cc @gyfora @wangyang0918

wangyang0918:
Before starting the review, I am wondering whether the application cluster could also benefit from this. Then users does not need to use the pod template to download the user jars if they are supported internally.

```
kind: FlinkDeployment
metadata:
  namespace: default
  name: basic-example
spec:
  image: flink:1.14
  flinkVersion: v1_14
  flinkConfiguration:
  ... ...
  job:
    jarURI: https://repo1.maven.org/maven2/org/apache/flink/flink-examples-streaming_2.12/1.14.3/flink-examples-streaming_2.12-1.14.3.jar
    parallelism: 2
    upgradeMode: stateless
```

Aitozi:
@wangyang0918 Do you mean to make flink fetch different source type aritifacts in the `KubernetesApplicationClusterEntrypoint` and `KubernetesTaskExecutorRunner` ?  I think it's feasible and it can indeed simplify the use of the application mode deployment

Aitozi:
Do you mean to make flink fetch different source type aritifacts in the `KubernetesApplicationClusterEntrypoint` and `KubernetesTaskExecutorRunner` ?  I think it's feasible and it can indeed simplify the use of the application mode deployment

wangyang0918:
@Aitozi It will be great and makes things easier if the upstream Flink could support remote user jars directly. Benefit from YARN distributed cache, the YARN application mode have already support HDFS user jars. But this is not the top priority since we could get this done in the flink-kubernetes-operator project first.

Given that users specify a remote user jar via `.spec.job.jarURI`, then the operator could inject an init-container to download the jars automatically.

Let's narrow down the scope of FLINK-27161 to only support fetching remote user jars for session job. And create a follow-up ticket to support remote user jars for application cluster. I will start to review this PR today.

Aitozi:
Thanks for your explanation , I get your meaning. We could support session job first and then use the way of injecting init container as a bridge before the flink has ability to fetch the remote jars. 
Besides, I remember there are already some tickets to discuss to enable flink can ship remote fs in k8s mode. I think we can continue these work after this have been done.

[ttps://issues.apache.org/jira/browse/FLINK-20867](https://issues.apache.org/jira/browse/FLINK-20867)  Support ship files from local/remote path to k8s pod
https://issues.apache.org/jira/browse/FLINK-26647  Can not add extra config files on native Kubernetes 
https://issues.apache.org/jira/browse/FLINK-26897  Provide an additional way to add dependency files to k8s

flink-kubernetes-operator/pom.xml
```diff
@@ -135,6 +141,13 @@ under the License.
             <version>${fabric8.version}</version>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-test-utils-junit</artifactId>
+            <version>${flink.version}</version>
+            <scope>test</scope>
+        </dependency>
```
wangyang0918, line 150: It seems that we do not need this dependency.

flink-kubernetes-operator/pom.xml
```diff
@@ -75,6 +75,12 @@ under the License.
             <version>${flink.version}</version>
         </dependency>
 
+        <dependency>
```
wangyang0918, line 78: We might need to update the notice file when introducing new dependency.

flink-kubernetes-operator/pom.xml
```diff
@@ -270,6 +283,13 @@ under the License.
                             <version>${flink.version}</version>
                             <outputDirectory>${plugins.tmp.dir}/flink-metrics-statsd</outputDirectory>
                         </artifactItem>
+                        <artifactItem>
```
wangyang0918, line 286: It is strange to only add the oss filesystem here. We need to bundle all the internal fs implementation or none.

wangyang0918 (changes requested):
This PR works well and I just left some minor comments.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/artifact/FileSystemBasedArtifactFetcher.java
```diff
@@ -0,0 +1,50 @@
...
+            new FileSystemBasedArtifactFetcher();
+
+    @Override
+    public File fetch(String uri, File targetDir) throws Exception {
+        org.apache.flink.core.fs.Path source = new org.apache.flink.core.fs.Path(uri);
+        var start = System.currentTimeMillis();
+        FileSystem fileSystem = source.getFileSystem();
+        String fileName = source.getName();
+        File targetFile = new File(targetDir, fileName);
+        FileUtils.copyToFile(fileSystem.open(source), targetFile);
```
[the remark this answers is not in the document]
Aitozi: Nice catch, fixed

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/artifact/HttpArtifactFetcher.java
```diff
@@ -0,0 +1,48 @@
...
+    public static final Logger LOG = LoggerFactory.getLogger(HttpArtifactFetcher.class);
+    public static final HttpArtifactFetcher INSTANCE = new HttpArtifactFetcher();
+
+    @Override
+    public File fetch(String uri, File targetDir) throws Exception {
+        var start = System.currentTimeMillis();
+        URL url = new URL(uri);
+        String fileName = FilenameUtils.getName(url.getFile());
+        File targetFile = new File(targetDir, fileName);
+        FileUtils.copyToFile(new URL(uri).openStream(), targetFile);
```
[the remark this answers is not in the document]
Aitozi: fixed

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/OperatorConfigOptions.java
```diff
@@ -97,4 +97,10 @@ public class OperatorConfigOptions {
                     .withDescription(
                             "The timeout for deployments to become ready/stable "
                                     + "before being rolled back if rollback is enabled.");
+
+    public static final ConfigOption<String> OPERATOR_USER_JAR_BASE_DIR =
```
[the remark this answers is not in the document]
Aitozi: Yes, But the CI seems not find it 😄

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
```diff
@@ -204,12 +209,14 @@ private JarRunResponseBody runJar(
         }
     }
 
-    private JarUploadResponseBody uploadJar(FlinkSessionJob sessionJob, Configuration conf)
+    private JarUploadResponseBody uploadJar(
+            FlinkSessionJob sessionJob, FlinkDeployment sessionCluster, Configuration conf)
             throws Exception {
-        Path path = jarResolver.resolve(sessionJob.getSpec().getJob().getJarURI());
+        String targetDir = artifactManager.generateJarDir(sessionCluster, sessionJob);
+        File jarFile = artifactManager.fetch(sessionJob.getSpec().getJob().getJarURI(), targetDir);
```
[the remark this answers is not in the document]
Aitozi: I also think about it, The good thing is that we can do no special thing for the local filesystem when `fetch` and `delete` (If no copy, we may can't delete after submit too). Do you think we need to handle the difference now ?

flink-kubernetes-operator/pom.xml
Aitozi: Nice catch, removed

flink-kubernetes-operator/pom.xml
Aitozi: As discussed in the mail list, I think we should avoid bundle all the fs jars It will make the image bigger and not necessary. I intend to add here as an example for the user to follow up the guide, But I think I can remove it now and clarify how to extend it in the doc.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/OperatorConfigOptions.java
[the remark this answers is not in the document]
Aitozi: Does the option change have to be copy to `configuration.md` manually ?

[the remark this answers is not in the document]
Aitozi: Added

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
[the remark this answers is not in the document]
wangyang0918: Let's keep the current behavior and do some improvements later if necessary.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/artifact/ArtifactManagerTest.java
```diff
@@ -0,0 +1,134 @@
...
+                        tempDir.toString());
+        Assertions.assertTrue(file.exists());
+        Assertions.assertEquals(tempDir.toString(), file.getParentFile().toString());
+    }
+
+    @Test
+    public void testHttpFetch() throws Exception {
+        HttpServer httpServer = null;
+        try {
+            httpServer = HttpServer.create(new InetSocketAddress(1234), 0);
```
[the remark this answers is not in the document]
Aitozi: Fixed by start with retry

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/OperatorConfigOptions.java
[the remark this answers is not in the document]
wangyang0918: Not in this PR, but we could improve to generate the document automatically just like in Flink.

flink-kubernetes-operator/pom.xml
Aitozi: Removed

flink-kubernetes-operator/pom.xml
Aitozi: The flink-core seems already included. 

```
- org.apache.flink:flink-core:1.14.4
```

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/OperatorConfigOptions.java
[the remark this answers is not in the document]
Aitozi: OK, I have add the option to the `configuration.md`

flink-kubernetes-operator/pom.xml
wangyang0918: Yes. We could remove it now and document it with `kustomize`.

Aitozi: Get it. I will add it in the followup doc ticket

Aitozi:
@wangyang0918 Thanks for your review, I have addressed your comments, Please take a look again.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/OperatorConfigOptions.java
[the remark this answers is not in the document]
Aitozi: > Not in this PR, but we could improve to generate the document automatically just like in Flink.

Created a ticket for this https://issues.apache.org/jira/browse/FLINK-27334

flink-kubernetes-operator/pom.xml
wangyang0918: What I mean is the transitive dependencies? Do they have already included in the NOTICE file?

flink-kubernetes-operator/pom.xml
wangyang0918: We just could add the document in FLINK-27270.

Aitozi: Yes

flink-kubernetes-operator/pom.xml
Aitozi: Since the dependency of `flink-clients` already include the `flink-core`, So I remove the duplicated declaration here

docs/content/docs/operations/configuration.md
```diff
@@ -62,3 +62,4 @@ To learn more about metrics and logging configuration please refer to the dedica
 | kubernetes.operator.observer.flink.client.timeout     |     10s    |  Duration    | The timeout for the observer to wait the flink rest client to return.            |
 | kubernetes.operator.reconciler.flink.cancel.job.timeout     |     1min    |  Duration    | The timeout for the reconciler to wait for flink to cancel job.            |
 | kubernetes.operator.reconciler.flink.cluster.shutdown.timeout     |     60s    |  Duration    | The timeout for the reconciler to wait for flink to shutdown cluster.           |
+| kubernetes.operator.user.artifacts.base.dir     |     60s    |  String |     The base dir to put the session job artifacts.           |
```
wangyang0918, line 65: The default value is not `60s`.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/artifact/ArtifactManager.java
```diff
@@ -0,0 +1,75 @@
...
+        createIfNotExists(targetDir);
+        URI uri = new URI(jarURI);
+        if ("http".equals(uri.getScheme()) || "https".equals(uri.getScheme())) {
+            return HttpArtifactFetcher.INSTANCE.fetch(jarURI, targetDir);
+        } else {
+            return FileSystemBasedArtifactFetcher.INSTANCE.fetch(jarURI, targetDir);
+        }
+    }
+
+    public String generateJarDir(FlinkDeployment flinkApp, FlinkSessionJob sessionJob) {
```
wangyang0918, line 65: We need to enforce the session cluster and session jobs are in the same namespace. Right?

Then maybe the `baseDir`  + `/` + `sessionJob.getMetadata().getNamespace()` + `/` + `sessionJob.getSpec().getClusterId()` + `/` + `sessionJob.getMetadata().getName()` is enough.

Aitozi: Oh yes, I will simplify it.

docs/content/docs/operations/configuration.md
Aitozi: Thanks , will fix

Aitozi:
I have updated, please take a look again @wangyang0918

wangyang0918 (approved):
+1 for merging.

PR #169, opened 2022-04-16
[FLINK-27269] Clean up the jar file after submitting the job

Aitozi:
This pr is meant to clean up the jar file after submitting the job to avoid the jar files exploding after serval upgrade

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
@@ -58,6 +58,8 @@ import org.apache.flink.runtime.rest.messages.job.savepoints.SavepointTriggerHea
 import org.apache.flink.runtime.rest.messages.job.savepoints.SavepointTriggerMessageParameters;
 import org.apache.flink.runtime.rest.messages.job.savepoints.SavepointTriggerRequestBody;
 import org.apache.flink.runtime.rest.util.RestConstants;
+import org.apache.flink.runtime.webmonitor.handlers.JarDeleteHeaders;
+import org.apache.flink.runtime.webmonitor.handlers.JarDeleteMessageParameters;
 import org.apache.flink.runtime.webmonitor.handlers.JarRunHeaders;
 import org.apache.flink.runtime.webmonitor.handlers.JarRunMessageParameters;
 import org.apache.flink.runtime.webmonitor.handlers.JarRunRequestBody;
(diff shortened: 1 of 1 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
```diff
@@ -234,6 +238,22 @@ private JarUploadResponseBody jarUpload(FlinkSessionJob sessionJob, Configuratio
         }
     }
 
+    private void jarDelete(Configuration conf, String jarId) {
```
gyfora, line 241: nit: could we please change the name to `deleteJar`?

gyfora (approved):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
Aitozi: Fixed

wangyang0918:
I am not sure whether deleting the submitted jars could actually work, which means freeing up the disk spaces. Because the classes are still in use in a dedicated user classloader. See https://github.com/apache/flink/blob/master/flink-clients/src/main/java/org/apache/flink/client/program/PackagedProgram.java#L144.

I will have a manual verification soon.

Aitozi:
@wangyang0918 Thanks for pointing this, I will also take a look then.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
```diff
@@ -234,6 +238,22 @@ private JarUploadResponseBody jarUpload(FlinkSessionJob sessionJob, Configuratio
...
+            JarDeleteHeaders headers = JarDeleteHeaders.getInstance();
+            JarDeleteMessageParameters parameters = headers.getUnresolvedMessageParameters();
+            parameters.jarIdPathParameter.resolve(jarId);
+            clusterClient
+                    .sendRequest(headers, parameters, EmptyRequestBody.getInstance())
+                    .get(
+                            operatorConfiguration.getFlinkClientTimeout().toSeconds(),
+                            TimeUnit.SECONDS);
+        } catch (Exception e) {
+            LOG.error("Failed to delete the jar.", e);
```
wangyang0918, line 253: We need to print the `jarId` when failed.

wangyang0918 (approved):
Using `lsof | grep delete` in JobManager pod and do not find any leaked the jars after deletion. It seems that `userCodeClassLoader.close()` in `PackagedProgram#close()` has worked as expected.

+1 for merging.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
Aitozi: Fixed

Aitozi:
> Using `lsof | grep delete` in JobManager pod and do not find any leaked the jars after deletion. It seems that `userCodeClassLoader.close()` in `PackagedProgram#close()` has worked as expected.
> 
> +1 for merging.

👍🏻 Nice way to verify, thanks

PR #170, opened 2022-04-17
[FLINK-27235] Publish Flink k8s Operator Helm Charts via Github Actions

GezimSejdiu:
Hi Flink k8s Operator lovers, 

this PR tends to add the helm releaser workflow (via Github Actions) so that it creates a release package of the operator when any changes are applied into https://github.com/apache/flink-kubernetes-operator/tree/main/helm/flink-kubernetes-operator.

By default, it pushes/stores helm charts into girhub-pages, but there is also an alternative to host them elsewhere, in our case at https://downloads.apache.org/flink/ :  (see here for more details: https://github.com/helm/chart-releaser-action#example-using-custom-config ).

Please, review and do let me know if there is something else needed.

Best regards, 
Gezim

### DIFF
--- a/.github/workflows/helm-charts.yaml
+++ b/.github/workflows/helm-charts.yaml
@@ -1,0 +1,35 @@
+name: Release Charts
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Configure Git
+        run: |
+          git config user.name "$GITHUB_ACTOR"
+          git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
+
+      - name: Install Helm
+        uses: azure/setup-helm@v1
+        with:
+          version: v3.8.1
+
+      - name: Run chart-releaser
+        uses: helm/chart-releaser-action@v1.4.0
+        with:
+          charts_dir: helm
+          charts_repo_url: https://downloads.apache.org/flink/
+        env:
+          CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+          CR_OWNER: "${{github.repository_owner}}"
+          CR_GIT_BASE_URL: "https://api.github.com/"


wangyang0918:
@GezimSejdiu Thanks for working on this. I have two concerns about this PR.
* `https://downloads.apache.org/flink/` is not a right place to host the snapshot helm chart release. Only the released version could be found there. So I think we need to publish the snapshot helm charts via github packages.
* I expect only when some changes happen in `helm/flink-kubernetes-operator`, then it needs to publish the helm charts again. Other java code changes should not trigger this.

.github/workflows/helm-charts.yaml
```diff
@@ -0,0 +1,35 @@
...
+          git config user.name "$GITHUB_ACTOR"
+          git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
+
+      - name: Install Helm
+        uses: azure/setup-helm@v1
+        with:
+          version: v3.8.1
+
+      - name: Run chart-releaser
+        uses: helm/chart-releaser-action@v1.4.0
```
MartijnVisser, line 28: Are we sure that ASF Infra has whitelisted and approved using this Github Action?

wangyang0918: Thanks for raising up your concerns. Do you mean that publishing the helm chart to github packages does not conform to the Apache bylaws? I am not an expert in this area :)

BTW, we do not mean to change the official release process. The released helm charts are still hosted in `https://downloads.apache.org/flink/ `.

cc @gyfora

MartijnVisser: I think it has been approved by INFRA recently, see https://issues.apache.org/jira/browse/INFRA-21731 and also https://infra.apache.org/github-actions-policy.html

gyfora: I think this should be fine as it is approved by the apache infra. This is only for convenience of exposing the different helm chart versions on github. As @wangyang0918 said this is not part of the official release process.

.github/workflows/helm-charts.yaml
```diff
@@ -0,0 +1,35 @@
...
+        with:
+          charts_dir: helm
+          charts_repo_url: https://downloads.apache.org/flink/
```
gyfora, line 31: This should point to the github pages for flink-kubernetes-operator instead i think

.github/workflows/helm-charts.yaml
MartijnVisser: Since Infra has whitelisted it, there shouldn't be an issue. It does make sense to pin the commit hash of the action (like the policy asks). The concern is not so much related to the release process, but such action could be used for credential stealing etc.

GezimSejdiu:
Hey @wangyang0918,

thanks a lot for your review on this.

Reg.: 
> * `https://downloads.apache.org/flink/` is not a right place to host the snapshot helm chart release. Only the released version could be found there. So I think we need to publish the snapshot helm charts via github packages.

Right, I wasn't aware of that and thus I will update to only be released within github-pages as https://github.com/helm/chart-releaser#chart-releaser point out if we do not want to host them elsewhere. For that to be possible we will need to have gh-pages branch (as pointed here: https://github.com/helm/chart-releaser-action#pre-requisites, point 2 and 3). Um, reg. `github-packages` -- this Github Actions do not publish them as Github Package but rather as a normal Releases (see here as an example: https://github.com/GezimSejdiu/flink-kubernetes-operator/releases and in order this release be usable -- it has to also be hosted somewhere e.g. gh-pages is the default one used from the plugin but it also provides alternatives (as I was trying to do it so that we publish that into flink. downloads).

> * I expect only when some changes happen in `helm/flink-kubernetes-operator`, then it needs to publish the helm charts again. Other java code changes should not trigger this.

Yes, that is true. It only triggers when anything has been changed into `helm` folder e..g there may be another helm chart (take `flink-job-deployment` -- creating Flink Job Deployment within Helm Charts).  

Let me know if: 
- we are going to use Github Pages as a hosting platform for the helm-chart (SNAPSHOT) so that we can also use it for fast prototying/testing? -- by that I can already remove the `charts_repo_url: https://downloads.apache.org/flink/` which then will opt in to use the default one `gh-pages`.

.github/workflows/helm-charts.yaml
GezimSejdiu: Hey @gyfora , 

that is true -- if we let to the defaults it points to github-pages (we do not have it currently here) and the idea was to use this as an alternative -- which will also host SNAPSHOTS (I got to know that it is only meant for releases and not also holding snapshots). So, I will update the chart-releaser-action to point to github-pages and let us also agree if we want to have gh-pages for SNAPSHOTs. My idea was to use only one hosting platform for both so that we do not need to change anything withing documentation but rather just use : 

```bash
helm repo add flink-operator-repo https://downloads.apache.org/flink/flink-kubernetes-operator-1.0-SNAPSHOT/
```
But I do also agree with the process and how that is done and thus we can also keep the documentation up-to-date when using snapshot refers to gh-pages and when using a proper release to `https://downloads.apache.org/flink/`

wangyang0918:
@GezimSejdiu Thanks for your explanation. I still have some questions before we could merge this PR.

* IIUC, we need to create a `gh-pages`, which will store the published charts.
* I am not sure whether this PR could work when some changes happen in `helm/flink-kubernetes-operator` while the version(1.0-SNAPSHOT) does not change.
* We might need to update the [quick start](https://nightlies.apache.org/flink/flink-kubernetes-operator-docs-main/docs/try-flink-kubernetes-operator/quick-start/) for main branch to use github snapshot release.

.github/workflows/helm-charts.yaml
wangyang0918: If only pushing to `main` branch, not the pull requests, will trigger the chart-releaser-action, do you think we still have the risk of credential stealing?

MartijnVisser: The attack vector is that the workflow could be changed/compromised to perform an unwanted/unverified change, such as outputting credentials to a log file or submitting it to a 3rd party command & control server. So yes, the risk would still be there, only (much) lower. If you pin the workflow with a SHA commit, you would mitigate it even more (because any change in the workflow would only happen in a newly created commit, which you then wouldn't include in your action run)

MartijnVisser:
For context, perhaps it also makes sense to look how Apache Airflow is doing this. I would recommend checking out their repositories and also https://cwiki.apache.org/confluence/display/INFRA/Github+Actions+to+DockerHub. Probably relevant from this page:

```UPDATE: Apache Airflow has changed the process since - there is no more automated push to DockerHub - all the CI images are only automatically published in GHCR.IO for the purpose of development and CI.  Apache Airflow switched to manual process of publishing RC and RELASE images by the release manage (using semi-automation of course). Similarly as preparing the releases, building and pushing the images is now executed on the release manager-controlled hardware. Once the image is locally built it is pushed to DockerHub using personal authentication of the release manager.```

File: .github/workflows/helm-charts.yaml (hunk 1 above)
Comment (wangyang0918) [reply .github/workflows/helm-charts.yaml:28]: Thanks. I get your point.

**Comment by GezimSejdiu** [comment]
> 



> @GezimSejdiu Thanks for your explanation. I still have some questions before we could merge this PR.
> 
> * IIUC, we need to create a `gh-pages`, which will store the published charts.

Yes, if we do not want to host them into https://downloads.apache.org/flink/ then in order for the releaser to work it needs a site where it can also write the index file and store the published charts. Please, do let me know if you are going to create it as we also need to adapt/change the way this workflow works e.g. remove [chart_repo_url](https://github.com/apache/flink-kubernetes-operator/pull/170/commits/621303887b323cfac53ed6f95b2959298a0d894c#diff-4e6486e1da78f83d1007495745cc4bb5a27de1d2664d652564ab193d6634a81fR31).

> * I am not sure whether this PR could work when some changes happen in `helm/flink-kubernetes-operator` while the version(1.0-SNAPSHOT) does not change.
Right, the generated release is based on: 
```
      --release-name-template string   Go template for computing release names, using chart metadata (default "{{ .Name }}-{{ .Version }}")
``` 
and thus will check if the same version/release does exists. The idea here is that `main` is considered as a proper release when you push changes to the chart -- it needs a release. 
A workaround is to use: 
```
      --skip-existing                  Skip upload if release exists
``` 
but then it will not serve the purpose of "SNAPSHOTING"  -- so not sure how to handle this. Maybe we can ask the community at https://github.com/helm/chart-releaser to be able to allow `--override or similar? 

> * We might need to enrich the [quick start](https://nightlies.apache.org/flink/flink-kubernetes-operator-docs-main/docs/try-flink-kubernetes-operator/quick-start/) to document how to use helm chart snapshot release.

Yes, asap as this is pushed and we want to use it -- we will also have to update the guide on how to use the snapshot (e.g. I wanted to use it here: https://seatunnel.apache.org/docs/start/kubernetes -- to test the new version but wasn't able to pull it :( (of course installing it locally can also work).

**Comment by wangyang0918** [comment]
Creating a branch `gh-pages`, which will store the index file and the published charts, makes sense to me.

If the helm chart-releaser could support snapshot override, then this PR makes sense to me. Because users could test the new version easily.

```
helm repo add flink-operator-repo https://apache.github.io/flink-kubernetes-operator/
helm --version 1.0-SNAPSHOT install flink-kubernetes-operator flink-operator-repo/flink-kubernetes-operator
helm --version 1.0-SNAPSHOT upgrade flink-kubernetes-operator flink-operator-repo/flink-kubernetes-operator
```

cc @MartijnVisser @gyfora WDYT?

GezimSejdiu:
> If the helm chart-releaser could support snapshot override, then this PR makes sense to me. Because users could test the new version easily.

Indeed. That would be quite beneficial and also the steps you just listed on how to use the SNAPSHOT would be just great that we can already test/use the release (SNAPSHOT) before doing a proper release.

From what I can see and also looked into some discussions at https://github.com/helm/chart-releaser-action -- overriding the release doesn't seem supported :( and in our case, the skip flag doesn't really make sense (we want to have a continuous deployment of the SNAPSHOT, right). 

I can think of a workaround/scenario how to achieve this 'override' the release:
- move the default branch to `develop` so that it is obvious it is a SNAPSHOT branch and thus every merge to develop will mean will release a SNAPSHOT and add a step in the workflow that deletes the actual release (if present) as it generates it via: `"{{.Name }}-{{.Version }}"` and then when merged to the main/master branch we have to avoid this step and consider that as a proper release (or even do not let it run for main as anyway the process you seem to follow is by doing a manual release process, and therefore we will not interfere with the main/official release) 

What do you think of this idea? Do you think it is feasible?

GezimSejdiu:
Hi @wangyang0918 ,

I tried a solution on how to simulate the re-deployment of the SNAPSHOT (but I didn't like it myself as it is a bit tricky and as such I will not push it but rather wait if there is such an option from the official chart-releaser-action).

I did also comment on this issue: https://github.com/helm/chart-releaser-action/issues/8#issuecomment-1114057516 there and till we have something I will close the issue for now.

We may still come back to this when there are more news about it.

Once again, I thank you all for such great feedback and also for being so constructive. Keep doing a great job.

Best regards, 
Gezim

PR #171, opened 2022-04-18
[FLINK-27289]Do some optimizations for "FlinkService#stopSessionCluster"

spoon-lz:
In the "FlinkService#stopSessionCluster" method, if 'deleteHaData=true', the 'FlinkUtils#waitForClusterShutdown' method will be called twice.
So we should only execute 'FlinkUtils#waitForClusterShutdown' again with 'deleteHaData=false',This will reduce two visits to the kubernetes cluster

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
@@ -404,7 +404,6 @@ public class FlinkService {
     public void stopSessionCluster(
             ObjectMeta objectMeta, Configuration conf, boolean deleteHaData, long shutdownTimeout) {
         FlinkUtils.deleteCluster(objectMeta, kubernetesClient, deleteHaData, shutdownTimeout);
-        FlinkUtils.waitForClusterShutdown(kubernetesClient, conf, shutdownTimeout);
     }
 
     public void triggerSavepoint(
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/SessionReconciler.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/SessionReconciler.java
@@ -94,6 +94,10 @@ public class SessionReconciler extends AbstractDeploymentReconciler {
                 effectiveConfig,
                 false,
                 operatorConfiguration.getFlinkShutdownClusterTimeout().toSeconds());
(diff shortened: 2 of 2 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
```diff
@@ -225,7 +225,9 @@ public Optional<String> cancelJob(
     public void stopSessionCluster(
             FlinkDeployment deployment, Configuration conf, boolean deleteHaData) {
         FlinkUtils.deleteCluster(deployment, kubernetesClient, deleteHaData);
-        FlinkUtils.waitForClusterShutdown(kubernetesClient, conf);
+        if (!deleteHaData) {
```
wangyang0918, line 228: Could we simply remove the `FlinkUtils.waitForClusterShutdown(kubernetesClient, conf)` here and call it directly in the `SessionReconciler#upgradeSessionCluster`?

Further more, we may do not need the `FlinkService#stopSessionCluster` and use `FlinkUtils.deleteCluster` instead.

wangyang0918 (changes requested):
Thanks for creating this PR. I left a comment and please have a look.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
spoon-lz: FlinkService#stopSessionCluster has been removed

wangyang0918:
@spoon-lz Could you please rebase the latest main branch and resolve the conflicts.

spoon-lz:
@wangyang0918  Rebase is complete

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
```diff
@@ -401,12 +400,6 @@ public Optional<String> cancelSessionJob(
         return savepointOpt;
     }
 
-    public void stopSessionCluster(
```
wangyang0918, line 404: After you rebase the latest master, I am afraid that we still need the `stopSessionCluster`, which simply call the `FlinkUtils.deleteCluster` in the production code. And we could still keep the testing implementation.

spoon-lz: `stopSessionCluster` is back now

wangyang0918 (approved):
+1

PR #172, opened 2022-04-19
When do job types support Python? Checkpoint automatically executed and supports S3 compatibility? Have time to plan? Thank you!

caesar168:
When do job types support Python? Checkpoint automatically executed and supports S3 compatibility? Have time to plan? Thank you!


bgeng777:
Hi @caesar168, thanks for your interest. I believe it would be much better if you can follow the [support](https://github.com/apache/flink-kubernetes-operator#support) guideline to post your questions to the mailing list.
Besides, it would also help others in the community to answer your questions in the mailing list if you can provide more context(e.g. your expected/current usage of pyflink job with k8s).

PR #173, opened 2022-04-19
[FLINK-27310] Fix FlinkOperatorITCase

bgeng777:
- Fix FlinkOperatorITCase that not sets JM replica correctly.
- Due to the [doc of failsafe plugin](https://maven.apache.org/surefire/maven-failsafe-plugin/), use the `mvn verify` to replace `mvn integration-test` so the `post-integration-test` phase can be executed and as a result, if the IT fails, the maven build will fail.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,12 +73,12 @@ jobs:
       - name: Tests in flink-kubernetes-operator
         run: |
           cd flink-kubernetes-operator
-          mvn integration-test -Dit.skip=false
+          mvn verify -Dit.skip=false
           cd ..
       - name: Tests in flink-kubernetes-webhook
         run: |
           cd flink-kubernetes-webhook
-          mvn integration-test -Dit.skip=false
+          mvn verify -Dit.skip=false
           cd ..
       - name: Stop the operator
         run: |
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,4 @@ buildNumber.properties
 
 .idea
 *.iml
+*.DS_Store
--- a/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/FlinkOperatorITCase.java
+++ b/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/FlinkOperatorITCase.java
@@ -114,6 +114,7 @@ public class FlinkOperatorITCase {
         resource.setCpu(1);
         JobManagerSpec jm = new JobManagerSpec();
         jm.setResource(resource);
+        jm.setReplicas(1);
         spec.setJobManager(jm);
         TaskManagerSpec tm = new TaskManagerSpec();
         tm.setResource(resource);
--- a/flink-kubernetes-operator/src/test/resources/log4j2-test.properties
+++ b/flink-kubernetes-operator/src/test/resources/log4j2-test.properties
@@ -34,7 +34,7 @@
 # limitations under the License.
 ################################################################################
 
-rootLogger.level = OFF
+rootLogger.level = INFO
 rootLogger.appenderRef.console.ref = ConsoleAppender
 
 # Log all infos to the console
--- a/pom.xml
+++ b/pom.xml
@@ -137,6 +137,7 @@ under the License.
                     <execution>
                         <goals>
                             <goal>integration-test</goal>
+                            <goal>verify</goal>
                         </goals>
                     </execution>
                 </executions>


bgeng777:
cc @wangyang0918

.github/workflows/ci.yml
```diff
@@ -73,7 +73,7 @@ jobs:
       - name: Tests in flink-kubernetes-operator
         run: |
           cd flink-kubernetes-operator
-          mvn integration-test -Dit.skip=false
+          mvn verify -Dit.skip=false
```
wangyang0918, line 76: I am afraid the `org.apache.flink.kubernetes.operator.FlinkOperatorITCase` is not triggered by using `mvn verify`. Could you please confirm this?

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/FlinkOperatorITCase.java
```diff
@@ -114,6 +114,7 @@ private static FlinkDeployment buildSessionCluster() {
         resource.setCpu(1);
         JobManagerSpec jm = new JobManagerSpec();
         jm.setResource(resource);
+        jm.setReplicas(1);
```
wangyang0918, line 117: Why we have to set the replicas explicitly?

flink-kubernetes-operator/src/test/resources/log4j2-test.properties
```diff
@@ -34,7 +34,7 @@
 # limitations under the License.
 ################################################################################
 
-rootLogger.level = OFF
+rootLogger.level = INFO
```
wangyang0918, line 37: Using `INFO` log level makes debugging failed tests easier. Hope it does not have other side effects.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/FlinkOperatorITCase.java
bgeng777: Without this setting, `replicas` in `JobManagerSpec` will be zero be default. That's why the IT test failed.
So do you mean we should change the default value of `replicas` in the class? It seems to be a better solution.

wangyang0918: Yes. I believe the `replicas` is not a field which should be configured explicitly. In most cases, no matter the HA enabled or not, they do not need to set this. Unless they really want a standby JobManager to get faster recovery.

Maybe we could also remove the `replicas` in the examples and `e2e-tests`.

.gitignore
```diff
@@ -35,3 +35,4 @@ buildNumber.properties
 
 .idea
 *.iml
+*.DS_Store
```
wangyang0918, line 38: Why do we need this change?

bgeng777: I notice that when running `git add -A` command, some DS_Store files generated by Mac will be added too. I believe we do not want these files and adding it to gitigore helps developers. Do you have other concerns?

.github/workflows/ci.yml
wangyang0918: Could you please share a failed CI when using `mvn verify`?

.gitignore
wangyang0918: I am just curious how the `DS_Store` files are generated.

.github/workflows/ci.yml
bgeng777: Sure, you can see failed CI [here](https://github.com/bgeng777/flink-kubernetes-operator/runs/6089734117?check_suite_focus=true#step:10:2083)

.gitignore
bgeng777: Typically, it is because I opened some files(e.g. the CI test report) of this github project with `Finder` in my local machine.

.github/workflows/ci.yml
```diff
@@ -73,7 +73,7 @@ jobs:
...
           cd ..
       - name: Tests in flink-kubernetes-webhook
```
wangyang0918, line 78: Could we also use the `mvn verify` in "Tests in flink-kubernetes-webhook" even though we do not have ITCase now?

.gitignore
wangyang0918: nit: Keep the empty line at end of file.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/FlinkOperatorITCase.java
wangyang0918: Would you like to change the default value of `replicas` in this PR or it will be done in a follow-up ticket?

bgeng777: I checked the usage of `replicas` and I think it is better to leave it to another ticket. The change is not big but I hope to make this PR more concentrated on CI workflow improvement.

.github/workflows/ci.yml
bgeng777: Yes, it is a good suggestion, I believe it is fine to use `verify` on webhook as well. I have updated.

wangyang0918 (approved):
+1 for merging.

PR #176, opened 2022-04-20
[FLINK-27279] Extract common status interfaces

Aitozi:
This pr is meant to extract common status interfaces to simplify the ReconcileUtils

**The brief change log**

1. Introduce the `CommonView` and `SpecView` to describe the common part between the `FlinkDeployment` and `FlinkSessionJob`
2. Use the `ReconciliationStatus` for both `FlinkDeployment` and `FlinkSessionJob`

Verified by the current tests.


Aitozi:
I also try to make the ReconcileUtils to a new Object to handle the `reconciledSucceed`, `reconcileError` and `savepointSucceed`. But I have to pass the object everywhere like `controller` and `reconcilerFactory` and `reconciler`. So I keep use the static way but unify the handle for  `FlinkDeployment` and `FlinkSessionJob`

Please let me know if you have better suggestion cc @gyfora @wangyang0918 thanks

gyfora:
@Aitozi seems like the FlinkDeploymentStatus and FlinkSessionJobStatus also have a lot in common (the only diff is that FlinkDeploymentStatus has an extra JMDeploymentStatus field)

Aitozi:
@gyfora Thanks for your suggestion, I will try as you suggested

Aitozi:
Due to the java type erasure, I can't not completly eliminate the `FlinkDeploymentReconciliationStatus` and `FlinkSessionJobReconciliationStatus` because I can't get the class at runtime 😔.
The new interface `ReconcileTarget` is responsible for get the common view of the custom resource. If we have special reconcile update logic for different custom resource, we can extend the interface to make it happen. What do you think about the current implementation @gyfora

gyfora:
Thank you, I will check this later today :)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconcileTarget.java
```diff
@@ -0,0 +1,62 @@
...
+     * @return the current reconciliation status.
+     */
+    ReconciliationStatus<SPEC> getReconcileStatus();
+
+    /**
+     * Let the target custom resource handle the reconciliation error.
+     *
+     * @param error The error to be handled.
+     */
+    void handleError(@Nullable String error);
```
gyfora, line 61: I think these methods should be part of the CommonStatus interface instead. Simply have the ReonciliationStatus<SPEC> and error fields there and use get/set

gyfora: That way the ReconcileTerget interface could be also removed and use `CustomResource<SPEC extends SpecView, STATUS extends CommonStatus>`

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/status/ReconciliationStatus.java
```diff
@@ -54,19 +49,20 @@ public class ReconciliationStatus {
     private ReconciliationState state = ReconciliationState.DEPLOYED;
 
     @JsonIgnore
-    public FlinkDeploymentSpec deserializeLastReconciledSpec() {
-        return ReconciliationUtils.deserializedSpecWithVersion(
-                lastReconciledSpec, FlinkDeploymentSpec.class);
+    public abstract Class<SPEC> getSpecClass();
```
gyfora, line 52: You could also simply add a constructor that takes the spec class (you have that when you call initStatus by calling getSpec().getClass()). Then you would not need 2 subclasses just to implement this.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconcileTarget.java
Aitozi: Good suggestion, I have refactor based on your suggestion, please take a look again.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/status/ReconciliationStatus.java
Aitozi: I have tried this way, But if I add a field to record the spec class in `ReconciliationStatus` (making it `@JsonIgnore` ), It will cause the [CRDGenerator stack overflow ](https://github.com/fabric8io/kubernetes-client/issues/4092). So I kept it as it now

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/spec/SpecView.java
```diff
@@ -0,0 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.kubernetes.operator.crd.spec;
+
+/** The common view of the spec. */
+public interface SpecView {
```
gyfora, line 21: I think it would be better to call this SpecWithJob :)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/status/CommonStatus.java
```diff
@@ -0,0 +1,41 @@
...
+@NoArgsConstructor
+public abstract class CommonStatus<SPEC extends SpecView> {
+
+    /** Last observed status of the Flink job on Application/Session cluster. */
+    private JobStatus jobStatus = new JobStatus();
+
+    /** Error information about the FlinkDeployment/FlinkSessionJob. */
+    private String error;
+
+    public abstract ReconciliationStatus<SPEC> getReconciliationStatus();
```
gyfora, line 40: Missing javadoc comment

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/spec/SpecView.java
gyfora, line 21: Could this be an abstract class instead called `FlinkSpecBase` or `AbstractFlinkSpec` which could contain the job field (with the lombok annotation) that both deployment and sessionjobspec would extend?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/status/ReconciliationStatus.java
gyfora: interesting, I will take a look at this later :)

docs/content/docs/custom-resource/reference.md
```diff
@@ -160,26 +160,41 @@ This page serves as a full reference for FlinkDeployment custom resource definit
...
 
+### CommonStatus
+**Class**: org.apache.flink.kubernetes.operator.crd.status.CommonStatus
+
+**Description**: Last observed common status of the Flink deployment/Flink SessionJob.
+
+| Parameter | Type | Docs |
+| ----------| ---- | ---- |
+| jobStatus | org.apache.flink.kubernetes.operator.crd.status.JobStatus | Last observed status of the Flink job on Application/Session cluster. |
+| error | java.lang.String | Error information about the FlinkDeployment/FlinkSessionJob. |
```
gyfora, line 171: The inheritence structure breaks the (not very sophisticated) doc generation logic. We should either fix it here or open a blocker jira to track this

gyfora (commented):
Added some last minor comments. Thanks @Aitozi it looks really good now.

docs/content/docs/custom-resource/reference.md
Aitozi: I also confused here, what‘s the expect content here? Do we need to display the `FlinkDeploymentReconciliationStatus` fileds include (in the parent) directly and skip the abstract classes

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/status/CommonStatus.java
Aitozi: fixed

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/spec/SpecView.java
Aitozi: Nice suggestion, It looks more naturally now :).

docs/content/docs/custom-resource/reference.md
Aitozi: I would open another PR after this to fix this problem soon .

gyfora: I think from the users perspective it would be best to skip any abstract classes and only show the “child/leaf” classes in the hierarchy with all fields from the superclasses .

Aitozi: Get, I will try to make up the content in this way

PR #177, opened 2022-04-21
[FLINK-27303][FLINK-27309] Introduce FlinkConfigManager for efficient config management

gyfora:
This PR improves/reworks the configuration management of the oprator. Problems with the current approach:
  - Configuration and FlinkOperatorConfigurations are passed around everywhere
  - Every component generates effective configurations on their own (without consistency which can cause problems)
  - Configs are generated again and again creating a large number of temporary files in every reconcile iteration
  - Configuration is completely static and impossible to change without operator restart
  
To try to tackle all these problems at once I have introduced the `FlinkConfigManager` object which is responsible for:
 - Generating configuration for deployment / cluster interactions (observe)
 - The config generation logic is consistently aware of currently deployed spec (this is important due to rollback logic)
 - Caching the generated configs for a given specification to speed things up and avoid extra tmp files
 - Allow modifying default configurations on the fly by watching ConfigMap changes
 
While these changes touch almost all components, it is mostly cosmetic. The core improvements are located in the `FlinkConfigManager` class and which config is requested in the different parts of the reconcile logic.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
@@ -21,6 +21,7 @@ import org.apache.flink.configuration.ConfigOption;
 import org.apache.flink.configuration.Configuration;
 import org.apache.flink.configuration.CoreOptions;
 import org.apache.flink.configuration.DeploymentOptions;
+import org.apache.flink.configuration.DeploymentOptionsInternal;
 import org.apache.flink.configuration.JobManagerOptions;
 import org.apache.flink.configuration.MemorySize;
 import org.apache.flink.configuration.PipelineOptions;
@@ -32,10 +33,13 @@ import org.apache.flink.kubernetes.operator.crd.spec.FlinkDeploymentSpec;
 import org.apache.flink.kubernetes.operator.crd.spec.Resource;
 import org.apache.flink.kubernetes.operator.crd.spec.UpgradeMode;
 import org.apache.flink.streaming.api.environment.ExecutionCheckpointingOptions;
+import org.apache.flink.util.FileUtils;
 import org.apache.flink.util.StringUtils;
 
 import io.fabric8.kubernetes.api.model.Pod;
 import io.fabric8.kubernetes.client.internal.SerializationUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.io.IOException;
@@ -54,13 +58,17 @@ import static org.apache.flink.kubernetes.utils.Constants.CONFIG_FILE_LOGBACK_NA
 
 /** Builder to get effective flink config from {@link FlinkDeployment}. */
 public class FlinkConfigBuilder {
+
+    private static final Logger LOG = LoggerFactory.getLogger(FlinkConfigBuilder.class);
+
+    protected static final String GENERATED_FILE_PREFIX = "flink_op_generated_";
+    protected static final Duration DEFAULT_CHECKPOINTING_INTERVAL = Duration.ofMinutes(5);
+
     private final String namespace;
     private final String clusterId;
     private final FlinkDeploymentSpec spec;
     private final Configuration effectiveConfig;
     private final boolean forDeployment;
-
-    protected static final Duration DEFAULT_CHECKPOINTING_INTERVAL = Duration.ofMinutes(5);
 
     protected FlinkConfigBuilder(
             FlinkDeployment deployment, Configuration flinkConfig, boolean forDeployment) {
@@ -287,7 +295,7 @@ public class FlinkConfigBuilder {
 
     private static String createLogConfigFiles(String log4jConf, String logbackConf)
             throws IOException {
-        File tmpDir = Files.createTempDirectory("conf").toFile();
+        File tmpDir = Files.createTempDirectory(GENERATED_FILE_PREFIX + "conf_").toFile();
 
         if (log4jConf != null) {
             File log4jConfFile = new File(tmpDir.getAbsolutePath(), CONFIG_FILE_LOG4J_NAME);
@@ -303,9 +311,37 @@ public class FlinkConfigBuilder {
     }
 
     private static String createTempFile(Pod podTemplate) throws IOException {
-        final File tmp = File.createTempFile("podTemplate_", ".yaml");
+        final File tmp = File.createTempFile(GENERATED_FILE_PREFIX + "podTemplate_", ".yaml");
         Files.write(tmp.toPath(), SerializationUtils.dumpAsYaml(podTemplate).getBytes());
         tmp.deleteOnExit();
         return tmp.getAbsolutePath();
+    }
+
+    protected static void cleanupTmpFiles(Configuration configuration) {
+        configuration
+                .getOptional(KubernetesConfigOptions.KUBERNETES_POD_TEMPLATE)
+                .ifPresent(FlinkConfigBuilder::deleteSilentlyIfGenerated);
+        configuration
+                .getOptional(KubernetesConfigOptions.JOB_MANAGER_POD_TEMPLATE)
+                .ifPresent(FlinkConfigBuilder::deleteSilentlyIfGenerated);
+        configuration
+                .getOptional(KubernetesConfigOptions.TASK_MANAGER_POD_TEMPLATE)
+                .ifPresent(FlinkConfigBuilder::deleteSilentlyIfGenerated);
+        configuration
+                .getOptional(DeploymentOptionsInternal.CONF_DIR)
+                .ifPresent(FlinkConfigBuilder::deleteSilentlyIfGenerated);
+    }
+
+    private static void deleteSilentlyIfGenerated(String file) {
+        try {
+            File localFile = new File(file);
+            if (!localFile.getName().startsWith(FlinkConfigBuilder.GENERATED_FILE_PREFIX)) {
+                return;
+            }
+            LOG.debug("Deleting tmp config file {}", localFile);
+            FileUtils.deleteFileOrDirectory(localFile);
+        } catch (Exception err) {
+            LOG.error("Could not clean up file " + file, err);
+        }
     }
 }
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
@@ -68,29 +68,24 @@ public class FlinkConfigBuilder {
     private final String clusterId;
     private final FlinkDeploymentSpec spec;
     private final Configuration effectiveConfig;
-    private final boolean forDeployment;
 
-    protected FlinkConfigBuilder(
-            FlinkDeployment deployment, Configuration flinkConfig, boolean forDeployment) {
+    protected FlinkConfigBuilder(FlinkDeployment deployment, Configuration flinkConfig) {
         this(
                 deployment.getMetadata().getNamespace(),
                 deployment.getMetadata().getName(),
                 deployment.getSpec(),
-                flinkConfig,
-                forDeployment);
+                flinkConfig);
     }
 
     protected FlinkConfigBuilder(
             String namespace,
             String clusterId,
             FlinkDeploymentSpec spec,
-            Configuration flinkConfig,
-            boolean forDeployment) {
+            Configuration flinkConfig) {
         this.namespace = namespace;
         this.clusterId = clusterId;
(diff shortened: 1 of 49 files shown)


gyfora:
cc @Aitozi @bgeng777 @morhidi @tweise @wangyang0918

wangyang0918:
Just a quick note: instead of creating a watch on the ConfigMap, I prefer to simply use `ScheduledExecutorService#scheduleAtFixedRate` to detect default configuration file change on the fly. It is similar to `monitorInterval` in log4j. Because naked watch is not stable and could be closed in many cases, we need to use informer instead. But the informer might be an over-kill.

gyfora:
> Just a quick note: instead of creating a watch on the ConfigMap, I prefer to simply use `ScheduledExecutorService#scheduleAtFixedRate` to detect default configuration file change on the fly. It is similar to `monitorInterval` in log4j. Because naked watch is not stable and could be closed in many cases, we need to use informer instead. But the informer might be an over-kill.

Makes sense @wangyang0918 thats a simple change to make it more robust

gyfora:
Reworked the watcher logic and added some more tests

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
```diff
@@ -55,31 +54,45 @@
...
+            String namespace,
+            String clusterId,
+            FlinkDeploymentSpec spec,
+            Configuration flinkConfig,
+            boolean forDeployment) {
+        this.namespace = namespace;
+        this.clusterId = clusterId;
         this.spec = spec;
         this.effectiveConfig = new Configuration(flinkConfig);
+        this.forDeployment = forDeployment;
```
Aitozi, line 85: Do we have to add comments for `forDeployment`. It seems mainly to save the create temp template file Right ?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
```diff
@@ -177,19 +192,21 @@ public FlinkConfigBuilder applyJobManagerSpec() throws IOException {
...
+    protected FlinkConfigBuilder applyTaskManagerSpec() throws IOException {
         if (spec.getTaskManager() != null) {
             setResource(spec.getTaskManager().getResource(), effectiveConfig, false);
-            setPodTemplate(
-                    spec.getPodTemplate(),
-                    spec.getTaskManager().getPodTemplate(),
-                    effectiveConfig,
-                    false);
+            if (forDeployment) {
+                setPodTemplate(
```
Aitozi, line 199: The temp file may still keep for a long time during the operator running, because it clean up relies on the `deleteOnExit`. Can we expire the temp files when the cache are expired ?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigManager.java
```diff
@@ -0,0 +1,214 @@
...
+        KubernetesClient kubeClient;
+
+        public ConfigUpdater(KubernetesClient kubeClient) {
+            this.kubeClient = kubeClient;
+        }
+
+        public void run() {
+            try {
+                LOG.debug("Checking for config update changes...");
+                var newConf = readConfiguration(kubeClient);
```
Aitozi, line 200: Why not read the config directly from the mounted dir. I think we do not have to interact with the configMap directly. Same at the the `FlinkConfigManager` construct

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
gyfora: I will try to add that, good idea

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigManager.java
gyfora: Are the mounted files updated automatically if the configmap changes?

Aitozi: As I know, it is

gyfora: nice, that will make this much simpler I agree

gyfora: I fixed this and verified that it works correctly

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
Aitozi: I think over it again, It may bring the sync problem if the config is just used to deploy and the async replace is scheduled. Maybe the cache lifecycle can be align to the job lifecycle, which means we can cleanup the cache and config in the cleanup method ?  WDYT ?

gyfora: I have added cleanup for the temp files. Also with this I removed the `forDeployment` logic because it is not really necessary anymore (temp files will be cleaned up anyways.)

Set the expiration after access to 1 hour to avoid any problems with cleanup during deployment.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
gyfora: I removed this already

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
gyfora: But you are right this could still be a problem if the clean up is triggered based on the configmap update, I will think about it a little

gyfora: @Aitozi I have pushed a fix for this, instead of invalidating the cache I bump a config version long field. This will avoid race conditions between default config updates and temp file cleanup

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigManager.java
```diff
@@ -0,0 +1,183 @@
...
+                    key.f2,
+                    objectMapper.convertValue(key.f3, FlinkDeploymentSpec.class),
+                    defaultConfig);
+        } catch (Exception e) {
+            throw new RuntimeException("Failed to load configuration", e);
+        }
+    }
+
+    private void scheduleConfigWatcher() {
+        var checkInterval = defaultConfig.get(OPERATOR_DYNAMIC_CONFIG_CHECK_INTERVAL);
```
Aitozi, line 154: nit: we can directly use `defaultConfig.get(OPERATOR_DYNAMIC_CONFIG_CHECK_INTERVAL).toMillis();`

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigManager.java
```diff
@@ -0,0 +1,183 @@
...
+    private static final ObjectMapper objectMapper = new ObjectMapper();
+
+    private static final int MAX_CACHE_SIZE = 1000;
+    private static final Duration CACHE_TIMEOUT = Duration.ofMinutes(30);
+
+    private volatile Configuration defaultConfig;
+    private volatile FlinkOperatorConfiguration operatorConfiguration;
+    private final AtomicLong defaultConfigVersion = new AtomicLong(0);
+
+    private final LoadingCache<Tuple4<Long, String, String, ObjectNode>, Configuration> cache;
```
Aitozi, line 64: Can we replace the key to a simple inner class instead of `Tuple4`, it may make the meaning more clear

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
Aitozi: It looks good now, +1 for this otherwise two other minor comments.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigManager.java
gyfora: fixed

Aitozi (approved):
+1

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigManager.java
```diff
@@ -0,0 +1,193 @@
...
+/** Configuration manager for the Flink operator. */
+public class FlinkConfigManager {
+
+    private static final Logger LOG = LoggerFactory.getLogger(FlinkConfigManager.class);
+    private static final ObjectMapper objectMapper = new ObjectMapper();
+
+    private static final int MAX_CACHE_SIZE = 1000;
+    private static final Duration CACHE_TIMEOUT = Duration.ofMinutes(30);
+
+    private volatile Configuration defaultConfig;
```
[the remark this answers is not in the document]
gyfora: it's the default config onto which the `FlinkDeployment` flinkConfiguration sections are applied.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractDeploymentReconciler.java
```diff
@@ -39,27 +39,22 @@ public abstract class AbstractDeploymentReconciler implements Reconciler<FlinkDe
...
-        this.operatorConfiguration = operatorConfiguration;
-        this.defaultConfig = defaultConfig;
+        this.configManager = configManager;
     }
 
     @Override
     public DeleteControl cleanup(FlinkDeployment flinkApp, Context context) {
-        Configuration effectiveConfig = FlinkUtils.getEffectiveConfig(flinkApp, defaultConfig);
-        return shutdownAndDelete(flinkApp, effectiveConfig);
+        return shutdownAndDelete(flinkApp, configManager.getObserveConfig(flinkApp));
```
bgeng777, line 57: The older code builds config based on current `flinkApp` but IIUC, here `getObserveConfig`  will use `lastReconciledSpec` or `lastStableSpec`. Is this change on purpose?

gyfora: I tried to consistently change it everywhere where we interact with the running job. Because that was technically deployed with either the lastReconciled or lastStable spec as you see it here.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/sessionjob/FlinkSessionJobReconciler.java
```diff
@@ -119,38 +116,39 @@ public void reconcile(FlinkSessionJob flinkSessionJob, Context context) throws E
                 if (desiredJobState == JobState.RUNNING) {
                     LOG.info("Upgrading/Restarting running job, suspending first...");
                 }
-                stateAfterReconcile = suspendJob(flinkSessionJob, upgradeMode, effectiveConfig);
+                stateAfterReconcile = suspendJob(flinkSessionJob, upgradeMode, deployedConfig);
             }
             if (currentJobState == JobState.SUSPENDED && desiredJobState == JobState.RUNNING) {
                 if (upgradeMode == UpgradeMode.STATELESS) {
-                    submitAndInitStatus(flinkSessionJob, effectiveConfig, null);
+                    submitAndInitStatus(flinkSessionJob, deployedConfig, null);
```
bgeng777, line 123: I think here we should use `getDeployConfig` to submit a new job instead of using `getObserveConfig`?

gyfora: I think this is correct, we need to get the config that is running on the cluster at the moment. Which is the observeConfig

PR #183, opened 2022-04-25
[FLINK-27397] Improve the CrdReferenceDoclet to handle the abstract c…

Aitozi:
…lass


gyfora:
This is great :) works like a charm 👍

PR #186, opened 2022-04-27
[FLINK-27425] Default job service account needs access to services to work with Kafka

mnuttall:
As per https://issues.apache.org/jira/browse/FLINK-27425 this resolves the error, 

```
[ERROR] Could not execute SQL statement. Reason:
io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: GET at: https://10.96.0.1/api/v1/namespaces/default/services/basic-example-rest. Message: Forbidden!Configured service account doesn't have access. Service account may have been revoked. services "basic-example-rest" is forbidden: User "system:serviceaccount:default:flink" cannot get resource "services" in API group "" in the namespace "default".
```

seen when running a Flink job that uses the Kafka connector against a default install.

### DIFF
--- a/helm/flink-kubernetes-operator/templates/rbac.yaml
+++ b/helm/flink-kubernetes-operator/templates/rbac.yaml
@@ -82,6 +82,7 @@ rules:
     resources:
       - pods
       - configmaps
+      - services
     verbs:
       - '*'
   - apiGroups:


morhidi:
Can you please attach the stack trace? I don't what is trying to access the `basic-example-rest` service

mnuttall:
Hi @morhidi, after further testing, I only see this problem in a very limited situation:

- create a session cluster using a FlinkDeployment with no `spec.job` stanza
- exec into the cluster pod
- run bin/sql-client.sh
- create a TABLE backed by a kafka topic and execute a SELECT against it

At this point, the RBAC error shown above appears inline within the sql-client, but no exceptions appear either within the sql-client or the pod logs.

However, the Role does not need to be changed in order for the same to work on an application cluster with the SQL submitted via the Table API.

Should it be possible to exec into a session cluster's job manager and run and submit jobs via the sql-client?

morhidi:
Running the sql client from your host against an exposed session cluster should work I guess. Is that something you've tried already?

Aitozi:
The directly submitted SQL scripts will not managed by the operator.

Aitozi:
@mnuttall I think what you need is the ability to submit SQL job to the session cluster, Right ? If yes, there is an ongoing [ticket](https://issues.apache.org/jira/browse/FLINK-27009) tracking this now , you can take a look there.

mnuttall:
Update - yes, I've connected an sql-client to a session cluster and successfully run jobs against it, using Kafka. 

```
apiVersion: flink.apache.org/v1alpha1
kind: FlinkDeployment
metadata:
  namespace: default
  name: flink-test
spec:
  image: flink:1.14.4-scala_2.11-java11
  flinkVersion: v1_14
  flinkConfiguration:
  serviceAccount: flink
  jobManager:
    replicas: 1
    resource:
      memory: "2048m"
      cpu: 1
  taskManager:
    resource:
      memory: "2048m"
      cpu: 1
```

I'm going to close this PR and issue since it's not enough to get the sql-client _within_ the session cluster to work, and not required for application clusters.

PR #187, opened 2022-04-28
[FLINK-27443] Create standalone mode parameters and decorators for JM and TMs

usamj:
This PR creates decorators and parameters which contain information needed to create JM and TM deployments in standalone mode. These will then be used by the Standalone ClusterDescriptor to actually create the resources needed for a standalone Flink Cluster. 

The `flink-kubernetes-standalone-cluster` is introduced as the parameters and decorators integrate with their similar counterparts in the main Flink package which use a different Fabric8 version that is shaded. 

Note: This is a draft as the FLIP is still under discussion

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,6 +38,7 @@ ENV FLINK_HOME=/opt/flink
 ENV OPERATOR_VERSION=1.1-SNAPSHOT
 ENV OPERATOR_JAR=flink-kubernetes-operator-$OPERATOR_VERSION-shaded.jar
 ENV WEBHOOK_JAR=flink-kubernetes-webhook-$OPERATOR_VERSION-shaded.jar
+ENV STANDALONE_CLUSTER_JAR=flink-kubernetes-standalone-$OPERATOR_VERSION.jar
 ENV FLINK_KUBERNETES_SHADED_JAR=flink-kubernetes-shaded-$OPERATOR_VERSION.jar
 
 WORKDIR /flink-kubernetes-operator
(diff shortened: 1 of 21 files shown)


morhidi:
@wangyang0918 @gyfora @mbalassi can you please kick the WF process, it's pending approval? Shall we wait for the 1.0.0 release cut before adding this feature?

gyfora:
@usamj i think we need to vote on the flip before we can start with this properly

gyfora:
Could you start the voting thread?

usamj:
Sure, I've created the voting thread

flink-kubernetes-standalone-cluster/src/main/java/org/apache/flink/kubernetes/operator/kubeclient/decorators/UserLibMountDecorator.java
```diff
@@ -0,0 +1,81 @@
...
+
+import static org.apache.flink.util.Preconditions.checkNotNull;
+
+/**
+ * Mount the Flink User Lib directory to enable Flink to pick up a Jars defined in
+ * pipeline.classpaths. Used for starting standalone application clusters
+ */
+public class UserLibMountDecorator extends AbstractKubernetesStepDecorator {
+    private static final String USER_LIB_VOLUME = "user-lib-dir";
+    private static final String USER_LIB_PATH = "/opt/flink/usrlib";
```
gyfora, line 39: Wouldn't this simply come from the user provided image?

gyfora (commented):
In general there seem to be some overlap with the decorators that already exist in Flink. It will be good to see this in use :) 

I left some minor comments.

flink-kubernetes-standalone-cluster/src/main/java/org/apache/flink/kubernetes/operator/kubeclient/decorators/UserLibMountDecorator.java
usamj: This can come from user provided image but the images in the flink docker don't have this directory created. The reason it needs to be created is that if it isn't created then configuration defined user classpaths are ignored (https://github.com/apache/flink/blob/b904c948fb17f57b81a8e794cc67718c432bbcaa/flink-clients/src/main/java/org/apache/flink/client/program/DefaultPackagedProgramRetriever.java#L119)

gyfora (changes requested):
Added a minor comment to simplify module naming, and after that I think we can merge this

usamj:
Updated the module names

gyfora:
This can be closed I think, it is included in your other PR

PR #189, opened 2022-04-28
[FLINK-27422] Do not create temporary pod template files for JobManager and TaskManager if not configured explicitly

SteNicholas:
It doesn't need to create temporary pod template files for JobManager and TaskManager if it is not configured explicitly via `.spec.JobManagerSpec.podTemplate` or `.spec.TaskManagerSpec.podTemplate`.

**The brief change log**

- `FlinkConfigBuilder#setPodTemplate` adds the check to create temporary pod template files for JobManager and TaskManager when it is configured explicitly via `.spec.JobManagerSpec.podTemplate` or `.spec.TaskManagerSpec.podTemplate`.


SteNicholas:
@wangyang0918 , PTAL.

PR #191, opened 2022-05-03
[FLINK-27303] Improve config cache settings + add cleanup

gyfora:
This adds the missing cleanup logic + makes size/expiration configurable

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigManager.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigManager.java
@@ -43,6 +43,7 @@ import org.slf4j.LoggerFactory;
 import java.time.Duration;
 import java.util.Set;
 import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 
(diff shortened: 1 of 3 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigManager.java
```diff
@@ -87,8 +89,15 @@ public Configuration load(Key k) {
                                 });
 
         updateDefaultConfig(defaultConfig);
+        ScheduledExecutorService executorService = Executors.newSingleThreadScheduledExecutor();
+        executorService.scheduleWithFixedDelay(
```
wangyang0918, line 93: Calling the `Cache#cleanUp()` periodically and cleaning up the tmp files in the `removalListener` could fix the issue in FLINK-27303. But actually the pod template, config, log tmp files are useless after the successful submission.

Do you think we could clean up the tmp files more aggressive? Imagine the operator is managing thousand of jobs, the current implementation will still have so many tmp files.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigManager.java
```diff
@@ -87,8 +89,15 @@ public Configuration load(Key k) {
                                 });
 
         updateDefaultConfig(defaultConfig);
+        ScheduledExecutorService executorService = Executors.newSingleThreadScheduledExecutor();
```
wangyang0918, line 92: It seems that the `executorService` is never shutdown. It might not be a major problem because the `FlinkConfigManager` has the same lifecycle with whole process.

gyfora: I thought about this but the operator doesn't really offer lifecycle proper lifecycle methods.

wangyang0918 (approved):
+1 for merging

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigManager.java
wangyang0918: You are right. We could not simply delete the tmp files in case of roll-back.

PR #193, opened 2022-05-04
[FLINK-27458] Expose allowNonRestoredState flag in JobSpec

Aitozi:
This PR introduces the `allowNonRestoredState` in `JobSpec` to let the session job can enable/disable it individually.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
@@ -217,8 +217,10 @@ public class FlinkConfigBuilder {
                 effectiveConfig.set(
                         CoreOptions.DEFAULT_PARALLELISM, spec.getJob().getParallelism());
             }
-            if (spec.getJob().isAllowNonRestoredState()) {
-                effectiveConfig.set(SavepointConfigOptions.SAVEPOINT_IGNORE_UNCLAIMED_STATE, true);
+            if (spec.getJob().getAllowNonRestoredState() != null) {
+                effectiveConfig.set(
+                        SavepointConfigOptions.SAVEPOINT_IGNORE_UNCLAIMED_STATE,
+                        spec.getJob().getAllowNonRestoredState());
             }
         } else {
             effectiveConfig.set(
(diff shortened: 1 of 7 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/spec/JobSpec.java
```diff
@@ -65,4 +65,7 @@ public class JobSpec {
 
     /** Upgrade mode of the Flink job. */
     @EqualsAndHashCode.Exclude private UpgradeMode upgradeMode = UpgradeMode.STATELESS;
+
+    /** Allow checkpoint state that cannot be mapped to any job vertex in tasks. */
+    @EqualsAndHashCode.Exclude private boolean allowNonRestoredState = false;
```
gyfora, line 70: I think we should not exclude this from the equals comparison

Aitozi: I think this option only affect the next upgrade behavior whether to recover from the unknown state mapping and do not have to trigger a direct upgrade after user change this. So I think it should be same with the upgradeMode.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
```diff
@@ -223,6 +224,15 @@ protected FlinkConfigBuilder applyJobOrSessionSpec() throws URISyntaxException {
         return this;
     }
 
+    protected FlinkConfigBuilder applyAllowNonRestoredState() {
+        if (spec.getJob() != null) {
```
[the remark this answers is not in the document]
Aitozi: Fixed

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
```diff
@@ -216,6 +217,9 @@ protected FlinkConfigBuilder applyJobOrSessionSpec() throws URISyntaxException {
                 effectiveConfig.set(
                         CoreOptions.DEFAULT_PARALLELISM, spec.getJob().getParallelism());
             }
+            if (spec.getJob().isAllowNonRestoredState()) {
+                effectiveConfig.set(SavepointConfigOptions.SAVEPOINT_IGNORE_UNCLAIMED_STATE, true);
```
gyfora, line 221: I think we could make the `allowNonRestoredState` Boolean instead of boolean and set it into the config regardless of the value if it's not null. 

With the current logic if the user defines true in the flinkConfiguration and sets it to false in the jobspec we would not use it.

gyfora: we could also add a test for this to make sure the spec takes precedence

Aitozi: Good point, fixed

PR #196, opened 2022-05-06
[FLINK-27500] Validation errors should not block reconciliation

gyfora:
This PR solves the issue detailed in https://issues.apache.org/jira/browse/FLINK-27500 by reconciling using the previous (valid) spec if validation fails for some reason.

This allows rollback/jm-recovery operations to proceed even if an invalid CR was submitted in thea meantime

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
@@ -43,6 +43,8 @@ import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.fasterxml.jackson.databind.node.TextNode;
 import io.fabric8.kubernetes.client.CustomResource;
 import io.javaoperatorsdk.operator.api.reconciler.UpdateControl;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nullable;
 
@@ -52,6 +54,8 @@ import java.util.Objects;
 
 /** Reconciliation utilities. */
 public class ReconciliationUtils {
+
+    private static final Logger LOG = LoggerFactory.getLogger(ReconciliationUtils.class);
 
     private static final ObjectMapper objectMapper = new ObjectMapper();
 
@@ -66,14 +70,20 @@ public class ReconciliationUtils {
 
         var clonedSpec = ReconciliationUtils.clone(spec);
         AbstractFlinkSpec lastReconciledSpec = reconciliationStatus.deserializeLastReconciledSpec();
+        boolean upgrading = false;
         if (lastReconciledSpec != null && lastReconciledSpec.getJob() != null) {
             Long oldSavepointTriggerNonce = lastReconciledSpec.getJob().getSavepointTriggerNonce();
             clonedSpec.getJob().setSavepointTriggerNonce(oldSavepointTriggerNonce);
+            if (clonedSpec.getJob().getState() == JobState.RUNNING
+                    && stateAfterReconcile == JobState.SUSPENDED) {
+                upgrading = true;
+            }
             clonedSpec.getJob().setState(stateAfterReconcile);
         }
         reconciliationStatus.serializeAndSetLastReconciledSpec(clonedSpec);
         reconciliationStatus.setReconciliationTimestamp(System.currentTimeMillis());
-        reconciliationStatus.setState(ReconciliationState.DEPLOYED);
+        reconciliationStatus.setState(
+                upgrading ? ReconciliationState.UPGRADING : ReconciliationState.DEPLOYED);
 
         if (spec.getJob() != null && spec.getJob().getState() == JobState.SUSPENDED) {
             // When a job is suspended by the user it is automatically marked stable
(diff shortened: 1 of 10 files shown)


gyfora:
cc @Aitozi @morhidi

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
```diff
@@ -255,4 +257,34 @@ public static boolean shouldRollBack(
...
+        ReconciliationUtils.updateForReconciliationError(deployment, newErr);
+
+        var lastReconciledSpec =
+                deployment.getStatus().getReconciliationStatus().deserializeLastReconciledSpec();
+        if (lastReconciledSpec == null) {
+            // Validation failed before anything was deployed, nothing to do
+            return false;
+        } else {
+            // We need to observe/reconcile using the last version of the deployment spec
+            deployment.setSpec(lastReconciledSpec);
```
Aitozi, line 281: This means we will change the spec when it generates the validation error, will this break the principle that not change the spec directly ?

Aitozi: I have a doubt that can we enforce user to use the operator with the webhook? I think it will make the things much easier, because it will not apply the unexpected spec succeed. Do you know whether there is strong reason for users to disable the webhook?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
```diff
@@ -255,4 +257,34 @@ public static boolean shouldRollBack(
                 .minus(readinessTimeout)
                 .isAfter(Instant.ofEpochMilli(reconciliationStatus.getReconciliationTimestamp()));
     }
+
+    public static <
+                    SPEC extends AbstractFlinkSpec,
+                    STATUS extends CommonStatus<SPEC>,
+                    CR extends CustomResource<SPEC, STATUS>>
+            boolean applyValidationErrorAndResetSpec(
```
Aitozi, line 265: nit: for the return `boolean` I think we'd better document for it, Otherwise we have to read the code to know what it mean

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
```diff
@@ -62,14 +66,20 @@ void updateForSpecReconciliationSuccess(
 
         var clonedSpec = ReconciliationUtils.clone(spec);
         AbstractFlinkSpec lastReconciledSpec = reconciliationStatus.deserializeLastReconciledSpec();
+        boolean upgrading = false;
         if (lastReconciledSpec != null && lastReconciledSpec.getJob() != null) {
             Long oldSavepointTriggerNonce = lastReconciledSpec.getJob().getSavepointTriggerNonce();
             clonedSpec.getJob().setSavepointTriggerNonce(oldSavepointTriggerNonce);
+            if (clonedSpec.getJob().getState() == JobState.RUNNING
```
Aitozi, line 73: What about add a method to judge whether there is on-going upgrading ?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
gyfora: The spec is never changed. We always change it back to the original one and we only update the status.

that’s the whole idea :)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
```diff
@@ -112,21 +122,11 @@ public static <T> T clone(T object) {
     public static <CR extends CustomResource> UpdateControl<CR> toUpdateControl(
             CR originalCopy, CR current) {
 
-        UpdateControl<CR> updateControl;
-        if (!Objects.equals(originalCopy.getSpec(), current.getSpec())) {
-            throw new UnsupportedOperationException(
-                    "Detected spec change after reconcile, this probably indicates a bug.");
-        }
+        current.setSpec(originalCopy.getSpec());
```
gyfora, line 125: @Aitozi this is what I was referring to

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
gyfora: I wish we could always use the webhook but even in our environments it’s not always possible due to access limitations.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
Aitozi: thanks, I get your meaning now

Aitozi (approved):
LGTM

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
Aitozi: Get it

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
gyfora: I would prefer to keep it this way for now as it's only a simple enum check

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
```diff
@@ -305,4 +307,46 @@ public static boolean isJobRunning(FlinkDeploymentStatus status) {
...
+     * @param deployment The current deployment to be reconciled
+     * @param validationError Validation error encountered for the current spec
+     * @return True if the spec was reset and reconciliation can continue. False if nothing to
+     *     reconcile.
+     */
+    public static <
+                    SPEC extends AbstractFlinkSpec,
+                    STATUS extends CommonStatus<SPEC>,
+                    CR extends CustomResource<SPEC, STATUS>>
+            boolean applyValidationErrorAndResetSpec(CR deployment, String validationError) {
```
Aitozi, line 329: Hi @gyfora, I have another concern here. For the sessionJob, if the validationError comes from the change by the session cluster spec, this solution may not work, Right ?

gyfora: Could you give me an example where this would not work?

PR #198, opened 2022-05-09
[FLINK-27261] Disable `web.cancel.enable` for session cluster

SteNicholas:
In [FLINK-27154](https://issues.apache.org/jira/browse/FLINK-27154), we disable `web.cancel.enable` for application cluster. We should also do this for session cluster.

**The brief change log**

- Sets `web.cancel.enable` to false at default in `FlinkConfigBuilder` to avoid users accidentally cancelling jobs.


SteNicholas:
> Could we also disable web submission as part of this PR? I think that would be nice :)

@gyfora I have also disabled web submission but the `test_sessionjob_ha.sh` e2e test has compatibility problem in Flink 1.13.

SteNicholas:
@gyfora, the behaviour that the web submission disables will cause the failure for the uploading and running the job. Hence I don't disable web submission in this PR.

gyfora:
Ok, thanks for checking :)

gyfora (approved):
LGTM 👍

wangyang0918 (approved):
+1 for merging.

PR #199, opened 2022-05-10
[FLINK-27551] Update status manually instead of relying on updatecontrol

gyfora:
This PR reworks how status updates of Flink resources are updated in kubernetes. This is necessary due to https://github.com/java-operator-sdk/java-operator-sdk/issues/1198

Based on offline discussion with the JOSDK team this seems to be the safest short term solution.

The caching of the latest status is necessary to avoid race condition between spec updates and status patches that are made by us.

Tests have been updated with some init logic to enable patching them through the kubernetes client + removed the reliance of mutating status.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
@@ -18,8 +18,10 @@
 package org.apache.flink.kubernetes.operator.controller;
 
 import org.apache.flink.api.common.JobStatus;
+import org.apache.flink.api.java.tuple.Tuple2;
 import org.apache.flink.kubernetes.operator.config.FlinkConfigManager;
 import org.apache.flink.kubernetes.operator.crd.FlinkDeployment;
+import org.apache.flink.kubernetes.operator.crd.status.FlinkDeploymentStatus;
 import org.apache.flink.kubernetes.operator.crd.status.JobManagerDeploymentStatus;
 import org.apache.flink.kubernetes.operator.exception.DeploymentFailedException;
 import org.apache.flink.kubernetes.operator.exception.ReconciliationException;
@@ -49,6 +51,7 @@ import org.slf4j.LoggerFactory;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 
 /** Controller that runs the main reconcile loop for Flink deployments. */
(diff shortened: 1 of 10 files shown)


gyfora:
cc @morhidi @wangyang0918

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
```diff
@@ -86,26 +92,29 @@ public FlinkDeploymentController(
     @Override
     public DeleteControl cleanup(FlinkDeployment flinkApp, Context context) {
         LOG.info("Deleting FlinkDeployment");
+        OperatorUtils.updateStatusFromCache(flinkApp, statusCache);
```
morhidi, line 95: @gyfora This means to me we never give a chance to update the status outside the operator, correct?

gyfora: I think the user cannot update the status either. But in a way yes this means that only the Flink Operator can update the status that is correct.

Later when this is improved in the java operator sdk (version 3 ) we could rework this if we think it's important but this should not be a serious limiation.

wangyang0918:
I want to confirm what is the long term solution. Do we want to keep the manual status update or leverage the java-operator-sdk v3 to avoid the conflicts?

The manual status update has a downside that the CR status could be inconsistent the real status. I think that is why we need the `statusCache` and have to update the CR status before every reconciliation. I just feel that it makes things more complicated. I know it also has the upside that we could update status in any time and do not need to re-trigger reconciliation. So I am not pretty sure which is the better one.

gyfora:
Thanks for raising this @wangyang0918 
To be very specific here, the `statusCache` is needed because the java-operator-sdk-framework does not know about the status update for already "scheduled" reconcile operations so it might give us a stale one. The `statusCache` is always consistent with the real status.

I think the we need both features in the long run:
 1. Robust status updates
 2. Persist status mid-reconciliation
 
V3 of the operator sdk will definitely solve 1. and they might actually solve 2. eventually for us. There is a larger design change that we should do after the release  to help with 2:

Instead of storing non-user-facing information into the status we could use an extra configmap per FlinkDeployment. This is the supported approach by the operator framework for storing peristent states for the reconciliation logic. This would already work today so we don't need to wait for any additional feature. Then we could simply move out things like lastReconciledSpec, reconciliationStatus etc that are only important for the operator and update them on the fly.

So to conclude my thoughts:

Long term solution would be a combination of moving to v3 and coming up with something nicer for the on-the-fly status updates which are necessary to keep some things consistent and clean.

wangyang0918:
Using the java-operator-sdk v3 for robust status updates and an extra ConfigMap per FlinkDeployment for the on-the-fly status updates makes a lot of sense to me.

Given that the robust status updates could be covered re-trigger a reconciliation. We could upgrade the java-operator-sdk version later. Why do we not introduce the extra ConfigMap now for the on-the-fly status updates?

gyfora:
@wangyang0918 I would not rush the ConfigMap change for the following reasons:

 - It's not completely clear to me what should go into status and what in the configmap
 - I am not even sure if we want to do this or not, status updates should be fine in theory and the operator framework might actually support on the fly status updates if we want
 
 I think this requires a FLIP and some careful consideration/feedback. We can introduce this from v1beta1 -> v1 if we want in a backward compatible way.

wangyang0918:
I get your point and will finish the review today. Let's keep an eye on the java-operator-sdk to see whether on-the-fly status updates could also be supported :)

wangyang0918 (approved):
Very nice PR. I have no more comments.

+1 for merging.

PR #200, opened 2022-05-10
[FLINK-27495] Observe last savepoint status directly from cluster

gyfora:
This improves the savepoint tracking logic by observing the last savepoint/checkpoint directly from the cluster for terminal job states.

This is required for the correct functioning of LAST_STATE and SAVEPOINT upgrades in terminal states:
 - Job fatally failed
 - Job failed
 - Operator got restarted during a savepoint upgrade

It also requires users to set checkpoint directory for savepoint/last-state modes + sets externalized checkpoints config by default. The checkpoint directory is necessary to get externally addressable checkpoints.

I also improved the Status patching flow by factoring it into a StatusHelper class which allows us to update the status also from the observer easily after observing a new savepoint. This is important for consistency of the operator logic


gyfora:
cc @wangyang0918 @tweise @morhidi

gyfora:
The CI failure seems to be related to Flink base version mismatch, I will have to iterate on this after the 1.15 version bump is merged

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
```diff
@@ -528,6 +538,71 @@ public void triggerSavepoint(
...
+                                                            NonPersistentMetadataCheckpointStorageLocation
+                                                                    .EXTERNAL_POINTER))
+                            .max(
+                                    Comparator.comparingLong(
+                                            cp ->
+                                                    cp.get(CheckpointStatistics.FIELD_NAME_ID)
+                                                            .asLong()))
+                            .map(
+                                    cp ->
+                                            new Savepoint(
```
wangyang0918, line 587: I am not pretty sure whether we could wrap the checkpoint as the `Savepoint` and store in the `.status.SavepointInfo`. It is a little misleading since savepoint is usually triggered by user manually and is consistent between Flink major versions.

gyfora: I decided to do this because when we record this we are actually in a special scenario. Terminal job, and it's only a checkpoint when the job failed/finished (otherwise it would be a savepoint due to stopwithsavepoint).

I decided to put it in savepointinfo for simplicity from the operator side to avoid introducing new status fields and keeping the logic simple.

The savpoint info in any case is not the real source of truth because anything can happen that prevents us from recording information so I think this is fair. With the savepoint history feature this will be improved further I believe

examples/basic-checkpoint-ha.yaml
```diff
@@ -21,11 +21,13 @@ kind: FlinkDeployment
 metadata:
   name: basic-checkpoint-ha-example
 spec:
+  serviceAccount: flink
```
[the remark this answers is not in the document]
gyfora: yea, I fixed this in parallel with someone else :)

wangyang0918 (approved):
Verified locally the checkpoint of failed job could be stored into `.status.jobStatus.savepointInfo`

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
wangyang0918: Given that the `.status.jobStatus.savepointInfo` not the real source of truth, it is reasonable to also store the checkpoint into this field.

It will be great if we could also describe this in the documentation :)

morhidi: I suggest we call it lastStateInfo, or add another field for the checkpoint.

wangyang0918: `lastStateInfo` looks better and then we could have a `snapshotType` field(e.g. checkpoint, savepoint).

gyfora: After an offline discussion with @morhidi we agreed to keep it as is, and work on moving this out from the status to a configmap (after the release) as it's very much internal at this point and not really meant to be for user use.

PR #201, opened 2022-05-10
[FLINK-27499] Bump base Flink version to 1.15.0

SteNicholas:
With the 1.15.0 release out, we should bump our Flink dependency to 1.15 if this does not interfere with the 1.14 compatibility.

**The brief change log**

- Bumps the `flink.version` in `pom.xml` to 1.15.0.
- Bumps the flink version of flink dependencies to 1.15.0.
- Fixs the conflicts of the callers for the `JarRunRequestBody`, `SavepointTriggerRequestBody` and. `ClusterClient`.


SteNicholas:
@gyfora , sorry for missing the copied configs. I have removed the copied configs `SHUTDOWN_ON_APPLICATION_FINISH` and `SUBMIT_FAILED_JOB_ON_APPLICATION_ERROR`. PTAL.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
```diff
@@ -192,7 +194,8 @@ private JarRunResponseBody runJar(
                             job.getParallelism() > 0 ? job.getParallelism() : null,
                             jobID,
                             job.getAllowNonRestoredState(),
-                            savepoint);
+                            savepoint,
+                            RestoreMode.DEFAULT);
```
gyfora, line 198: Based on the CI we should only set the RestoreMode if version is newer than 1.14 otherwise set null

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
```diff
@@ -349,7 +352,8 @@ public void cancelJob(FlinkDeployment deployment, UpgradeMode upgradeMode) throw
                                             .stopWithSavepoint(
                                                     Preconditions.checkNotNull(jobId),
                                                     false,
-                                                    savepointDirectory)
+                                                    savepointDirectory,
+                                                    SavepointFormatType.DEFAULT)
```
gyfora, line 356: Only set formatType if version newer than 1.14 otherwise set null

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
```diff
@@ -447,7 +451,11 @@ public Optional<String> cancelSessionJob(
                     try {
                         String savepoint =
                                 clusterClient
-                                        .stopWithSavepoint(jobID, false, savepointDirectory)
+                                        .stopWithSavepoint(
+                                                jobID,
+                                                false,
+                                                savepointDirectory,
+                                                SavepointFormatType.DEFAULT)
```
gyfora, line 458: Only set formatType if version newer than 1.14 otherwise set null

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
```diff
@@ -491,7 +499,11 @@ public void triggerSavepoint(
                             .sendRequest(
                                     savepointTriggerHeaders,
                                     savepointTriggerMessageParameters,
-                                    new SavepointTriggerRequestBody(savepointDirectory, false))
+                                    new SavepointTriggerRequestBody(
+                                            savepointDirectory,
+                                            false,
+                                            SavepointFormatType.DEFAULT,
```
gyfora, line 505: Only set formatType if version newer than 1.14 otherwise set null

gyfora (changes requested):
I think we need to make some adjustments for compatibility reasons :)

gyfora:
Probably the easiest would be to copy the request body class from Flink 1.15 and add the `@JsonInclude(Include.NON_NULL)` on the class and modify so that we are allowed to set null for pre 1.15 clusters

gyfora:
@SteNicholas let me know how I could help, it would be nice to merge this tomorrow

gyfora:
https://github.com/apache/flink-kubernetes-operator/pull/205

SteNicholas:
> Probably the easiest would be to copy the request body class from Flink 1.15 and add the `@JsonInclude(Include.NON_NULL)` on the class and modify so that we are allowed to set null for pre 1.15 clusters

Thanks for @gyfora to fork this pull request. IMO, the `SavepointTriggerRequestBody` and `StopWithSavepointRequestBody` are added temporarily to address compatibility problem for 1.15.0 version. Right? In addition to this method, there is no other elegant way. Do I need to open the ticket here to fix the problems of these two classes in 1.15.1?

SteNicholas:
> Probably the easiest would be to copy the request body class from Flink 1.15 and add the `@JsonInclude(Include.NON_NULL)` on the class and modify so that we are allowed to set null for pre 1.15 clusters

Thanks for @gyfora to fork this pull request. IMO, the `SavepointTriggerRequestBody` and `StopWithSavepointRequestBody` are added temporarily to address compatibility problem for 1.15.0 version. Right? In addition to this method, there is no other elegant way. Do I need to open the ticket here to fix the problems of these two classes in 1.15.1?

SteNicholas:
Thanks for @gyfora to fork this pull request and I have copied the `SavepointTriggerRequestBody` and `StopWithSavepointRequestBody` in this pull request. IMO, the `SavepointTriggerRequestBody` and `StopWithSavepointRequestBody` are added temporarily to address compatibility problem for 1.15.0 version. Right? In addition to this method, there is no other elegant way. Do I need to open the ticket here to fix the problems of these two classes in 1.15.1?

cc @wangyang0918

.github/workflows/ci.yml
```diff
@@ -41,10 +41,10 @@ jobs:
...
         run: |
           set -o pipefail; mvn clean install -Pgenerate-docs | tee ./mvn.log; set +o pipefail
-          if [[ $(grep -c "overlapping classes" ./mvn.log) -gt 0 ]];then
-            echo "Found overlapping classes, please fix"
-            exit 1
-          fi
+          #if [[ $(grep -c "overlapping classes" ./mvn.log) -gt 0 ]];then
+          #  echo "Found overlapping classes, please fix"
+          #  exit 1
+          #fi
```
gyfora, line 47: It would be nice to specifically ignore only the errors we introduced by those 2 classes and not turn off this completely.

.github/workflows/ci.yml
```diff
@@ -123,7 +123,7 @@ jobs:
       - name: Build with Maven
         run: |
           set -o pipefail; mvn clean install -DskipTests | tee ./mvn.log; set +o pipefail
-          if [[ $(grep -c "overlapping classes" ./mvn.log) -gt 0 ]];then
```
gyfora, line 126: I think overlapping classes are shown per jar. So this change might hide some other overlapping classes in different jars

flink-kubernetes-operator/src/main/resources/META-INF/NOTICE
```diff
@@ -70,25 +70,25 @@ This project bundles the following dependencies under the Apache Software Licens
...
-- org.apache.flink:flink-rpc-core:1.14.4
-- org.apache.flink:flink-runtime:1.14.4
-- org.apache.flink:flink-shaded-asm-7:7.1-14.0
-- org.apache.flink:flink-shaded-force-shading:14.0
-- org.apache.flink:flink-shaded-guava:30.1.1-jre-14.0
-- org.apache.flink:flink-shaded-jackson:2.12.4-14.0
-- org.apache.flink:flink-shaded-netty:4.1.65.Final-14.0
-- org.apache.flink:flink-shaded-zookeeper-3:3.4.14-14.0
-- org.apache.flink:flink-streaming-java_2.12:1.14.4
+- org.apache.flink:flink-annotations:1.15.0
```
wangyang0918, line 73: We might need to review the bundled transitive dependencies since we bump the Flink version.

wangyang0918:
We might need a discussion for whether to add `@JsonInclude(JsonInclude.Include.NON_NULL)` for specific `RequestBody` implementation. Just like other clients(e.g. ZooKeeper Curator, Yarn client), Flink client with new version could not access old cluster may be somewhat reasonable.

.github/workflows/ci.yml
```diff
@@ -42,8 +42,10 @@ jobs:
         run: |
           set -o pipefail; mvn clean install -Pgenerate-docs | tee ./mvn.log; set +o pipefail
           if [[ $(grep -c "overlapping classes" ./mvn.log) -gt 0 ]];then
```
wangyang0918, line 44: Maybe we could count the overlapping classes like following.
```
cat ./mvn.log | grep -E -v 'flink-runtime-.*.jar, flink-kubernetes-operator-.*.jar define 2 overlapping classes' | grep -c "overlapping classes" -
```

gyfora: I like this

wangyang0918 (approved):
+1

PR #203, opened 2022-05-11
[FLINK-27337] Prevent session cluster to be deleted when there are ru…

Aitozi:
…nning jobs

This PR is meant to ensure all the session jobs are deleted before cleanup the session cluster.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkControllerConfig.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkControllerConfig.java
@@ -37,4 +37,9 @@ public class FlinkControllerConfig<CR extends HasMetadata> extends AnnotationCon
     public Set<String> getNamespaces() {
         return watchedNamespaces;
     }
+
+    @Override
+    public Set<String> getEffectiveNamespaces() {
+        return watchedNamespaces;
+    }
 }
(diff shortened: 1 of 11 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkControllerConfig.java
```diff
@@ -37,4 +37,9 @@ public FlinkControllerConfig(Reconciler<CR> reconciler, Set<String> watchedNames
     public Set<String> getNamespaces() {
         return watchedNamespaces;
     }
+
+    @Override
+    public Set<String> getEffectiveNamespaces() {
```
Aitozi, line 42: This PR change the behavior of the effectiveNamespaces, it will make it easy to know what is the current effective namespace direct from the config. By this, we can easily to create `InformerManger` in the webhook module and do not have to consider what is the current effective namespaces in the operator.

Aitozi:
cc @gyfora @wangyang0918

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/SessionReconciler.java
```diff
@@ -134,4 +142,31 @@ protected void shutdown(FlinkDeployment deployment) {
...
+                        .getIndexer()
+                        .byIndex(OperatorUtils.CLUSTER_ID_INDEX, flinkApp.getMetadata().getName());
+        if (!sessionJobs.isEmpty()) {
+            var error =
+                    String.format(
+                            "The session jobs %s should be deleted first",
+                            sessionJobs.stream()
+                                    .map(job -> job.getMetadata().getName())
+                                    .collect(Collectors.toList()));
+            // TODO generate error events for this
```
Aitozi, line 160: I think events will be better here to help user know what happen. But, I have problem when generate the events here, I found that the example in `DeploymentFailedException#asEvent` can not work (It seems lack some of required parameters.)  I will improve this after figuring it out.

gyfora (approved):

Aitozi:
force pushed to solve conflicts

PR #204, opened 2022-05-11
[Flink 27329] Add default value of replica of JM pod and not declare it in example yamls

bgeng777:
1. Revisit default values in org.apache.flink.kubernetes.operator.crd.spec (details is [here])(https://issues.apache.org/jira/browse/FLINK-27329?page=com.atlassian.jira.plugin.system.issuetabpanels%3Aall-tabpanel)
2. Set default value of `crd.spec.Resource#cpu` to `1.0`
3. Set default value of  `crd.spec.FlinkDeploymentSpec#serviceAccount` to `flink`
4. Set default value of  `crd.spec.JobManagerSpec#replicas` to 1 and remove its usage in examples.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultValidator.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultValidator.java
@@ -81,7 +81,8 @@ public class DefaultValidator implements FlinkResourceValidator {
                 validateJobSpec(spec.getJob(), effectiveConfig),
                 validateJmSpec(spec.getJobManager(), effectiveConfig),
                 validateTmSpec(spec.getTaskManager()),
-                validateSpecChange(deployment, effectiveConfig));
+                validateSpecChange(deployment, effectiveConfig),
+                validateServiceAccount(spec.getServiceAccount()));
     }
 
     private static Optional<String> firstPresent(Optional<String>... errOpts) {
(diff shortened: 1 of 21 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/spec/FlinkDeploymentSpec.java
```diff
@@ -45,7 +45,7 @@ public class FlinkDeploymentSpec extends AbstractFlinkSpec {
     private String imagePullPolicy;
 
     /** Kubernetes service used by the Flink deployment. */
-    private String serviceAccount;
+    private String serviceAccount = "flink";
```
gyfora, line 48: I suggest to leave it on `null` and throw a validation error if not set.

bgeng777: thanks for the suggestion. It makes sense to me to add validation check so that serviceAccount will be a required field.
For users who use helm and set a different serviceAccount, they are supposed to be aware of using correct service account.
For users who use helm and use the default serviceAccount, we may add some comments in examples or detailed validation error info to alert users to set the correct serviceAccount(i.e. `flink`).

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultValidator.java
```diff
@@ -385,4 +386,12 @@ private Optional<String> validateSpecChange(FlinkSessionJob sessionJob) {
 
         return Optional.empty();
     }
+
+    private Optional<String> validateServiceAccount(String serviceAccount) {
+        if (serviceAccount == null) {
+            return Optional.of(
+                    "serviceAccount must be defined. If you use helm, its value should be the same with the name of jobServiceAccount.");
```
gyfora, line 393: I think it would be better to say `spec.serviceAccount must be defined. ...` so the user knows where exactly it's missing.

gyfora (changes requested):
Added a minor comment, I think you also need to regenerate the docs :)

wangyang0918 (approved):
+1

PR #205, opened 2022-05-11
[fork][FLINK-27499] Bump base Flink version to 1.15.0

gyfora:
This is a for of @SteNicholas 's work to eliminate some outstanding issues


gyfora:
@SteNicholas I managed to Fix your PR with a simple trick, had to disable the overlapping class test.
I simply copied the Flink 1.15 request bodys and removed the required formatType + added include nonnull

cc @wangyang0918

gyfora:
I know that this is not a pretty solution but anything else involves 5 times as much class / method copying from Flink. The RestClusterClient is not very extensible :/

PR #206, opened 2022-05-12
[FLINK-26639] Publish snapshot artifacts nightly

mbalassi:
I was inspired by https://github.com/apache/spark/pull/30623, we have to merge it before I get to try it out since it needs to be on the apache account to work. I might still need to request the necessary secrets to be available (we already have some available for the docs upload).

### DIFF
--- a/.github/workflows/publish_snapshot.yml
+++ b/.github/workflows/publish_snapshot.yml
@@ -1,0 +1,66 @@
+################################################################################
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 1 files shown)


.github/workflows/publish_snapshot.yml
```diff
@@ -0,0 +1,66 @@
...
+          
+          set -o pipefail; mvn clean deploy -Dgpg.skip -DskipTests -Papache-release | tee ./mvn.log; set +o pipefail
+          if [[ $(grep -c "overlapping classes" ./mvn.log) -gt 0 ]];then
+            echo "Found overlapping classes, please fix"
+            exit 1
+          fi
+          if [[ $(git diff HEAD | wc -l) -gt 0 ]];then
+            echo "Please generate the java doc via 'mvn clean install -DskipTests -Pgenerate-docs' again"
+            exit 1
+          fi
```
gyfora, line 64: I think it would be better not to have these checks duplicated in all CI runs (we also might change some with the 1.15 bump)

mbalassi: Fair, I will refactor it to a common function.

gyfora: That is also an option, I was suggesting to simply remove it. If it runs once somewhere else that is enough to generally fail the CI.

mbalassi: Ok, going for the cheap option then.

PR #207, opened 2022-05-12
[FLINK-27036] Exclude final release tags from docker build

mbalassi:
These should be tagged manually from the latest rc instead.


gyfora (approved):
looks good 👍

PR #209, opened 2022-05-12
[FLINK-27270] Add document of session job operations

Aitozi:


docs/content/docs/custom-resource/overview.md
```diff
@@ -146,9 +148,51 @@ For Session clusters the operator only provides very basic management and monito
...
+```
+
+### FlinkSessionJob spec overview
+
+The spec contains the information to submit a session job to the session cluster. Mostly, it will define at least the following fields:
+
+ - deploymentName: The name of the target session cluster's CR
+ - job: The Spec of the Session Job
+
+The job spec is similar field with the application job, but the jarURI can support different source. It supports the http and filesystem based
```
morhidi, line 175: wording: The job specification has the same structure in FlinkSessionJobs and FlinkDeployments, but in FlinkSessionJobs the jarUri can contain remote sources too.

morhidi (commented):
LGTM, added minor suggestions

Aitozi:
@morhidi Thanks for your suggestion, I have addressed your comments, PTAL again.

morhidi (commented):
LGTM, thanks

PR #211, opened 2022-05-13
[FLINK-27600] Add minimal jar coping and deleting logs

haoxins:
Because we delete the jar after submitting the job, it's better to leave some logs for the end users

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
@@ -281,11 +281,13 @@ public class FlinkService {
                                     .toSeconds(),
                             TimeUnit.SECONDS);
         } finally {
+            LOG.debug("Deleting the jar file {}", jarFile);
             FileUtils.deleteFileOrDirectory(jarFile);
         }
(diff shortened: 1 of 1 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/artifact/FileSystemBasedArtifactFetcher.java
```diff
@@ -39,6 +39,7 @@ public File fetch(String uri, File targetDir) throws Exception {
         FileSystem fileSystem = source.getFileSystem();
         String fileName = source.getName();
         File targetFile = new File(targetDir, fileName);
+        LOG.info("Coping file from {} to {}", source, targetFile);
         try (var inputStream = fileSystem.open(source)) {
             FileUtils.copyToFile(inputStream, targetFile);
         }
```
Aitozi, line 45: I think the line below this already log this. It's in `debug` level.

haoxins: removed and leave the others as `debug` also

Aitozi (approved):
LGTM

PR #213, opened 2022-05-14
[FLINK-27612] Generate warning events when deleting the session cluste…

Aitozi:
…r with running jobs

Manually test results:

image

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/SessionReconciler.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/SessionReconciler.java
@@ -161,10 +161,10 @@ public class SessionReconciler extends AbstractDeploymentReconciler {
             if (EventUtils.createOrUpdateEvent(
                     kubernetesClient,
                     flinkApp,
-                    EventUtils.TYPE_WARNING,
+                    EventUtils.Type.Warning,
                     "Cleanup",
                     error,
-                    EventUtils.COMPONENT_OPERATOR)) {
+                    EventUtils.Component.Operator)) {
                 LOG.warn(error);
             }
             return DeleteControl.noFinalizerRemoval()
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/SessionReconciler.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/SessionReconciler.java
@@ -161,10 +161,10 @@ public class SessionReconciler extends AbstractDeploymentReconciler {
             if (EventUtils.createOrUpdateEvent(
                     kubernetesClient,
                     flinkApp,
-                    EventUtils.Type.Warning.name(),
+                    EventUtils.TYPE_WARNING,
                     "Cleanup",
                     error,
-                    EventUtils.Component.Operator.name())) {
+                    EventUtils.COMPONENT_OPERATOR)) {
                 LOG.warn(error);
             }
             return DeleteControl.noFinalizerRemoval()
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/SessionReconciler.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/SessionReconciler.java
@@ -28,6 +28,7 @@ import org.apache.flink.kubernetes.operator.crd.status.ReconciliationStatus;
 import org.apache.flink.kubernetes.operator.informer.InformerManager;
 import org.apache.flink.kubernetes.operator.reconciler.ReconciliationUtils;
 import org.apache.flink.kubernetes.operator.service.FlinkService;
+import org.apache.flink.kubernetes.operator.utils.EventUtils;
 import org.apache.flink.kubernetes.operator.utils.FlinkUtils;
 import org.apache.flink.kubernetes.operator.utils.IngressUtils;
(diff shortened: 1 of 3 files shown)


gyfora:
I think this is a very nice change, would be great to reuse the utility for the deployment related events that we send also to avoid duplication.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/EventUtils.java
```diff
@@ -0,0 +1,98 @@
...
+                                    new ObjectReferenceBuilder()
+                                            .withKind(target.getKind())
+                                            .withUid(target.getMetadata().getUid())
+                                            .withName(target.getMetadata().getName())
+                                            .withNamespace(target.getMetadata().getNamespace())
+                                            .build())
+                            .withType(type)
+                            .withReason(reason)
+                            .withFirstTimestamp(ZonedDateTime.now().format(formatter))
+                            .withLastTimestamp(ZonedDateTime.now().format(formatter))
```
gyfora, line 76: I think `Instant.now().toString()` should do the trick also

Aitozi: Fixed

Aitozi:
> I think this is a very nice change, would be great to reuse the utility for the deployment related events that we send also to avoid duplication.

Yes, I notice that `TODO`, we could improve that in another PR

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/SessionReconciler.java
```diff
@@ -157,8 +159,23 @@ public DeleteControl cleanup(FlinkDeployment flinkApp, Context context) {
...
+            var events =
+                    EventUtils.getEvents(flinkApp, kubernetesClient).getItems().stream()
+                            .filter(event -> event.getMessage().equals(error))
+                            .collect(Collectors.toList());
+            Event lastEvent = null;
+            if (events.isEmpty()) {
+                LOG.warn(error);
+            } else {
+                lastEvent = events.get(0);
+            }
```
morhidi, line 171: Can we move this logic into the `EventUtils`?

morhidi: Here is an example for reference: https://github.com/EnMasseProject/enmasse/blob/master/k8s-api/src/main/java/io/enmasse/k8s/api/KubeEventLogger.java

gyfora: +1 for this suggestion, it makes for a nice self-contained logic that will be very easy to use everywhere.

Aitozi: Cool, I will update accordingly

Aitozi:
@morhidi I have addressed your comments, please take a look again. cc @gyfora

gyfora:
@Aitozi I think we should also adopt the naming scheme of the other solution that the @morhidi linked.
This avoids using generated event names and allows us to find them easily by ns/name directly instead of listing and filtering.

Aitozi:
I refactor based on the example, PTAL again @gyfora

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/EventUtils.java
```diff
@@ -0,0 +1,109 @@
...
+    /** The type of the events. */
+    public enum Type {
+        Normal,
+        Warning
+    }
+
+    /** The component of events. */
+    public enum Component {
+        Operator
+    }
```
gyfora, line 39: I think if we define these as Enums, we should use the enum values directly in the methods. Otherwise they could be String constants

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/EventUtils.java
```diff
@@ -0,0 +1,109 @@
...
+package org.apache.flink.kubernetes.operator.utils;
+
+import io.fabric8.kubernetes.api.model.EventBuilder;
+import io.fabric8.kubernetes.api.model.HasMetadata;
+import io.fabric8.kubernetes.api.model.ObjectReferenceBuilder;
+import io.fabric8.kubernetes.client.KubernetesClient;
+
+import java.time.Instant;
+
+/** The util to generate an event for the target resource. */
```
gyfora, line 27: We should add a note here that the core logic is copied/reused from https://github.com/EnMasseProject/enmasse/blob/master/k8s-api/src/main/java/io/enmasse/k8s/api/KubeEventLogger.java

gyfora (approved):
Looks very good!

Added 2 minor comments after that it should be good to go :) 

@morhidi could add the missing logic for the deployment controller afterwards

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/EventUtils.java
Aitozi: You are right, Fixed

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/EventUtils.java
Aitozi: Fixed

PR #214, opened 2022-05-14
[FLINK-27614] Use informer in webhook to avoid query apiserver at eac…

Aitozi:
…h update

This PR is meant to enhance the webhook logic to use the informer to avoid query apiserver during each update event

image


gyfora (approved):
Looks good!

wangyang0918 (approved):
+1 for merging.

PR #215, opened 2022-05-14
[FLINK-27163] Add label for the session job to help list the session …

Aitozi:
…jobs in the same session cluster

This PR is meant to add label for the session job to help to list session jobs under the same session.

image

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/sessionjob/FlinkSessionJobReconciler.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/sessionjob/FlinkSessionJobReconciler.java
@@ -42,12 +42,14 @@ import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nullable;
 
+import java.util.HashMap;
 import java.util.Optional;
 
 /** The reconciler for the {@link FlinkSessionJob}. */
 public class FlinkSessionJobReconciler implements Reconciler<FlinkSessionJob> {
 
     private static final Logger LOG = LoggerFactory.getLogger(FlinkSessionJobReconciler.class);
+    public static final String LABEL_SESSION_CLUSTER = "target.session.name";
 
     private final FlinkConfigManager configManager;
     private final KubernetesClient kubernetesClient;
@@ -75,6 +77,24 @@ public class FlinkSessionJobReconciler implements Reconciler<FlinkSessionJob> {
         Optional<FlinkDeployment> flinkDepOptional =
                 OperatorUtils.getSecondaryResource(
                         flinkSessionJob, context, configManager.getOperatorConfiguration());
+
+        var labels = flinkSessionJob.getMetadata().getLabels();
+        if (labels == null) {
+            labels = new HashMap<>();
+        }
+        if (!flinkSessionJob
+                .getSpec()
+                .getDeploymentName()
+                .equals(labels.get(LABEL_SESSION_CLUSTER))) {
+            labels.put(LABEL_SESSION_CLUSTER, flinkSessionJob.getSpec().getDeploymentName());
+            flinkSessionJob.getMetadata().setLabels(labels);
+            kubernetesClient
+                    .resources(FlinkSessionJob.class)
+                    .inNamespace(flinkSessionJob.getMetadata().getNamespace())
+                    .withName(flinkSessionJob.getMetadata().getName())
+                    .patch(flinkSessionJob);
+            return;
+        }
 
         // if session cluster is not ready, we can't do reconcile for the job.
         if (!helper.sessionClusterReady(flinkDepOptional)) {
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/sessionjob/FlinkSessionJobReconciler.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/sessionjob/FlinkSessionJobReconciler.java
@@ -49,7 +49,7 @@ import java.util.Optional;
 public class FlinkSessionJobReconciler implements Reconciler<FlinkSessionJob> {
 
     private static final Logger LOG = LoggerFactory.getLogger(FlinkSessionJobReconciler.class);
-    public static final String LABEL_SESSION_CLUSTER = "target.session.name";
+    public static final String LABEL_TARGET_SESSION = "target.session";
 
     private final FlinkConfigManager configManager;
     private final KubernetesClient kubernetesClient;
@@ -78,23 +78,7 @@ public class FlinkSessionJobReconciler implements Reconciler<FlinkSessionJob> {
                 OperatorUtils.getSecondaryResource(
                         flinkSessionJob, context, configManager.getOperatorConfiguration());
 
-        var labels = flinkSessionJob.getMetadata().getLabels();
-        if (labels == null) {
-            labels = new HashMap<>();
-        }
-        if (!flinkSessionJob
-                .getSpec()
-                .getDeploymentName()
-                .equals(labels.get(LABEL_SESSION_CLUSTER))) {
-            labels.put(LABEL_SESSION_CLUSTER, flinkSessionJob.getSpec().getDeploymentName());
-            flinkSessionJob.getMetadata().setLabels(labels);
-            kubernetesClient
-                    .resources(FlinkSessionJob.class)
-                    .inNamespace(flinkSessionJob.getMetadata().getNamespace())
-                    .withName(flinkSessionJob.getMetadata().getName())
-                    .patch(flinkSessionJob);
-            return;
-        }
+        patchInternalLabel(flinkSessionJob);
 
         // if session cluster is not ready, we can't do reconcile for the job.
         if (!helper.sessionClusterReady(flinkDepOptional)) {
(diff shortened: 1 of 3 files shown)


morhidi:
LGTM overall. nit: Shouldn't we use a shorter label, e.g: 
```
k get sessionjob -l target=basic-session-cluster
```

morhidi:
On second thought we could use something like this for filtering, aren't we?
```
k get --field-selector spec.deploymentName=basic-session-cluster
```

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/sessionjob/FlinkSessionJobReconciler.java
```diff
@@ -76,6 +78,24 @@ public void reconcile(FlinkSessionJob flinkSessionJob, Context context) throws E
                 OperatorUtils.getSecondaryResource(
                         flinkSessionJob, context, configManager.getOperatorConfiguration());
 
+        var labels = flinkSessionJob.getMetadata().getLabels();
```
morhidi, line 81: I suggest to extract this into a utility method (we could use computeIfAbsent)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/sessionjob/FlinkSessionJobReconciler.java
```diff
@@ -76,6 +78,24 @@ public void reconcile(FlinkSessionJob flinkSessionJob, Context context) throws E
...
+        if (labels == null) {
+            labels = new HashMap<>();
+        }
+        if (!flinkSessionJob
+                .getSpec()
+                .getDeploymentName()
+                .equals(labels.get(LABEL_SESSION_CLUSTER))) {
+            labels.put(LABEL_SESSION_CLUSTER, flinkSessionJob.getSpec().getDeploymentName());
+            flinkSessionJob.getMetadata().setLabels(labels);
+            kubernetesClient
```
morhidi, line 91: Can't we use an update control for this?

Aitozi:
> k get sessionjob --field-selector spec.deploymentName=basic-session-cluster

I tried this way before, but the custom resource seems do not support field selector

```
kubectl get sessionjob --field-selector spec.deploymentName=basic-session-cluster
Error from server (BadRequest): Unable to find "flink.apache.org/v1beta1, Resource=flinksessionjobs" that match label selector "", field selector "spec.deploymentName=basic-session-cluster": field label not supported: spec.deploymentName

kubectl version --short
Flag --short has been deprecated, and will be removed in the future. The --short output will become the default.
Client Version: v1.24.0
Kustomize Version: v4.5.4
Server Version: v1.23.3
```

After search the related issues, I think the custom resource can only support field selector by `metadata.namespace` and `metadata.name`

https://github.com/kubernetes/kubernetes/pull/53345

Aitozi:
> LGTM overall. nit: Shouldn't we use a shorter label, e.g:
> 
> ```
> k get sessionjob -l target=basic-session-cluster
> ```
> 
> It's easier to remember :)

Good idea, but the `target` is a bit too generic, I lean to rename it to `target.session`

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/sessionjob/FlinkSessionJobReconciler.java
Aitozi: The reconciler do not return `UpdateControl` now, If we want to use the update control for it, we should adjust the interface first to return the update control :/ . We have just decide not to do [this](https://issues.apache.org/jira/browse/FLINK-26915) now. I'm not sure whether we need to break it for this PR.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/sessionjob/FlinkSessionJobReconciler.java
Aitozi: Moved to a dedicated method.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/sessionjob/FlinkSessionJobReconciler.java
morhidi: Got it, but this is the first use case that we modify something else then the status. We might need to reconsider this again. Updating the CR deep in the stack sounds worse to me.

morhidi:
> > k get sessionjob --field-selector spec.deploymentName=basic-session-cluster
> 
> I tried this way before, but the custom resource seems do not support field selector
> 
> ```
> kubectl get sessionjob --field-selector spec.deploymentName=basic-session-cluster
> Error from server (BadRequest): Unable to find "flink.apache.org/v1beta1, Resource=flinksessionjobs" that match label selector "", field selector "spec.deploymentName=basic-session-cluster": field label not supported: spec.deploymentName
> 
> kubectl version --short
> Flag --short has been deprecated, and will be removed in the future. The --short output will become the default.
> Client Version: v1.24.0
> Kustomize Version: v4.5.4
> Server Version: v1.23.3
> ```
> 
> After search the related issues, I think the custom resource can only support field selector by `metadata.namespace` and `metadata.name`
> 
> [kubernetes/kubernetes#53345](https://github.com/kubernetes/kubernetes/pull/53345)


> > k get sessionjob --field-selector spec.deploymentName=basic-session-cluster
> 
> I tried this way before, but the custom resource seems do not support field selector
> 
> ```
> kubectl get sessionjob --field-selector spec.deploymentName=basic-session-cluster
> Error from server (BadRequest): Unable to find "flink.apache.org/v1beta1, Resource=flinksessionjobs" that match label selector "", field selector "spec.deploymentName=basic-session-cluster": field label not supported: spec.deploymentName
> 
> kubectl version --short
> Flag --short has been deprecated, and will be removed in the future. The --short output will become the default.
> Client Version: v1.24.0
> Kustomize Version: v4.5.4
> Server Version: v1.23.3
> ```
> 
> After search the related issues, I think the custom resource can only support field selector by `metadata.namespace` and `metadata.name`
> 
> [kubernetes/kubernetes#53345](https://github.com/kubernetes/kubernetes/pull/53345)

Yeah, I was almost sure you've tried it, but I thought it was worth asking :)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/sessionjob/FlinkSessionJobReconciler.java
Aitozi: Yes, agree with it. I'm ok to rework the interface to work out it. Also cc @gyfora @wangyang0918

morhidi: We hit some race conditions anyway with the status updates, so it maybe good as is with the java-operator-sdk v2 and we can change it later.

morhidi:
+1
For the record we need this functionality due: [Enable arbitrary CRD field selectors](https://github.com/kubernetes/kubernetes/issues/53459)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/sessionjob/FlinkSessionJobReconciler.java
Aitozi: As I know we have manually patch the status after https://github.com/apache/flink-kubernetes-operator/pull/199 . So this change do not bring the resourceVersion conflicts, I think. Do you mean keep it as now ?

morhidi: Let's see what @gyfora and @wangyang0918 think/suggest.

wangyang0918:
TBH, I do not like to do the CR `.metadata` or `.spec` updates in the reconciler. Especially, this is only for easier filter, not for operator internal use case. In the production environment, I believe the users could do this manually or covered by external job management tools.

I also have another feeling this is the responsibility of [mutating admission webhook](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#mutatingadmissionwebhook).

morhidi:
> TBH, I do not like to do the CR `.metadata` or `.spec` updates in the reconciler. Especially, this is only for easier filter, not for operator internal use case. In the production environment, I believe the users could do this manually or covered by external job management tools.
> 
> I also have another feeling this is the responsibility of [mutating admission webhook](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#mutatingadmissionwebhook).

I've been also thinking about putting this into the webhook or maybe using the label exclusively for linking the two CRs.

Aitozi:
@wangyang0918 Regarding why we need this: although we can manage the job with external job management tools, but we also may need to operate in command line. In the case to operate on all the jobs of a session cluster, eg: delete all the session jobs before shutdown the cluster, it will be useful I think.

I think the mutating admission webhook is a better way. But one shortcoming is it's optional. If disabled, it will not work

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/sessionjob/FlinkSessionJobReconciler.java
gyfora: if you don't explicitly set the resourceversion to `null` before patching you might still get an error.

gyfora:
I am also leaning toward -1 on the approach in this PR as so far we have explicitly avoided making changes to the metadata/spec for other features.

gyfora:
I think mutating webhooks provide a good alternative solution but I would prefer to revisit this issue after the release.

Aitozi:
Thanks for your good suggestion @morhidi @wangyang0918 @gyfora . I will try to implement it with mutating webhook. It's ok to me to leave it for next release

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/sessionjob/FlinkSessionJobReconciler.java
Aitozi: yeah, but I think this operation is retriable

gyfora:
Should we close this PR for now?

PR #216, opened 2022-05-15
[FLINK-27435] Savepoint History

tweise:
Add savepoint history to status, with age and count based savepoint expiration. Example:
```
Status:
  Job Manager Deployment Status:  READY
  Job Status:
    Job Id:    7cedf7c39c986e583dbebd372be3a8d3
    Job Name:  State machine job
    Savepoint Info:
      Last Savepoint:
        Location:    file:/flink-data/savepoints/savepoint-7cedf7-0fb4b50d72b6
        Time Stamp:  1652671155388
      Savepoint History:
        Location:         file:/flink-data/savepoints/savepoint-7cedf7-ab75503094b5
        Time Stamp:       1652670977699
        Location:         file:/flink-data/savepoints/savepoint-7cedf7-1c12ca67aa68
        Time Stamp:       1652671008035
        Location:         file:/flink-data/savepoints/savepoint-7cedf7-0fb4b50d72b6
        Time Stamp:       1652671124981
      Trigger Id:         20342ead12977c31384dfd83cda82377
      Trigger Timestamp:  1652671114901
    Start Time:           1652670821756
    State:                RUNNING
    Update Time:          1652671155360
  Reconciliation Status:
```

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
@@ -143,6 +143,19 @@ public class KubernetesOperatorConfigOptions {
                     .withDescription(
                             "Whether to enable recovery of missing/deleted jobmanager deployments.");
 
+    public static final ConfigOption<Integer> OPERATOR_SAVEPOINT_HISTORY_MAX_COUNT =
+            ConfigOptions.key("kubernetes.operator.savepoint.history.max.count")
+                    .intType()
+                    .defaultValue(10)
+                    .withDescription("Maximum number of savepoint history entries to retain.");
+
+    public static final ConfigOption<Duration> OPERATOR_SAVEPOINT_HISTORY_MAX_AGE =
+            ConfigOptions.key("kubernetes.operator.savepoint.history.max.age")
+                    .durationType()
+                    .defaultValue(Duration.ofHours(24))
+                    .withDescription(
+                            "Maximum age for savepoint history entries to retain. Due to lazy clean-up, the most recent savepoint may live longer than the max age.");
+
     public static final ConfigOption<Map<String, String>> JAR_ARTIFACT_HTTP_HEADER =
             ConfigOptions.key("kubernetes.operator.user.artifacts.http.header")
                     .mapType()
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/SavepointObserver.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/SavepointObserver.java
@@ -17,6 +17,7 @@
 
 package org.apache.flink.kubernetes.operator.observer;
 
+import org.apache.flink.annotation.VisibleForTesting;
 import org.apache.flink.api.common.JobID;
 import org.apache.flink.api.common.JobStatus;
 import org.apache.flink.configuration.Configuration;
@@ -34,6 +35,9 @@ import org.apache.flink.kubernetes.operator.utils.StatusHelper;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.time.Duration;
+import java.util.Iterator;
+import java.util.List;
 import java.util.Optional;
 
 /** An observer of savepoint progress. */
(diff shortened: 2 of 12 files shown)


morhidi:
Please provide the missing docs too, hint:
```
mvn clean install -Pgenerate-docs
```
then commit the doc changes

gyfora (approved):
Looks good!

Please rebase on `main`  + generate the docs to make CI pass

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
```diff
@@ -143,6 +143,18 @@ public class KubernetesOperatorConfigOptions {
                     .withDescription(
                             "Whether to enable recovery of missing/deleted jobmanager deployments. False by default for Flink 1.14, true for newer Flink version.");
 
+    public static final ConfigOption<Integer> OPERATOR_SAVEPOINT_HISTORY_MAX_COUNT =
+            ConfigOptions.key("kubernetes.operator.savepoint.history.max.count")
+                    .intType()
+                    .defaultValue(10)
+                    .withDescription("Maximum number of savepoint history entries to retain.");
+
+    public static final ConfigOption<Duration> OPERATOR_SAVEPOINT_HISTORY_MAX_AGE =
```
wangyang0918, line 152: Since we are performing a lazy clean-up, the savepoint may live longer than the max age.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/SavepointObserver.java
```diff
@@ -131,9 +136,56 @@ private Optional<String> observeTriggeredSavepointProgress(
 
         LOG.info("Savepoint status updated with latest completed savepoint info");
         currentSavepointInfo.updateLastSavepoint(savepointFetchResult.getSavepoint());
+        updateSavepointHistory(
+                currentSavepointInfo, savepointFetchResult.getSavepoint(), deployedConfig);
```
gyfora, line 140: I think we should also update the savepoint history when we shut down using savepoint. (Or in case of failure of the shutdown operation as part of the `observeLatestSavepoint`)

tweise: The job needs to be running for savepoint cleanup to succeed.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
tweise: I added this to the doc.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/SavepointObserver.java
gyfora: That makes sense but the savepoint itself could still be recorded in the history and cleaned up later after upgrade

tweise:
@gyfora added recording of final savepoint. Tested by suspending deployment.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/SavepointObserver.java
```diff
@@ -131,15 +136,72 @@ private Optional<String> observeTriggeredSavepointProgress(
...
     private void observeLatestSavepoint(
             SavepointInfo savepointInfo, String jobID, Configuration deployedConfig) {
         try {
             flinkService
                     .getLastCheckpoint(JobID.fromHexString(jobID), deployedConfig)
-                    .ifPresent(savepointInfo::updateLastSavepoint);
+                    .ifPresent(
+                            sp -> {
+                                savepointInfo.updateLastSavepoint(sp);
+                                updateSavepointHistory(savepointInfo, sp, deployedConfig, false);
```
gyfora, line 203: I am afraid this will only work for Flink 1.15 and it's not enough. You need to also add this to `FlinkService#cancelJob` where it would be recorded regardless of Flink version.

There is no absolute guarantee that it will be recorded either way but having it in both places eliminates the bulk of the possibilities :)

gyfora: You could cover this quite nicely with a FlinkVersion parameterized tests. The TestingFlinkService behaves according to the Flink version (keeping the Flink deployment around in Finished state after cancel for 1.15 and removing for earlier versions)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
```diff
@@ -431,11 +431,12 @@ public void cancelJob(FlinkDeployment deployment, UpgradeMode upgradeMode) throw
...
-                location ->
-                        deploymentStatus
-                                .getJobStatus()
-                                .getSavepointInfo()
-                                .setLastSavepoint(Savepoint.of(location)));
+                location -> {
+                    Savepoint sp = Savepoint.of(location);
+                    deploymentStatus.getJobStatus().getSavepointInfo().setLastSavepoint(sp);
+                    // this is required for Flink < 1.15
+                    deploymentStatus.getJobStatus().getSavepointInfo().addSavepointToHistory(sp);
```
wangyang0918, line 438: We also need to do this for `FlinkSessionJobReconciler`.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/SavepointObserver.java
```diff
@@ -131,15 +135,63 @@ private Optional<String> observeTriggeredSavepointProgress(
...
+            }
+        }
+    }
+
+    private void disposeSavepointQuietly(Savepoint sp, Configuration conf) {
+        try {
+            LOG.info("Disposing savepoint {}", sp);
+            flinkService.disposeSavepoint(sp.getLocation(), conf);
+        } catch (Exception e) {
+            // savepoint dispose error should nota affect the deployment
```
wangyang0918, line 180: typo: nota?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
tweise: done!

gyfora:
I merged this. 
Added some [missing tests](https://github.com/apache/flink-kubernetes-operator/commit/ba06ce8e49d95b50dd50e9950a8e96bbe4d5c707) that now cover this as part of our integration tests.  

You can see, that they would have caught the problem of not registering the savepoint in history on cancel for the right flink versions:

Screenshot 2022-05-17 at 07 42 39

PR #218, opened 2022-05-17
[hotfix] Do not use nulls in status for savepoint trigger and cluster info

gyfora:
Seems like there is a problem with the fabric8 patch logic and it doesn't allow us to set status fields to "null" and thus remove them.

This meant that savepoint trigger information would not be cleared correctly.

I will still investigate to see if there is a potential simpler workaround. 
The problem seems to be caused by 
https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/Serialization.java#L53
https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/utils/serialization/UnmatchedFieldTypeModule.java

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/status/SavepointInfo.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/status/SavepointInfo.java
@@ -38,10 +38,10 @@ public class SavepointInfo {
     private Savepoint lastSavepoint;
 
     /** Trigger id of a pending savepoint operation. */
-    private String triggerId;
+    private String triggerId = "";
 
     /** Trigger timestamp of a pending savepoint operation. */
-    private Long triggerTimestamp;
+    private Long triggerTimestamp = 0L;
 
     /** List of recent savepoints. */
     private List<Savepoint> savepointHistory = new ArrayList<>();
@@ -52,8 +52,8 @@ public class SavepointInfo {
     }
 
     public void resetTrigger() {
-        this.triggerId = null;
-        this.triggerTimestamp = null;
+        this.triggerId = "";
+        this.triggerTimestamp = 0L;
     }
 
     public void updateLastSavepoint(Savepoint savepoint) {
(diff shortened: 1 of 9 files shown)


gyfora:
@wangyang0918  do you have any alternative idea maybe? :)

Aitozi:
Does it caused by it used the `PatchType#JSON_MERGE` which will delete the key when it's value is null ?

gyfora:
@Aitozi it should delete the key when the value is null that is correct. But it doesn’t seem to work as it drops the null values before making the request to k8s.

you can verify by calling: Serialization.asJson(resource)

Aitozi:
You mean it didn't keep the null value, so the patch for the field is skipped right ? @gyfora

gyfora:
Exactly, the path request sent to k8s should contain the null values so merge would remove those keys. Unfortunately fabric8 applies some custom plugin to the objectmapper that forces it to drop nulls (with a simple objectmapper they would be there).

So the request sent to k8s will not contian nulls so it is impossible to remove keys from the status (which is what we want when we set something to null)

Aitozi:
Thanks, I think I get what you mean, I tried a way to custom the serializer may be can pass the null 

```
    @Test
    public void test() {
        var application = TestUtils.buildApplicationCluster();

        Serialization.jsonMapper().registerModule(new TestModule());
        System.out.println(Serialization.asJson(application));
    }

    public class TestModule extends SimpleModule {

        public TestModule() {
            super("CustomTestModule", VersionUtil.versionFor(SavepointInfo.class));
            addSerializer(SavepointInfo.class, new SavepointInfoSerializer());
        }
    }


public class SavepointInfoSerializer extends JsonSerializer<SavepointInfo> {

    @Override
    public void serialize(
            SavepointInfo savepointInfo,
            JsonGenerator jsonGenerator,
            SerializerProvider serializerProvider)
            throws IOException {
        jsonGenerator.writeStartObject();
        if (savepointInfo.getTriggerId() == null) {
            jsonGenerator.writeStringField("triggerId", null);
        }
    }
}

```

generated:

```
"savepointInfo":{
       "triggerId":null
 }
```

gyfora:
@Aitozi that would mean we need to write custom serializers instead of simply avoiding nulls (and we also rely on the implementation of the fabric8 serialization logic which makes it more unsafe)

I think avoiding nulls is simpler and feels less hacky.

Aitozi:
Yes, I think your solution looks good to me :) , custom serializer is a burden. +1 from my side

Aitozi (approved):

wangyang0918 (approved):
Avoiding null makes more sense to me. +1

wangyang0918:
BTW, I have to admin that setting the value to `null` in json/yaml is not a best practice.

morhidi:
+1 It looked scary for the first read. Happy to see an elegant solution for this.

PR #220, opened 2022-05-17
[FLINK-27665] Optimize event triggering on DeploymentFailedExceptions

morhidi:
- Separated the event related logic from the DeploymentFailedExceptions
- Introduced a new component type for JobManagerDeployment, reporting DeploymentFailedExceptions
- Reduced the event types to Warning and Normal only see [API docs](https://github.com/kubernetes/api/blob/master/events/v1/types.go#L84)


morhidi:
cc @wangyang0918 @Aitozi

Aitozi (approved):
LGTM

PR #223, opened 2022-05-18
[hotfix][FLINK-27572] Harden HA meta checking logic

gyfora:
Improvements to FLINK-27572 based on some corner-cases we observed with @morhidi around turning HA on and off

Specifically it fixes:
 - Improves the isHaDataAvailable logic by checking both deployed and new configuration for HA enabled in addition to validating the configmaps. This is necessary to not detect garbage + make sure that HA data is actually used on restore
 - Records upgrade mode actually used for suspend into the status, this is important for 2 cases:
  1. When switching from savepoint -> last-state + ha, we record savpeoint upgrade mode which eliminates HA data validation on restore
  2. When upgrading an unhealthy job with savepoint mode (using ha + last-state) we record last-state in the status and actually check on recovery that the HA metadata is present

Tests have been improved to cover these cases


gyfora:
cc @morhidi @wangyang0918 @Aitozi

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkUtils.java
```diff
@@ -245,9 +245,19 @@ public static void deleteJobGraphInKubernetesHA(
...
 
     public static boolean isHaMetadataAvailable(
-            Configuration conf, KubernetesClient kubernetesClient) {
-        String clusterId = conf.get(KubernetesConfigOptions.CLUSTER_ID);
-        String namespace = conf.get(KubernetesConfigOptions.NAMESPACE);
+            Configuration observeConfig,
+            Configuration deployConfig,
+            KubernetesClient kubernetesClient) {
+
+        if (!FlinkUtils.isKubernetesHAActivated(observeConfig)
```
wangyang0918, line 252: It is a hard time for me to understand why we are verifying the HA enabled again here. AFAIK, we should already done this in the validator.

After more consideration, I find it is useful in `ApplicationReconciler#getAvailableUpgradeMode`. Maybe we should not mix the HA enabled verification and HA metadata existence. It will also make some logging/exception confusing.

wangyang0918, line 252: It is a hard time for me to understand why we are verifying the HA enabled again here. AFAIK, we should already done this in the validator.

After more consideration, I find it is useful in `ApplicationReconciler#getAvailableUpgradeMode`. Maybe we should not mix the HA enabled verification and HA metadata existence. It will also make some logging/exception confusing.

wangyang0918:
> 1. When switching from savepoint -> last-state + ha, we record savpeoint upgrade mode which eliminates HA data validation on restore

I think you mean we need to record `savepoint` upgrade mode when switching from stateless + non-ha -> last-state + ha here. For savepoint -> last-state + ha, we already record `savepoint` upgrade mode in the status. Right?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkUtils.java
gyfora: you are right, I will try to clean this up and keep it in the getAvailableUpgrade mode where it covers a special cornercase for the savpeoint upgrade mode

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
```diff
@@ -100,46 +103,49 @@ public void reconcile(FlinkDeployment flinkApp, Context context) throws Exceptio
...
                     LOG.info("Upgrading/Restarting running job, suspending first...");
                 }
-                flinkService.cancelJob(flinkApp, upgradeMode);
+                Optional<UpgradeMode> availableUpgradeMode =
+                        getAvailableUpgradeMode(flinkApp, deployConfig);
+                if (availableUpgradeMode.isEmpty()) {
+                    return;
+                }
+                // We must record the upgrade mode used to the status later
+                desiredJobSpec.setUpgradeMode(availableUpgradeMode.get());
```
gyfora, line 126: @wangyang0918 this is the part I was referring to. It is very important to know what upgradeMode we used during suspend to be able to verify/not-verify HA metadata. 

If you are switching from without HA (from savepoint/stateless) the available upgrade mode is savepoint. In this case we record savepoint in the lastReconciledSpec upgrademode field so on restore we know not to enforce HA metadata validation because it cannot possibly be there.

gyfora:
Thanks @wangyang0918 for the input, I removed the unnecessary redundant HA config checks, this made this change much simpler.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
wangyang0918: My bad. You are right. Both changing from savepoint/stateless without HA to last-state, we need to record the `savepoint` upgrade mode to the status.

wangyang0918: My bad. You are right. Both changing from savepoint/stateless without HA to last-state, we need to record the `savepoint` upgrade mode to the status.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
```diff
@@ -128,18 +130,18 @@ public void reconcile(FlinkDeployment flinkApp, Context context) throws Exceptio
...
+                        // We decide to enforce HA based on how job was previously suspended
+                        lastReconciledSpec.getJob().getUpgradeMode() == UpgradeMode.LAST_STATE);
                 stateAfterReconcile = JobState.RUNNING;
             }
             ReconciliationUtils.updateForSpecReconciliationSuccess(flinkApp, stateAfterReconcile);
             IngressUtils.updateIngressRules(
                     deployMeta, currentDeploySpec, deployConfig, kubernetesClient);
         } else if (ReconciliationUtils.shouldRollBack(
-                flinkService, reconciliationStatus, deployConfig)) {
+                flinkService, reconciliationStatus, observeConfig)) {
```
wangyang0918, line 141: Do we need to use the `deployConfig` here? For example, we change the image along with `kubernetes.operator.deployment.rollback.enabled` from `false` to `true`, the rollback should also work.

wangyang0918, line 141: Do we need to use the `deployConfig` here? For example, we change the image along with `kubernetes.operator.deployment.rollback.enabled` from `false` to `true`, the rollback should also work.

gyfora: it does not really make a difference which one we use here. The only case when they would not be the same is if we have already rolled back :)

gyfora: when you change the spec, that's always an upgrade (not a rollback) so this code would not be called

wangyang0918 (approved):
+1 for merging.

PR #224, opened 2022-05-18
[FLINK-27008] Document the configurable parameters of the helm chart and their default values

SteNicholas:
All the configurable parameters of the helm chart and their default values are needed to document, which could be put in the helm section.

**The brief change log**

- Documents all the configurable parameters of the helm chart and their default values.


wangyang0918 (approved):
+1

PR #227, opened 2022-05-19
[FLINK-27685] Scale subresource prototype

gyfora:
This is a proof of concept of introducing Kubernetes scaling capabilities and support for using the Horizontal Pod Autoscaler for application deployments.

**Background:** 
https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#scale-subresource
https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/

**Scaling for Flink applications:**
Our initial idea was to use the parallelism as the scaling unit/parameter. This is however not a really good match for the Kubernetes scaling behaviour, where you have a concept of replicas and watch pods/containers to decide whether to scale up or down. By using parallelism you will end up with confusing terminology and half used task managers (where not all the slots are taken)

We decided it might be better to use the number of taskmanager pods as the replica count for scaling purposes as it more closely matches what is actually happening under the hood

**In this PR:**
We introduce `taskManager.replicas` in the spec, which is by default `null`. When defined it takes precedence over the job parallelism setting and the parallelism is computed using `replicas * numTaskSlots`.

We also expose the taskmanager replicas together with the pod label selector in the status (this is necessary for the scale subresource). 

After setting the correct annotations the generate subresource looks like this:

```
subresources:
      scale:
        labelSelectorPath: .status.taskManager.labelSelector
        specReplicasPath: .spec.taskManager.replicas
        statusReplicasPath: .status.taskManager.replicas
```

With this we can now crate a job and hpa:
image

The HorizontalPodAutoscaler policy will continuously monitor the TM pods and scale the number up/down.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkUtils.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkUtils.java
@@ -19,7 +19,9 @@ package org.apache.flink.kubernetes.operator.utils;
 
 import org.apache.flink.api.common.JobStatus;
 import org.apache.flink.configuration.Configuration;
+import org.apache.flink.configuration.CoreOptions;
 import org.apache.flink.configuration.HighAvailabilityOptions;
+import org.apache.flink.configuration.TaskManagerOptions;
 import org.apache.flink.kubernetes.configuration.KubernetesConfigOptions;
 import org.apache.flink.kubernetes.highavailability.KubernetesHaServicesFactory;
 import org.apache.flink.kubernetes.kubeclient.decorators.ExternalServiceDecorator;
@@ -277,5 +279,11 @@ public class FlinkUtils {
 
     public static boolean clusterShutdownDisabled(FlinkDeploymentSpec spec) {
         return spec.getFlinkVersion().isNewerVersionThan(FlinkVersion.v1_14);
+    }
+
+    public static int getNumTaskManagers(Configuration conf) {
+        int parallelism = conf.get(CoreOptions.DEFAULT_PARALLELISM);
+        int taskSlots = conf.get(TaskManagerOptions.NUM_TASK_SLOTS);
+        return (parallelism + taskSlots - 1) / taskSlots;
     }
 }
--- a/docs/content/docs/custom-resource/reference.md
+++ b/docs/content/docs/custom-resource/reference.md
@@ -149,6 +149,7 @@ This page serves as a full reference for FlinkDeployment custom resource definit
 | Parameter | Type | Docs |
 | ----------| ---- | ---- |
 | resource | org.apache.flink.kubernetes.operator.crd.spec.Resource | Resource specification for the TaskManager pods. |
+| replicas | java.lang.Integer | Number of TaskManager replicas. If defined, takes precedence over parallelism |
 | podTemplate | io.fabric8.kubernetes.api.model.Pod | TaskManager pod template. It will be merged with FlinkDeploymentSpec.podTemplate. |
 
 ### UpgradeMode
@@ -188,6 +189,7 @@ This page serves as a full reference for FlinkDeployment custom resource definit
 | clusterInfo | java.util.Map<java.lang.String,java.lang.String> | Config information from running clusters. |
 | jobManagerDeploymentStatus | org.apache.flink.kubernetes.operator.crd.status.JobManagerDeploymentStatus | Last observed status of the JobManager deployment. |
 | reconciliationStatus | org.apache.flink.kubernetes.operator.crd.status.FlinkDeploymentReconciliationStatus | Status of the last reconcile operation. |
+| taskManager | org.apache.flink.kubernetes.operator.crd.status.TaskManagerInfo | Information about the TaskManagers for the scale subresource. |
 
 ### FlinkSessionJobReconciliationStatus
 **Class**: org.apache.flink.kubernetes.operator.crd.status.FlinkSessionJobReconciliationStatus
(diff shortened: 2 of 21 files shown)


jeesmon:
@gyfora Thanks for the prototype. We are very much interested in this feature for kube native deployment. Does this work like reactive-mode? Do we need any other special configuration for this to work? Like `scheduler-mode=reactive`.

gyfora:
@jeesmon as you can see in the example I provided this works with the `FlinkDeployment` resource which the operator manages using the Flink Native integration.

You don't need to set anything else, only use the taskManager.replicas config instead of setting the parallelism for your `FlinkDeployment` resource. This will work in a similar way to how the reactive mode works with the only difference is that scaling is done through the kubernetes api or a HPA.

In this version the operator does not create the HPA policy, it is definitely out of scope for this initial version and even later we have to discuss how far we want to go :)

jeesmon:
This is perfect, hope it will get approved and merged soon. As with reactive-mode, this will work only for `FlinkDeployment` and not for `FlinkSessionJob`, right? Thanks.

jeesmon:
Yes, I can build locally and try. We just started exploring this operator and one of the hurdles to go with operator was lack of autoscaling support in native integration. This will definitely help with that concern. I will keep you updated. Thanks.

jeesmon:
@gyfora I was able to run the `basic-scaling` example with your changes locally and it seems to be working fine. One thing I noticed is parallelism getting updated even when new pods are in Pending state. For example, I manually scaled up `FlinkDeployment` using `kubectl scale` and I didn't have enough resources to run all replicas of TMs in my minkube cluster. But operator seemed to have adjusted parallelism based on taskManager.replicas without checking the status of TM pods. Not sure it's an expected behavior but just wanted to point it out.

gyfora:
@jeesmon That is completely expected. When you scale up the operator performs a stateful upgrade to the new parallelism.
The job is resubmitted and the native integration takes care of creating the new taskmanagers. If you don't have enough resources your job won't be able to start.

jeesmon:
@gyfora Considering a scale event can potentially block resuming a job (due to new task managers in Pending state because of worker node provisioning delay or other scheduling issues), do you agree a standalone mode + reactive is a safer approach for auto scaling? Just trying to compare pros and cons. Thanks.

gyfora:
@jeesmon in some sense I agree.

But task manager pods can fail for various reasons and you need to make sure you can provision them reliably if you are running real time applications.

So if you have capacity problems in the cluster , maybe autoscaler is not a good idea in general.

gyfora:
But in general you are right the reactive mode should be quicker and more reliable in shaky conditions. We can probably support that in the future once standalone mode support is added

gyfora:
Also you cannot delete the CRD if you have flinkdeployment object those need to be deleted first

gyfora:
@wangyang0918 @mbalassi @Aitozi @morhidi  do you think we should merge this as a first iteration or discuss further on the mailing list? I feel like the general sentiment is that this approach seems to be reasonable :)

rr-sushant-choudhary:
@gyfora Deleting Flink deployment object itself is stuck, it's not deleting the resource. Flink Operator Pod is running and doesn't show any error logs too on both webhook and operator container.

gyfora:
@rr-sushant-choudhary try restarting the operator , I believe this is not related to this particular change. Feel free to ping me in email/slack if you get stuck. There are also ways of force deleting things if you google it.

let’s try to keep this off from this PR discussion if you don’t mind :)

Aitozi:
@gyfora Thanks for driving this topic, I will take a closer look on it tomorrow

Aitozi (commented):
Hi @gyfora , I have went through the implementation code. I think it's a good try to use HPA in flink job. I have two concerns about it:

- Can we calculate the accurate replicas by the spec#parallelism. eg: if use setParallelism in the jar, it will have higher priority, I think.
- This model is bundled with the predefined taskmanager resource spec. Will it work with the FineGrained resource policy (although it has not been supported now) . Certainly it's out of the scope of this PR, But I think we could discuss it further. 

Also looking forward for other guys ideas on this.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkUtils.java
```diff
@@ -278,4 +280,10 @@ public static boolean isKubernetesHAActivated(Configuration configuration) {
     public static boolean clusterShutdownDisabled(FlinkDeploymentSpec spec) {
         return spec.getFlinkVersion().isNewerVersionThan(FlinkVersion.v1_14);
     }
+
+    public static int getNumTaskManagers(Configuration conf) {
+        int parallelism = conf.get(CoreOptions.DEFAULT_PARALLELISM);
+        int taskSlots = conf.get(TaskManagerOptions.NUM_TASK_SLOTS);
+        return (parallelism + taskSlots - 1) / taskSlots;
```
Aitozi, line 287: what's the meaning of `(parallelism + taskSlots - 1) / taskSlots` ?

Aitozi: IMO, it should be `parallelism / taskSlots`

gyfora: In java integer division always rounds down. 5/2 would result in 2 instead of rounding up to 3 what we need to compute here. `(parallelism + taskSlots - 1) / taskSlots` is a nice trick to divide parallelism/taskslots while rounding up.

gyfora:
@Aitozi thank you for raising these concerns. I think the main question is, why would anyone use autoscaling/HPA (and expect it to work) if they hardcode parallelism or have very fine grained resource requirements. I think if they have these problems the autoscaling logic is really not for them and I think it's very hard to solve it nicely. The general assumption for HPA is homogeneous resources.

The hardcoding / overwriting of configuration from within your program is a concern in general when using the operator. Not just for scaling but any manual spec change too we should advise users to only use the spec to configure these settings.

Other than this, since the current behaviour is not impacted by this change, people can decide whether to use HPA or not. :)

morhidi:
I agree with @gyfora here. Let's add this feature without overthinking it too much and see what we can bring out of it.

Aitozi:
>  The general assumption for HPA is homogeneous resources.

I totally agree with your point, and we should let user have the same expect from it

PR #228, opened 2022-05-19
[FLINK-27686] Only patch status on change

gyfora:
Check whether status changed before applying patch to reduce load on the kubernetes API.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/StatusHelper.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/StatusHelper.java
@@ -20,8 +20,9 @@ package org.apache.flink.kubernetes.operator.utils;
 
 import org.apache.flink.api.java.tuple.Tuple2;
 import org.apache.flink.kubernetes.operator.crd.status.CommonStatus;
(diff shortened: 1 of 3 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/StatusHelper.java
```diff
@@ -64,6 +67,14 @@ public <T extends CustomResource<?, STATUS>> void patchAndCacheStatus(T resource
         // in the meantime
         resource.getMetadata().setResourceVersion(null);
 
+        ObjectNode newStatus = objectMapper.convertValue(resource.getStatus(), ObjectNode.class);
+        ObjectNode previousStatus = statusCache.put(getKey(resource), newStatus);
+
+        if (previousStatus == null || newStatus.equals(previousStatus)) {
```
Aitozi, line 73: why here do not need to patch status when the previousStatus is null ?

gyfora: Very good catch you are right @Aitozi

Aitozi (approved):
Looks good

PR #229, opened 2022-05-20
[hotfix] Change version of documentation in update_branch_version.sh

wangyang0918:
We also need to change version of documentation in `update_branch_version.sh`.

### DIFF
--- a/tools/releasing/update_branch_version.sh
+++ b/tools/releasing/update_branch_version.sh
@@ -64,7 +64,11 @@ perl -pi -e "s#^ENV OPERATOR_VERSION=.*#ENV OPERATOR_VERSION=${NEW_VERSION}#" Do
 # change Helm chart version info
 perl -pi -e "s#^version: .*#version: ${NEW_VERSION}#" helm/flink-kubernetes-operator/Chart.yaml
 perl -pi -e "s#^appVersion: .*#appVersion: ${NEW_VERSION}#" helm/flink-kubernetes-operator/Chart.yaml
-
+#change version of documentation
+cd docs
+perl -pi -e "s#^  Version = .*#  Version = \"${NEW_VERSION}\"#" config.toml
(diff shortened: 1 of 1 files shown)


wangyang0918:
cc @gyfora

tools/releasing/update_branch_version.sh
```diff
@@ -64,7 +64,11 @@ perl -pi -e "s#^ENV OPERATOR_VERSION=.*#ENV OPERATOR_VERSION=${NEW_VERSION}#" Do
 # change Helm chart version info
 perl -pi -e "s#^version: .*#version: ${NEW_VERSION}#" helm/flink-kubernetes-operator/Chart.yaml
 perl -pi -e "s#^appVersion: .*#appVersion: ${NEW_VERSION}#" helm/flink-kubernetes-operator/Chart.yaml
-
+#change version of documentation
+cd docs
+perl -pi -e "s#^  Version = .*#  Version = \"${NEW_VERSION}\"#" config.toml
+perl -pi -e "s#^  VersionTitle = .*#  VersionTitle = \"${NEW_VERSION}\"#" config.toml
+cd ..
```
gyfora, line 71: I don't think we should do this here. The docs are built from the release branch (which otherwise has the snapshot version). I think we need to update the config manually when cutting the release branch like you did

gyfora: On a second thought, it wouldnt really hurt (it's currently also out of date on the main branch). For rc branches it should not change anything if the release branch is already updated correctly

wangyang0918: I do not mean to update the release branch, but the main branch. It is still `1.0-SNAPSHOT`, not `1.1-SNAPSHOT`.

https://github.com/apache/flink-kubernetes-operator/blob/main/docs/config.toml#L37

PR #236, opened 2022-05-22
[hotfix] Fix last checkpoint observe with empty history

gyfora:
Fixes a very important cornercase where latest checkpoint wasn't observed correctly after job restore recovery.

The current logic relies on being able to observe the latest checkpoint/savepoint correctly in case of a terminal job state and setting it in the status.

Previously this was done completely based on the checkpoint history, without accounting for the possibility that the history may be empty if the job fails after restore before the first checkpoint.

At the moment this bug would cause jobs to fall back to an invalid earlier checkpoint in some cases.


gyfora:
@wangyang0918 @Aitozi @morhidi this is a critical fix that should be reviewd & merged before the first RC

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
```diff
@@ -633,8 +623,26 @@ public Optional<Savepoint> getLastCheckpoint(JobID jobId, Configuration conf) th
...
+
             if (!latestCheckpointOpt.isPresent()) {
-                LOG.warn("Could not find any externally addressable checkpoints.");
+                LOG.warn("Could not find any checkpoints.");
+            }
+
+            if (latestCheckpointOpt.isPresent()
+                    && latestCheckpointOpt
+                            .get()
+                            .equals(
```
wangyang0918, line 640: Comparing `Savepoint` with a `String`.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/CheckpointHistoryWrapper.java
```diff
@@ -19,24 +19,43 @@
...
 @Data
 @NoArgsConstructor
 public class CheckpointHistoryWrapper implements ResponseBody {
 
-    public static final String FIELD_NAME_HISTORY = "history";
-
     @JsonProperty(FIELD_NAME_HISTORY)
     private List<ObjectNode> history;
+
+    @JsonProperty(FIELD_NAME_LATEST_CHECKPOINTS)
```
wangyang0918, line 48: I am not sure whether we could always use the field `latest` instead of `history`.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
gyfora: Good catch I was planning to add a test for this anyways :)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/CheckpointHistoryWrapper.java
gyfora: We could but we still need to check whether latest savepoint or latest checkpoint is newer in that case

wangyang0918 (approved):
+1 for merging.

Verified locally after restoring and also triggering a savepoint. It works well.

PR #237, opened 2022-05-23
[FLINK-27257] Flink kubernetes operator triggers savepoint failed because of not all tasks running

SteNicholas:
At present Flink kubernetes operator triggers savepoint failed because of not all tasks running. If the exception is that not all required tasks are currently running, then continue fetch savepoint until it is successfully triggered.

**The brief change log**
- `fetchSavepointInfo` adds the check whether the job status is RUNNING and the exception is that not all required tasks are currently running and continue to fetch savepoint for this situation.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
@@ -49,6 +49,7 @@ import org.apache.flink.kubernetes.operator.exception.DeploymentFailedException;
 import org.apache.flink.kubernetes.operator.observer.SavepointFetchResult;
 import org.apache.flink.kubernetes.operator.reconciler.ReconciliationUtils;
 import org.apache.flink.kubernetes.operator.utils.FlinkUtils;
+import org.apache.flink.runtime.checkpoint.CheckpointException;
 import org.apache.flink.runtime.client.JobStatusMessage;
 import org.apache.flink.runtime.highavailability.nonha.standalone.StandaloneClientHAServices;
 import org.apache.flink.runtime.jobgraph.RestoreMode;
@@ -110,6 +111,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 
+import static org.apache.flink.api.common.JobStatus.RECONCILING;
 import static org.apache.flink.kubernetes.operator.config.FlinkConfigBuilder.FLINK_VERSION;
 
 /** Service for submitting and interacting with Flink clusters and jobs. */
@@ -121,6 +123,9 @@ public class FlinkService {
     private final ArtifactManager artifactManager;
     private final FlinkConfigManager configManager;
     private final ExecutorService executorService;
+
+    private static final String CHECKPOINT_FAILURE_REASON =
+            "Not all required tasks are currently running.";
 
     public FlinkService(KubernetesClient kubernetesClient, FlinkConfigManager configManager) {
         this.kubernetesClient = kubernetesClient;
@@ -608,7 +613,7 @@ public class FlinkService {
     }
 
     public SavepointFetchResult fetchSavepointInfo(
-            String triggerId, String jobId, Configuration conf) {
+            String triggerId, String jobId, String jobStatus, Configuration conf) {
         LOG.info("Fetching savepoint result with triggerId: " + triggerId);
         try (RestClusterClient<String> clusterClient =
                 (RestClusterClient<String>) getClusterClient(conf)) {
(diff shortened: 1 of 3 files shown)


SteNicholas:
@wangyang0918, PTAL.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
```diff
@@ -618,32 +623,51 @@ public SavepointFetchResult fetchSavepointInfo(
...
-
-            if (response.get().resource().getLocation() == null) {
-                if (response.get().resource().getFailureCause() != null) {
-                    LOG.error(
-                            "Failure occurred while fetching the savepoint result",
-                            response.get().resource().getFailureCause());
-                    return SavepointFetchResult.error(
-                            response.get().resource().getFailureCause().toString());
-                } else {
+            while (true) {
```
morhidi, line 626: We cannot block the execution here, better to wait another reconcile loop

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
```diff
@@ -618,32 +623,51 @@ public SavepointFetchResult fetchSavepointInfo(
...
+                                EmptyRequestBody.getInstance());
+
+                if (response.get() == null || response.get().resource() == null) {
                     return SavepointFetchResult.pending();
                 }
+
+                if (response.get().resource().getLocation() == null) {
+                    // If the failure cause contains 'Not all required tasks are currently
+                    // running.', then continue until savepoint is successfully fetched.
+                    if (response.get().resource().getFailureCause() != null) {
```
morhidi, line 640: Once we get back the error `Aborting checkpoint. Failure reason: Not all required tasks are currently running.`, the checkpoint will never succeed.

morhidi (commented):
Thanks for the PR. Had a quick look, added some pointers

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
morhidi: Actually this error happens when the tasks are coming up. Checking the job status only, even when it is RUNNING state won't be sufficient.

morhidi:
@wangyang0918 we have the detailed JOB state: ```
```
{
  "jobs": [
    {
      "jid": "00000000000000000000000000000000",
      "name": "State machine job",
      "state": "RUNNING",
      "start-time": 1653471451258,
      "end-time": -1,
      "duration": 1666751,
      "last-modification": 1653472755540,
      "tasks": {
        "total": 4,
        "created": 0,
        "scheduled": 0,
        "deploying": 0,
        "running": 4,
        "finished": 0,
        "canceling": 0,
        "canceled": 0,
        "failed": 0,
        "reconciling": 0,
        "initializing": 0
      }
    }
  ]
}
```

The Rest Client doesn't seem to map everything https://github.com/apache/flink/blob/ffca99d5f369c21182b729ce5abe082ffa40d579/flink-clients/src/main/java/org/apache/flink/client/program/rest/RestClusterClient.java#L611

We can create our own listJobs method, and we could have an accurate state.

SteNicholas:
> The detailed JOB state:
> 
> ```
> {
>   "jobs": [
>     {
>       "jid": "00000000000000000000000000000000",
>       "name": "State machine job",
>       "state": "RUNNING",
>       "start-time": 1653471451258,
>       "end-time": -1,
>       "duration": 1666751,
>       "last-modification": 1653472755540,
>       "tasks": {
>         "total": 4,
>         "created": 0,
>         "scheduled": 0,
>         "deploying": 0,
>         "running": 4,
>         "finished": 0,
>         "canceling": 0,
>         "canceled": 0,
>         "failed": 0,
>         "reconciling": 0,
>         "initializing": 0
>       }
>     }
>   ]
> }
> ```
> 
> The Rest Client doesn't seem to map everything https://github.com/apache/flink/blob/ffca99d5f369c21182b729ce5abe082ffa40d579/flink-clients/src/main/java/org/apache/flink/client/program/rest/RestClusterClient.java#L611
> 
> We can create our own listJobs method, and we could have an accurate state. @wangyang0918 WDYT?

@morhidi IMO, the `listJobs` interface should return the task execution states.  With the task execution states, whether the job is actually running could be checked. In addtion, I'm confusing how to create own `listJobs` method. Do you mean that introduces `listJobsForOperator` interface in `ClusterClient`?

morhidi:
> > The detailed JOB state:
> > ```
> > {
> >   "jobs": [
> >     {
> >       "jid": "00000000000000000000000000000000",
> >       "name": "State machine job",
> >       "state": "RUNNING",
> >       "start-time": 1653471451258,
> >       "end-time": -1,
> >       "duration": 1666751,
> >       "last-modification": 1653472755540,
> >       "tasks": {
> >         "total": 4,
> >         "created": 0,
> >         "scheduled": 0,
> >         "deploying": 0,
> >         "running": 4,
> >         "finished": 0,
> >         "canceling": 0,
> >         "canceled": 0,
> >         "failed": 0,
> >         "reconciling": 0,
> >         "initializing": 0
> >       }
> >     }
> >   ]
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The Rest Client doesn't seem to map everything https://github.com/apache/flink/blob/ffca99d5f369c21182b729ce5abe082ffa40d579/flink-clients/src/main/java/org/apache/flink/client/program/rest/RestClusterClient.java#L611
> > We can create our own listJobs method, and we could have an accurate state. @wangyang0918 WDYT?
> 
> @morhidi IMO, the `listJobs` interface should return the task execution states. With the task execution states, whether the job is actually running could be checked. In addtion, I'm confusing how to create own `listJobs` method. Do you mean that introduces `listJobsForOperator` interface in `ClusterClient`?

No, we could do something like this in the `FlinkService`:

```
 public Collection<JobDetails> getJobDetails(Configuration conf) throws Exception {
        try (RestClusterClient<String> clusterClient =
                     (RestClusterClient<String>) getClusterClient(conf)) {

            var response =
                    clusterClient
                            .sendRequest(
                                    JobsOverviewHeaders.getInstance(),
                                    EmptyMessageParameters.getInstance(),
                                    EmptyRequestBody.getInstance())
                            .get(
                                    configManager
                                            .getOperatorConfiguration()
                                            .getFlinkClientTimeout()
                                            .toSeconds(),
                                    TimeUnit.SECONDS);

            return response.getJobs();
        }
    }
```

and use it instead of the current `Collection<JobStatusMessage> listJobs(Configuration conf)`. There's no reason to wait for the RestClient changes from the Flink runtime.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
SteNicholas: @morhidi, the main idea is that if the job status is running but the error `Aborting checkpoint. Failure reason: Not all required tasks are currently running.`, we could retry to trigger checkpoint until it's successfully triggered. IMO,  this error could happen when certain tasks are initialized.

SteNicholas:
> Collection<JobStatusMessage> listJobs(Configuration conf)

@morhidi +1. I understood your point. Above `getJobDetails` interface could help to check whether the job is running.
@wangyang0918, WDYT?

wangyang0918:
For Flink batch jobs, we do not require all the tasks running to indicate that the Flink job is running. So it does not make sense when we want to use `getJobDetails` to verify whether the job is running. But I agree it could be used to check whether all the tasks are running before triggering a savepoint.

gyfora:
@SteNicholas @wangyang0918 @morhidi I don't really think that checking the state will solve the fundamental problem here.

We cannot transactionally check and trigger the savepoint at the same time so there will always be cases when we thought we can trigger but the savepoint fails immediately.

The robust approach would be re-triggering the failed savepointing attempt upon encountering this error in the observer.

gyfora:
I am working on supporting periodic savepoint triggering, my PR will continue some improvements to the savepoint triggering logic.

I suggest we close this PR for now and revisit after that, as it should be much easier to tackle then.

SteNicholas:
@gyfora , ok. I will close the pull request.

PR #240, opened 2022-05-24
[FLINK-27747] Use tar instead of helm package when creating source release

wangyang0918:
`helm package` command will remove the apache license header in the `Chart.yaml`. Given that we already update the version in `update_branch_version`, it is unnecessary to specify the `--app-version` and `--version` again.

So we could simply use `tar` to prepare the helm package.


wangyang0918:
cc @mbalassi @gyfora WDYT?

gyfora (approved):
I think the change makes sense and simplifies the script also.

+1

PR #241, opened 2022-05-24
FLINK-27746: Remove COPY .git in dockerfile

tisonkun:
As build process doesn't depend on those files.

Signed-off-by: tison <wander4096@gmail.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,7 +36,6 @@ COPY $DOCS_DIR/pom.xml ./$DOCS_DIR/
 COPY $OPERATOR_DIR/src ./$OPERATOR_DIR/src
 COPY $WEBHOOK_DIR/src ./$WEBHOOK_DIR/src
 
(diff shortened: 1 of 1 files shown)


morhidi:
cc @SteNicholas I guess this was part of FLINK-27124] Flink Kubernetes operator prints starting logs with corrent version

tisonkun:
@morhidi thanks for pointing this out. Then I think @SteNicholas can handle this whole story properly. Closed as lack of background.

Although, an insight is that we may use another approach to generate git.property since the released tarball normally excludes `.git` artifacts (or we'd like to include them?).

morhidi:
cc @wangyang0918 WDYT?

wangyang0918:
I lean to skip copying the `.git` directory if not exist. Then the version in log will be `UNKNOWN` if building the image with source release.

PR #243, opened 2022-05-24
[FLINK-27746] Flink kubernetes operator docker image could not build with source release

SteNicholas:
Flink kubernetes operator docker image could not build with source release at present. The `.git` directory should be copied when the directory exists.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,7 +36,8 @@ COPY $DOCS_DIR/pom.xml ./$DOCS_DIR/
 COPY $OPERATOR_DIR/src ./$OPERATOR_DIR/src
 COPY $WEBHOOK_DIR/src ./$WEBHOOK_DIR/src
 
-COPY .git ./.git
+# Copy the .git directory when the directory exists.
+COPY *.git ./.git
 COPY tools ./tools
 
 RUN --mount=type=cache,target=/root/.m2 mvn -ntp clean install -pl !flink-kubernetes-docs -DskipTests=$SKIP_TESTS


Dockerfile
```diff
@@ -36,7 +36,7 @@ COPY $DOCS_DIR/pom.xml ./$DOCS_DIR/
 COPY $OPERATOR_DIR/src ./$OPERATOR_DIR/src
 COPY $WEBHOOK_DIR/src ./$WEBHOOK_DIR/src
 
-COPY .git ./.git
+COPY *.git ./.git
```
tisonkun, line 39: I'm unsure whether this line can possibly pollute unexpected files. At least we should add a comment about this tricky - it's a workaround since Dockerfile doesn't support conditional copy directly.

SteNicholas: @tisonkun, I have removed the `.git` directory and built image successfully. The change will copy the `.git` directory when the directory exists and not copy this directory when the `.git` directory doesn't exist.

wangyang0918 (approved):
It seems that we do not have a better way. https://localcoder.org/conditional-copy-add-in-dockerfile

Maybe we could revisit how to get the commit id after the 1.0.0 release.

tisonkun:
@wangyang0918 I'm thinking of `COPY . .` with a `.dockerignore` file. But yes, we can investigate better way after release 1.0.0.

wangyang0918:
I have created a new ticket FLINK-27759 to follow the potential improvement.

PR #245, opened 2022-05-26
[FLINK-27804] Do not observe cluster/job mid upgrade

gyfora:
UPGRADING state means we have either deleted the deployment for last-state upgrades or observed the result of the cancel with savepoint operation.

No need to further observe anything


wangyang0918:
Does this PR could work when the operator failed before patching the status, but after the job is canceled with savepoint successfully?

gyfora:
Yes @wangyang0918 as it would only be in UPGRADING state if patching succeeded after the cancel-with-savepoint operation

wangyang0918:
> Does this PR could work when the operator failed before patching the status, but after the job is canceled with savepoint successfully?

When the JobManager pod exactly failed over at the same time, we will never get the savepoint. But this is out the scope of this PR.

wangyang0918 (approved):
+1

gyfora:
Merging this for now, there is definitely follow up work to be done in the future to handle cases when the actual deployment step fails during an upgrade which is generally not handled at the moment

PR #246, opened 2022-05-27
[FLINK-27788] Adding annotation to k8 operator Pod

zeus1ammon:


helm/flink-kubernetes-operator/templates/flink-operator.yaml
```diff
@@ -34,6 +34,12 @@ spec:
         {{- include "flink-operator.selectorLabels" . | nindent 8 }}
       annotations:
         kubectl.kubernetes.io/default-container: {{ .Chart.Name }}
+      {{- $keyExist := .Values.operatorPodDeployment | default dict -}}
```
mbalassi, line 37: This could be just `if .Values.operatorPodDeployment.annotations`, right?
We are using the same convention throughout the helm chart.

helm/flink-kubernetes-operator/values.yaml
```diff
@@ -30,6 +30,9 @@ image:
 rbac:
   create: true
 
+operatorPodDeployment:
```
mbalassi, line 33: I find the name confusing. Is it for the pod or the deployment?

mbalassi (changes requested):
@zeus1ammon do you think it would make sense to add labels too if you are touching this anyway?

helm/flink-kubernetes-operator/values.yaml
zeus1ammon: it is for Pod annotation. I had named it this way, as we have only a helm template for deployment, but we can change it. Please let me know, if there is any suggestion

zeus1ammon:
> @zeus1ammon do you think it would make sense to add labels too if you are touching this anyway?

As labels are cosmetic and not functional like annotation (at least in case of our operator, as we may rarely end up running multiple operator in a cluster), we may not need to pass custom values in most cases.  However, if we envision a scenario, where customer would need to pass custom values for labels, apart from the default ones passed in helper.tpl (which should be mostly sufficient in most of the use cases to identify the operator pods via the default labels), we can plan to add "labels" too. Please let me know your thoughts

helm/flink-kubernetes-operator/templates/flink-operator.yaml
gyfora: How about using a construct like this?

```
{{- if index (.Values.defaultConfiguration) "flink-conf.yaml" }}
  {{- index (.Values.defaultConfiguration) "flink-conf.yaml" | nindent 4 -}}
{{- end }}
```

helm/flink-kubernetes-operator/values.yaml
gyfora: Then we could call it simply `operatorPod`

gyfora (changes requested):

mbalassi:
@zeus1ammon do you think you can update the PR based on @gyfora's suggestions this week?

zeus1ammon:
Sure. Sorry I was not actively checking the comments. Will address & test
all the comments and commit it back in the next couple of days.
thanks,
Jagan

mbalassi:
Looking forward to it @zeus1ammon.

zeus1ammon:
> Looking forward to it @zeus1ammon.

@mbalassi I have addressed the comments., please have a look

mbalassi (approved):
Thanks, @zeus1ammon. Merging.

PR #247, opened 2022-05-28
[FLINK-27668] Document dynamic operator configuration

SteNicholas:
The Kubernetes operator now support dynamic config changes through the operator ConfigMap.This feature is not documented properly and it should be added to the operations/configuration page.

**The brief change log**

- Document dynamic operator configuration in the `operations/configuration` page.

### DIFF
--- a/docs/content/docs/operations/configuration.md
+++ b/docs/content/docs/operations/configuration.md
@@ -60,16 +60,7 @@ Verify whether dynamic operator configuration updates is enabled via the `deploy
 2022-05-28 13:08:29,222 o.a.f.k.o.c.FlinkConfigManager [INFO ] Enabled dynamic config updates, checking config changes every PT5M
 ```
 
-When the interval for the controller to reschedule the reconcile process need to change with 30 seconds,  the ConfigMap can be directly edited via `kubectl patch` or `kubectl edit` command. The `kubernetes.operator.reconciler.reschedule.interval` config option, of which default value is 60 seconds, can directly change to 30s in the `defaultConfiguration` section:
-
-```yaml
-defaultConfiguration:
-  create: true
-  append: false
-  flink-conf.yaml: |+
-    # Flink Config Overrides
-    kubernetes.operator.reconciler.reschedule.interval: 30 s
-```
+To change config values dynamically the ConfigMap can be directly edited via `kubectl patch` or `kubectl edit` command. For example to change the reschedule interval you can override `kubernetes.operator.reconciler.reschedule.interval`.
 
 Verify whether the config value of `kubernetes.operator.reconciler.reschedule.interval` is updated to 30 seconds via the `deploy/flink-kubernetes-operator` log has:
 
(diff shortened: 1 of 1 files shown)


SteNicholas:
@gyfora , PTAL.

docs/content/docs/operations/configuration.md
```diff
@@ -50,11 +50,50 @@ defaultConfiguration:
...
+When the interval for the controller to reschedule the reconcile process need to change with 30 seconds, `kubernetes.operator.reconciler.reschedule.interval` in the `defaultConfiguration` section can directly change to 30s:
+
+```
+defaultConfiguration:
+  create: true
+  append: false
+  flink-conf.yaml: |+
+    # Flink Config Overrides
+    kubernetes.operator.reconciler.reschedule.interval: 30 s
+```
```
gyfora, line 83: I think this is not a good example because if you try to update the configmap through the Helm chart (using helm-upgrade) nothing will happen because helm only checks if the configmap is present. It will not update it.

We should simply show an example of editing the configmap directly.

docs/content/docs/operations/configuration.md
```diff
@@ -50,11 +50,50 @@ defaultConfiguration:
 
 To learn more about metrics and logging configuration please refer to the dedicated [docs page]({{< ref "docs/operations/metrics-logging" >}}).
 
+## Dynamic Operator Configuration
+
+The Kubernetes operator supports dynamic config changes through the operator ConfigMaps. Dynamic operator configuration can be enabled by setting `kubernetes.operator.dynamic.config.enabled`  to true. Time interval for checking dynamic config changes is specified by `kubernetes.operator.dynamic.config.check.interval` of which default value is 5 minutes.
```
gyfora, line 55: We should say that it is enabled by default, and can be **disabled** with the config.

docs/content/docs/operations/configuration.md
```diff
@@ -50,11 +50,39 @@ defaultConfiguration:
...
+When the interval for the controller to reschedule the reconcile process need to change with 30 seconds,  the ConfigMap can be directly edited via `kubectl patch` or `kubectl edit` command. The `kubernetes.operator.reconciler.reschedule.interval` config option, of which default value is 60 seconds, can directly change to 30s in the `defaultConfiguration` section:
+
+```yaml
+defaultConfiguration:
+  create: true
+  append: false
+  flink-conf.yaml: |+
+    # Flink Config Overrides
+    kubernetes.operator.reconciler.reschedule.interval: 30 s
+```
```
gyfora, line 72: Did you try this? I think updating the Helm values won't upgrade the configmap (if the operator is already running)

So please  change this to simply:

```
To change config values dynamically the ConfigMap can be directly edited via `kubectl patch` or `kubectl edit` command. For example to change the reschedule interval you can override `kubernetes.operator.reconciler.reschedule.interval`.
```

PR #248, opened 2022-05-30
[FLINK-27834] Copy all files except those filtered by .dockerignore

tisonkun:
Signed-off-by: tison <wander4096@gmail.com>

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,32 +1,23 @@
 # Compiled class file
-*.class
+**/*.class
 
 # Log file
-*.log
-
-# BlueJ files
-*.ctxt
+**/*.log
 
 # Package Files
-*.jar
-*.war
-*.nar
-*.ear
-*.zip
-*.tar.gz
-*.rar
+**/*.jar
+**/*.war
+**/*.nar
+**/*.ear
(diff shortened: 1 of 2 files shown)


tisonkun:
cc @morhidi @SteNicholas

.dockerignore
```diff
@@ -0,0 +1,32 @@
...
+*.jar
+*.war
+*.nar
+*.ear
+*.zip
+*.tar.gz
+*.rar
+
+### Maven template
+target/
```
wangyang0918, line 20: Could this line exclude all the target directories of submodules?

wangyang0918: It seems that the `.dockerignore` is a little different from `.gitignore`. https://docs.docker.com/engine/reference/builder/#dockerignore-file

Maybe we need to use `**/target`, `**/*.iml`, etc.

tisonkun: You're right. Will update.

tisonkun: Updated 97fab7e

gyfora (approved):

tisonkun:
Thanks for your reviews!

PR #249, opened 2022-05-31
[FLINK-26179] Support for periodic savepoints

gyfora:
This PR contains a few key improvements for savepoint triggering / management:

 1. Unify and simplify savepoint triggering between application and session jobs
 2. Simplify savepoint info update logic
 3. Add trigger type information to savepoints (we can also add new features to the cleanup logic based on this as a next step)
 4. Introduce periodic savepoint triggering
 5. Improve test coverage

### DIFF
--- a/docs/layouts/shortcodes/generated/kubernetes_operator_config_configuration.html
+++ b/docs/layouts/shortcodes/generated/kubernetes_operator_config_configuration.html
@@ -75,6 +75,12 @@
             <td>The interval before a savepoint trigger attempt is marked as unsuccessful.</td>
         </tr>
         <tr>
+            <td><h5>kubernetes.operator.periodic.savepoint.interval</h5></td>
+            <td style="word-wrap: break-word;">0 ms</td>
+            <td>Duration</td>
+            <td>Interval at which periodic savepoints will be triggered. The triggering schedule is not guaranteed, savepoints will be triggered as part of the regular reconcile loop.</td>
+        </tr>
+        <tr>
             <td><h5>kubernetes.operator.reconciler.flink.cancel.job.timeout</h5></td>
             <td style="word-wrap: break-word;">1 min</td>
             <td>Duration</td>
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
@@ -163,4 +163,12 @@ public class KubernetesOperatorConfigOptions {
                     .withDescription(
                             "Custom HTTP header for HttpArtifactFetcher. The header will be applied when getting the session job artifacts. "
                                     + "Expected format: headerKey1:headerValue1,headerKey2:headerValue2.");
+
+    public static final ConfigOption<Duration> PERIODIC_SAVEPOINT_INTERVAL =
+            ConfigOptions.key("kubernetes.operator.periodic.savepoint.interval")
+                    .durationType()
+                    .defaultValue(Duration.ZERO)
+                    .withDescription(
+                            "Interval at which periodic savepoints will be triggered. "
+                                    + "The triggering schedule is not guaranteed, savepoints will be triggered as part of the regular reconcile loop.");
 }
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/status/SavepointTriggerType.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/status/SavepointTriggerType.java
@@ -15,26 +15,16 @@
  * limitations under the License.
  */
 
-package org.apache.flink.kubernetes.operator.observer;
+package org.apache.flink.kubernetes.operator.crd.status;
 
-import lombok.Value;
-
-/** Result of a fetch savepoint operation. */
-@Value
-public class SavepointFetchResult {
-    String location;
-    boolean pending;
-    String error;
-
-    public static SavepointFetchResult error(String error) {
-        return new SavepointFetchResult(null, false, error);
-    }
-
-    public static SavepointFetchResult pending() {
-        return new SavepointFetchResult(null, true, null);
-    }
-
-    public static SavepointFetchResult completed(String location) {
-        return new SavepointFetchResult(location, false, null);
-    }
+/** Savepoint trigger mechanism. */
+public enum SavepointTriggerType {
+    /** Savepoint manually triggered by changing the savepointTriggerNonce. */
+    MANUAL,
+    /** Savepoint periodically triggered by the operator. */
+    PERIODIC,
+    /** Savepoint triggered during stateful upgrade. */
+    UPGRADE,
+    /** Savepoint trigger mechanism unknown, such as savepoint retrieved directly from Flink job. */
+    UNKNOWN
 }
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
@@ -286,17 +286,15 @@ public class ReconciliationUtils {
                         == JobManagerDeploymentStatus.MISSING);
     }
 
-    public static boolean isJobInTerminalState(FlinkDeploymentStatus status) {
+    public static boolean isJobInTerminalState(CommonStatus<?> status) {
         var jobState = status.getJobStatus().getState();
         return org.apache.flink.api.common.JobStatus.valueOf(jobState).isGloballyTerminalState();
     }
 
-    public static boolean isJobRunning(FlinkDeploymentStatus status) {
-        JobManagerDeploymentStatus deploymentStatus = status.getJobManagerDeploymentStatus();
-        return deploymentStatus == JobManagerDeploymentStatus.READY
-                && org.apache.flink.api.common.JobStatus.RUNNING
-                        .name()
-                        .equals(status.getJobStatus().getState());
+    public static boolean isJobRunning(CommonStatus<?> status) {
+        return org.apache.flink.api.common.JobStatus.RUNNING
+                .name()
+                .equals(status.getJobStatus().getState());
     }
 
     /**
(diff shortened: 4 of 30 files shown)


gyfora:
cc @Aitozi @morhidi

morhidi:
Thanks @gyfora looks promising. I'll have a look later today.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
```diff
@@ -163,4 +163,12 @@ public class KubernetesOperatorConfigOptions {
...
                             "Custom HTTP header for HttpArtifactFetcher. The header will be applied when getting the session job artifacts. "
                                     + "Expected format: headerKey1:headerValue1,headerKey2:headerValue2.");
+
+    public static final ConfigOption<Duration> PERIODIC_SAVEPOINT_INTERVAL =
+            ConfigOptions.key("kubernetes.operator.periodic.savepoint.interval")
+                    .durationType()
+                    .noDefaultValue()
+                    .withDescription(
+                            "Interval at which periodic savepoints will be triggered. "
+                                    + "The triggering schedule is not guaranteed an will be done as part of the regular reschedule loop.");
```
morhidi, line 173: nit: reschedule loop->reconcile loop,  sounds better to me here

docs/layouts/shortcodes/generated/kubernetes_operator_config_configuration.html
```diff
@@ -74,6 +74,12 @@
             <td>Duration</td>
             <td>The interval before a savepoint trigger attempt is marked as unsuccessful.</td>
         </tr>
+        <tr>
+            <td><h5>kubernetes.operator.periodic.savepoint.interval</h5></td>
```
morhidi, line 78: Do we expect this property being set globally?

gyfora: Users can choose to set it globally or override on a per job level, it works both ways.

gyfora: I can change this so setting it to 0 would turn it off explicitly

gyfora:
@morhidi I ran into this issue also with the disposal. I think it's unrelated to my change this is what prompted me to try to improve the dispose logic a little. I will dig a little further into this tomorrow

morhidi:
@gyfora I was comparing the savepoint history with the ones on the dashboard. Does it make sense to display it the same way e.g. latest on top, human readable timestamp?

Aitozi (commented):
Thanks @gyfora It looks good, I left some minor comments.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
```diff
@@ -653,7 +655,10 @@ public SavepointFetchResult fetchSavepointInfo(
     public void disposeSavepoint(String savepointPath, Configuration conf) throws Exception {
         try (RestClusterClient<String> clusterClient =
                 (RestClusterClient<String>) getClusterClient(conf)) {
-            clusterClient.disposeSavepoint(savepointPath);
+            clusterClient.sendRequest(
```
Aitozi, line 658: I check the `disposeSavepoint` internal implementation, why current way is better?

docs/layouts/shortcodes/generated/kubernetes_operator_config_configuration.html
```diff
@@ -74,6 +74,12 @@
...
+            <td style="word-wrap: break-word;">(none)</td>
+            <td>Duration</td>
+            <td>Interval at which periodic savepoints will be triggered. The triggering schedule is not guaranteed an will be done as part of the regular reschedule loop.</td>
```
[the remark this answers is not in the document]
gyfora: fixing

Sushant20:
@gyfora Related question: If periodic savepoint is enabled, how do we restore the job with specific savepoint path as part of native integration operator? There is something similar available on other operator page: https://github.com/GoogleCloudPlatform/flink-on-k8s-operator/blob/master/docs/savepoints_guide.md#starting-a-job-from-a-savepoint

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/SavepointUtils.java
```diff
@@ -17,40 +17,146 @@
...
+                && SavepointUtils.savepointInProgress(jobStatus)) {
+            jobStatus.getSavepointInfo().resetTrigger();
+            LOG.error("Job is not running, cancelling savepoint operation");
+            EventUtils.createOrUpdateEvent(
+                    client,
+                    resource,
+                    EventUtils.Type.Warning,
+                    "SavepointError",
+                    "Savepoint failed for savepointTriggerNonce: "
+                            + resource.getSpec().getJob().getSavepointTriggerNonce(),
```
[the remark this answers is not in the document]
gyfora: 👍

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultValidator.java
```diff
@@ -197,6 +197,12 @@ private Optional<String> validateJobSpec(JobSpec job, Map<String, String> confMa
                         String.format(
                                 "Savepoint could not be manually triggered for the running job while config key[%s] is not set",
                                 CheckpointingOptions.SAVEPOINT_DIRECTORY.key()));
+            } else if (configuration.contains(
+                    KubernetesOperatorConfigOptions.PERIODIC_SAVEPOINT_INTERVAL)) {
+                return Optional.of(
```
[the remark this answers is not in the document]
gyfora: They are not that bad :) at least it gives a clear error

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
```diff
@@ -640,8 +642,8 @@ public SavepointFetchResult fetchSavepointInfo(
                 }
             }
             Savepoint savepoint =
-                    new Savepoint(
-                            System.currentTimeMillis(), response.get().resource().getLocation());
+                    Savepoint.of(
+                            response.get().resource().getLocation(), SavepointTriggerType.UNKNOWN);
```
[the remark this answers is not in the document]
gyfora: I will actually change the fetch method to return only the location which is the only thing it actually gets.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
gyfora: The disposeSavepoint method of the RestClusterClient, first triggers the dispose action and then keeps polling Flink until it completes.

The way we use it, this is completely unnecessary as we do not handle/retry dispose actions from our side anyways. This way we can save some rest api calls.

However we lose the error logs unfortunately if the dispose actually failed.

gyfora:
@Aitozi I fixed your comments

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
```diff
@@ -286,17 +286,15 @@ private static boolean jmMissingForRunningDeployment(FlinkDeployment deployment)
...
     }
 
-    public static boolean isJobRunning(FlinkDeploymentStatus status) {
-        JobManagerDeploymentStatus deploymentStatus = status.getJobManagerDeploymentStatus();
-        return deploymentStatus == JobManagerDeploymentStatus.READY
-                && org.apache.flink.api.common.JobStatus.RUNNING
-                        .name()
-                        .equals(status.getJobStatus().getState());
+    public static boolean isJobRunning(CommonStatus<?> status) {
+        return org.apache.flink.api.common.JobStatus.RUNNING
```
SteNicholas, line 295: Could this indeed represent wheter the job is running? IMO, this should check with the task execution states in the job.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/status/SavepointTriggerType.java
```diff
@@ -0,0 +1,30 @@
...
+/** Savepoint trigger mechanism. */
+public enum SavepointTriggerType {
+    /** Savepoint manually triggered by changing the savepointTriggerNonce. */
+    MANUAL,
+    /** Savepoint periodically triggered by the operator. */
+    PERIODIC,
+    /** Savepoint triggered during stateful upgrade. */
+    UPGRADE,
+    /** Savepoint trigger mechanism unknown. */
+    UNKNOWN
```
SteNicholas, line 29: Which case is the `UNKNOWN` used for?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
gyfora: I agree that this should be improved, but completely out of scope for this PR :) 

I will open a follow up jira.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/status/SavepointTriggerType.java
gyfora: When the mechanism is not know, for example any savepoint triggered before this update will not have this information.

Or in cases when the savepoint info is retrieved from the flink cluster directly

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/SavepointUtils.java
```diff
@@ -17,40 +17,153 @@
...
+                                                .getJob()
+                                                .getSavepointTriggerNonce());
+        if (triggerNonceChanged) {
+            return Optional.of(SavepointTriggerType.MANUAL);
+        }
+
+        var savepointInterval =
+                conf.getOptional(KubernetesOperatorConfigOptions.PERIODIC_SAVEPOINT_INTERVAL)
+                        .map(Duration::toMillis)
+                        .orElse(0L);
```
SteNicholas, line 116: Could the default value of `PERIODIC_SAVEPOINT_INTERVAL ` be 0 milliseconds?

gyfora: makes sense

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/status/SavepointTriggerType.java
SteNicholas: @gyfora, are the above cases above mentioned need to add in comments or document to illustrate?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
Aitozi: Got it.

Aitozi (approved):

SteNicholas (approved):
LGTM.

PR #252, opened 2022-06-03
[FLINK-27889] fix: Catch the error when last reconciled spec is null

Miuler:
[[](https://github.com/apache/flink-kubernetes-operator/commit/1f47f33485f0af8d1367b55b98dccbe12db452c2)[FLINK-27889](https://issues.apache.org/jira/browse/FLINK-27889)[] fix: Catch the error when last reconciled spec is null](https://github.com/apache/flink-kubernetes-operator/commit/1f47f33485f0af8d1367b55b98dccbe12db452c2)

### DIFF
--- a/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentControllerTest.java
+++ b/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentControllerTest.java
@@ -276,6 +276,14 @@ public class FlinkDeploymentControllerTest {
 
     @ParameterizedTest
     @EnumSource(FlinkVersion.class)
+    public void cleanUpDeploymentWithError(FlinkVersion flinkVersion) {
+        FlinkDeployment flinkDeployment = TestUtils.buildApplicationCluster();
+        var deleteControl = testController.cleanup(flinkDeployment, context);
+        assertNotNull(deleteControl);
+    }
+
+    @ParameterizedTest
+    @EnumSource(FlinkVersion.class)
     public void verifyUpgradeFromSavepoint(FlinkVersion flinkVersion) {
         FlinkDeployment appCluster = TestUtils.buildApplicationCluster(flinkVersion);
         appCluster.getSpec().getJob().setUpgradeMode(UpgradeMode.SAVEPOINT);
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
@@ -324,7 +324,12 @@ public class ApplicationReconciler extends AbstractDeploymentReconciler {
     @Override
     @SneakyThrows
     protected void shutdown(FlinkDeployment flinkApp) {
-        flinkService.cancelJob(flinkApp, UpgradeMode.STATELESS);
+        var status = flinkApp.getStatus();
+        if (status.getReconciliationStatus().getLastReconciledSpec() == null) {
+            flinkService.deleteClusterDeployment(flinkApp.getMetadata(), status, true);
+        } else {
+            flinkService.cancelJob(flinkApp, UpgradeMode.STATELESS);
+        }
     }
 
     private void triggerSavepoint(FlinkDeployment deployment) throws Exception {


gyfora:
@Miuler could you please provide a detailed description of what this change does specifically and what is the error scenario that we are handling here?

This is a very critical and somewhat complex part of the operator logic so we should try to be aware of the full impact of these changes.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
```diff
@@ -354,8 +355,54 @@ protected ClusterClient<String> getClusterClient(Configuration config) throws Ex
...
-        var deploymentStatus = deployment.getStatus();
+        Optional<String> savepointOpt = Optional.empty();
+
+        Configuration conf = null;
+        FlinkDeploymentStatus deploymentStatus = deployment.getStatus();
+        try {
+            conf = configManager.getObserveConfig(deployment);
+            savepointOpt = cancelJob(deployment, upgradeMode, conf, deploymentStatus);
+        } catch (LastReconciledSpecException e) {
+            LOG.warn("Last reconciled spec is null", e);
```
gyfora, line 366: Why would we cancel the job if last reconciled spec is null? that means it was never deployed

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/exception/LastReconciledSpecException.java
```diff
@@ -0,0 +1,25 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.kubernetes.operator.exception;
+
+/** Exception when last status is null. */
+public class LastReconciledSpecException extends RuntimeException {
```
gyfora, line 21: Seems like you are using exception for flow control instead of explicitly checking the last reconciled spec, can we avoid this?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
```diff
@@ -354,8 +355,54 @@ protected ClusterClient<String> getClusterClient(Configuration config) throws Ex
...
+        }
+
+        switch (upgradeMode) {
```
gyfora, line 369: Could you please not refactor the method? Seems like a lot of unnecessary changes, it’s hard to see what actually changed in the logic

gyfora:
In addition I would like to kindly ask you to not include any refactorings with this fix initially. This will help us understand the problem your solution better.

After that if we collectively feel that the cancel method needs refactoring we can discuss. At the current stage I think that method is fine as is, but maybe your fix will bring out some new currently overlooked aspects of this logic.

Miuler:
> @Miuler could you please provide a detailed description of what this change does specifically and what is the error scenario that we are handling here?
> 
> This is a very critical and somewhat complex part of the operator logic so we should try to be aware of the full impact of these changes.

Excuse me, I commente in Jira the error, 

Is a real error, in my cluster, I ended up with 4 deployments in certification (UAT) but without any pod, and every time the cluster restarted it tried to delete, and then it died. Without any solution.

I refactored with intention the show is the solution, but I can do a squash in a single commit.

The intention is to divide it into two parts, the first to try to connect to the cluster and delete it and then delete the deployment, in order to catch the error when the last spec is null and only delete the deployment. And I use a Exception but.

And use a custom exception, because you already threw a RuntimeException, I just wrapped it to distinguish it from any other JVM RuntimeException, and couldn't handle it because it would change the behavior of FlinkConfigManager#getObserveConfig which is used in several places

Miuler:
image

This log was sent to me by sysadmin, and when I do the `kubectl get flinkdeployment` command, I get:

```
kubectl get flinkdeployments -A
NAMESPACE       NAME                      AGE
flink-x-01   cosmosdb         9h
flink-x-01   cosmosdb-test1   8h
flink-x-02   cosmosdb-test2   76m
```

and none can start or erase.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
Miuler: For some reason the deployment was corrupted, and could never finish starting, and it is impossible to delete it, impossible.

This happened to me several times, in different namespaces.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/exception/LastReconciledSpecException.java
Miuler: For some reason the deployment was corrupted, and could never finish starting, and it is impossible to delete it, impossible.

This happened to me several times, in different namespaces.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
Miuler: The method would be enormous. It seemed like a good practice to me to separate the behavior of deleting the cluster using the flink client and, on the other hand, deleting the kubernetes deployment. At first it is more complicated but then it becomes easier to read

gyfora:
Do you have a yaml that I can use to reproduce this? I am still not convinced that we need to refactor etc

gyfora:
And please don’t get me wrong, all I am trying to do is keep this as simple as possible :) I have already spent quite a lot of time simplifying these methods and the flow to make it more understandable

Miuler:
> Do you have a yaml that I can use to reproduce this? I am still not convinced that we need to refactor etc

In the test is the real yaml, but hide some names, the yaml initial I don't have it, but I think a FlinkDeployment whitout job was created first, I think they didn't delete it, I'm not sure, and then they added a job inside the FlinkDeployment, but with jarURI with http and it gave an error, was tried to fix it and pass it to local, but They couldn't anymore, I'm not sure if they tried to directly remove the kebertenes pod and deploy, only that in the end they left the flink unstable, and this can happen, users do everything

gyfora:
@Miuler i will take a look later tonight to see what is the minimal fix that we need. Then we can discuss if we want to refactor the cancel in a separate ticket maybe

gyfora:
@Miuler I looked into this a little and this is the only thing we need to change for this to work (and make your test to pass):

```
diff --git a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
index 442206f9..f5beddbb 100644
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
@@ -324,7 +324,12 @@ public class ApplicationReconciler extends AbstractDeploymentReconciler {
     @Override
     @SneakyThrows
     protected void shutdown(FlinkDeployment flinkApp) {
-        flinkService.cancelJob(flinkApp, UpgradeMode.STATELESS);
+        var status = flinkApp.getStatus();
+        if (status.getReconciliationStatus().getLastReconciledSpec() == null) {
+            flinkService.deleteClusterDeployment(flinkApp.getMetadata(), status, true);
+        } else {
+            flinkService.cancelJob(flinkApp, UpgradeMode.STATELESS);
+        }
     }

     private void triggerSavepoint(FlinkDeployment deployment) throws Exception {
```

This is also a more straightforward fix because the main problem is calling cancel job on something that was never deployed (it doesn't really make sense in the first place so no wonder it was broken)

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentControllerTest.java
```diff
@@ -274,6 +278,89 @@ public void verifyInProgressDeploymentWithCrashLoopBackoff() throws Exception {
         validatingResponseProvider.assertValidated();
     }
 
+    @ParameterizedTest
+    @EnumSource(FlinkVersion.class)
+    public void cleanUpDeploymentWithError(FlinkVersion flinkVersion)
+            throws JsonProcessingException {
```
gyfora, line 284: I think this test case is bit of an overkill.
To reproduce the logic and validate the fix it could simply be replaced with the following 2 lines:

```
        FlinkDeployment flinkDeployment = TestUtils.buildApplicationCluster();
        testController.cleanup(flinkDeployment, context);
```

We don't really need the error state, json etc because thats not related to the actual issue at hand I think. Also the rest of the method tests deployments, upgrades etc that are already tested in other methods in this class.

Miuler: ok

Miuler:
> @Miuler I looked into this a little and this is the only thing we need to change for this to work (and make your test to pass):
> 
> ```
> diff --git a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
> index 442206f9..f5beddbb 100644
> --- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
> +++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
> @@ -324,7 +324,12 @@ public class ApplicationReconciler extends AbstractDeploymentReconciler {
>      @Override
>      @SneakyThrows
>      protected void shutdown(FlinkDeployment flinkApp) {
> -        flinkService.cancelJob(flinkApp, UpgradeMode.STATELESS);
> +        var status = flinkApp.getStatus();
> +        if (status.getReconciliationStatus().getLastReconciledSpec() == null) {
> +            flinkService.deleteClusterDeployment(flinkApp.getMetadata(), status, true);
> +        } else {
> +            flinkService.cancelJob(flinkApp, UpgradeMode.STATELESS);
> +        }
>      }
> 
>      private void triggerSavepoint(FlinkDeployment deployment) throws Exception {
> ```
> 
> This is also a more straightforward fix because the main problem is calling cancel job on something that was never deployed (it doesn't really make sense in the first place so no wonder it was broken)

ok, I thinks it is better. :D

Miuler:
@gyfora I already corrected the code

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigManager.java
```diff
@@ -134,6 +135,11 @@ public Configuration getDeployConfig(ObjectMeta objectMeta, FlinkDeploymentSpec
     public Configuration getObserveConfig(FlinkDeployment deployment) {
         var deployedSpec = ReconciliationUtils.getDeployedSpec(deployment);
         if (deployedSpec == null) {
+            try {
+                LOG.debug("deployment: {}", objectMapper.writeValueAsString(deployment));
+            } catch (JsonProcessingException e) {
+                LOG.debug("deployment: {}", deployment);
+            }
```
gyfora, line 142: I think this debug log does not add any value here. Could we please remove it?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
```diff
@@ -229,14 +229,12 @@ private JarRunResponseBody runJar(
                                     ? RestoreMode.DEFAULT
                                     : null);
             LOG.info("Submitting job: {} to session cluster.", jobID.toHexString());
+            var clientTimeout =
+                    configManager.getOperatorConfiguration().getFlinkClientTimeout().toSeconds();
+            LOG.debug("clientTimeout: {}", clientTimeout);
```
gyfora, line 234: The client timeout configuration is static and already logged on info level when the operator starts. We should remove the logging here.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentControllerTest.java
```diff
@@ -274,6 +279,15 @@ public void verifyInProgressDeploymentWithCrashLoopBackoff() throws Exception {
...
+        FlinkDeployment flinkDeployment = TestUtils.buildApplicationCluster();
+        val deleteControl = testController.cleanup(flinkDeployment, context);
```
gyfora, line 287: I think you accidentally used lombok `val` instead of `var`

gyfora (changes requested):
Looks good, I have added a few minor comments.

Also based on your feedback I opened a follow up jira for improving the cancelJob method (https://issues.apache.org/jira/browse/FLINK-27896). If you would like to take that ticket, I am happy to work together with you to make this nicer :)

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentControllerTest.java
Miuler: Can I use val in the tests? I prefer to use final instead of a mutable variable.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
Miuler: ok

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentControllerTest.java
gyfora: We generally don’t use this and also avoid final variables in these cases as it doesn’t affect anything only adds boilerplate on these cases.

also we should restrict Lombok usage where it adds significant tangible value

gyfora: final variables inside methods are a bit of a controversy in Java. I would say the flink best practice is to only use this in performance critical low level code

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
Miuler: I have a doubt, what is the properti for this clinetTimeout?

I add this in the values.yaml
```
  flink-conf.yaml: |+
    # Flink Config Overrides
    client.timeout: 4 MINUTE
```

and in my pod I see this
```
exec -ti  migration-cosmosdb-wape-5578f9948c-9t6xm  -- bash

root@migration-cosmosdb-wape-5578f9948c-9t6xm:/opt/flink# grep time conf/flink-conf.yaml
client.timeout: 4 MINUTE
```

but in my log I see this
```
2022-06-04 12:20:30,505 o.a.f.k.o.c.FlinkConfigManager [INFO ] Updating default configuration to {blob.server.port=6124, taskmanager.memory.process.size=1728m, client.timeout=4 MINUTE, jobmanager.memory.process.size=1600m, jobmanager.rpc.port=6123, taskmanager.rpc.port=6122, queryable-state.proxy.ports=6125, paralle
lism.default=2, taskmanager.numberOfTaskSlots=2, kubernetes.operator.metrics.reporter.slf4j.interval=5 MINUTE, kubernetes.operator.observer.progress-check.interval=5 s, kubernetes.operator.metrics.reporter.slf4j.factory.class=org.apache.flink.metrics.slf4j.Slf4jReporterFactory, kubernetes.operator.reconciler.resched
ule.interval=15 s}
...
...
2022-06-04 12:20:30,762 i.j.o.a.c.ExecutorServiceManager [DEBUG] Initialized ExecutorServiceManager executor: class java.util.concurrent.ThreadPoolExecutor, timeout: 10
...
...
2022-06-04 12:28:12,124 o.a.f.k.o.s.FlinkService       [DEBUG][flink-wape-02/migration-cosmosdb-wape-sessionjob] clientTimeout: 10
...
...
2022-06-04 12:26:52,685 i.j.o.p.e.ReconciliationDispatcher [ERROR][flink-wape-02/migration-cosmosdb-wape-sessionjob] Error during event processing ExecutionScope{ resource id: CustomResourceID{name='migration-cosmosdb-wape-sessionjob', namespace='flink-wape-02'}, version: null} failed.
org.apache.flink.kubernetes.operator.exception.ReconciliationException: org.apache.flink.util.FlinkRuntimeException: java.util.concurrent.TimeoutException
    at org.apache.flink.kubernetes.operator.controller.FlinkSessionJobController.reconcile(FlinkSessionJobController.java:117)
    at org.apache.flink.kubernetes.operator.controller.FlinkSessionJobController.reconcile(FlinkSessionJobController.java:59)
    at io.javaoperatorsdk.operator.processing.Controller$2.execute(Controller.java:101)
    at io.javaoperatorsdk.operator.processing.Controller$2.execute(Controller.java:76)
    at io.javaoperatorsdk.operator.api.monitoring.Metrics.timeControllerExecution(Metrics.java:34)
    at io.javaoperatorsdk.operator.processing.Controller.reconcile(Controller.java:75)
    at io.javaoperatorsdk.operator.processing.event.ReconciliationDispatcher.reconcileExecution(ReconciliationDispatcher.java:143)
    at io.javaoperatorsdk.operator.processing.event.ReconciliationDispatcher.handleReconcile(ReconciliationDispatcher.java:109)
    at io.javaoperatorsdk.operator.processing.event.ReconciliationDispatcher.handleDispatch(ReconciliationDispatcher.java:74)
    at io.javaoperatorsdk.operator.processing.event.ReconciliationDispatcher.handleExecution(ReconciliationDispatcher.java:50)
    at io.javaoperatorsdk.operator.processing.event.EventProcessor$ControllerExecution.run(EventProcessor.java:349)
    at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
    at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
    at java.base/java.lang.Thread.run(Unknown Source)
Caused by: org.apache.flink.util.FlinkRuntimeException: java.util.concurrent.TimeoutException
    at org.apache.flink.kubernetes.operator.service.FlinkService.runJar(FlinkService.java:240)
    at org.apache.flink.kubernetes.operator.service.FlinkService.submitJobToSessionCluster(FlinkService.java:198)
    at org.apache.flink.kubernetes.operator.reconciler.sessionjob.FlinkSessionJobReconciler.submitAndInitStatus(FlinkSessionJobReconciler.java:164)
    at org.apache.flink.kubernetes.operator.reconciler.sessionjob.FlinkSessionJobReconciler.reconcile(FlinkSessionJobReconciler.java:88)
    at org.apache.flink.kubernetes.operator.reconciler.sessionjob.FlinkSessionJobReconciler.reconcile(FlinkSessionJobReconciler.java:48)
    at org.apache.flink.kubernetes.operator.controller.FlinkSessionJobController.reconcile(FlinkSessionJobController.java:115)
    ... 13 more
Caused by: java.util.concurrent.TimeoutException
    at java.base/java.util.concurrent.CompletableFuture.timedGet(Unknown Source)
    at java.base/java.util.concurrent.CompletableFuture.get(Unknown Source)
    at org.apache.flink.kubernetes.operator.service.FlinkService.runJar(FlinkService.java:237)
    ... 18 more
```

I see my `client.timeout=4 MINUTE` but also a `clientTimeout 10` seconds?

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentControllerTest.java
Miuler: ok

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigManager.java
Miuler: Ok

Miuler:
@gyfora I already revert the commit where I add logs, I only have the doubt of the timeout.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
Miuler: > I have a doubt, what is the properti for this clinetTimeout?
> 
> I add this in the values.yaml
> 
> ```
>   flink-conf.yaml: |+
>     # Flink Config Overrides
>     client.timeout: 4 MINUTE
> ```
> 
> and in my pod I see this
> 
> ```
> exec -ti  migration-cosmosdb-wape-5578f9948c-9t6xm  -- bash
> 
> root@migration-cosmosdb-wape-5578f9948c-9t6xm:/opt/flink# grep time conf/flink-conf.yaml
> client.timeout: 4 MINUTE
> ```
> 
> but in my log I see this
> 
> ```
> 2022-06-04 12:20:30,505 o.a.f.k.o.c.FlinkConfigManager [INFO ] Updating default configuration to {blob.server.port=6124, taskmanager.memory.process.size=1728m, client.timeout=4 MINUTE, jobmanager.memory.process.size=1600m, jobmanager.rpc.port=6123, taskmanager.rpc.port=6122, queryable-state.proxy.ports=6125, paralle
> lism.default=2, taskmanager.numberOfTaskSlots=2, kubernetes.operator.metrics.reporter.slf4j.interval=5 MINUTE, kubernetes.operator.observer.progress-check.interval=5 s, kubernetes.operator.metrics.reporter.slf4j.factory.class=org.apache.flink.metrics.slf4j.Slf4jReporterFactory, kubernetes.operator.reconciler.resched
> ule.interval=15 s}
> ...
> ...
> 2022-06-04 12:20:30,762 i.j.o.a.c.ExecutorServiceManager [DEBUG] Initialized ExecutorServiceManager executor: class java.util.concurrent.ThreadPoolExecutor, timeout: 10
> ...
> ...
> 2022-06-04 12:28:12,124 o.a.f.k.o.s.FlinkService       [DEBUG][flink-wape-02/migration-cosmosdb-wape-sessionjob] clientTimeout: 10
> ...
> ...
> 2022-06-04 12:26:52,685 i.j.o.p.e.ReconciliationDispatcher [ERROR][flink-wape-02/migration-cosmosdb-wape-sessionjob] Error during event processing ExecutionScope{ resource id: CustomResourceID{name='migration-cosmosdb-wape-sessionjob', namespace='flink-wape-02'}, version: null} failed.
> org.apache.flink.kubernetes.operator.exception.ReconciliationException: org.apache.flink.util.FlinkRuntimeException: java.util.concurrent.TimeoutException
>     at org.apache.flink.kubernetes.operator.controller.FlinkSessionJobController.reconcile(FlinkSessionJobController.java:117)
>     at org.apache.flink.kubernetes.operator.controller.FlinkSessionJobController.reconcile(FlinkSessionJobController.java:59)
>     at io.javaoperatorsdk.operator.processing.Controller$2.execute(Controller.java:101)
>     at io.javaoperatorsdk.operator.processing.Controller$2.execute(Controller.java:76)
>     at io.javaoperatorsdk.operator.api.monitoring.Metrics.timeControllerExecution(Metrics.java:34)
>     at io.javaoperatorsdk.operator.processing.Controller.reconcile(Controller.java:75)
>     at io.javaoperatorsdk.operator.processing.event.ReconciliationDispatcher.reconcileExecution(ReconciliationDispatcher.java:143)
>     at io.javaoperatorsdk.operator.processing.event.ReconciliationDispatcher.handleReconcile(ReconciliationDispatcher.java:109)
>     at io.javaoperatorsdk.operator.processing.event.ReconciliationDispatcher.handleDispatch(ReconciliationDispatcher.java:74)
>     at io.javaoperatorsdk.operator.processing.event.ReconciliationDispatcher.handleExecution(ReconciliationDispatcher.java:50)
>     at io.javaoperatorsdk.operator.processing.event.EventProcessor$ControllerExecution.run(EventProcessor.java:349)
>     at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
>     at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
>     at java.base/java.lang.Thread.run(Unknown Source)
> Caused by: org.apache.flink.util.FlinkRuntimeException: java.util.concurrent.TimeoutException
>     at org.apache.flink.kubernetes.operator.service.FlinkService.runJar(FlinkService.java:240)
>     at org.apache.flink.kubernetes.operator.service.FlinkService.submitJobToSessionCluster(FlinkService.java:198)
>     at org.apache.flink.kubernetes.operator.reconciler.sessionjob.FlinkSessionJobReconciler.submitAndInitStatus(FlinkSessionJobReconciler.java:164)
>     at org.apache.flink.kubernetes.operator.reconciler.sessionjob.FlinkSessionJobReconciler.reconcile(FlinkSessionJobReconciler.java:88)
>     at org.apache.flink.kubernetes.operator.reconciler.sessionjob.FlinkSessionJobReconciler.reconcile(FlinkSessionJobReconciler.java:48)
>     at org.apache.flink.kubernetes.operator.controller.FlinkSessionJobController.reconcile(FlinkSessionJobController.java:115)
>     ... 13 more
> Caused by: java.util.concurrent.TimeoutException
>     at java.base/java.util.concurrent.CompletableFuture.timedGet(Unknown Source)
>     at java.base/java.util.concurrent.CompletableFuture.get(Unknown Source)
>     at org.apache.flink.kubernetes.operator.service.FlinkService.runJar(FlinkService.java:237)
>     ... 18 more
> ```
> 
> I see my `client.timeout=4 MINUTE` but also a `clientTimeout 10` seconds?

All my pipelines are multiplied by this timeout :(

gyfora: We should probably improve these config names but this is currently controlled by:
```
kubernetes.operator.observer.flink.client.timeout
```

gyfora:
Merged this to main manually

PR #253, opened 2022-06-03
[FLINK-27892] More than 1 secondary resource related to primary

morhidi:
Creating the primary cache indexers using composite keys [name,namespace] in EventSourceUtils. Added a sample job to the comments to reproduce the issue: https://issues.apache.org/jira/browse/FLINK-27892. 

Note: Added a one liner method to emphasise what we are doing, can be removed if needed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,9 +95,11 @@ jobs:
           - description: 'Default configuration'
             namespace: default
             extraArgs: ''
+            tests: "test_application_kubernetes_ha.sh test_application_operations.sh test_sessionjob_kubernetes_ha.sh test_sessionjob_operations.sh"
           - description: 'WatchNamespaces enabled'
             namespace: flink
             extraArgs: '--create-namespace --set "watchNamespaces={default,flink}"'
+            tests: "test_application_kubernetes_ha.sh test_application_operations.sh test_sessionjob_kubernetes_ha.sh test_sessionjob_operations.sh test_multi_sessionjob.sh"
         versions:
           - image: flink:1.13
             flinkVersion: v1_13
@@ -153,9 +155,10 @@ jobs:
           sed -i "s/image: flink:.*/image: ${{ matrix.versions.image }}/" e2e-tests/data/*.yaml
           sed -i "s/flinkVersion: .*/flinkVersion: ${{ matrix.versions.flinkVersion }}/" e2e-tests/data/*.yaml
           git diff HEAD
-          ls e2e-tests/test_*.sh | while read script_test;do \
-            echo "Running $script_test"
-            bash $script_test || exit 1
+          TESTS="${{ matrix.config.tests }}"
+          for test in $TESTS;do \
+            echo "Running e2e-tests/$test"
+            bash e2e-tests/$test || exit 1
           done
           git reset --hard
       - name: Stop the operator
--- a/e2e-tests/data/multi-sessionjob.yaml
+++ b/e2e-tests/data/multi-sessionjob.yaml
@@ -1,0 +1,186 @@
+################################################################################
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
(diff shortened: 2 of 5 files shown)


morhidi:
cc @Aitozi

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/EventSourceUtils.java
```diff
@@ -113,4 +133,8 @@ public static InformerEventSource<Deployment, FlinkDeployment> getDeploymentInfo
 
         return new InformerEventSource<>(configuration, context);
     }
+
+    private static String compsiteKey(String name, String namespace) {
```
Aitozi, line 137: Is it `composite` ?

Aitozi (commented):
@morhidi The fix looks good, left one minor comments. 
BTW, do you think whether we need to add some tests to guard the `EventSourceUtils` work as expected ?

gyfora:
I agree it would be great to add some tests for this @morhidi . I know we didn't have tests before either but it's a bit concerning to be able to change such core logic without any test changes/failures.

morhidi:
Thanks folks, sure thing! Tho I'm off this week. I'll take care of this once I'm back next week.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/EventSourceUtils.java
morhidi: yup

morhidi:
@Aitozi @gyfora I ended up adding some extra e2e tests for multi namespace session job deployments. We don't have too much resources on a minikube cluster but I managed to squeeze in the 2 x (1 session cluster + 1 session job) we need here. PTAL

Aitozi (approved):
Thanks for updating, +1 % a format comment

.github/workflows/ci.yml
```diff
@@ -153,7 +153,82 @@ jobs:
...
+      - name: Cache local Maven repository
+        uses: actions/cache@v3
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
+      - name: Build with Maven
+        run: |
+          set -o pipefail; mvn clean install -DskipTests | tee ./mvn.log; set +o pipefail
```
Aitozi, line 200: Do we still need `Build with Maven` in the `e2e_multi_ci` ?

.github/workflows/ci.yml
```diff
@@ -153,7 +153,82 @@ jobs:
...
+        run: |
+          kubectl get pods -A
+          kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v1.7.1/cert-manager.yaml
+          kubectl -n cert-manager wait --all=true --for=condition=Available --timeout=300s deploy
+      - name: Build image
+        run: |
+          export SHELL=/bin/bash
+          export DOCKER_BUILDKIT=1
+          eval $(minikube -p minikube docker-env)
+          docker build --progress=plain --no-cache -f ./Dockerfile -t flink-kubernetes-operator:ci-latest --progress plain .
```
Aitozi, line 219: nit: Is there some way to avoid building twice image

e2e-tests/data/multi-sessionjob.yaml
```diff
@@ -0,0 +1,186 @@
...
+apiVersion: networking.k8s.io/v1
+kind: IngressClass
+metadata:
+  annotations:
+    ingressclass.kubernetes.io/is-default-class: "true"
+  labels:
+    app.kubernetes.io/component: controller
+  name: nginx
+spec:
+  controller: k8s.io/ingress-nginx
```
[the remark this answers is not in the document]
morhidi: aww not again, 'Preferences > Editor > General > Ensure line feed at file end on save' FTW :D

.github/workflows/ci.yml
morhidi: I'll try to simply add the tests to the existing e2e_ci workflow, we can probably optimize the tests to run parallel without building the images, I'll open a follow up if I succeed

.github/workflows/ci.yml
morhidi: we do it with every combination, we could probably avoid it, but that needs some extra efforts

morhidi:
Thanks for the review @Aitozi. I also added a follow up for speeding up the e2e tests. https://issues.apache.org/jira/browse/FLINK-28059

PR #254, opened 2022-06-03
[FLINK-27812] Support Dynamic Change of Watched Namespaces

morhidi:
This PR enables handling namespace changes via Dynamically Operator Configuration through the config option `kubernetes.operator.watched.namespaces`. Watched namespaces originally were defined using environment variables, thus namespaces changes required a POD restart always. With the current feature Helm upgrades won't trigger an automatic POD restart on namespace changes anymore. The feature is disabled by default, can be enabled by setting `kubernetes.operator.dynamic.namespaces.enabled=true` 

Suggestions for testing:

```
helm install flink-kubernetes-operator helm/flink-kubernetes-operator --set "watchNamespaces={default}"
helm upgrade flink-kubernetes-operator helm/flink-kubernetes-operator --set "watchNamespaces={default,flink}"
```
The helm upgrade will update the config map, but it takes some time until the new configs are propagated in the operator. Worth setting the `kubernetes.operator.dynamic.config.check.interval: 10 s` for local testing.

I have also moved the InformerManager into the webhook module since it is only used there, and opened a [FR](https://github.com/java-operator-sdk/java-operator-sdk/issues/1286) in JOSDK to get rid of it on long term.

I would be happy to get your feedback on this @wangyang0918 @Aitozi @gyfora @tweise


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
```diff
@@ -187,4 +188,18 @@ public class KubernetesOperatorConfigOptions {
...
+                    .defaultValue(Constants.WATCH_ALL_NAMESPACES)
+                    .withDescription(
+                            "Comma separated list of namespaces the operator monitors for custom resources. Defaults to "
+                                    + Constants.WATCH_ALL_NAMESPACES);
+
+    public static final ConfigOption<Boolean> OPERATOR_DYNAMIC_NAMESPACES_ENABLED =
+            ConfigOptions.key("kubernetes.operator.dynamic.namespaces.enabled")
+                    .booleanType()
+                    .defaultValue(false)
+                    .withDescription("Enables dynamic change of namespaces. Defaults to false");
```
gyfora, line 204: nit:  `Enables dynamic change of watched/monitored namespaces`

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigManager.java
```diff
@@ -109,14 +117,19 @@ public Configuration getDefaultConfig() {
 
     @VisibleForTesting
     public void updateDefaultConfig(Configuration newConf) {
-        if (newConf.equals(defaultConfig)) {
+        if (defaultConfig != null
+                && newConf != null
```
gyfora, line 121: Do we need these nullchecks here? Why would they ever be null?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigManager.java
```diff
@@ -109,14 +117,19 @@ public Configuration getDefaultConfig() {
...
             LOG.info("Default configuration did not change, nothing to do...");
             return;
         }
 
         LOG.info("Updating default configuration to {}", newConf);
-        this.operatorConfiguration =
-                FlinkOperatorConfiguration.fromConfiguration(newConf, namespaces);
+        this.operatorConfiguration = FlinkOperatorConfiguration.fromConfiguration(newConf);
+
+        if (this.operatorConfiguration.getDynamicNamespacesEnabled()) {
```
gyfora, line 130: should we check if thhe watched namespaces actually changed?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/informer/InformerManager.java
```diff
@@ -104,4 +100,24 @@ Map<String, SharedIndexInformer<CR>> createRunnableInformer(
             return informers;
         }
     }
+
+    public void setNameSpaces(Set<String> watchedNamespaces) {
```
gyfora, line 104: nit: `setNameSpaces` -> `setNamespaces`

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
morhidi: makes sense

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigManager.java
morhidi: Yes, this is needed

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigManager.java
morhidi: Sure, I'll add an extra check here

morhidi:
Thank you for your review @gyfora I've addressed your comments.

PR #257, opened 2022-06-06
[hotfix] Remove incorrect savepoint path check

gyfora:
The validation for valid lastSavepoint path is obsolete and can actually cause issues when trying to resume a job after a last-state suspend operation.

With the current logic upgradeMode mostly affects how the job is suspended and not how they are restored, as the way the job was suspended directly affects what state is available for restore.

The only exception is stateless restore which directly forces empty state, otherwise the job will always use the last state available (in other words, last-state restore is implicit)


gyfora:
cc @wangyang0918

PR #258, opened 2022-06-06
[chore] Change the session cluster not found log level to error

haoxins:


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/sessionjob/SessionJobHelper.java
```diff
@@ -54,7 +54,7 @@ public boolean sessionClusterReady(Optional<FlinkDeployment> flinkDeploymentOpt)
                 return true;
             }
         } else {
-            logger.info("Session cluster deployment is not found");
+            logger.error("Session cluster deployment is not found");
```
gyfora, line 57: I think INFO/WARN is more appropriate as this is completely expected to happen and it's not really an error. @Aitozi what do you think?

Aitozi: +1

Aitozi (approved):

PR #260, opened 2022-06-07
[FLINK-27497] Track terminal job states in the observer

Aitozi:
This PR improve the logic in `JobStatusObserver`
- generate an event after each job state changed. 
- fetch the JobResult for `FAILED` job and forward the error to the CR status

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
@@ -18,8 +18,10 @@
 package org.apache.flink.kubernetes.operator.observer;
 
 import org.apache.flink.configuration.Configuration;
+import org.apache.flink.kubernetes.operator.crd.AbstractFlinkResource;
 import org.apache.flink.kubernetes.operator.crd.status.JobStatus;
 import org.apache.flink.kubernetes.operator.service.FlinkService;
+import org.apache.flink.kubernetes.operator.utils.EventUtils;
 import org.apache.flink.runtime.client.JobStatusMessage;
 
 import org.slf4j.Logger;
@@ -34,6 +36,7 @@ import java.util.concurrent.TimeoutException;
 public abstract class JobStatusObserver<CTX> {
 
     private static final Logger LOG = LoggerFactory.getLogger(JobStatusObserver.class);
+    private static final int MAX_ERROR_STRING_LENGTH = 512;
     private final FlinkService flinkService;
 
     public JobStatusObserver(FlinkService flinkService) {
@@ -43,11 +46,13 @@ public abstract class JobStatusObserver<CTX> {
     /**
      * Observe the status of the flink job.
      *
-     * @param jobStatus The job status to be observed.
+     * @param resource The custom resource to be observed.
      * @param deployedConfig Deployed job config.
      * @return If job found return true, otherwise return false.
      */
-    public boolean observe(JobStatus jobStatus, Configuration deployedConfig, CTX ctx) {
+    public boolean observe(
+            AbstractFlinkResource<?, ?> resource, Configuration deployedConfig, CTX ctx) {
+        var jobStatus = resource.getStatus().getJobStatus();
         LOG.info("Observing job status");
         var previousJobStatus = jobStatus.getState();
         List<JobStatusMessage> clusterJobStatuses;
@@ -63,19 +68,13 @@ public abstract class JobStatusObserver<CTX> {
         }
 
         if (!clusterJobStatuses.isEmpty()) {
-            Optional<String> targetJobStatus = updateJobStatus(jobStatus, clusterJobStatuses);
-            if (targetJobStatus.isEmpty()) {
+            Optional<JobStatusMessage> targetJobStatusMessage =
+                    filterTargetJob(jobStatus, clusterJobStatuses);
+            if (targetJobStatusMessage.isEmpty()) {
                 jobStatus.setState(org.apache.flink.api.common.JobStatus.RECONCILING.name());
                 return false;
             } else {
-                if (targetJobStatus.get().equals(previousJobStatus)) {
-                    LOG.info("Job status ({}) unchanged", previousJobStatus);
-                } else {
-                    LOG.info(
-                            "Job status successfully updated from {} to {}",
-                            previousJobStatus,
-                            targetJobStatus.get());
-                }
+                updateJobStatus(resource, targetJobStatusMessage.get(), deployedConfig);
             }
             return true;
         } else {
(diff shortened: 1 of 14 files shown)


Aitozi:
cc @gyfora @wangyang0918

Aitozi:
I am not sure which error info will be more suitable (the full stack vs the simplified message as below ) . 

```
error: 'org.apache.flink.runtime.JobException: Recovery is suppressed by NoRestartBackoffTimeStrategy'
```

I prefer to show the full stack here, which will be more explicit . The drawback is will pollute the resource object

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/ApplicationObserver.java
```diff
@@ -82,11 +79,14 @@ protected boolean observeFlinkCluster(
...
 
+        var previousJobState = jobStatus.getState();
         boolean jobFound =
                 jobStatusObserver.observe(
-                        jobStatus,
+                        flinkApp.getStatus(),
                         deployedConfig,
                         new ApplicationObserverContext(flinkApp, context, deployedConfig));
+        EventUtils.generateEventsOnJobStatusChanged(
+                kubernetesClient, previousJobState, jobStatus, flinkApp);
```
gyfora, line 89: I think this method should be called in the JobStatusObserver directly to avoid having it in multiple places. There you already have previousJobState etc.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
```diff
@@ -94,14 +91,64 @@ private void ifRunningMoveToReconciling(JobStatus jobStatus, String previousJobS
...
+
+        if (clusterJobStatus.getJobState() == org.apache.flink.api.common.JobStatus.FAILED) {
+            try {
+                var result =
+                        flinkService.requestJobResult(deployedConfig, clusterJobStatus.getJobId());
+                result.getSerializedThrowable()
+                        .ifPresent(
+                                t -> {
+                                    var error = t.getFullStringifiedStackTrace();
+                                    if (error != null && !error.equals(status.getError())) {
```
gyfora, line 141: `!error.equals(status.getError())` check not necessary

gyfora (changes requested):
Added 1-2 minor comments but looks good :) 

Regarding the error, I would be in favor of shorter error messages in the status. If the job is FAILED the user can always check the logs and need to investigate anyways as it's a terminal state.

Even if we decide for the full stack trace we need to trim it to a max size to be safe.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
Aitozi: This is meant to avoid the duplicated log in operator

gyfora: oh I see, sry

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/ApplicationObserver.java
Aitozi: I have tried that way, but it relies on the resource object, It will have to touch some more class to pass it.

gyfora: Wouldn't it be enough to simply change the jobstatus observer to get the full resource object instead of the status only?

Aitozi: Good suggestion, I have pushed the commit, PTAL again

Aitozi:
> Added 1-2 minor comments but looks good :)
> 
> Regarding the error, I would be in favor of shorter error messages in the status. If the job is FAILED the user can always check the logs and need to investigate anyways as it's a terminal state.
> 
> Even if we decide for the full stack trace we need to trim it to a max size to be safe.

I keep the stack trace with a limited size of 0.5k to give the as much as error info in status

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
```diff
@@ -31,23 +36,29 @@
...
      * Observe the status of the flink job.
      *
-     * @param jobStatus The job status to be observed.
+     * @param resource The custom resource to be observed.
      * @param deployedConfig Deployed job config.
      * @return If job found return true, otherwise return false.
      */
-    public boolean observe(JobStatus jobStatus, Configuration deployedConfig, CTX ctx) {
+    public boolean observe(
+            AbstractFlinkResource<SPEC, STATUS> resource, Configuration deployedConfig, CTX ctx) {
```
gyfora, line 60: Do we need the new generic parameters? I think AbstractFlinkResource<?, ?> would be enough and we also dont need to change the class signature that way

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
```diff
@@ -31,23 +36,29 @@
...
 
     private static final Logger LOG = LoggerFactory.getLogger(JobStatusObserver.class);
+    private static final int MAX_ERROR_STRING_LENGTH = 512;
     private final FlinkService flinkService;
+    private final KubernetesClient kubernetesClient;
 
-    public JobStatusObserver(FlinkService flinkService) {
+    public JobStatusObserver(FlinkService flinkService, KubernetesClient client) {
         this.flinkService = flinkService;
+        this.kubernetesClient = client;
```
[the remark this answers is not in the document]
Aitozi: hah, thanks for remind. I did't notice that just now.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
Aitozi: Replaced with generic parameters

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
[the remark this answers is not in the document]
Aitozi: Removed

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
```diff
@@ -94,14 +95,84 @@ private void ifRunningMoveToReconciling(JobStatus jobStatus, String previousJobS
...
+        var jobStatus = resource.getStatus().getJobStatus();
+        var previousJobStatus = jobStatus.getState();
+
+        jobStatus.setState(clusterJobStatus.getJobState().name());
+        jobStatus.setJobName(clusterJobStatus.getJobName());
+        jobStatus.setJobId(clusterJobStatus.getJobId().toHexString());
+        jobStatus.setStartTime(String.valueOf(clusterJobStatus.getStartTime()));
+
+        if (jobStatus.getState().equals(previousJobStatus)) {
+            LOG.info("Job status ({}) unchanged", previousJobStatus);
```
gyfora, line 129: Should we return after this line?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
Aitozi: thanks, looks more simple

Aitozi:
I have optimized the method, please take a look again. @gyfora

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
Aitozi: Fixed

PR #261, opened 2022-06-08
[FLINK-27961] Include the resource uid when generate the event name

Aitozi:
Currently the event name do not include the uid of the target resource. If a resource is recreated, it will be associated with the former object's events. It's not expected and will be confusing with the empty events when describe the resource.


gyfora (approved):
Good catch

PR #264, opened 2022-06-09
[FLINK-27913] Remove savepointHistoryMaxCount and savepointHistoryMaxAge from FlinkOperatorConfiguration

SteNicholas:
Currently `savepointHistoryMaxCount` and `savepointHistoryMaxAge` is part of the `FlinkOperatorConfiguration` class which means that users cannot override this from their user deployment. `savepointHistoryMaxCount` and `savepointHistoryMaxAge` should be removed and get directly from the effective config. A max allowed value for these configurations that is configured on the FlinkOperatorConfiguration level so that users cannot infinitely grow the status size and cause problems for the k8s api should be introduced.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkOperatorConfiguration.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkOperatorConfiguration.java
@@ -40,7 +40,8 @@ public class FlinkOperatorConfiguration {
     Duration flinkCancelJobTimeout;
     Duration flinkShutdownClusterTimeout;
     String artifactsBaseDir;
-    Configuration operatorConfig;
+    Integer savepointHistoryCountThreshold;
+    Duration savepointHistoryAgeThreshold;
 
     public static FlinkOperatorConfiguration fromConfiguration(
             Configuration operatorConfig, Set<String> watchedNamespaces) {
@@ -78,6 +79,15 @@ public class FlinkOperatorConfiguration {
                 operatorConfig.get(
                         KubernetesOperatorConfigOptions.OPERATOR_USER_ARTIFACTS_BASE_DIR);
 
+        Integer savepointHistoryCountThreshold =
+                operatorConfig.get(
+                        KubernetesOperatorConfigOptions
+                                .OPERATOR_SAVEPOINT_HISTORY_MAX_COUNT_THRESHOLD);
+        Duration savepointHistoryAgeThreshold =
+                operatorConfig.get(
+                        KubernetesOperatorConfigOptions
+                                .OPERATOR_SAVEPOINT_HISTORY_MAX_AGE_THRESHOLD);
+
         String flinkServiceHostOverride = null;
         if (EnvUtils.get("KUBERNETES_SERVICE_HOST") == null) {
             // not running in k8s, simplify local development
@@ -95,6 +105,7 @@ public class FlinkOperatorConfiguration {
                 flinkCancelJobTimeout,
                 flinkShutdownClusterTimeout,
                 artifactsBaseDir,
-                operatorConfig);
+                savepointHistoryCountThreshold,
+                savepointHistoryAgeThreshold);
     }
 }
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkOperatorConfiguration.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkOperatorConfiguration.java
@@ -40,8 +40,8 @@ public class FlinkOperatorConfiguration {
     Duration flinkCancelJobTimeout;
     Duration flinkShutdownClusterTimeout;
     String artifactsBaseDir;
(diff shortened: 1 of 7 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/SavepointObserver.java
```diff
@@ -148,13 +149,38 @@ void cleanupSavepointHistory(
...
+                        KubernetesOperatorConfigOptions.OPERATOR_SAVEPOINT_HISTORY_MAX_COUNT);
+        if (countThreshold != null && maxCount > countThreshold) {
+            LOG.warn(
+                    "Maximum number of savepoint history entries to retain changes to {}. The value of '{}' exceeds the value of '{}'",
+                    countThreshold,
+                    KubernetesOperatorConfigOptions.OPERATOR_SAVEPOINT_HISTORY_MAX_COUNT.key(),
+                    KubernetesOperatorConfigOptions.OPERATOR_SAVEPOINT_HISTORY_COUNT_THRESHOLD
+                            .key());
+            maxCount = countThreshold;
+        }
```
gyfora, line 165: Can we refactor this pattern into a utility method? 
```
public static <T extends Comparable<T>> T getValueWithThreshold(ConfigOption<T> option, ConfigOption<T> threshold, Configuraiton config) {
...
}
```

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
```diff
@@ -145,12 +145,25 @@ public class KubernetesOperatorConfigOptions {
...
                     .defaultValue(10)
                     .withDescription("Maximum number of savepoint history entries to retain.");
 
+    public static final ConfigOption<Duration> OPERATOR_SAVEPOINT_HISTORY_AGE_THRESHOLD =
+            ConfigOptions.key("kubernetes.operator.savepoint.history.age.threshold")
+                    .durationType()
+                    .noDefaultValue()
+                    .withDescription(
+                            "Age threshold for savepoint history entries to retain. Due to lazy clean-up, the most recent savepoint may live longer than the max age.");
+
```
gyfora, line 166: I think the new config keys should be:
`OPERATOR_SAVEPOINT_HISTORY_MAX_AGE.key() + ".threshold"`
`OPERATOR_SAVEPOINT_HISTORY_MAX_COUNT.key() + ".threshold"`

To make this straightforward to users and admins

gyfora (changes requested):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkOperatorConfiguration.java
```diff
@@ -40,8 +40,7 @@ public class FlinkOperatorConfiguration {
     Duration flinkCancelJobTimeout;
     Duration flinkShutdownClusterTimeout;
     String artifactsBaseDir;
-    int savepointHistoryMaxCount;
-    Duration savepointHistoryMaxAge;
+    Configuration operatorConfig;
```
gyfora, line 43: I would prefer to keep the concrete thershold values as fields here to keep this consistent

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/ConfigOptionUtils.java
```diff
@@ -0,0 +1,58 @@
...
+     * @param thresholdOption The threshold option.
+     * @param effectiveConfig The effective config.
+     * @param operatorConfig The operator config.
+     * @return The value of {@link ConfigOption} with threshold.
+     */
+    public static <T extends Comparable<T>> T getValueWithThreshold(
+            ConfigOption<T> configOption,
+            ConfigOption<T> thresholdOption,
+            Configuration effectiveConfig,
+            Configuration operatorConfig) {
```
gyfora, line 44: instead of passing the `thresholdOption` + `operatorConfig` we could simply pass the threshold value directly that we get from the operatorconfig to make this simpler

gyfora (changes requested):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/ConfigOptionUtils.java
```diff
@@ -0,0 +1,58 @@
...
+     * @param configOption The config option.
+     * @param thresholdOption The threshold option.
+     * @param effectiveConfig The effective config.
+     * @param operatorConfig The operator config.
+     * @return The value of {@link ConfigOption} with threshold.
+     */
+    public static <T extends Comparable<T>> T getValueWithThreshold(
+            ConfigOption<T> configOption,
+            ConfigOption<T> thresholdOption,
+            Configuration effectiveConfig,
```
gyfora, line 43: you could simply call the variable conf/config this is just a utility method

PR #265, opened 2022-06-11
[FLINK-27613] Add label for the session job to help list the session …

Aitozi:
…jobs in the same session cluster

This PR is based on https://github.com/apache/flink-kubernetes-operator/pull/215 but use the mutator webhook to do the same thing


flink-kubernetes-webhook/src/main/java/org/apache/flink/kubernetes/operator/admission/Utils.java
```diff
@@ -0,0 +1,34 @@
...
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.fabric8.kubernetes.api.model.HasMetadata;
+
+/** Admission utils. */
+public class Utils {
+
+    private static final ObjectMapper mapper = new ObjectMapper();
+
+    public static <T> T convertToTargetType(HasMetadata resource, Class<T> targetType)
```
gyfora, line 29: I think we don’t need this utility class, ObjectMapper already has a convertValue method that does exactly this.

flink-kubernetes-webhook/src/main/java/org/apache/flink/kubernetes/operator/admission/mutator/FlinkMutator.java
```diff
@@ -0,0 +1,70 @@
...
+                    return sessionJob;
+                } catch (Exception e) {
+                    throw new RuntimeException(e);
+                }
+            }
+        }
+        return resource;
+    }
+
+    private void patchInternalLabel(FlinkSessionJob flinkSessionJob) {
```
gyfora, line 56: I think it would be clearer to call it simply `setSessionTargetLabel`

helm/flink-kubernetes-operator/templates/webhook.yaml
```diff
@@ -109,4 +109,36 @@ webhooks:
         operator: In
         values: [{{- range .Values.watchNamespaces }}{{ . | quote }},{{- end}}]
   {{- end }}
-  {{- end }}
+---
```
gyfora, line 112: I think we should have a separate config flag to enable this. We should handle the validating and mutating webhooks independently of each other as some envs might want one but not the other.

gyfora (changes requested):

flink-kubernetes-webhook/src/main/java/org/apache/flink/kubernetes/operator/admission/Utils.java
Aitozi: 👍🏻, Fixed

PR #266, opened 2022-06-12
[FLINK-27960] Make the apt-get updating optional

Aitozi:
when developing, we could use to speed up

```
docker build --build-arg FAST=true . -t <repo>/flink-kubernetes-operator:latest
```


mbalassi (commented):
@Aitozi since you are introducing this anyway we could do it similarly to the way disable maven tests in the default docker build, but still run them during CI:
https://github.com/apache/flink-kubernetes-operator/blob/main/.github/workflows/ci.yml#L66

I like the fast flag potentially controlling both, up to you. 😃

Aitozi:
@mbalassi Thanks for reviewing

> I like the fast flag potentially controlling both

But the `SKIP_TESTS` and `FAST` have different default value. The `FAST` should be true to trigger update when building the artifact image (for the security reason https://issues.apache.org/jira/browse/FLINK-27728). So the default value need to be true, Right?

mbalassi:
@Aitozi apologies if I was unclear, for the sake of clarity of explanation let us introduce the `SKIP_OS_UPDATE` flag (you can choose to call this `FAST` or name the "shared" control flag `FAST` if you think having that makes sense).

1. By default / for dev builds:
```
SKIP_TESTS: true
SKIP_OS_UPDATE: true
```
So that if you choose to use `FAST` to control both it can be set to `true` by default.

2. For the CI / automated builds:
```
SKIP_TESTS: false
SKIP_OS_UPDATE: false
```
So that if you choose to use `FAST` to control both then the CI could set this flag to `false`.

If you choose to not introduce a shared flag, then I slightly prefer calling the flag you introduce here `SKIP_OS_UPDATE`.

Aitozi:
@mbalassi Thanks for your detail explanation. I finally choose to use different options 😄 and set the default `SKIP_OS_UPDATE` to false, It's for the security consideration :). 
The main reason I want to use two options is that we can skip tests and not skip update os when building the final artifacts

morhidi:
+1 on the feature @Aitozi. IMO we can go with `SKIP_OS_UPDATE=true` by default, that would affect local builds only.

morhidi:
We can use the `SKIP_OS_UPDATE=false` in the CI workflow two. Not sure how much it adds to the build process but on OS upgrade can pull in some regression too.

Aitozi:
Hi @morhidi do you mean I do not add the in `.github/workflows/docker-bake.hcl` ? 

```
  args = {
    SKIP_OS_UPDATE = false
  }
```

Is this file will be used to build the release artifacts?  I think we have to make sure the release artifacts are enable the OS update

morhidi:
> Hi @morhidi do you mean I do not need to add the in `.github/workflows/docker-bake.hcl` ?
> 
> ```
>   args = {
>     SKIP_OS_UPDATE = false
>   }
> ```
> 
> Is this file will be used to build the release artifacts? I think we have to make sure the release artifacts are enable the OS update

I meant this one https://github.com/apache/flink-kubernetes-operator/blob/1dd9850fb37d4f7d572f46b247470bda13fb3660/.github/workflows/ci.yml#L66. So that the e2e test run with the configuration that we'll release.

Aitozi:
@morhidi Get your meaning, make sense to me. Updated

morhidi:
LGTM, thanks!

PR #269, opened 2022-06-14
[FLINK-27688] Add pluggable FlinkResourceListener interface

gyfora:
This PR introduces a new interface called FlinkResourceListener that allows operator users to listen to resource status changes and events triggered by the operator itself.

This provides a key integration point for platforms built on top of the operators. Similar to custom validators or metrics reporters, FlinkResourceListener implementations are also loaded using the built in plugin mechanism.

To listeners are integrated into the StatusHelper and the newly introduced EventHelper objects which will seamlessly forward all updates/events without added complexity.

The PR also improves event triggering and simplifies some test cases that were affected by this change.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/listener/ListenerUtils.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/listener/ListenerUtils.java
@@ -1,0 +1,137 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.kubernetes.operator.listener;
+
+import org.apache.flink.annotation.VisibleForTesting;
+import org.apache.flink.api.java.tuple.Tuple2;
+import org.apache.flink.configuration.ConfigConstants;
+import org.apache.flink.configuration.ConfigOptions;
+import org.apache.flink.configuration.Configuration;
+import org.apache.flink.configuration.CoreOptions;
+import org.apache.flink.configuration.DelegatingConfiguration;
+import org.apache.flink.core.plugin.PluginUtils;
+import org.apache.flink.kubernetes.operator.config.FlinkConfigManager;
+import org.apache.flink.kubernetes.operator.utils.FlinkUtils;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/** Flink resource listener utilities. */
+public class ListenerUtils {
+
+    private static final Logger LOG = LoggerFactory.getLogger(FlinkUtils.class);
+
+    private static final String PREFIX = "kubernetes.operator.plugins.listeners.";
+    private static final String SUFFIX = ".class";
+    private static final Pattern PTN =
+            Pattern.compile(Pattern.quote(PREFIX) + "([\\S&&[^.]]*)" + Pattern.quote(SUFFIX));
+    private static final List<String> EXTRA_PARENT_FIRST_PATTERNS =
+            List.of("io.fabric8", "com.fasterxml");
+
+    /**
+     * Load {@link FlinkResourceListener} implementations from the plugin directory. Only listeners
+     * that are explicitly named and configured will be enabled.
+     *
+     * <p>Config format: kubernetes.operator.plugins.listeners.test.class: com.myorg.MyListener
+     * kubernetes.operator.plugins.listeners.test.k1: v1
+     *
+     * @param configManager {@link FlinkConfigManager} to access plugin configurations.
+     * @return Enabled listeners.
+     */
+    public static Collection<FlinkResourceListener> discoverListeners(
+            FlinkConfigManager configManager) {
+        var listeners = new ArrayList<FlinkResourceListener>();
+        var conf = getListenerBaseConf(configManager);
+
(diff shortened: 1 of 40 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/listener/ListenerUtils.java
```diff
@@ -0,0 +1,127 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(FlinkUtils.class);
+
+    private static final String PREFIX = "kubernetes.operator.plugins.listeners.";
+    private static final String SUFFIX = ".class";
+    private static final Pattern PTN =
+            Pattern.compile(Pattern.quote(PREFIX) + "([\\S&&[^.]]*)" + Pattern.quote(SUFFIX));
+    private static final List<String> EXTRA_PARENT_FIRST_PATTERNS =
+            List.of("io.fabric8", "com.fasterxml");
+
+    public static Collection<FlinkResourceListener> discoverListeners(
```
tweise, line 57: javadoc to summarize how we find listeners?

tweise, line 57: javadoc to summarize how we find listeners?

gyfora: Will do, I will also add a docs page once we agreed on the initial interface.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/EventHelper.java
```diff
@@ -0,0 +1,104 @@
...
+import io.fabric8.kubernetes.api.model.Event;
+import io.fabric8.kubernetes.client.KubernetesClient;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Collection;
+import java.util.function.BiConsumer;
+
+/** Helper class for status management and updates. */
+public class EventHelper {
```
tweise, line 34: `EventHelper` is a bit awkward from a naming perspective, since this is a pluggable component that needs to be injected everywhere. I see that you want to separate the creation of the event from the listener. Would `EventEmitter` be a fit for this?

gyfora: I am a bit torn here, this class doesn't emit events on its own. So maybe `EventUtil` would be better if we dont like the `helper` name. We also have `StatusHelper` now that could also be renamed to `StatusUtil`. These classes create an interface between the operator and the kubernetes client so we can control these interactions better.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/StatusHelper.java
```diff
@@ -122,4 +144,45 @@ public <T extends CustomResource<?, STATUS>> void removeCachedStatus(T resource)
...
+                    listeners.forEach(
+                            listener -> {
+                                try {
+                                    if (resource instanceof FlinkDeployment) {
+                                        listener.onDeploymentStatusUpdate(ctx);
+                                    } else {
+                                        listener.onSessionJobStatusUpdate(ctx);
+                                    }
+                                } catch (Exception e) {
+                                    LOG.error(
```
tweise, line 179: Shouldn't this propagate the exception since it is an error that needs to fixed vs. swallowed? Otherwise WARN would be more appropriate.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/EventHelper.java
tweise: Maybe we can look at it from the client perspective rather than how it is implemented? It functions as a kind of "sink" for events.

tweise:
@gyfora overall LGTM. Some details to be discussed. A bit more work on the docs maybe. I could not easily find a description that the listeners will provide a way to emit the events in addition to storing them in k8s, which is key to the k8s native integration.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/EventHelper.java
```diff
@@ -0,0 +1,104 @@
...
+        this.eventListener = eventListener;
+    }
+
+    public boolean triggerEvent(
+            AbstractFlinkResource<?, ?> resource,
+            EventUtils.Type type,
+            String reason,
+            String message,
+            EventUtils.Component component) {
+        return EventUtils.createOrUpdateEvent(
```
tweise, line 53: It would also be possible to handle the k8s event create/update as listener implementation, but perhaps that is overkill when the intention is to keep it hard wired?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/EventHelper.java
morhidi: It is called EventRecorder in kubebuilder:
https://book-v1.book.kubebuilder.io/beyond_basics/creating_events.html

morhidi: We could probably request some JOSDK API support for this feature. What do you think @csviri?

gyfora: `EventRecorder` and `StatusRecorder` sounds perfect, thanks @morhidi , will make this change

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/StatusHelper.java
gyfora: You are probably right, it might be a key integration point, propagating the error makes sense

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/EventHelper.java
gyfora: I feel that would be slightly weird because the listener is actually listening to the events :) Technically it could be done to create the event and add a hardwired listener that sends it to kubernetes but I don't think it would make the logic easier to understand.

gyfora:
@tweise @morhidi updated the PR according to the comments:
 - Rename EventHelper -> EventRecorder, StatusHelper -> StatusRecorded
 - Propagate listener errors
 - Add doc page for implementing listeners

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/EventHelper.java
csviri: `EventReconder` makes sense for sure, I think `StatusReconder` we discussed before that it should be rather an other resource where the state is recorded (in general). 
But please create an issue that describes the uses cases and/or motivation for JOSDK! Eventually EventRecorder will be supported for sure. We can discuss the other. 

If you think the `FlinkResourceListener` makes sense to generalize, please describe the motivation behind that too.

PR #270, opened 2022-06-16
[FLINK-28087]Add validation for the meta.name of FlinkDeployment CR.

Grypse:
The meta.name of FlinkDeployment CR must be match the regex:
` '[a-z]([-a-z0-9]*[a-z0-9])?'`

By RRC-1035, a DNS-1035 label must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?').

The rest service name of Flink Cluster uses the meta.name of FlinkDeployment CR, So the value of meta.name must follow the convention of DNS-1035 label.

To avoid operator repeatedly creating and destroying flink clusters due to invalid service name, So I suggest add validation for the meta.name of FlinkDeployment CR.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultValidator.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultValidator.java
@@ -109,7 +109,7 @@ public class DefaultValidator implements FlinkResourceValidator {
         if (!matcher.matches()) {
             return Optional.of(
                     String.format(
-                            "The FlinkDeployment meta.name: %s is a invalid value, and must start and end with an alphanumeric character (e.g. 'my-name',  or '123-abc'), and the length must be no more than 45 characters.",
+                            "The FlinkDeployment name: %s is invalid, must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123'), and the length must be no more than 45 characters.",
                             name));
         }
         return Optional.empty();
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultValidator.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultValidator.java
@@ -48,10 +48,13 @@ import javax.annotation.Nullable;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 /** Default validator implementation for {@link FlinkDeployment}. */
 public class DefaultValidator implements FlinkResourceValidator {
-
+    private static final Pattern DEPLOYMENT_NAME_PATTERN =
+            Pattern.compile("[a-z]([-a-z\\d]*[a-z\\d])?");
     private static final String[] FORBIDDEN_CONF_KEYS =
             new String[] {
                 KubernetesConfigOptions.NAMESPACE.key(), KubernetesConfigOptions.CLUSTER_ID.key()
@@ -77,6 +80,7 @@ public class DefaultValidator implements FlinkResourceValidator {
             effectiveConfig.putAll(spec.getFlinkConfiguration());
         }
         return firstPresent(
+                validateDeploymentName(deployment.getMetadata().getName()),
                 validateFlinkVersion(spec.getFlinkVersion()),
                 validateFlinkDeploymentConfig(effectiveConfig),
                 validateIngress(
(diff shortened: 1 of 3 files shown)


Grypse:
Please help me review this PR. @gyfora

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultValidator.java
```diff
@@ -100,6 +104,21 @@ private static Optional<String> firstPresent(Optional<String>... errOpts) {
         return Optional.empty();
     }
 
+    private Optional<String> validateDeploymentName(String name) {
+        if (name == null) {
+            return Optional.of("FlinkDeployment name must be define.");
```
gyfora, line 109: The null check is unnecessary, this will never be null on kubernetes

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultValidator.java
```diff
@@ -100,6 +104,21 @@ private static Optional<String> firstPresent(Optional<String>... errOpts) {
...
+        } else {
+            Matcher matcher = DEPLOYMENT_NAME_PATTERN.matcher(name);
+            if (!matcher.matches()) {
+                return Optional.of(
+                        String.format(
+                                "The FlinkDeployment meta.name: %s is a invalid value, and must start and end with an alphanumeric character (e.g. 'my-name',  or '123-abc', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?'",
```
gyfora, line 115: Minor correction:

```
The FlinkDeployment name: %s is invalid. It must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name', or 'abc-123')
```

gyfora (changes requested):
Added some minor comments.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultValidator.java
Grypse: Ok, I removed the null check.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultValidator.java
Grypse: Corrected!

gyfora: can you actually take the string that I wrote? I think you only removed the end but I improved the whole message.
meta.name -> name
a invalid value -> invalid.

Fix example: 123-abc does not match the regex, it is not valid

gyfora:
@Grypse we actually just hit another related issue today with long deployment names:
```
java.lang.IllegalArgumentException: kubernetes.cluster-id must be no more than 45 characters
```
While you are at it could you please add a validation for this too to your regex? This way we can avoid a new ticket/PR :)

Grypse:
> 

OK，No problem.
I will add length validation to the regex of the value of meta.name.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultValidator.java
Grypse: I change the message to this:
`The FlinkDeployment name: %s is invalid, must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123'), and the length must be no more than 45 characters.`
More detailed message have been added, including what to start and end with and length requirements.
What do u think?

gyfora: Yes please change the message to what you wrote (you did not change it to this :D)

gyfora: Can you please just copy-paste this into the message so we can avoid a few other iterations on this? 
```
The FlinkDeployment name: %s is invalid, must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name', or 'abc-123'), and the length must be no more than 45 characters.
```

Grypse: Changed.

Grypse:
@gyfora  Why the check of test_ci is error? I ran the command ` mvn verify -Dit.skip=false `  is succeed  on local.

PR #271, opened 2022-06-17
[FLINK-27820] Handle deployment errors on observe

gyfora:
This change allows the operator to recover from errors during and directly after Flink cluster submissions. One common cause for this would be a temporary unavailability of the kubernetes API which would prevent us from updating the status after deployment.

Before this change this would lead to a fatal error where the operator would try to submit the flink cluster again and again (even though it is already deployed)

To solve this case the following changes were introduced:

- LastReconciledSpec now contains the CR generation for the deployed spec. During upgrades this is recorded during the first SUSPEND step of the upgrade operation
- The CR generation is also added as an annotation to the Flink Cluster Deployment object
- The default ReonciliationStatus for new deployments is now UPGRADING (previously it was DEPLOYED)
- In the observer, if the reconciliation status is UPGRADING (new or upgrading deployments) we check whether the Deployment is there and if so we compare the generation annotation. If it matches to the target generation, we know it was a succesful upgrade so we upgrade the status

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
@@ -18,7 +18,6 @@
 package org.apache.flink.kubernetes.operator.controller;
 
 import org.apache.flink.api.common.JobStatus;
-import org.apache.flink.api.java.tuple.Tuple2;
 import org.apache.flink.kubernetes.operator.config.FlinkConfigManager;
 import org.apache.flink.kubernetes.operator.crd.FlinkDeployment;
 import org.apache.flink.kubernetes.operator.crd.status.FlinkDeploymentStatus;
@@ -53,7 +52,6 @@ import org.slf4j.LoggerFactory;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
-import java.util.concurrent.ConcurrentHashMap;
 
 /** Controller that runs the main reconcile loop for Flink deployments. */
 @ControllerConfiguration()
@@ -73,8 +71,6 @@ public class FlinkDeploymentController
     private final MetricManager<FlinkDeployment> metricManager;
     private final StatusRecorder<FlinkDeploymentStatus> statusRecorder;
     private final EventRecorder eventRecorder;
-    private final ConcurrentHashMap<Tuple2<String, String>, FlinkDeploymentStatus> statusCache =
-            new ConcurrentHashMap<>();
 
     public FlinkDeploymentController(
             FlinkConfigManager configManager,
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
@@ -34,7 +34,6 @@ import org.apache.flink.kubernetes.operator.utils.EventUtils;
 import org.apache.flink.kubernetes.operator.utils.StatusRecorder;
 import org.apache.flink.kubernetes.operator.validation.FlinkResourceValidator;
 
-import io.fabric8.kubernetes.client.KubernetesClient;
 import io.javaoperatorsdk.operator.api.reconciler.Cleaner;
 import io.javaoperatorsdk.operator.api.reconciler.Context;
 import io.javaoperatorsdk.operator.api.reconciler.ControllerConfiguration;
@@ -63,7 +62,6 @@ public class FlinkDeploymentController
     private static final Logger LOG = LoggerFactory.getLogger(FlinkDeploymentController.class);
 
     private final FlinkConfigManager configManager;
-    private final KubernetesClient kubernetesClient;
 
     private final Set<FlinkResourceValidator> validators;
     private final ReconcilerFactory reconcilerFactory;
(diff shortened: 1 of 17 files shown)


gyfora:
cc @Aitozi @morhidi @wangyang0918

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
```diff
@@ -73,8 +71,6 @@
     private final MetricManager<FlinkDeployment> metricManager;
     private final StatusRecorder<FlinkDeploymentStatus> statusRecorder;
     private final EventRecorder eventRecorder;
-    private final ConcurrentHashMap<Tuple2<String, String>, FlinkDeploymentStatus> statusCache =
```
morhidi, line 76: nit: we don't use the kubernetesClient here either

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/AbstractDeploymentObserver.java
```diff
@@ -252,6 +257,60 @@ private void onMissingDeployment(FlinkDeployment deployment) {
...
+    /**
+     * Checks a deployment that is currently in the UPGRADING state whether it was already deployed
+     * but we simply miss the status information. After comparing the target resource generation
+     * with the one from the possible deployment if they match we update the status to the already
+     * DEPLOYED state.
+     *
+     * @param flinkDep Flink resource to check.
+     * @param context Context for reconciliation.
+     */
+    private void checkIfAlreadyUpgraded(FlinkDeployment flinkDep, Context context) {
```
morhidi, line 269: How about calling this method as 'syncWithDeployment` or similar. How problematic it is to introduce a new DEPLOYING state?

gyfora: Well this check specifically checks if an UPGRADING deployment is already upgraded. Calling it sync would give the impression that it does something more.

Also it's not really problematic to introduce a new DEPLOYING state but it would add no value so I would prefer not do it.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
gyfora: good point, will remove that too

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
```diff
@@ -229,24 +227,32 @@ private static boolean upgradeStarted(
...
-            return objectMapper.treeToValue(objectNode, specClass);
+
+            return Tuple2.of(objectMapper.treeToValue(objectNode, specClass), internalMeta);
         } catch (JsonProcessingException e) {
             throw new RuntimeException("Could not deserialize spec, this indicates a bug...", e);
         }
     }
 
-    public static String writeSpecWithCurrentVersion(Object spec) {
+    public static String writeSpecWithMeta(
```
morhidi, line 249: How about introducing a wrapper object here instead of modifying the objectnode?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/AbstractDeploymentObserver.java
morhidi: The way I understand the method catches up with initial deployments too, and the UPGRADING state covers both status. Distinguishing two states DEPLOYING and UPGRADING makes sense to me, or calling the method `checkIfAlreadyDeployedOrUpgraded`. But maybe I don't clearly understand the logic here.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
gyfora: makes sense

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/AbstractDeploymentObserver.java
gyfora: If I could I would simply rename UPGRADING -> DEPLOYING. But since that's not an option in this case I would prefer to simply keep it called UPGRADING and use that state also for initial first deployments.

gyfora:
Fixed your comments @morhidi , please take a look :)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/AbstractDeploymentObserver.java
```diff
@@ -252,6 +257,56 @@ private void onMissingDeployment(FlinkDeployment deployment) {
...
+                    Map<String, String> annotations = deployment.getMetadata().getAnnotations();
+                    if (annotations == null) {
+                        return;
+                    }
+                    Long deployedGeneration =
+                            Optional.ofNullable(annotations.get(FlinkUtils.CR_GENERATION_LABEL))
+                                    .map(Long::valueOf)
+                                    .orElse(-1L);
+
+                    // For first deployments we get the generation from the metadata directly
```
morhidi, line 283: nit: after moving the logic to a utility method, this comment is seemingly redundant

gyfora: yes, forgot this one

morhidi:
> Fixed your comments @morhidi , please take a look :)
LGTM, thanks!

PR #273, opened 2022-06-22
[FLINK-28166] Configurable Automatic Retries on Error

morhidi:
JOSDK defaults to the following retry configuration (exponential backoff)
```
GenericRetry.defaultLimitedExponentialRetry()
    .setInitialInterval(2000)
    .setIntervalMultiplier(1.5D)
    .setMaxAttempts(5);
```
This PR makes it configurable via the following operator config options:
```
kubernetes.operator.retry.initial.interval: 5 s
kubernetes.operator.retry.interval.multiplier: 2
kubernetes.operator.retry.max.attempts: 10
```

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
@@ -194,13 +194,32 @@ public class KubernetesOperatorConfigOptions {
                     .stringType()
                     .defaultValue(Constants.WATCH_ALL_NAMESPACES)
                     .withDescription(
-                            "Comma separated list of namespaces the operator monitors for custom resources. Defaults to "
-                                    + Constants.WATCH_ALL_NAMESPACES);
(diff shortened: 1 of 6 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
```diff
@@ -203,4 +203,28 @@ public class KubernetesOperatorConfigOptions {
...
+                            "Interval multiplier of automatic reconcile retries on recoverable errors. Defaults to "
+                                    + 2.0);
+
+    public static final ConfigOption<Integer> OPERATOR_RETRY_MAX_ATTEMPTS =
+            ConfigOptions.key("kubernetes.operator.retry.max.attempts")
+                    .intType()
+                    .defaultValue(10)
+                    .withDescription(
+                            "Max attempts of automatic reconcile retries on recoverable errors. Defaults to "
+                                    + 10);
```
gyfora, line 229: You don't need to specify default values in the description itself. It is generated automatically in the HTML

gyfora (commented):
Added a minor comment, otherwise looks good :)

PR #280, opened 2022-06-25
[FLINK-28008] Can not get secondary resource from context after operator restart

morhidi:
- bumping JOSDK version v3.0.2 -> v3.0.3
- add primaryToSecondary mapper for session jobs
- removing the workaround for setting namespaces (not needed anymore from JOSDK v3.0.3+)
- Changing config dumps into a one-liner

### DIFF
--- a/flink-kubernetes-operator/src/main/resources/log4j2.properties
+++ b/flink-kubernetes-operator/src/main/resources/log4j2.properties
@@ -24,3 +24,6 @@ appender.console.name = ConsoleAppender
 appender.console.type = CONSOLE
 appender.console.layout.type = PatternLayout
 appender.console.layout.pattern = %style{%d}{yellow} %style{%-30c{1.}}{cyan} %highlight{[%-5level] [%X{resource.namespace}.%X{resource.name}] %msg%n%throwable}
+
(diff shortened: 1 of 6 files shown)


morhidi:
cc @Aitozi

flink-kubernetes-operator/src/main/resources/log4j2.properties
```diff
@@ -24,3 +24,6 @@ appender.console.name = ConsoleAppender
 appender.console.type = CONSOLE
 appender.console.layout.type = PatternLayout
 appender.console.layout.pattern = %style{%d}{yellow} %style{%-30c{1.}}{cyan} %highlight{[%-5level] [%X{resource.namespace}.%X{resource.name}] %msg%n%throwable}
+
+logger.conf.name = org.apache.flink.configuration.GlobalConfiguration
+logger.conf.level = WARN
```
gyfora, line 29: Why did you change this? Isn't it helpful to see what exact config is loaded by the operator?

gyfora: makes sense :)

gyfora (approved):
Looks good, will merge this after the config log fix

PR #281, opened 2022-06-27
[FLINK-26891] Record important deployment events

tweise:
Record events that may be of interest to users when looking at the deployment with describe. Generally, these will be conversions of existing log messages. I added couple examples for illustration, will add more like submit or trigger/complete savepoint if there is agreement on the approach.

image

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
@@ -30,7 +30,6 @@ import org.apache.flink.kubernetes.operator.reconciler.ReconciliationUtils;
 import org.apache.flink.kubernetes.operator.reconciler.deployment.ReconcilerFactory;
 import org.apache.flink.kubernetes.operator.utils.EventRecorder;
 import org.apache.flink.kubernetes.operator.utils.EventSourceUtils;
-import org.apache.flink.kubernetes.operator.utils.EventUtils;
 import org.apache.flink.kubernetes.operator.utils.StatusRecorder;
 import org.apache.flink.kubernetes.operator.validation.FlinkResourceValidator;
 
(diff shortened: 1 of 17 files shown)


morhidi:
Nice job, @tweise! I was wondering if we could/should make the Reason field part of the FlinkResourceListener API? If we do so a call control layer could leverage it and could build some business logic on top of it. Even if we don't make it part of the API yet an enum type with the possible reason values would enforce some naming convention and consistency. What do you think?

gyfora (commented):
In addition to what Mayas said, I think with the reconciler unification merged, we should try to move the Event generation logic to `AbstractFlinkReconciler` and `AbstractJobReconciler` as much as possible to get a conistent behavior across resource types.

tweise:
> Nice job, @tweise! I was wondering if we could/should make the Reason field part of the FlinkResourceListener API? If we do so a call control layer could leverage it and could build some business logic on top of it. Even if we don't make it part of the API yet an enum type with the possible reason values would enforce some naming convention and consistency. What do you think?

Great suggestion. We could make the reason codes an enum similar to component and type. How about we then also change the signature of `triggerEvent` to use the enum instead of string?

morhidi:
> > Nice job, @tweise! I was wondering if we could/should make the Reason field part of the FlinkResourceListener API? If we do so a call control layer could leverage it and could build some business logic on top of it. Even if we don't make it part of the API yet an enum type with the possible reason values would enforce some naming convention and consistency. What do you think?
> 
> Great suggestion. We could make the reason codes an enum similar to component and type. How about we then also change the signature of `triggerEvent` to use the enum instead of string?

Sounds good!

gyfora (commented):
The PR looks pretty good!

My only question whether we should trigger events before or after executing the actions using the Flink service.

For example now we trigger a suspended event before trying to cancel the job. So it might happen that the event is there but the job is not suspended. Of course we could have the reverse situation if we changed the order. Just wanted to discuss this explicitly.

What do you think @tweise ?

tweise:
> The PR looks pretty good!
> 
> My only question whether we should trigger events before or after executing the actions using the Flink service.
> 
> For example now we trigger a suspended event before trying to cancel the job. So it might happen that the event is there but the job is not suspended. Of course we could have the reverse situation if we changed the order. Just wanted to discuss this explicitly.
> 
> What do you think @tweise ?

It should be OK to have the event logged first. If the subsequent action fails, we have a trace that it was attempted and any subsequent error would yield anther event so that there should not be any confusion.

I pulled the reason codes into `EventRecorder`. (I did not add them to the listener as the listener does not actually reference them and it would also be a bit unusual to have them there from the operator code caller perspective.)

@gyfora @morhidi PTAL

PR #283, opened 2022-06-28
[FLINK-28228] Never skip generations when observing already upgraded deployment

gyfora:
This PR makes some major improvements to the spec upgrade and initial deployment flow.

Most importantly, everytime before deploying a new spec (initial deployment or during upgrade) we patch the status and record the spec being upgraded to in the `lastReconciledSpec` with `UPGRADING` ReconciliationState.

This allow us to always detect when an UPGRADE was successful even if the subsequent status upgrade failed for any reason.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/AbstractDeploymentObserver.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/AbstractDeploymentObserver.java
@@ -263,7 +263,7 @@ public abstract class AbstractDeploymentObserver implements Observer<FlinkDeploy
      */
     private void checkIfAlreadyUpgraded(FlinkDeployment flinkDep, Context context) {
         var status = flinkDep.getStatus();
-        if (status.getReconciliationStatus().getLastReconciledSpec() == null) {
+        if (status.getReconciliationStatus().isFirstDeployment()) {
             return;
         }
         Optional<Deployment> depOpt = context.getSecondaryResource(Deployment.class);
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/AbstractDeploymentObserver.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/AbstractDeploymentObserver.java
@@ -278,21 +278,35 @@ public abstract class AbstractDeploymentObserver implements Observer<FlinkDeploy
                     Long upgradeTargetGeneration =
                             ReconciliationUtils.getUpgradeTargetGeneration(flinkDep);
 
+                    Long currentSpecGeneration = flinkDep.getMetadata().getGeneration();
+
                     if (deployedGeneration.equals(upgradeTargetGeneration)) {
-                        logger.info(
-                                "Last reconciled generation is already deployed, setting reconciliation status to "
-                                        + ReconciliationState.DEPLOYED);
+                        logger.info("Pending upgrade is already deployed, updating status.");
 
                         var firstDeploy =
                                 status.getReconciliationStatus().getLastReconciledSpec() == null;
-                        var conf =
-                                firstDeploy
-                                        ? configManager.getDeployConfig(
-                                                flinkDep.getMetadata(), flinkDep.getSpec())
-                                        : configManager.getObserveConfig(flinkDep);
+                        if (firstDeploy) {
+                            ReconciliationUtils.updateForSpecReconciliationSuccess(
+                                    flinkDep,
+                                    JobState.RUNNING,
+                                    configManager.getDeployConfig(
+                                            flinkDep.getMetadata(), flinkDep.getSpec()));
+                        } else {
+                            ReconciliationUtils.updateStatusForAlreadyUpgraded(flinkDep);
+                        }
+
+                        status.getJobStatus()
+                                .setState(org.apache.flink.api.common.JobStatus.RECONCILING.name());
+                        status.setJobManagerDeploymentStatus(JobManagerDeploymentStatus.DEPLOYING);
+                    } else if (deployedGeneration.equals(currentSpecGeneration)) {
+                        logger.info(
+                                "Deployment is already upgraded to latest spec, updating status.");
 
                         ReconciliationUtils.updateForSpecReconciliationSuccess(
-                                flinkDep, JobState.RUNNING, conf);
+                                flinkDep,
+                                JobState.RUNNING,
+                                configManager.getDeployConfig(
+                                        flinkDep.getMetadata(), flinkDep.getSpec()));
                         status.getJobStatus()
                                 .setState(org.apache.flink.api.common.JobStatus.RECONCILING.name());
                         status.setJobManagerDeploymentStatus(JobManagerDeploymentStatus.DEPLOYING);
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/AbstractDeploymentObserver.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/AbstractDeploymentObserver.java
@@ -262,8 +262,11 @@ public abstract class AbstractDeploymentObserver implements Observer<FlinkDeploy
      * @param context Context for reconciliation.
      */
     private void checkIfAlreadyUpgraded(FlinkDeployment flinkDep, Context context) {
-        Optional<Deployment> depOpt = context.getSecondaryResource(Deployment.class);
         var status = flinkDep.getStatus();
+        if (status.getReconciliationStatus().getLastReconciledSpec() == null) {
+            return;
+        }
+        Optional<Deployment> depOpt = context.getSecondaryResource(Deployment.class);
         depOpt.ifPresent(
                 deployment -> {
                     Map<String, String> annotations = deployment.getMetadata().getAnnotations();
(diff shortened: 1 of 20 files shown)


gyfora:
cc @Aitozi

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/AbstractDeploymentObserver.java
```diff
@@ -279,21 +279,35 @@ private void checkIfAlreadyUpgraded(FlinkDeployment flinkDep, Context context) {
...
+                                    configManager.getDeployConfig(
+                                            flinkDep.getMetadata(), flinkDep.getSpec()));
+                        } else {
+                            ReconciliationUtils.updateStatusForAlreadyUpgraded(flinkDep);
+                        }
+
+                        status.getJobStatus()
+                                .setState(org.apache.flink.api.common.JobStatus.RECONCILING.name());
+                        status.setJobManagerDeploymentStatus(JobManagerDeploymentStatus.DEPLOYING);
+                    } else if (deployedGeneration.equals(currentSpecGeneration)) {
```
Aitozi, line 302: I think we have a problem here, we can only do the thing when the deployedGeneration equal to the `currentSpecGeneration` or the `upgradeTargetGeneration`. If they are not equal, then we do not know how to bring the job do the desired version.

Aitozi: What about also record the updateTargetSpec before execute the upgrade or submit. So that, we can always have the right base to compare. And it will not changed by user. In the observer, we compare it with the deployedGeneration. WDYT?

gyfora: this only affects the first deployment scenario so it’s not a critical optimization but I can see how complicated it would be to add

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
```diff
@@ -352,22 +382,33 @@ ErrorStatusUpdateControl<R> toErrorStatusUpdateControl(
...
                         .deserializeLastReconciledSpecWithMeta();
 
         if (lastSpecWithMeta == null || lastSpecWithMeta.f1 == null) {
             // For first deployments and when migrating from before this feature simply return
             // current generation
-            return deployment.getMetadata().getGeneration();
+            return resource.getMetadata().getGeneration();
         }
 
         return lastSpecWithMeta.f1.get("metadata").get("generation").asLong(-1L);
```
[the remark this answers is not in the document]
gyfora: Let’s do this next time when we add new information there. I will add an improvement ticket

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/AbstractDeploymentObserver.java
gyfora: I will check this today

gyfora:
@Aitozi I reworked this PR so that it covers all the discussed corner cases and also Session clusters (which was previously uncovered).

The main change I introduced is to always put the upgrade target spec into the lastReconciledSpec before submitting a deployment. 

This turned into a bigger change but actually makes the whole flow more robust.

morhidi:
I'm testing the current logic locally and hit:
```
flink-kubernetes-operator 2022-06-30 11:11:28,759 o.a.f.k.o.o.d.ApplicationObserver [WARN ][default/basic-checkpoint-ha-example] Running deployment generation -1 doesn't match upgrade ta │
│ rget generation 4.
```
I deleted the JM deployment manually, when the deployment came up again it doesn't contain the generation from the CR.

gyfora:
@morhidi that is expected and should not cause any problems

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/AbstractDeploymentObserver.java
```diff
@@ -262,8 +262,11 @@ private void onMissingDeployment(FlinkDeployment deployment) {
      * @param context Context for reconciliation.
      */
     private void checkIfAlreadyUpgraded(FlinkDeployment flinkDep, Context context) {
-        Optional<Deployment> depOpt = context.getSecondaryResource(Deployment.class);
         var status = flinkDep.getStatus();
+        if (status.getReconciliationStatus().getLastReconciledSpec() == null) {
```
morhidi, line 266: Would could probably introduce an isFirstDeployment() getter to the DepoymentStatus. I found this check in the code in multiple places.

Aitozi:
I'm still checking the whole logic, I will give you some feedback tomorrow morning @gyfora

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
```diff
@@ -99,15 +103,20 @@ public final void reconcile(CR cr, Context ctx) throws Exception {
...
             deploy(
                     cr,
                     spec,
                     status,
                     deployConfig,
                     Optional.ofNullable(spec.getJob()).map(JobSpec::getInitialSavepointPath),
                     false);
-            ReconciliationUtils.updateForSpecReconciliationSuccess(
-                    cr, JobState.RUNNING, deployConfig);
+            ReconciliationUtils.updateStatusForDeployedSpec(cr, deployConfig);
```
Aitozi, line 119: If we throw exception in the deploy, the resource may not be truly deployed. So, I think the `firstDeployment` condition is not full. IMO, It should includes :

- lastReconciledSpec == null
- Or state in the `ReconciliationState.UPGRADING` and `JobManagerDeploymentStatus.Missing`

Aitozi: I'm second to @morhidi  to make the `isFirstDeploy` as a dedicated methods.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractJobReconciler.java
```diff
@@ -106,19 +107,27 @@ protected void reconcileSpecChange(
             // We must record the upgrade mode used to the status later
             currentDeploySpec.getJob().setUpgradeMode(availableUpgradeMode.get());
             cancelJob(resource, availableUpgradeMode.get(), observeConfig);
-            newState = JobState.SUSPENDED;
+            if (desiredJobState == JobState.RUNNING) {
+                ReconciliationUtils.updateStatusBeforeDeploymentAttempt(resource, deployConfig);
```
Aitozi, line 111: Is this redundant, we do not do upgrade in this branch. This will be done in the `if (currentJobState == JobState.SUSPENDED && desiredJobState == JobState.RUNNING)` I think.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/SessionReconciler.java
```diff
@@ -133,7 +135,16 @@ protected void rollback(FlinkDeployment deployment, Context ctx, Configuration o
         FlinkDeploymentSpec rollbackSpec = reconciliationStatus.deserializeLastStableSpec();
         Configuration rollbackConfig =
                 configManager.getDeployConfig(deployment.getMetadata(), rollbackSpec);
-        upgradeSessionCluster(deployment, rollbackSpec, rollbackConfig);
+
+        deleteSessionCluster(deployment, observeConfig);
+        deploy(
```
Aitozi, line 140: Why in the rollback we do not have to record the target upgrade status ? I think the rollback should deploy the stable spec with the corresponding generation, WDYT

Aitozi: I also found that the general flow of the deploy become the 
- record the target upgrade/firstDeploy in status
- deploy
- update the deployed spec

So I think we could include it in the `deploy` directly

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/AbstractDeploymentObserver.java
Aitozi: +1, and I think the condition should be improve Since the lastReconciliationStatus will be materialized before the `deploy` so there is the case it was not really finished. We should take this into account.

Aitozi (commented):
I think the PR is in good shape now, left some thoughts about some cases in my mind. Please  take a look @gyfora

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
gyfora: @Aitozi it is only safe to directly deploy if the lastReconciledSpec was null, otherwise it might be suspended due to a spec change etc. I think the current logic is correct.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/SessionReconciler.java
gyfora: The rollback logic doesn't assume anything about the generation etc so there is not really a need to record it.

Also we cannot put this in `deploy` directly because deploy is used for spec update, JM recovery, and rollback. And the target generation will only be useful for spec change.

gyfora:
@morhidi @Aitozi I tried to address the comments. I only added the isFirstDeployment check utility method.

While I partially agreed with some other comments regarding conditions, generation handling for rollbacks etc. but  I found no easy way to incoporate those suggestions in a way that would make the code structure or the functionality better.

Even before the comments I have thought about those cases but I think the current logic would work reliably.

morhidi:
> @morhidi @Aitozi I tried to address the comments. I only added the isFirstDeployment check utility method.
> 
> While I partially agreed with some other comments regarding conditions, generation handling for rollbacks etc. but I found no easy way to incoporate those suggestions in a way that would make the code structure or the functionality better.
> 
> Even before the comments I have thought about those cases but I think the current logic would work reliably.

Thanks @gyfora appreciate your efforts. Although the generation based deployment validation is a good step towards making sure we have running exactly what we intended to I still have a feeling we need to simplify the logic here. Just thinking out loud: How about generating a hash based on all the spec fields that should trigger an upgrade and we mark the deployment with it. If we find the hash is different we simply do an upgrade?

gyfora:
@morhidi i think that’s more complicated that simplified :) 

what would be the added benefit if the current approach already guarantees consistency?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
Aitozi: I think the problem around the upgrade can be fixed by this PR. The only weak part is about the firstDeploy. Because the upgrade failure will be taken over by the check during the next reconciler for 

```
 boolean specChanged =
                reconciliationStatus.getState() == ReconciliationState.UPGRADING
                        || !currentDeploySpec.equals(lastReconciledSpec);
```

But the firstDeploy do not have the same thing. What about introduce the `INITIALIZING` for the firstDeploy to distinguish between the upgrade ?

gyfora: Because this case is covered I would not add a new ReconciliationState unless necessary

Aitozi: Then, we can know whether it is an firstDeploy by `lastReconciledSpec == null ||  reconciliationStatus.getState() == ReconciliationState.INITIALIZING`

gyfora: @Aitozi I just manually verified this again by throwing errors directly before / after first deploy. And it both works.

Aitozi (approved):
@gyfora Thanks for your explanation. I have no other concerns. Very nice work, +1

PR #284, opened 2022-06-28
[tooling] Use 7 letter hash in release scripts

gyfora:
Newer git clients seem to return 8 letter hashes sometimes by default. This fix ensures that our release scripts always uses only the first 7 letters.


tisonkun:
@gyfora out of curiosity, it's a big deal whether we hold 7 or 8 letter hash in release script? Or just for reproducing stably?

mbalassi (approved):
LGTM

PR #285, opened 2022-06-29
[FLINK-26955] Add CRD compatibility check to maven build

gyfora:
Add automated CRD compatibility check logic to the maven build comparing against the 1.0.0 release.

Manually verified + unit tests added for the cases covered by the checker

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/CrdCompatibilityChecker.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/CrdCompatibilityChecker.java
@@ -29,7 +29,22 @@ import java.net.URL;
 import java.util.ArrayList;
 import java.util.List;
 
-/** Utility for checking backward CRD compatibility. */
+/**
+ * Utility for checking backward CRD compatibility.
+ *
+ * <p>The aim is to provide a very strict compatibility check logic initially that can be improved
+ * over time with compatible type changes if necessary by CRD updates
+ *
+ * <p>It performs the following checks:
+ *
+ * <ul>
+ *   <li>No property removed from any object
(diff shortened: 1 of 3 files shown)


gyfora:
cc @bgeng777 @morhidi

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/CrdCompatibilityChecker.java
```diff
@@ -0,0 +1,160 @@
...
+import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.net.URL;
+import java.util.ArrayList;
+import java.util.List;
+
+/** Utility for checking backward CRD compatibility. */
```
tweise, line 32: Maybe add a summary here on how compatibility is checked?

morhidi:
Thanks @gyfora for the PR. Having such a tool overall is a great idea, it'll force us to think twice before introducing any changes to the CRD. I second @tweise's suggestion to agree formally on a few bullets that we add to the CRD checker, e.g.: https://www.faun.dev/c/stories/dineshparvathaneni/kubernetes-crd-versioning-for-operator-developers/

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/CrdCompatibilityChecker.java
gyfora: Done

morhidi:
+1 LGTM, as we understand more about CRD version compatibility we should add/modify the rules here. Nice job!

PR #288, opened 2022-06-30
[FLINK-27975] Remove unnecessary RBAC rules from operator

jeesmon:
Removed un-used apiGroup and resources from required RBAC

Signed-off-by: Jeesmon Jacob <jjeesmon@vmware.com>

### DIFF
--- a/helm/flink-kubernetes-operator/templates/rbac.yaml
+++ b/helm/flink-kubernetes-operator/templates/rbac.yaml
@@ -22,24 +22,23 @@ RBAC rules used to create the operator (cluster)role based on the scope
 {{- define "flink-operator.rbacRules" }}
 rules:
   - apiGroups:
-      - flink-operator
-    resources:
-      - "*"
-    verbs:
-      - "*"
-  - apiGroups:
       - ""
     resources:
       - pods
       - services
-      - endpoints
-      - persistentvolumeclaims
       - events
       - configmaps
       - secrets
-      - nodes
     verbs:
       - "*"
+{{- if .Values.rbac.nodesRule.create }}
+  - apiGroups:
+    - ""
+    resources:
+      - nodes
+    verbs:
+      - list
+{{- end }}
   - apiGroups:
       - apps
     resources:
--- a/docs/content/docs/operations/helm.md
+++ b/docs/content/docs/operations/helm.md
@@ -64,6 +64,7 @@ The configurable parameters of the Helm chart and which default values as detail
 | image.pullPolicy | The image pull policy of flink-kubernetes-operator. | IfNotPresent |
 | image.tag | The image tag of flink-kubernetes-operator. | latest |
 | rbac.create | Whether to enable RBAC to create for said namespaces. | true |
+| rbac.nodesRule.create | Whether to add RBAC rule to list nodes which is needed for rest-service exposed as NodePort type. | false |
 | operatorPod.annotations | Custom annotations to be added to the operator pod (but not the deployment). | |
 | operatorPod.labels | Custom labels to be added to the operator pod (but not the deployment). | |
 | operatorServiceAccount.create | Whether to enable operator service account to create for flink-kubernetes-operator. | true |
--- a/helm/flink-kubernetes-operator/values.yaml
+++ b/helm/flink-kubernetes-operator/values.yaml
@@ -29,6 +29,11 @@ image:
 
 rbac:
   create: true
+  # kubernetes.rest-service.exposed.type: NodePort requires
+  # list permission for nodes at the cluster scope.
+  # Set create to true if you are using NodePort type.
+  nodesRule:
+    create: false
 
 operatorPod:
   annotations: {}


jeesmon:
@mbalassi @gyfora Could you please review?

gyfora:
Could you please share how you came to the conclusion of removing these specific ones?
I am just trying to understand how to verify this properly :)

jeesmon:
I think this is another place to look for the kube resources that flink is using: https://github.com/apache/flink/tree/master/flink-kubernetes/src/main/java/org/apache/flink/kubernetes/kubeclient/resources

 I see management of only the following resources in native integration:
 
image

Out of which only Service/Pod/ConfigMap are the only API resources.

helm/flink-kubernetes-operator/templates/rbac.yaml
```diff
@@ -21,23 +21,14 @@ RBAC rules used to create the operator (cluster)role based on the scope
...
-    resources:
-      - "*"
-    verbs:
-      - "*"
   - apiGroups:
       - ""
     resources:
       - pods
       - services
-      - endpoints
```
morhidi, line 35: Not sure about the endpoints, we create quite a bunch:
```
(minikube:default)  ➜  flink-kubernetes-operator git:(FLINK-28228) k get endpoints
NAME                               ENDPOINTS                         AGE
basic-checkpoint-ha-example-rest   172.17.0.9:8081                   8s
basic-ingress                      172.17.0.7:6124,172.17.0.7:6123   16s
basic-ingress-rest                 172.17.0.7:8081                   16s
flink-operator-webhook-service     172.17.0.6:9443                   3m36s
```
Does the service permissions include this?

helm/flink-kubernetes-operator/templates/rbac.yaml
```diff
@@ -21,23 +21,14 @@ RBAC rules used to create the operator (cluster)role based on the scope
...
-      - persistentvolumeclaims
       - events
       - configmaps
       - secrets
-      - nodes
```
morhidi, line 40: I remember we needed it either when playing with Ingress or the `kubernetes.rest-service.exposed.type` fail to recall. Worth double-checking both before removing.

helm/flink-kubernetes-operator/templates/rbac.yaml
```diff
@@ -21,23 +21,14 @@ RBAC rules used to create the operator (cluster)role based on the scope
...
-      - "*"
-    verbs:
-      - "*"
   - apiGroups:
       - ""
     resources:
       - pods
       - services
-      - endpoints
-      - persistentvolumeclaims
```
morhidi, line 36: The helm chart functionality can be extended with kustomize templates, and PVCs could be a useful to mount plugins and what not for the Operator. Although I'm generally in favour of least privilege principle, I would probably won't remove this one. WDYT?

helm/flink-kubernetes-operator/templates/rbac.yaml
```diff
@@ -21,23 +21,14 @@ RBAC rules used to create the operator (cluster)role based on the scope
 */}}
 {{- define "flink-operator.rbacRules" }}
 rules:
-  - apiGroups:
```
morhidi, line 24: +1 This block seemingly here for no good, can be removed.

helm/flink-kubernetes-operator/templates/rbac.yaml
jeesmon: @morhidi Even if we are adding PVC to operator, operator is not creating/managing PVC. PVC is created outside of operator as I understood. So mounting a persistentVolumeClaim to operator deployment doesn't require this RBAC. `e2e-tests/data/sessionjob-cr.yaml` is a good example for this and that test is passed without the RBAC. Please correct me if I'm wrong.

helm/flink-kubernetes-operator/templates/rbac.yaml
jeesmon: I will double check that config and get back.

helm/flink-kubernetes-operator/templates/rbac.yaml
jeesmon: As far as I can see, `services` RBAC covers it. Operator or flink native integration doesn't seem to be creating or managing endpoints directly. But I will double check.

helm/flink-kubernetes-operator/templates/rbac.yaml
morhidi: Good point, can be removed then

helm/flink-kubernetes-operator/templates/rbac.yaml
morhidi: Discussed this offline with @mbalassi we don't manage endpoint directly, thus this one can be removed too

jeesmon: Great, thanks

helm/flink-kubernetes-operator/templates/rbac.yaml
jeesmon: @morhidi You are right, `kubernetes.rest-service.exposed.type: NodePort` requires `list` permission on `nodes` at the cluster scope.

```
2022-07-01 13:30:20,742 i.j.o.p.e.ReconciliationDispatcher [ERROR][flink-27975/basic-example] Error during event processing ExecutionScope{ resource id: ResourceID{name='basic-example', namespace='flink-27975'}, version: 16843355} failed.
org.apache.flink.kubernetes.operator.exception.ReconciliationException: org.apache.flink.kubernetes.shaded.io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: GET at: https://10.96.0.1/api/v1/nodes. Message: Forbidden!Configured service account doesn't have access. Service account may have been revoked. nodes is forbidden: User "system:serviceaccount:flink-27975:flink-operator" cannot list resource "nodes" in API group "" at the cluster scope.
```

This brings another point that `kubernetes.rest-service.exposed.type: NodePort` will throw error for namespace scoped operator as it requires `list` permission at the cluster scope for `nodes`.

I will update the PR with `nodes` rbac and some comments. Thanks for your pointers.

@gyfora FYI

jeesmon:
> Thanks @jeesmon, looks good. I agree with removing endpoints and pvcs - hope you can confirm that we can also remove nodes. If it is only sometimes needed we could make a config value for it in the helm chart.

@mbalassi Yes, working on an update for `nodes` rules. Thanks.

jeesmon:
@morhidi @mbalassi I have updated the PR to conditionally create `nodes` rbac. Please check and let me know your feedback. Thanks.

jeesmon:
@morhidi Could you please trigger the workflow to make sure no CI issues?

mbalassi:
Great job, @jeesmon. Please update the Helm chart [docs](https://github.com/apache/flink-kubernetes-operator/blob/main/docs/content/docs/operations/helm.md) to reflect your change (unfortunately we do not generate this) and then I will merge your PR.

jeesmon:
@mbalassi Doc updated

mbalassi (approved):
Great work, merging.

mbalassi:
Now that all checks have passed I am merging.

PR #289, opened 2022-07-01
chore: move up imagePullSecrets field in the Helm values

haoxins:
When I added the`imagePullSecrets`, I found it now worked as expected.
The reason is that there exists an `imagePullSecrets` field at the bottom of the file that is not easy to notice.


morhidi:
LGTM

mbalassi (approved):
LGTM, merging.

PR #292, opened 2022-07-05
[FLINK-28385] Change the validator to return an error if the Jar URI is an empty string

haoxins:
This error may happen if I use Golang to submit the job with some bugs.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultValidator.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultValidator.java
@@ -178,7 +178,7 @@ public class DefaultValidator implements FlinkResourceValidator {
             return Optional.empty();
         }
 
-        if (job.getJarURI() == null) {
+        if (StringUtils.isNullOrWhitespaceOnly(job.getJarURI())) {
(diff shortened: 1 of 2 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultValidator.java
```diff
@@ -178,7 +178,7 @@ private Optional<String> validateJobSpec(
             return Optional.empty();
         }
 
-        if (job.getJarURI() == null) {
+        if (job.getJarURI() == null || job.getJarURI().equals("")) {
```
gyfora, line 181: This condition in this `if` check could be replaced by `StringUtils.isEmpty(job.getJarURI())`

gyfora (changes requested):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultValidator.java
haoxins: `isEmpty` not found, using `isNullOrWhitespaceOnly` instead

mbalassi:
Thanks @haoxins. Could you please add a simple test for this?

PR #293, opened 2022-07-05
[FLINK-28348] Add configurable flag to disable last-state fallback for savepoint upgrade

morhidi:
Introducing `kubernetes.operator.job.upgrade.last-state-fallback.enabled` variable to enable configurable fall-back to last-state upgrade when the job is not running but HA is available.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
@@ -230,4 +230,11 @@ public class KubernetesOperatorConfigOptions {
                     .defaultValue(10)
                     .withDescription(
                             "Max attempts of automatic reconcile retries on recoverable errors.");
+
+    public static final ConfigOption<Boolean> OPERATOR_JOB_UPGRADE_LAST_STATE_FALLBACK_ENABLED =
+            ConfigOptions.key("kubernetes.operator.job.upgrade.last-state-fallback.enabled")
+                    .booleanType()
+                    .defaultValue(true)
+                    .withDescription(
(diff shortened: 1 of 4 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
```diff
@@ -230,4 +230,10 @@ public class KubernetesOperatorConfigOptions {
                     .defaultValue(10)
                     .withDescription(
                             "Max attempts of automatic reconcile retries on recoverable errors.");
+
+    public static final ConfigOption<Boolean> OPERATOR_JOB_UPGRADE_LAST_STATE_FALLBACK_ENABLED =
+            ConfigOptions.key("kubernetes.operator.job.upgrade.last-state-fallback.enabled")
+                    .booleanType()
+                    .defaultValue(true)
+                    .withDescription("Enables last-state fallback for savepoint upgrade mode.");
```
gyfora, line 238: We should add a bit more detail here, when/why this would happen

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkUtils.java
```diff
@@ -281,6 +282,11 @@ public static boolean isKubernetesHAActivated(Configuration configuration) {
                 .equalsIgnoreCase(KubernetesHaServicesFactory.class.getCanonicalName());
     }
 
+    public static boolean isLastStateFallbackEnabled(Configuration configuration) {
+        return configuration.getBoolean(
+                KubernetesOperatorConfigOptions.OPERATOR_JOB_UPGRADE_LAST_STATE_FALLBACK_ENABLED);
+    }
+
```
gyfora, line 289: Do we need a utility method for this? The config.get is pretty self explanatory

gyfora (commented):
Added some minor comments :)

morhidi:
@gyfora addressed your comments, PTAL

PR #294, opened 2022-07-05
[FLINK-28350] Always use savepoint when switching between Flink versions

gyfora:
As HA metadata is not always compatible (for example Flink 1.14 <-> 1.15) we should always take a savepoint between version upgrades and not allow rollbacks.


gyfora:
cc @morhidi

morhidi:
The PR looks good @gyfora , cloud you please shed some light on why we don't allow rolling back on version changes?

gyfora:
@morhidi For rollbacks only HA metadata can be used (because if a savepoint can be taken then the job is running and stable so won't be rolled back). And between versions we cannot use HA metadata so rollbacks are not possible.

PR #295, opened 2022-07-05
[FLINK-27009] Add example for running SQL scripts using the operator

gyfora:
This PR introduces a complete example for running SQL Scripts directly using the operator. 

With this we provide users a straightforward way of running SQL scripts without relying on any new operator or upstream features. This should satisfy the general need.

### DIFF
--- a/examples/flink-sql-runner-example/README.md
+++ b/examples/flink-sql-runner-example/README.md
@@ -21,9 +21,12 @@ under the License.
 
 ## Overview
 
-This is an end-to-end example of running Flink SQL scripts directly using the Flink Kubernetes Operator.
+This is an end-to-end example of running Flink SQL scripts using the Flink Kubernetes Operator.
 
-SQL script submission might one day become a top level feature of the operator but due to current limitations some manual work is required to customise the logic for the given user environment.
+It is only intended to serve as a showcase of how Flink SQL can be executed on the operator and users are expected to extend the implementation and dependencies based on their production needs. 
+
+Currently, it is not planned to add direct API support for SQL submission to the Kubernetes operator due to the complexity of image and dependency management that is specific to each use-case.
+At the same time we are confident that using these examples as a starting point the operator would cover all user needs. If Apache Flink itself extends the SQL support for Application mode in the future, the operator will aim to support that.
 
 *What's in this example?*
 
--- a/examples/flink-sql-runner-example/README.md
+++ b/examples/flink-sql-runner-example/README.md
@@ -1,0 +1,77 @@
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+# Flink Kubernetes Operator SQL Example
+
+## Overview
+
+This is an end-to-end example of running Flink SQL scripts directly using the Flink Kubernetes Operator.
+
+SQL script submission might one day become a top level feature of the operator but due to current limitations some manual work is required to customise the logic for the given user environment.
+
+*What's in this example?*
+
+ 1. SQL Script runner Flink Java application
+ 2. DockerFile to build custom image with script runner + SQL scripts
+ 3. Example YAML for submitting scripts using the operator
+
+## How does it work?
+
+As Flink doesn't support submitting SQL scripts directly as jobs, we have created a simple Flink Java application that takes the user script and executes it using the `TableEnvironment#executeSql` method.
+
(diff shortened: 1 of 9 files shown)


examples/flink-sql-runner-example/README.md
```diff
@@ -0,0 +1,77 @@
...
+under the License.
+-->
+
+# Flink Kubernetes Operator SQL Example
+
+## Overview
+
+This is an end-to-end example of running Flink SQL scripts directly using the Flink Kubernetes Operator.
+
+SQL script submission might one day become a top level feature of the operator but due to current limitations some manual work is required to customise the logic for the given user environment.
```
tweise, line 26: Since we are not planning to include any sql runner logic into the operator baseline due to the dependency issues it would bring along, should we set the expectation that this is NOT going to be an operator feature? If Flink one day has a SQL support for application mode, then there should hopefully be nothing extra required in the operator.

tweise:
Thanks for contributing this. It's a great example how to customize the operator to support SQL but should not be confused with a future operator feature!

examples/flink-sql-runner-example/README.md
gyfora: Sounds reasonable, I will change the wording to not give the wrong impression

examples/flink-sql-runner-example/README.md
```diff
@@ -0,0 +1,77 @@
...
+kubectl apply -f sql-example.yaml
+```
+
+## Connectors and other extensions
+
+This example will only work with the very basic table and connector types out of the box, however enabling new ones is very easy.
+
+Simply find your [required connector](https://nightlies.apache.org/flink/flink-docs-master/docs/connectors/table/overview/) and add it as compile dependency to the `flink-sql-runner-example` project `pom.xml`. This will ensure that is packaged into the sql-runner fatjar and will be available for you on the cluster.
+
+Once you dive deeper you will quickly find that the SqlRunner implementation is very basic and might not cover your more advanced needs. Feel free to simply extend or customise the code as necessary for your requirements.
```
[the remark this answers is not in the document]
gyfora: sounds good

gyfora:
Cc @bgeng777

gyfora:
Addressed your comments regarding the readme @tweise

tweise (approved):

examples/flink-sql-runner-example/sql-example.yaml
```diff
@@ -0,0 +1,41 @@
...
+    resource:
+      memory: "2048m"
+      cpu: 1
+  taskManager:
+    resource:
+      memory: "2048m"
+      cpu: 1
+  job:
+    jarURI: local:///opt/flink/usrlib/sql-runner.jar
+    args: ["/opt/flink/usrlib/sql-scripts/simple.sql"]
```
SteNicholas, line 39: Could user directy put the mutiple SQL into the args? If yes, this could remove the 'Add your SQL script files under the `sql-scripts` directory' step.

gyfora: Good point, we could have an option to add the SQL directly here

SteNicholas:
@gyfora, could we add some examples to [flink-playgrounds](https://github.com/apache/flink-playgrounds)?

examples/flink-sql-runner-example/sql-example.yaml
gyfora: The problem is that Flink cannot really handle args with newline characters in it. So probably better to keep the script directly.

gyfora:
> @gyfora, could we add some examples to [flink-playgrounds](https://github.com/apache/flink-playgrounds)?

I guess you could add something there but probably not as part of this ticket :) What would you add?

SteNicholas:
> > @gyfora, could we add some examples to [flink-playgrounds](https://github.com/apache/flink-playgrounds)?
> 
> I guess you could add something there but probably not as part of this ticket :) What would you add?

End to end example to play the Flink Kubernetes Operator and the playground examples run with the Flink Kubernetes Operator.

examples/flink-sql-runner-example/sql-example.yaml
SteNicholas: For example: `args: ["CREATE TABLE blackhole_table (name STRING, age INT) WITH ('connector' = 'blackhole')", "INSERT INTO blackhole_table VALUES ('fred flintstone', 35), ('barney rubble', 32)"]`.

examples/flink-sql-runner-example/pom.xml
```diff
@@ -0,0 +1,134 @@
...
+			<scope>provided</scope>
+		</dependency>
+
+		<!-- Add connector dependencies here. They must be in the default scope (compile). -->
+
+		<!-- Example:
+
+		<dependency>
+			<groupId>org.apache.flink</groupId>
+			<artifactId>flink-connector-kafka</artifactId>
```
bgeng777, line 55: It may be better to use example with `flink-sql-connector-kafka` as in Table API, we usually use sql connectors instead of datastream connectors.

examples/flink-sql-runner-example/sql-example.yaml
bgeng777: I have tried Nicholas' suggestion in my previous POC. It can help users to reuse the sql runner images and for simple SQL statements, it is fine but for users writing complex sqls with multiple table options, it can be pretty easy to make mistakes without newline character. Maybe we can leave the work to users who want to customize the sql runner.

morhidi: `kubernetes.secrets` are auto-mounted to the pods, users can use this trick to mount their sql scripts to the running job automatically without building new images each time they improve their scripts.

morhidi:
Great work @gyfora LGTM!

examples/flink-sql-runner-example/pom.xml
gyfora: good point!

gyfora: Actually this is correct. The `flink-sql-connector-kafka` is just a fatjar of the kafka connector.

PR #297, opened 2022-07-06
[FLINK-28396] Updating config file with doc versions

doramarsal:
Updating config file with previous doc versions and updating version name


mbalassi (approved):
Thanks @doramarsal, merging.

PR #300, opened 2022-07-07
[FLINK-28438] Clear the previous error if session job is running or finished

haoxins:
Once an error occurs, the error filed will contain the message even if the job is running successfully at the end.
We should clear the error message when the session job is running or finished.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
@@ -157,7 +157,9 @@ public abstract class JobStatusObserver<CTX> {
             AbstractFlinkResource<?, ?> resource,
             JobStatusMessage clusterJobStatus,
             Configuration deployedConfig) {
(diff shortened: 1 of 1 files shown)


gyfora:
Please provide a clear description of the problem in the JIRA ticket and an overview of your solution in the PR description.

We cannot review it until then

haoxins:
> Please provide a clear description of the problem in the JIRA ticket and an overview of your solution in the PR description.
> 
> We cannot review it until then

updated

haoxins:
closing this because I don't have a quick path to reproduce it.

gyfora:
Sorry @haoxins I am bit confused.

Are you closing this because you think the ticket is not valid? or because you don't have time to write a test-case for it?

haoxins:
> Are you closing this because you think the ticket is not valid? or because you don't have time to write a test-case for it?

Not about the unit test 😂
I can't find a quick path to reproduce the error (The error message is `timeout`).
And because I'm using `v1` (not the latest version) on my cluster, so I cant make sure the error will occur with the latest code.

So, I closing this because I can't make sure the ticket is valid.

gyfora:
Ah okay :D 

Thank you for the clarification, I will keep an eye out if I ever see this error myself :)

PR #307, opened 2022-07-10
[FLINK-28187] Handle SessionJob errors on observe

Aitozi:

This PR is meant to use the same mechanism of the `AbstractDeploymentObserver`. It can recognize the lost job information now, based on the fixed jobId generator. 

## Brief change log

  - Use the `uid.hashcode` + `generation` to make up the `JobID`.
  - Detect the jobs in the session cluster with the same uid prefix and generation. 

## Verifying this change

- new tests to verify the change. 

## Documentation

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/sessionjob/SessionJobObserver.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/sessionjob/SessionJobObserver.java
@@ -33,7 +33,6 @@ import org.apache.flink.kubernetes.operator.reconciler.ReconciliationUtils;
 import org.apache.flink.kubernetes.operator.reconciler.sessionjob.SessionJobReconciler;
 import org.apache.flink.kubernetes.operator.service.FlinkService;
 import org.apache.flink.kubernetes.operator.utils.EventRecorder;
-import org.apache.flink.kubernetes.operator.utils.FlinkUtils;
 import org.apache.flink.kubernetes.operator.utils.SavepointUtils;
 import org.apache.flink.kubernetes.operator.utils.StatusRecorder;
 import org.apache.flink.runtime.client.JobStatusMessage;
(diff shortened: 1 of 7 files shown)


Aitozi:
cc @gyfora

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/sessionjob/SessionJobObserver.java
```diff
@@ -115,4 +131,62 @@ public void observe(FlinkSessionJob flinkSessionJob, Context context) {
...
+                        .getStatus()
+                        .getJobStatus()
+                        .setJobId(matchedJobs.get(0).toHexString());
+            } else {
+                LOG.warn(
+                        "Running job {}'s generation {} doesn't match upgrade target generation {}.",
+                        matchedJobID.toHexString(),
+                        deployedGeneration,
+                        upgradeTargetGeneration);
+            }
```
gyfora, line 189: I think it might make sense to throw an error at this point. Otherwise we get double submission because the new job will have a new jobid due to the different generation.

gyfora (changes requested):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/sessionjob/SessionJobObserver.java
Aitozi: I think we have a chance to get here:

1. The upgrade fails between the `UPGRADING` status recorded and deploy
2. The deployedGeneration will smaller than the targetGeneration

In this case, we have to wait for the next turn to deal with reconcileSpecChange. It will cancel the oldJobID in jobStatus and submit with new JobID.

Aitozi: For safe, I think we could throw an exception here if the matchedJobID do not equal the the id in the JobStatus too. WDTY ?

gyfora: Sorry I don’t completely understand.

when we are in UPGRADING state the old job is already suspended. We should never have upgrading state + still the old job running

Aitozi: You are right, Updated.

PR #311, opened 2022-07-11
[FLINK-28479] Add metrics for resource lifecycle state transitions

gyfora:

Introduce histogram metrics for tracking how long resource lifecycle state transitions take between the following states:

 - CREATED
 - SUSPENDED
 - UPGRADING
 - DEPLOYED
 - STABLE
 - ROLLING_BACK
 - ROLLED_BACK
 - FAILED

New metrics:

```
FlinkDeployment.Lifecycle.Transition.ResumeTimeSeconds: count=0, min=0, max=0, mean=NaN, ...
FlinkDeployment.Lifecycle.Transition.SuspendTimeSeconds: count=1, min=2, max=2, mean=2.0, ...
FlinkDeployment.Lifecycle.Transition.UpgradeTimeSeconds: count=1, min=33, max=33, mean=33.0, ...
FlinkDeployment.Lifecycle.Transition.StabilizationTimeSeconds: count=1, min=29, max=29, mean=29.0, ...
FlinkDeployment.Lifecycle.Transition.RollbackTimeSeconds: count=0, min=0, max=0, mean=NaN, ...
FlinkDeployment.Lifecycle.Transition.SubmissionTimeSeconds: count=1, min=1, max=1, mean=1.0, ...

FlinkDeployment.Lifecycle.State.STATE_NAME.Count: 0
```

## Brief change log

 - Introduce ResourceLifecycleState derived from the resource status
 - Add mechanism to track ResourceLifecycleState transitions
 - Create histogram metrics for select transitions
 - Add count metrics for each state
 - Add tests

## Verifying this change

New unit tests + manually verified on minikube

## Documentation

  - If yes, how is the feature documented? **[TODO]**


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/metrics/MetricManager.java
```diff
@@ -18,31 +18,37 @@
...
 
-import io.fabric8.kubernetes.client.CustomResource;
-
+import java.time.Clock;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
 /** Metric manager for Operator managed custom resources. */
-public class MetricManager<CR extends CustomResource<?, ?>> {
+public class MetricManager<CR extends AbstractFlinkResource<?, ?>> {
```
morhidi, line 31: Shall we call it `ResourceMetricManager`?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/metrics/lifecycle/LifecycleMetrics.java
```diff
@@ -0,0 +1,205 @@
...
+ */
+public class LifecycleMetrics<CR extends AbstractFlinkResource<?, ?>> {
+
+    public static final List<Transition> TRACKED_TRANSITIONS = getTrackedTransitions();
+
+    private final Map<Tuple2<String, String>, ResourceLifecycleMetricTracker> lifecycleTrackers =
+            new ConcurrentHashMap<>();
+    private final Set<String> namespaces = Collections.newSetFromMap(new ConcurrentHashMap<>());
+
+    private final int histogramWindowSize = 1000;
```
morhidi, line 62: Don't you think it worth introducing a global window size config for the histograms across the operator?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/metrics/MetricManager.java
gyfora: I think the name is fine for now.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/metrics/lifecycle/ResourceLifecycleMetricsTest.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    @Test
+    public void testLifecycleTracker() {
+        var histos = initHistos();
+        var lifecycleTracker =
+                new ResourceLifecycleMetricTracker(CREATED, Instant.ofEpochMilli(1000), histos);
+
+        long ts = 1000;
+        lifecycleTracker.onUpdate(UPGRADING, Instant.ofEpochMilli(ts += 1000));
```
morhidi, line 97: we could add a variable for `Instant.ofEpochMilli(ts += 1000)` here

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/metrics/lifecycle/LifecycleMetrics.java
gyfora: yes, that would be nice I agree

morhidi:
I love this PR @gyfora added some minor comments

gyfora:
> What is `FlinkDeployment.Lifecycle.State.STATE_NAME.Count: 0` ?

@morhidi `STATE_NAME` is a placeholder in this example. In practice it's gonna be one of CREATED, UPGRADING, DEPLOYED etc.

morhidi:
> > What is `FlinkDeployment.Lifecycle.State.STATE_NAME.Count: 0` ?
> 
> @morhidi `STATE_NAME` is a placeholder in this example. In practice it's gonna be one of CREATED, UPGRADING, DEPLOYED etc.



> > What is `FlinkDeployment.Lifecycle.State.STATE_NAME.Count: 0` ?
> 
> @morhidi `STATE_NAME` is a placeholder in this example. In practice it's gonna be one of CREATED, UPGRADING, DEPLOYED etc.

nit: you can update the description to FlinkDeployment.Lifecycle.State.<STATE_NAME>.Count = ... to make it obvious

gyfora:
cc @tweise @wangyang0918

tweise:
This is great and maybe also mention (docs!) that you are not only tracking transitions between adjacent states and how this can be used to track the upgrade latency directly.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/metrics/lifecycle/LifecycleMetrics.java
```diff
@@ -0,0 +1,217 @@
...
+                new Transition(ROLLING_BACK, ROLLED_BACK, true, TRANSITION_SUBMISSION));
+    }
+
+    /**
+     * Pojo for encapsulating state transitions and whether we should measure time from the
+     * beginning of from or since the last update.
+     */
+    @ToString
+    @RequiredArgsConstructor
+    protected static class Transition {
```
morhidi, line 211: Have you considered defining `Transition` as enum?

gyfora: There are 12 instances of Transition currently with not very well defined names. Also they are treated in the same way, so I feel making them an enum is not a real good fit here.

Do you have a concrete suggestion?

morhidi: We always measure the latency between two states, so it could be STABLE_STABLE, SUPENDED_STABLE, etc. The last param shows which metric they belong to, not really which transition.

morhidi: but I might be missing something

gyfora: Enums only make sense in cases where the different values have different meanings. (and that actually means you have logic built around them)

Transition objects are handled uniformly, you can add / remove new ones later it won't change anything.

I think enum is not a good fit here.

PR #315, opened 2022-07-12
[FLINK-28496] Label selector setting for operator

mbalassi:

Adds support for labelselector on Flink custom resources. This enables having multiple operators potentially watch the same namespace for blue/green deployment scenarios. When no selector is set the existing behavior of watching all resources of a namespace is reatined.

## Verifying this change

I have added a test for it verifying that we properly pass the labelselector configuration to JOSDK. Unfortunately the test fails when run together with a specific other test, have to fix this. It might make sense to add an e2e test to cover this too.

## Documentation

  - If yes, how is the feature documented? (via the configuration description)

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
@@ -234,6 +234,15 @@ public class KubernetesOperatorConfigOptions {
                             "Comma separated list of namespaces the operator monitors for custom resources.");
 
     @Documentation.Section(SECTION_SYSTEM)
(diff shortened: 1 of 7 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
```diff
@@ -233,6 +233,15 @@ public class KubernetesOperatorConfigOptions {
                     .withDescription(
                             "Comma separated list of namespaces the operator monitors for custom resources.");
 
+    @Documentation.Section(SECTION_SYSTEM)
+    public static final ConfigOption<String> OPERATOR_LABEL_SELECTOR =
```
SteNicholas, line 237: Please run `mvn package -Pgenerate-docs -pl flink-kubernetes-docs -nsu -DskipTests` to generate the config option in the configuration document.

PR #326, opened 2022-07-21
[FLINK-28616][doc] Make quickstart versions references dynamic based on config

gyfora:

Remove hardcoded versions from quickstart and set it based on site config.

## Verifying this change

Manually verified


morhidi:
nice! looks promising @gyfora

PR #338, opened 2022-08-04
[FLINK-28792] Upgrade to JOSDK 3.1.0

gyfora:

Upgrade josdk to 3.1.0

## Verifying this change

## Documentation

### DIFF
--- a/docs/layouts/shortcodes/generated/kubernetes_operator_config_configuration.html
+++ b/docs/layouts/shortcodes/generated/kubernetes_operator_config_configuration.html
@@ -112,7 +112,7 @@
         </tr>
         <tr>
             <td><h5>kubernetes.operator.reconcile.parallelism</h5></td>
-            <td style="word-wrap: break-word;">5</td>
+            <td style="word-wrap: break-word;">10</td>
             <td>Integer</td>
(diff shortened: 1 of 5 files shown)


gyfora:
cc @morhidi

docs/layouts/shortcodes/generated/kubernetes_operator_config_configuration.html
```diff
@@ -112,7 +112,7 @@
         </tr>
         <tr>
             <td><h5>kubernetes.operator.reconcile.parallelism</h5></td>
-            <td style="word-wrap: break-word;">5</td>
+            <td style="word-wrap: break-word;">10</td>
```
morhidi, line 115: why do we need to increase the parallelism?

gyfora: We use the default from the josdk we don’t set this value explicitly anywhere

morhidi: makes sense, didn't notice it

morhidi:
+1 for merging

PR #348, opened 2022-08-16
[docs] Adding helm-docs step in ci and README.md

kaskol10:

Auto-generate documentation from helm charts into markdown files. 

## Brief change log

- Added README.md done with helm-docs
- Added CI step to verify helm-docs changes are needed. 

## Documentation


gyfora:
Hi! Could you please explain why you think this is a good addition? In general we should try to keep the README small and link the docs instead.

kaskol10:
Hi! Thank you for your question! 

In the current project there is no [README from Helm charts](https://github.com/apache/flink-kubernetes-operator/tree/main/helm/flink-kubernetes-operator) and it's needed to deep into the values.yaml from the Helm Chart to check the different values or check the official documentation, for instance the table located [here](https://nightlies.apache.org/flink/flink-kubernetes-operator-docs-release-1.1/docs/operations/helm/#overriding-configuration-parameters-during-helm-install ) 

Using helm-docs tool would auto-generate documentation from helm charts into markdown easily and the CI step would avoid to push Helm changes without update the documentation. 

I think this change will increase the Helm project documentation with an easy step and it will allow to explore documentation near to the Helm Chart code. But I understand your point about link the docs

gyfora:
It would be great to integrate this into the regular docs but the problem I see is that this is not actually “documentation “ .

it only lists the possible config parameters but not what they do. 

Even if we don’t always document new parameters at least the current docs explain what they do :)

kaskol10:
The description could be added to the current values.yaml based on the [helm-docs documentation](https://github.com/norwoodj/helm-docs#helm-docs) and it would be pretty similar to the table description [here](https://nightlies.apache.org/flink/flink-kubernetes-operator-docs-release-1.1/docs/operations/helm/#overriding-configuration-parameters-during-helm-install). 

If you think it would be valuable to add a description for each parameter in the README to take this PR into account I can add comments to the [values.yaml](https://github.com/apache/flink-kubernetes-operator/blob/main/helm/flink-kubernetes-operator/values.yaml) to have the parameters description in the Helm Chart and in the README. 

Happy to collaborate if that makes sense to you

gyfora:
I think this makes sense but what about the following way:

1. Add the current parameter descriptions to the values yaml from the docs.
2. Remove current docs table
3. Generate docs table with the plugin
4. Add link to generated docs from main readme 

Also I was wondering is this plugin safe to use with the Apache GitHub credentials?

mbalassi:
@kaskol10 could you please address the above comments from @gyfora?

PR #351, opened 2022-08-22
[FLINK-28646] Handle scaling operation separately in reconciler/service

morhidi:

Enable reactive scaling for standalone deployments by detecting and handling spec changes that do not require full application upgrades.
The standalone integration for session clusters and application clusters allows for more efficient scaling operations when only the parallelism changes. We should distinguish this operation in the reconciler/service and implement this for standalone mode.

## Brief change log
- Added a new method `FlinkService.reactiveScale()` to handle reactive scaling for standalone deployments
- Added spec change detection that results a `DiffType`. Values can be:
  - `IGNORE` - the job is not upgraded
  - `SCALE` - if reactive mode is enable the job is rescaled, upgraded otherwise
  - `UPGRADE` - the job is upgraded
- Added `diff` package to the public operator API, classes marked with `@Experimental` 
## Verifying this change
- New unit tests where added to cover the functionality
- Verified manually by submitting `basic-reactive.yaml` and changing the `job.parallelism`
```
[INFO ] [default.basic-reactive-example] SCALE change(s) detected, starting reconciliation.
[INFO ] [default.basic-reactive-example] FlinkDeploymentSpec[job.parallelism=4] differs from FlinkDeploymentSpec[job.parallelism=2]
[INFO ] [default.basic-reactive-example] Scaling TM replicas: actual(1) -> desired(2)
[INFO ] [default.basic-reactive-example] Reactive scaling succeeded
```

## Documentation
  - If yes, how is the feature documented? Generated API doc

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
@@ -31,6 +31,7 @@ import org.apache.flink.kubernetes.operator.crd.status.JobManagerDeploymentStatu
 import org.apache.flink.kubernetes.operator.crd.status.ReconciliationState;
 import org.apache.flink.kubernetes.operator.reconciler.Reconciler;
 import org.apache.flink.kubernetes.operator.reconciler.ReconciliationUtils;
+import org.apache.flink.kubernetes.operator.reconciler.diff.DiffType;
 import org.apache.flink.kubernetes.operator.service.FlinkService;
 import org.apache.flink.kubernetes.operator.utils.EventRecorder;
 import org.apache.flink.kubernetes.operator.utils.FlinkUtils;
@@ -67,7 +68,8 @@ public abstract class AbstractFlinkResourceReconciler<
     protected final KubernetesClient kubernetesClient;
 
     public static final String MSG_SUSPENDED = "Suspending existing deployment.";
-    public static final String MSG_SPEC_CHANGED = "Detected spec change, starting reconciliation.";
+    public static final String MSG_SPEC_CHANGED =
+            "%s change(s) detected (%s), starting reconciliation.";
     public static final String MSG_ROLLBACK = "Rolling back failed deployment.";
     public static final String MSG_SUBMIT = "Starting deployment";
 
@@ -122,25 +124,33 @@ public abstract class AbstractFlinkResourceReconciler<
                 cr.getStatus().getReconciliationStatus().deserializeLastReconciledSpec();
         SPEC currentDeploySpec = cr.getSpec();
 
-        boolean specChanged =
-                reconciliationStatus.getState() == ReconciliationState.UPGRADING
-                        || !currentDeploySpec.equals(lastReconciledSpec);
-        var observeConfig = getObserveConfig(cr, ctx);
+        var specDiff = currentDeploySpec.diff(lastReconciledSpec);
+
         var flinkService = getFlinkService(cr, ctx);
+
+        boolean specChanged =
+                DiffType.IGNORE != specDiff.getType()
+                        || reconciliationStatus.getState() == ReconciliationState.UPGRADING;
+
+        var observeConfig = getObserveConfig(cr, ctx);
+
         if (specChanged) {
+
             if (checkNewSpecAlreadyDeployed(cr, deployConfig)) {
                 return;
             }
-            LOG.info(MSG_SPEC_CHANGED);
+
+            var specChangeMessage = String.format(MSG_SPEC_CHANGED, specDiff.getType(), specDiff);
+            LOG.info(specChangeMessage);
             if (reconciliationStatus.getState() != ReconciliationState.UPGRADING) {
                 eventRecorder.triggerEvent(
                         cr,
                         EventRecorder.Type.Normal,
                         EventRecorder.Reason.SpecChanged,
                         EventRecorder.Component.JobManagerDeployment,
-                        MSG_SPEC_CHANGED);
+                        specChangeMessage);
             }
-            reconcileSpecChange(cr, ctx, observeConfig, deployConfig);
+            reconcileSpecChange(cr, ctx, observeConfig, deployConfig, specDiff.getType());
         } else if (shouldRollBack(cr, observeConfig, flinkService)) {
             // Rollbacks are executed in two steps, we initiate it first then return
             if (initiateRollBack(status)) {
@@ -200,7 +210,11 @@ public abstract class AbstractFlinkResourceReconciler<
      * @throws Exception Error during spec upgrade.
      */
     protected abstract void reconcileSpecChange(
-            CR cr, Context<?> ctx, Configuration observeConfig, Configuration deployConfig)
+            CR cr,
+            Context<?> ctx,
+            Configuration observeConfig,
+            Configuration deployConfig,
+            DiffType diffType)
             throws Exception;
 
     /**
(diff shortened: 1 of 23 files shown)


morhidi:
cc @gyfora @tweise

docs/content/docs/custom-resource/reference.md
```diff
@@ -174,6 +174,79 @@ This page serves as a full reference for FlinkDeployment custom resource definit
...
+ https://github.com/apache/commons-lang/blob/master/src/main/java/org/apache/commons/lang3/builder/Diff.java
+
+| Parameter | Type | Docs |
+| ----------| ---- | ---- |
+| serialVersionUID | long |  |
+| fieldName | @lombok.NonNull java.lang.String |  |
+| left | T |  |
+| right | T |  |
+| type | @lombok.NonNull org.apache.flink.kubernetes.operator.crd.spec.diff.DiffType |  |
+
```
gyfora, line 192: These Diff related classes should not be part of the CRD reference. We should either move them to another package or change the generator logic.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/spec/diff/Diff.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+@Experimental
+@Value
+public class Diff<T> {
+    private static final long serialVersionUID = 1L;
+
+    @NonNull private final String fieldName;
+    private final T left;
+    private final T right;
+    @NonNull private final DiffType type;
```
gyfora, line 39: `private final` can be removed if `@Value` is used

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/spec/diff/DiffResult.java
```diff
@@ -0,0 +1,102 @@
...
+        return Collections.unmodifiableList(diffList);
+    }
+
+    public int getNumberOfDiffs() {
+        return diffList.size();
+    }
+
+    public DiffType getType() {
+        return type;
+    }
```
gyfora, line 70: We could use the `@Getter` for this

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/spec/diff/DiffResult.java
```diff
@@ -0,0 +1,102 @@
...
+        diffList.forEach(
+                diff -> {
+                    lhsBuilder.append(diff.getFieldName(), diff.getLeft());
+                    rhsBuilder.append(diff.getFieldName(), diff.getRight());
+                });
+
+        return String.format("%s differs from %s", lhsBuilder.build(), rhsBuilder.build());
+    }
+
+    private DiffType getSpechChangeType(List<Diff<?>> diffs) {
```
gyfora, line 92: should be static

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
```diff
@@ -93,4 +93,6 @@ void triggerSavepoint(
     PodList getJmPodList(FlinkDeployment deployment, Configuration conf);
 
     void waitForClusterShutdown(Configuration conf);
+
+    boolean reactiveScale(ObjectMeta meta, JobSpec jobSpec, Configuration conf);
```
gyfora, line 97: I think we should simply call this `scale` . This operation does not to be reactive, that's basically just one way to implement this for standalone.

gyfora: It could also return false by default to avoid having to implement it for non standalone clusters.

gyfora (changes requested):

morhidi:
I've addressed your comments @gyfora

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
```diff
@@ -122,16 +123,23 @@ public final void reconcile(CR cr, Context<?> ctx) throws Exception {
...
-        var flinkService = getFlinkService(cr, ctx);
+
         if (specChanged) {
+
             if (checkNewSpecAlreadyDeployed(cr, deployConfig)) {
                 return;
             }
-            LOG.info(MSG_SPEC_CHANGED);
+            LOG.info(String.format(MSG_SPEC_CHANGED, specDiff.getType()));
+            LOG.info(specDiff.toString());
```
gyfora, line 142: We should log this in a single line

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/reconciler/diff/SpecDiffTest.java
```diff
@@ -0,0 +1,171 @@
...
+        assertEquals(0, diff.getNumDiffs());
+
+        if (left instanceof FlinkDeploymentSpec) {
+            ((FlinkDeploymentSpec) right).setImage("flink:greatest");
+            assertEquals(DiffType.UPGRADE, left.diff(right).getType());
+            diff = left.diff(right);
+            assertEquals(DiffType.UPGRADE, diff.getType());
+            assertEquals(1, diff.getNumDiffs());
+        }
+
```
gyfora, line 99: It's a bit strange to use `ParameterizedTest` here because you actually have 3 different inner ifs for the 3 different params. I would personally prefer to have single test cases because the parameterization just breaks up the logic  (hides what is tested) and its harder to understand.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/reconciler/diff/SpecDiffTest.java
```diff
@@ -0,0 +1,171 @@
...
+        if (right.getJob() != null) {
+            right.getJob().setParallelism(100);
+            right.getJob().setUpgradeMode(UpgradeMode.LAST_STATE);
+            assertEquals(DiffType.UPGRADE, left.diff(right).getType());
+            diff = left.diff(right);
+            assertEquals(DiffType.UPGRADE, diff.getType());
+            assertEquals(3, diff.getNumDiffs());
+        }
```
gyfora, line 115: It would also be nice to include 1-2 more complex diff scenarios such us some inner array of the podTemplate (like env variables) have changed so that we get full confidence in the logic we copied

gyfora (changes requested):
Looks pretty good, added some minor comments for tests

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/reconciler/diff/SpecDiffTest.java
morhidi: Dropped the `ParameterizedTest` approach

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/reconciler/diff/SpecDiffTest.java
morhidi: Added some more advanced structures. Please note that we do a deep diff on Diffable objects only, the rest is a just plain equals.

gyfora (approved):
LGTM, I will give others (@tweise) some time to review before merging.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
morhidi: fixed

PR #356, opened 2022-08-28
[FLINK-29100] Relax upgrade checks to allow stateless restart when no stable spec is present

tweise:

Allow for upgrade of deployment with stateful upgrade mode that initially fails due to a non-recoverable error, just like it would be the case when the deployment ever reached stable spec.

## Brief change log

Check for presence of stable spec before executing logic that assumes existing state.

## Verifying this change

Added test for spec change before stable spec for all upgrade modes.

Tested manually by creating deployments with image pull error for SAVEPOINT and LAST_STATE upgrade modes. Verified that in either case spec update to valid image is reconciled and deployment becomes stable.

  - Dependencies (does it add or upgrade a dependency): (**no**)
  - The public API, i.e., is any changes to the `CustomResourceDescriptors`: (**no**)
  - Core observer or reconciler logic that is regularly executed: (**yes**)

## Documentation

  - Does this pull request introduce a new feature? (**no**)

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
@@ -27,6 +27,7 @@ import org.apache.flink.kubernetes.operator.crd.spec.FlinkDeploymentSpec;
 import org.apache.flink.kubernetes.operator.crd.spec.UpgradeMode;
 import org.apache.flink.kubernetes.operator.crd.status.FlinkDeploymentStatus;
 import org.apache.flink.kubernetes.operator.crd.status.JobManagerDeploymentStatus;
+import org.apache.flink.kubernetes.operator.crd.status.ReconciliationState;
 import org.apache.flink.kubernetes.operator.exception.DeploymentFailedException;
 import org.apache.flink.kubernetes.operator.reconciler.ReconciliationUtils;
 import org.apache.flink.kubernetes.operator.service.FlinkService;
@@ -98,12 +99,19 @@ public class ApplicationReconciler
                                 .OPERATOR_JOB_UPGRADE_LAST_STATE_FALLBACK_ENABLED)
                 && FlinkUtils.isKubernetesHAActivated(deployConfig)
                 && FlinkUtils.isKubernetesHAActivated(observeConfig)
-                && flinkService.isHaMetadataAvailable(deployConfig)
                 && !flinkVersionChanged(
                         ReconciliationUtils.getDeployedSpec(deployment), deployment.getSpec())) {
-            LOG.info(
-                    "Job is not running but HA metadata is available for last state restore, ready for upgrade");
-            return Optional.of(UpgradeMode.LAST_STATE);
+
+            if (!flinkService.isHaMetadataAvailable(deployConfig)) {
+                if (deployment.getStatus().getReconciliationStatus().getLastStableSpec() == null) {
+                    // initial deployment failure, reset to allow for spec change to proceed
+                    return resetOnMissingStableSpec(deployment, deployConfig);
+                }
+            } else {
+                LOG.info(
+                        "Job is not running but HA metadata is available for last state restore, ready for upgrade");
+                return Optional.of(UpgradeMode.LAST_STATE);
+            }
         }
 
         if (status.getJobManagerDeploymentStatus() == JobManagerDeploymentStatus.MISSING
@@ -118,6 +126,31 @@ public class ApplicationReconciler
         LOG.info(
                 "Job is not running yet and HA metadata is not available, waiting for upgradeable state");
         return Optional.empty();
+    }
+
+    private Optional<UpgradeMode> resetOnMissingStableSpec(
+            FlinkDeployment deployment, Configuration deployConfig) {
+        // initial deployment failure, reset to allow for spec change to proceed
+        flinkService.deleteClusterDeployment(
+                deployment.getMetadata(), deployment.getStatus(), false);
+        flinkService.waitForClusterShutdown(deployConfig);
+        if (!flinkService.isHaMetadataAvailable(deployConfig)) {
+            LOG.info(
+                    "Job never entered stable state. Clearing previous spec to reset for initial deploy");
+            // TODO: lastSpecWithMeta.f1.isFirstDeployment() is false
+            // ReconciliationUtils.clearLastReconciledSpecIfFirstDeploy(deployment);
+            deployment.getStatus().getReconciliationStatus().setLastReconciledSpec(null);
+            // UPGRADING triggers immediate reconciliation
+            deployment
+                    .getStatus()
+                    .getReconciliationStatus()
+                    .setState(ReconciliationState.UPGRADING);
+            return Optional.empty();
+        } else {
+            // proceed with upgrade if deployment succeeded between check and delete
+            LOG.info("Found HA state after deployment deletion, falling back to stateful upgrade");
+            return Optional.of(UpgradeMode.LAST_STATE);
+        }
     }
 
     @Override
--- a/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconcilerTest.java
+++ b/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconcilerTest.java
@@ -147,6 +147,11 @@ public class ApplicationReconcilerTest {
 
         deployment.getSpec().getJob().setUpgradeMode(UpgradeMode.LAST_STATE);
         deployment.getSpec().setRestartNonce(100L);
+        deployment
+                .getStatus()
+                .getReconciliationStatus()
+                .setLastStableSpec(
+                        deployment.getStatus().getReconciliationStatus().getLastReconciledSpec());
         flinkService.setHaDataAvailable(false);
         deployment.getStatus().getJobStatus().setState("RECONCILING");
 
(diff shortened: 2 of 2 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractJobReconciler.java
```diff
@@ -107,12 +107,18 @@ protected void reconcileSpecChange(
 
         JobState currentJobState = lastReconciledSpec.getJob().getState();
         JobState desiredJobState = currentDeploySpec.getJob().getState();
+        // check upgrade mode compatibility if deployment was ever stable
+        // an initially unsuccessful deployment can be reverted
+        var hasStableSpec =
+                resource.getStatus().getReconciliationStatus().getLastStableSpec() != null;
```
gyfora, line 113: I think this check is too weak. You might have completed checkpoints already at this point (if the job went into a restart loop after some checkpoints but before the operator considered it to be stable).

Also as we improve stability conditions it might become even weaker in the future

gyfora (changes requested):
I think only checking the existence of stable spec is too dangerous. The job can start up, take 2 checkpoints and go in a failure loop without the operator ever seeing it "stable".

Another problem with this approach is I think it introduces a bug with the `initialSavepointPath` logic. That field is only used in first deployment scenarios so with this change, it would be completely ignored the second time you try to fix your never running job.

I fixed a similar issue in this commit: https://github.com/apache/flink-kubernetes-operator/commit/ac21bc8fe148f6dc803988791224f792a66875ce

I think a slightly different approach would be a bit better:

1. Observer somehow marks the status as initial failure. It has to confirm that the jobmanager never actually started somehow.
2. In Reconciler if we get an upgrade and have initial failure, delete deployment & use `ReconciliationUtils.clearLastReconciledSpecIfFirstDeploy(flinkApp);`, reschedulre reconcile loop with 0 (like mid upgrade)
3. First deploy logic takes over submitting the new spec, with correct initial savepoint settings

tweise:
> I think only checking the existence of stable spec is too dangerous. The job can start up, take 2 checkpoints and go in a failure loop without the operator ever seeing it "stable".
> 
> Another problem with this approach is I think it introduces a bug with the `initialSavepointPath` logic. That field is only used in first deployment scenarios so with this change, it would be completely ignored the second time you try to fix your never running job.

Good catch. Looks like we are lacking test coverage for that.

> 
> I fixed a similar issue in this commit: [ac21bc8](https://github.com/apache/flink-kubernetes-operator/commit/ac21bc8fe148f6dc803988791224f792a66875ce)
> 
> I think a slightly different approach would be a bit better:
> 
> 1. Observer somehow marks the status as initial failure. It has to confirm that the jobmanager never actually started somehow.
> 2. In Reconciler if we get an upgrade and have initial failure, delete deployment & use `ReconciliationUtils.clearLastReconciledSpecIfFirstDeploy(flinkApp);`, reschedulre reconcile loop with 0 (like mid upgrade)

The reset process would need to be triggered by the spec change, not by observation of the pending deployment. What prompts the reset is that there is a spec change while the initial deployment has not been "reconciled". The pending deployment may succeed or not, operator cannot determine that. (Like in the situations that lead to the discovery of this issue, image pull error and entry point failure).

We can probably tighten the last spec change by also checking for presence of HA metadata. (`FlinkUtils.isKubernetesHAActivated(deployConfig) && FlinkUtils.isKubernetesHAActivated(observeConfig) && flinkService.isHaMetadataAvailable(deployConfig)`).

gyfora:
I am overall a bit skeptical whether we should add this "fix" or not, exactly because as you mentioned 

"The pending deployment may succeed or not, operator cannot determine that."

If this is true, I think we should not try to fix this because it cannot be done without possible introducing a dangerous race condition. Even if we add any check about isHA active etc it can still happen that the job starts just after we checked.

The user always have the possibility to delete and re-create initially failed deployments and redeploying it.

gyfora:
Maybe what we could do in these cases is to:

1. Delete deployment but keep HA metadata
2. Check if HA metadata exists (if it does the job started in the meantime)
3. If HA metadata doesn't exist, clear spec (reset to initial deployment)

gyfora:
Unfortunately even then it's possible that the job simply started but FAILED completely (that would delete HA metadata) and we cannot be sure whether checkpoints were taken. So again I am not completely sure about this...

tweise:
Why should this (from a users perspective) behave any different than a deployment after one checkpoint completed? I think we need a solution that does not require manual intervention. 

Wouldn't the HA check accomplish exactly that? To handle the very low probability race condition we check the HA metadata again after we deleted the deployment. If present, then the existing upgrade logic proceeds because the deployment had actually succeeded. Otherwise we reset by clearing the last reconciled spec as you had suggested above.

tweise:
Looks like there was a race condition here, too!

tweise:
> Unfortunately even then it's possible that the job simply started but FAILED completely (that would delete HA metadata) and we cannot be sure whether checkpoints were taken. So again I am not completely sure about this...

Yes, that could happen, at least theoretically! In practice though, wouldn't it be extremely rare that a job starts to checkpoint successfully and then goes into terminal failure (no longer attempting to recover from checkpoint) within the observation interval?

tweise:
@gyfora I'm testing the approach we discussed offline:
```
            if (!flinkService.isHaMetadataAvailable(deployConfig) && deployment.getStatus().getReconciliationStatus().getLastStableSpec() == null) {
                // initial deployment failure, reset to allow for spec change
                flinkService.deleteClusterDeployment(
                                        deployment.getMetadata(), deployment.getStatus(), false);
                flinkService.waitForClusterShutdown(deployConfig);
                // in case the deployment succeeded between check and delete, fall through to the upgrade path
                if (!flinkService.isHaMetadataAvailable(deployConfig)) {
                    LOG.info(
                            "Job never entered stable state. Clearing previous spec to reset for initial deploy");
                    // TODO: this doesn't work because lastSpecWithMeta.f1.isFirstDeployment() is false
                    // ReconciliationUtils.clearLastReconciledSpecIfFirstDeploy(deployment);
                    deployment.getStatus().getReconciliationStatus().setLastReconciledSpec(null);
                    // TODO: reschedule immediate reconciliation
                    return Optional.empty();
                }
            }
```

I need to check why `ReconciliationUtils.clearLastReconciledSpecIfFirstDeploy(deployment);` does not work, this could point to another issue.

tweise:
@gyfora PTAL

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
```diff
@@ -98,12 +99,39 @@ protected Optional<UpgradeMode> getAvailableUpgradeMode(
...
                 && FlinkUtils.isKubernetesHAActivated(observeConfig)
-                && flinkService.isHaMetadataAvailable(deployConfig)
                 && !flinkVersionChanged(
                         ReconciliationUtils.getDeployedSpec(deployment), deployment.getSpec())) {
-            LOG.info(
-                    "Job is not running but HA metadata is available for last state restore, ready for upgrade");
-            return Optional.of(UpgradeMode.LAST_STATE);
+
+            if (!flinkService.isHaMetadataAvailable(deployConfig)) {
+                if (deployment.getStatus().getReconciliationStatus().getLastStableSpec() == null) {
```
gyfora, line 106: Would be nice to encapsulate the logic inside this branch into a method/utility, there are quite a few things happening in there.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
```diff
@@ -98,12 +99,39 @@ protected Optional<UpgradeMode> getAvailableUpgradeMode(
...
+                    // initial deployment failure, reset to allow for spec change to proceed
+                    flinkService.deleteClusterDeployment(
+                            deployment.getMetadata(), deployment.getStatus(), false);
+                    flinkService.waitForClusterShutdown(deployConfig);
+                    // in case the deployment succeeded between check and delete, fall through to
+                    // the upgrade path
+                    if (!flinkService.isHaMetadataAvailable(deployConfig)) {
```
gyfora, line 113: I think there should be an `else` branch which returns LAST_STATE (if metadata was actually present after shutdown)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
```diff
@@ -98,12 +99,39 @@ protected Optional<UpgradeMode> getAvailableUpgradeMode(
...
+                        LOG.info(
+                                "Job never entered stable state. Clearing previous spec to reset for initial deploy");
+                        // TODO: lastSpecWithMeta.f1.isFirstDeployment() is false
```
gyfora, line 116: Would be nice to clear up this TODO

gyfora (changes requested):
Added some minor comments but otherwise the change looks good!

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
tweise: Indeed, and that also made it easier to move the logic to a separate method.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
tweise: done

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
tweise: I think there could be something wrong with the first deployment logic (or I don't understand it). But it would be better to deal with that outside of this PR.

tweise:
Created https://issues.apache.org/jira/browse/FLINK-29159

PR #357, opened 2022-08-29
[FLINK-29110] Use StatefulSet instead of Deployment to deploy JM and …

Grypse:

This pull request make an improvement for standalone mode to use StatefulSet instead of Deployment to deploy JM and TM to support mount a dynamically-created PersistentVolumeClaim.

## Brief change log
  - Use StatefulSet to instead of Deployment to depoly JM and TM
  - Introduce StandaloneKubernetesJobManagerSpecification to encapsulate JM specification info 
  - Add volumeClaimTemplates for JobManagerSpec and TaskManagerSpec
  - Add JOB_MANAGER_PVC_TEMPLATE and TASK_MANAGER_PVC_TEMPLATE ConfigOption to set jobmanager and taskmanager pvc template file separately.
  - Modify obersving logical for JM and TM StatefulSet.
  - Add statefulset rbac.
  - Modify docs/content/docs/custom-resource/reference.md to describe volumeClaimTemplates settings.
  - Add standalone cluster example with pvc settings.

## Verifying this change

This change is already covered by existing tests, such as .
 - Modified Fabric8FlinkStandaloneKubeClientTest to adapt StatefulSet creating test.
 - Modified KubernetesStandaloneClusterDescriptorTest to adapt StatefulSet to test standalone cluster deploying
 - Modified StandaloneFlinkServiceTest to adapt StatefulSet to test StandaloneFlinkService to deploy standalone cluster.
 - Modified StandaloneKubernetesJobManagerFactoryTest to adapt StatefulSet to test jobmanager spec.
 - Modified StandaloneKubernetesTaskManagerFactoryTest to adapt StatefulSet to test taskmanager spec

  - Added TestUtils.buildStandaloneSessionCluster() and TestUtils.buildStandaloneApplicationCluster() for standalone cluster tests.
  - Added SessionObserverTest.observeStandaloneSessionCluster to verify standalone session cluster observing logical

  - The public API, i.e., is any changes to the `CustomResourceDescriptors`: (yes )
  - Core observer or reconciler logic that is regularly executed: (yes )

## Documentation

  - Does this pull request introduce a new feature? (yes )
  - If yes, how is the feature documented? (docs )

### DIFF
--- a/examples/basic-checkpoint-ha.yaml
+++ b/examples/basic-checkpoint-ha.yaml
@@ -58,3 +58,4 @@ spec:
     upgradeMode: savepoint
     state: running
     savepointTriggerNonce: 0
+  mode: native
--- a/examples/basic-ingress.yaml
+++ b/examples/basic-ingress.yaml
@@ -39,3 +39,4 @@ spec:
   job:
     jarURI: local:///opt/flink/examples/streaming/StateMachineExample.jar
     parallelism: 2
+  mode: native
--- a/examples/basic-session-deployment-only.yaml
+++ b/examples/basic-session-deployment-only.yaml
@@ -34,3 +34,4 @@ spec:
     resource:
       memory: "2048m"
       cpu: 1
+  mode: native
--- a/examples/custom-logging.yaml
+++ b/examples/custom-logging.yaml
@@ -63,3 +63,4 @@ spec:
       logger.zookeeper.level = INFO
       logger.netty.name = org.apache.flink.shaded.akka.org.jboss.netty.channel.DefaultChannelPipeline
       logger.netty.level = OFF
+  mode: native
--- a/examples/pod-template.yaml
+++ b/examples/pod-template.yaml
@@ -82,3 +82,4 @@ spec:
     jarURI: local:///opt/flink/downloads/flink-examples-streaming.jar
     entryClass: org.apache.flink.streaming.examples.statemachine.StateMachineExample
     parallelism: 2
+  mode: native
--- a/examples/standalone_pvc.yaml
+++ b/examples/standalone_pvc.yaml
@@ -1,0 +1,81 @@
+################################################################################
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+# limitations under the License.
+################################################################################
+
+kind: FlinkDeployment
+metadata:
+  namespace: default
+  name: basic-example
+spec:
+  image: flink:1.15.1
+  flinkVersion: v1_15
+  flinkConfiguration:
+    taskmanager.numberOfTaskSlots: "2"
+  serviceAccount: flink
+  jobManager:
+    replicas: 1
+    resource:
+      memory: "2048m"
+      cpu: 1
+    volumeClaimTemplates:
+      - metadata:
+          name: log
+        spec:
+          accessModes: [ "ReadWriteOnce" ]
+          ### Please override with custom supported storageClassName
+          storageClassName: "alicloud-local-lvm"
+          resources:
+            requests:
+              storage: 10Gi
+    podTemplate:
+      apiVersion: v1
+      kind: Pod
+      metadata:
+        name: job-manager-pod-template
+      spec:
+        containers:
+          - name: flink-main-container
+            volumeMounts:
+              - name: log
+                mountPath: /opt/flink/log
+  taskManager:
+    replicas: 1
+    resource:
+      memory: "2048m"
+      cpu: 1
+    volumeClaimTemplates:
+      - metadata:
+          name: log
+        spec:
+          accessModes: [ "ReadWriteOnce" ]
+          ### Please override with custom supported storageClassName
+          storageClassName: "alicloud-local-lvm"
+          resources:
+            requests:
+              storage: 10Gi
+    podTemplate:
+      apiVersion: v1
+      kind: Pod
+      metadata:
+        name: task-manager-pod-template
+      spec:
+        containers:
+          - name: flink-main-container
+            volumeMounts:
+              - name: log
+                mountPath: /opt/flink/log
+  mode: standalone
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
@@ -46,6 +46,7 @@ import org.apache.flink.streaming.api.environment.ExecutionCheckpointingOptions;
 import org.apache.flink.util.FileUtils;
 import org.apache.flink.util.StringUtils;
 
+import io.fabric8.kubernetes.api.model.PersistentVolumeClaim;
 import io.fabric8.kubernetes.api.model.Pod;
 import io.fabric8.kubernetes.client.internal.SerializationUtils;
 import org.slf4j.Logger;
@@ -57,7 +58,9 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.nio.file.Files;
 import java.time.Duration;
+import java.util.ArrayList;
 import java.util.Collections;
+import java.util.List;
 
 import static org.apache.flink.configuration.DeploymentOptions.SHUTDOWN_ON_APPLICATION_FINISH;
 import static org.apache.flink.configuration.DeploymentOptions.SUBMIT_FAILED_JOB_ON_APPLICATION_ERROR;
@@ -214,6 +217,7 @@ public class FlinkConfigBuilder {
                         KubernetesConfigOptions.KUBERNETES_JOBMANAGER_REPLICAS,
                         spec.getJobManager().getReplicas());
             }
+            setPvcTemplate(spec.getJobManager().getVolumeClaimTemplates(), effectiveConfig, true);
         }
         return this;
     }
@@ -226,13 +230,13 @@ public class FlinkConfigBuilder {
                     spec.getTaskManager().getPodTemplate(),
                     effectiveConfig,
                     false);
-
             if (spec.getTaskManager().getReplicas() != null
                     && spec.getTaskManager().getReplicas() > 0) {
                 effectiveConfig.set(
                         StandaloneKubernetesConfigOptionsInternal.KUBERNETES_TASKMANAGER_REPLICAS,
                         spec.getTaskManager().getReplicas());
             }
+            setPvcTemplate(spec.getJobManager().getVolumeClaimTemplates(), effectiveConfig, false);
         }
 
         if (spec.getJob() != null
@@ -415,6 +419,33 @@ public class FlinkConfigBuilder {
         Files.write(tmp.toPath(), SerializationUtils.dumpAsYaml(podTemplate).getBytes());
         tmp.deleteOnExit();
         return tmp.getAbsolutePath();
+    }
+
+    private static void setPvcTemplate(
+            List<PersistentVolumeClaim> persistentVolumeClaims,
+            Configuration effectiveConfig,
+            boolean isJm)
+            throws IOException {
+        if (persistentVolumeClaims != null) {
+            final ConfigOption<String> pvcConfigOptions =
+                    isJm
+                            ? StandaloneKubernetesConfigOptionsInternal.JOB_MANAGER_PVC_TEMPLATE
+                            : StandaloneKubernetesConfigOptionsInternal.TASK_MANAGER_PVC_TEMPLATE;
+            effectiveConfig.setString(
+                    pvcConfigOptions, String.join(",", createPvcTempFile(persistentVolumeClaims)));
+        }
+    }
+
+    private static List<String> createPvcTempFile(List<PersistentVolumeClaim> pvcTemplates)
+            throws IOException {
+        List<String> pvcFilePaths = new ArrayList<>();
+        for (PersistentVolumeClaim pvc : pvcTemplates) {
+            final File tmp = File.createTempFile(GENERATED_FILE_PREFIX + "podTemplate_", ".yaml");
+            Files.write(tmp.toPath(), SerializationUtils.dumpAsYaml(pvc).getBytes());
+            tmp.deleteOnExit();
+            pvcFilePaths.add(tmp.getAbsolutePath());
+        }
+        return pvcFilePaths;
     }
 
     protected static void cleanupTmpFiles(Configuration configuration) {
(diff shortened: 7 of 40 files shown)


Grypse:
@gyfora @wangyang0918 Please help me review this PR.

Grypse:
@flinkbot run azure

gyfora:
Thank you for the PR @Grypse we will review this with @usamj

examples/standalone_pvc.yaml
```diff
@@ -0,0 +1,81 @@
...
+      kind: Pod
+      metadata:
+        name: task-manager-pod-template
+      spec:
+        containers:
+          - name: flink-main-container
+            volumeMounts:
+              - name: log
+                mountPath: /opt/flink/log
+  mode: standalone
```
gyfora, line 81: missing newline at end

examples/standalone_pvc.yaml
```diff
@@ -0,0 +1,81 @@
...
+    resource:
+      memory: "2048m"
+      cpu: 1
+    volumeClaimTemplates:
+      - metadata:
+          name: log
+        spec:
+          accessModes: [ "ReadWriteOnce" ]
+          ### Please override with custom supported storageClassName
+          storageClassName: "alicloud-local-lvm"
```
gyfora, line 40: Can we remove the reference to `alicloud` ?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
```diff
@@ -417,6 +421,33 @@ private static String createTempFile(Pod podTemplate) throws IOException {
...
+            final ConfigOption<String> pvcConfigOptions =
+                    isJm
+                            ? StandaloneKubernetesConfigOptionsInternal.JOB_MANAGER_PVC_TEMPLATE
+                            : StandaloneKubernetesConfigOptionsInternal.TASK_MANAGER_PVC_TEMPLATE;
+            effectiveConfig.setString(
+                    pvcConfigOptions, String.join(",", createPvcTempFile(persistentVolumeClaims)));
+        }
+    }
+
+    private static List<String> createPvcTempFile(List<PersistentVolumeClaim> pvcTemplates)
```
gyfora, line 439: We should add the newly created files to the list of cleaned up files under `cleanupTmpFiles` in this class

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkUtils.java
```diff
@@ -187,6 +187,18 @@ public static void setGenerationAnnotation(Configuration conf, Long generation)
...
+        if (generation == null) {
+            return;
+        }
+        var labels =
+                new HashMap<>(
+                        conf.getOptional(KubernetesConfigOptions.TASK_MANAGER_ANNOTATIONS)
+                                .orElse(Collections.emptyMap()));
+        labels.put(CR_GENERATION_LABEL, generation.toString());
+        conf.set(KubernetesConfigOptions.TASK_MANAGER_ANNOTATIONS, labels);
+    }
```
gyfora, line 200: I feel this method duplicates too much code, we could simply have a generic method that takes the TASK_MANAGER/JOBMANAGER_ANNOTATIONS config option as param to factor out the logic

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/AbstractDeploymentObserver.java
```diff
@@ -141,41 +148,19 @@ protected void observeJmDeployment(
...
-                // typically it takes a few seconds for the REST server to be ready
-                logger.info(
-                        "JobManager deployment port is ready, waiting for the Flink REST API...");
-                deploymentStatus.setJobManagerDeploymentStatus(
-                        JobManagerDeploymentStatus.DEPLOYED_NOT_READY);
+        if (KubernetesDeploymentMode.NATIVE.equals(flinkApp.getSpec().getMode())) {
+            Optional<Deployment> deployment = context.getSecondaryResource(Deployment.class);
+            if (deployment.isPresent()) {
+                jmDeploymentObserver(deployment.get(), effectiveConfig, deploymentStatus, flinkApp);
                 return;
```
gyfora, line 155: Why do we use StatefulSets for JobManager also? Seems to be an unnecessary complication without much benefits.

gyfora (changes requested):
I started reviewing this.

One bigger question:
Should we keep using `Deployment` for the jobmanager instead of `StatefulSet`? I think for Taskmanagers statefulset brings obvious benefits but for the JM it just makes the overall logic (observer etc.) much more complex.

cc @wangyang0918

examples/basic-checkpoint-ha.yaml
```diff
@@ -58,3 +58,4 @@ spec:
     upgradeMode: savepoint
     state: running
     savepointTriggerNonce: 0
+  mode: native
```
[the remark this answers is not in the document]
Grypse: Added!

examples/basic-ingress.yaml
```diff
@@ -39,3 +39,4 @@ spec:
   job:
     jarURI: local:///opt/flink/examples/streaming/StateMachineExample.jar
     parallelism: 2
+  mode: native
```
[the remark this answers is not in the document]
Grypse: Added!

examples/basic-session-deployment-only.yaml
```diff
@@ -34,3 +34,4 @@ spec:
     resource:
       memory: "2048m"
       cpu: 1
+  mode: native
```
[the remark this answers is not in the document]
Grypse: Added newline to end.

examples/custom-logging.yaml
```diff
@@ -63,3 +63,4 @@ spec:
       logger.zookeeper.level = INFO
       logger.netty.name = org.apache.flink.shaded.akka.org.jboss.netty.channel.DefaultChannelPipeline
       logger.netty.level = OFF
+  mode: native
```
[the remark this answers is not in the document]
Grypse: Added newline to the end.

examples/pod-template.yaml
```diff
@@ -82,3 +82,4 @@ spec:
     jarURI: local:///opt/flink/downloads/flink-examples-streaming.jar
     entryClass: org.apache.flink.streaming.examples.statemachine.StateMachineExample
     parallelism: 2
+  mode: native
```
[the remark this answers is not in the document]
Grypse: Added newline to the end.

examples/standalone_pvc.yaml
Grypse: Added newline to the end.

examples/standalone_pvc.yaml
Grypse: Just removed `alicloud` reference.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
Grypse: Added the newly created files to the list of cleaned up files to the method of `cleanupTmpFiles `

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkUtils.java
Grypse: I have taken the TASK_MANAGER/JOBMANAGER_ANNOTATIONS config option as param to refactor annotation generation

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/AbstractDeploymentObserver.java
Grypse: With StatefulSet for JobManager has the following advantages：

1. Avoid random string suffix of JM Pod, more intuitive identification of JM.
2. There is no need to manually create the PVC each time, which is very convenient when you need to create many flink clusters in production.
3. When JM replicas is greater than 1, the one-to-one correspondence between PVC and JM can be guaranteed.

gyfora: If these are the only advantages I would prefer Deployment:
 1. Native integration has the random suffix, not a problem you can select by label if necessary
 2. Not sure why you need a PVC for the JM
 3. There is not a good reason to have more than one 1 replica and even then PVC is not necessary anyways
 
 @wangyang0918

gyfora: Also looping @tweise here

gyfora:
My main concern with the Jobmanager StatefulSet change is that it introduces considerable complexity and branching in the core observer logic with very little benefit. I don't think PVCs are very important for JMs.

I would be in favor of keeping the Deployments for the JM component to be aligned with the Native implementation.

It would be great to have input from the others also as this is an important decision with major code implications :) :
@wangyang0918 @tweise @morhidi @usamj

usamj:
I agree with @gyfora. and aligns with what I originally thought when I saw the change. I don't think there is enough benefit to the JM being a StatefulSet

My other concern is whether we should only use StatefulSets for the TM's or whether using StatefulSets/Deployment for them should be a configurable option. The reason for this is that recovery time for a TM failing would be longer if it used StatefulSets. It could also result in issues and user action to fix if a TM pod was stuck in a Terminating state which could be unnecessary pain for users if they aren't using PVC's with their TM's

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/AbstractDeploymentObserver.java
Grypse: Another benefit:
when using a system with StatefulSet, allowing a simple HA solution like using a filesystem with PVC(such as SSD) can recover faster than remote storage(such as HDFS or OSS).

Grypse: It's very important to run JM with PVC for working with work-dir[1] to fast recovery.

1. https://nightlies.apache.org/flink/flink-docs-release-1.15/docs/deployment/resource-providers/standalone/working_directory/

gyfora: I think you can have PVC with the current podTemplate also right?

Also the JM doesn’t access much data during recovery , so even if it has to fetch it from remote storage it wouldn’t really have a big impact.

Grypse:
@usamj Maybe we can set three mode for flinkdeployment CR：

1. native
2. standalone_deployment
3. standalone_statefulset

two of these modes are available as standalone：

1. `standalone_deployment`
2. `standalone_statefulset`

And volumeClaimTemplates work only in `standalone_statefulset` mode.

Grypse:
I believe that configuring PVC for JM is necessary in some scenarios.
@gyfora As @usamj said，we can set a configurable option for standalone to choose Deployment or StatefulSet.

gyfora:
> @usamj Maybe we can set three mode for flinkdeployment CR：
> 
> 1. native
> 2. standalone_deployment
> 3. standalone_statefulset
> 
> two of these modes are available as standalone：
> 
> 1. standalone_deployment
> 2. standalone_statefulset
> 
> And volumeClaimTemplates work only in `standalone_statefulset` mode.

I think we should not complicate things even more :) If we want to make this configurable we could use a config option for it but maybe we dont need it.

@usamj would statefulset for TMs work for you? Or do you see any reasons why you would need to have this as a configurable option?

gyfora:
@Grypse I understand your concern regarding the JM. Let me try to address some of your points. Your biggest argument for using StatefulSets for JM is improved recovery time with a PVC.

In my experience JM recovery time is almost never an issue, and having PVC vs remote storage have almost no difference in speed for recovering checkpoint/ha metadata as they are not that large. Furthermore I think even with the current Deployment + podTemplate architecture you can configure a PVC if you really need to.

The Flink community also decided to use Deployment instead of StatefulSet for implementing the native integration and I don't remember people hitting any issues that (but of course I could be wrong here). Based on this I don't think we need to implement this right away.

In any case if we cannot agree on this we should discuss further on the dev mailing list to include the community.

I think we shouldn't downplay the cost of increased complexity in the operator. Instead of maintaining 2 parallel observer paths we could instead focus on making the Deployment observer logic more robust and better.

morhidi:
Hi Folks,

It would be really great to see what `faster` recovery means here, through an example. In general this sounds a fair benefit but before choosing a more complicated path here it would be great to see some obvious gains.

Would this solution benefit reactive scaling for example? Would this mean that in case of a TM failure it could recover before the job is being restarted with a lower parallelism? 

@Grypse do you have any comparison that would help us decide to embrace this feature?

Thanks,
Matyas

gyfora:
@morhidi I think this would have no impact on reactive scaling. I think the actual benefit is the possibility to configure PVCs for TMs which could help with checkpointing/local recovery in case of TM failures.

But as @usamj pointed out, it might actually negatively impact the TM startup time itself.

gyfora:
@Grypse this seems like a fairly large architecture decision that we underestimated at first. We would like to kindly ask you to prepare your proposal in a FLIP format and post it on the DEV mailing list.

Please describe the following points in the FLIP:

- What is the benefit of using StatefulSet for TaskManagers?
  - Does this only affect local recovery? What is the performance gain compared to Deployments?
  - What benefit do we have compared to the PVC available in the podTemplate? Seems like ReadWriteMany PVCs already provide the same benefits for Deployments
  - What is the implication of TaskManager startup time? Deployments start pods faster as they dont need to wait for termination
  - Should we make Deployment / StatefulSet configurable?

- What is the benefit of using StatefulSet for JobManagers?
  - What data would the jobmanager store in PVC? What is the performance gain?
  - Does this only affect state handles? That is very small data.
  - Why can't we do this already with the podTemplate?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/AbstractDeploymentObserver.java
wangyang0918: Sorry for join the discussion so late.

I share the same concern with @gyfora and @usamj . Having a PVC for JobManager does not take enough benefits now. IIRC, we does not have any state in the JobManager which needs to be recovered after failover. Using the StatefulSet also makes the JobManager has risk that could not be started if the previous one terminated too slow or stucked.

wangyang0918:
Having a FLIP to address the @gyfora's comments will make the discussion more efficient. Look forward to it with some testing data. For example, how long does it take to mount a PVC or terminate a stateful pod?

Grypse:
@gyfora @wangyang0918 @morhidi @usamj Thanks for yours review. I will make a FLIP to describe the proposal.

mbalassi:
@Grypse great, please post a link here to the FLIP after you posted it to the mailing list and the wiki for reference. Thanks.

gyfora:
@Grypse are you still working on this FLIP?

navinvishy:
Hi @gyfora , we have a situation where nodes are periodically taken down for maintenance, causing jobs to restart. A few restarts of a job within a few hours often result in an increasing consumer lag. It appears that the local task recovery techniques described [here](https://nightlies.apache.org/flink/flink-docs-release-1.17/docs/ops/state/large_state_tuning/#task-local-recovery) and [here](https://nightlies.apache.org/flink/flink-docs-master/docs/deployment/resource-providers/standalone/kubernetes/#enabling-local-recovery-across-pod-restarts) would address this. 

Since we use the flink kubernetes operator, I landed on this PR looking for ways to enable this in the operator. Happy to take this forward if necessary.

gyfora:
@navinvishy , I think this PR / approach does not provide a comprehensive solution as it only works for standalone mode which is strictly inferior to the native integration.

I think if you use the native integration with local recovery then if a TM goes down only the single new TM needs to download the state.

nviswanath-roblox:
> @navinvishy , I think this PR / approach does not provide a comprehensive solution as it only works for standalone mode which is strictly inferior to the native integration.
> 
> I think if you use the native integration with local recovery then if a TM goes down only the single new TM needs to download the state.

@gyfora from what I can tell, this will not work for node failures, would it? It seems like we would need a `StatefulSet` to make it work. I couldn't find a way to replace the Deployment with a StatefulSet with the operator.

PR #364, opened 2022-09-06
[FLINK-28272] Handle TLS Certificate Renewal in Webhook

gaborgsomogyi:

At the moment SSL context is not realoading the certificate when changed. In this PR I've added this functionality.

## Brief change log

* `FsWatchService` added with tests in `FsWatchServiceTest`
* `ReloadableSslContext` added
* Additional logging to see more

## Verifying this change

* Newly added unit tests
* Manually: https://gist.github.com/gaborgsomogyi/014054be6ef40326a2ade870f5f3dd02

## Documentation

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/fs/FileSystemWatchService.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/fs/FileSystemWatchService.java
@@ -1,0 +1,94 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.kubernetes.operator.fs;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.File;
+import java.nio.file.FileSystems;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.WatchEvent;
+import java.nio.file.WatchKey;
+import java.nio.file.WatchService;
+
+import static java.nio.file.StandardWatchEventKinds.ENTRY_CREATE;
+import static java.nio.file.StandardWatchEventKinds.ENTRY_DELETE;
+import static java.nio.file.StandardWatchEventKinds.ENTRY_MODIFY;
+import static java.nio.file.StandardWatchEventKinds.OVERFLOW;
+
+/** Service which is able to watch local filesystem directories. */
+public class FileSystemWatchService extends Thread {
+    private static final Logger LOG = LoggerFactory.getLogger(FileSystemWatchService.class);
+
+    private final String directoryPath;
+
+    public FileSystemWatchService(String directoryPath) {
+        if (!new File(directoryPath).isDirectory()) {
+            throw new IllegalArgumentException("Directory must exists: " + directoryPath);
+        }
+        this.directoryPath = directoryPath;
+    }
+
+    @Override
+    public void run() {
+        try (WatchService watcher = FileSystems.getDefault().newWatchService()) {
+            LOG.info("Starting watching path: " + directoryPath);
+            Path realDirectoryPath = Paths.get(directoryPath).toRealPath();
+            LOG.info("Path is resolved to real path: " + realDirectoryPath);
+            realDirectoryPath.register(watcher, ENTRY_CREATE, ENTRY_DELETE, ENTRY_MODIFY);
+            onWatchStarted(realDirectoryPath);
+
+            while (true) {
+                LOG.debug("Taking watch key");
+                WatchKey watchKey = watcher.take();
+                LOG.debug("Watch key arrived");
+                for (WatchEvent<?> watchEvent : watchKey.pollEvents()) {
+                    LOG.debug("Watch event count: " + watchEvent.count());
+                    if (watchEvent.kind() == OVERFLOW) {
+                        LOG.error("Filesystem events may have been lost or discarded");
+                        Thread.yield();
+                    } else if (watchEvent.kind() == ENTRY_CREATE) {
+                        onFileOrDirectoryCreated((Path) watchEvent.context());
+                    } else if (watchEvent.kind() == ENTRY_DELETE) {
+                        onFileOrDirectoryDeleted((Path) watchEvent.context());
+                    } else if (watchEvent.kind() == ENTRY_MODIFY) {
+                        onFileOrDirectoryModified((Path) watchEvent.context());
+                    } else {
+                        throw new IllegalStateException("Invalid event kind: " + watchEvent.kind());
+                    }
+                }
+                watchKey.reset();
+            }
+        } catch (InterruptedException e) {
+            LOG.info("Filesystem watcher interrupted");
+        } catch (Exception e) {
+            LOG.error("Filesystem watcher received exception and stopped: " + e);
+            throw new RuntimeException(e);
+        }
+    }
+
+    protected void onWatchStarted(Path realDirectoryPath) {}
+
+    protected void onFileOrDirectoryCreated(Path relativePath) {}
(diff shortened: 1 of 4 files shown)


gaborgsomogyi:
cc @morhidi @mbalassi

mbalassi:
Thanks, @gaborgsomogyi this is much needed.

The code LGTM, I would suggest to add a note to the documentation stating that this functionality exists. 😏 I do not see a clear cut best place for it, it might make sense to add an `Admission Control` subsection to the [architecture](https://nightlies.apache.org/flink/flink-kubernetes-operator-docs-main/docs/concepts/architecture/) page with a paragraph on the webhook functionality anyway and also include that the certificate is dynamically watched. What do you think @morhidi, @gaborgsomogyi?

gaborgsomogyi:
Good comment, tomorrow adding it unless somebody has better suggestion in terms if place.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/fs/FsWatchService.java
```diff
@@ -0,0 +1,90 @@
...
+import static java.nio.file.StandardWatchEventKinds.OVERFLOW;
+
+/** Service which is able to watch local filesystem directories. */
+public class FsWatchService extends Thread {
+    private static final Logger LOG = LoggerFactory.getLogger(FsWatchService.class);
+
+    private final String directoryPath;
+
+    public FsWatchService(String directoryPath) {
+        this.directoryPath = directoryPath;
```
SteNicholas, line 42: Should it check whether `directoryPath` exist? If not exist, it's unnecessary to run the `FsWatchService `.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/fs/FsWatchService.java
```diff
@@ -0,0 +1,90 @@
...
+            LOG.info("Filesystem watcher interrupted");
+        } catch (Exception e) {
+            LOG.error("Filesystem watcher received exception and stopped: " + e);
+            throw new RuntimeException(e);
+        }
+    }
+
+    protected void onWatchStarted(Path realDirectoryPath) {}
+
+    protected void onFileOrDirectoryCreated(Path relativePath) {}
```
SteNicholas, line 85: Could the `onFileOrDirectoryCreated `, `onFileOrDirectoryDeleted` and `onFileOrDirectoryModified` merge into one interface `onWatchedEvent(WatchEvent.Kind kind, Path relativePath)`?

flink-kubernetes-webhook/src/main/java/org/apache/flink/kubernetes/operator/admission/FlinkOperatorWebhook.java
```diff
@@ -138,19 +134,35 @@ private static SslContext createSslContext() throws Exception {
...
+        return reloadableSslContext;
+    }
 
-        return SslContextBuilder.forServer(kmf)
-                .ciphers(Http2SecurityUtil.CIPHERS, SupportedCipherSuiteFilter.INSTANCE)
-                .build();
+    private static void stopFsWatchService() throws InterruptedException {
+        if (fsWatchService != null) {
+            fsWatchService.interrupt();
+            fsWatchService.join();
```
SteNicholas, line 165: Should `fsWatchService` be set to `null`?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/fs/FsWatchService.java
```diff
@@ -0,0 +1,90 @@
...
+import java.nio.file.WatchKey;
+import java.nio.file.WatchService;
+
+import static java.nio.file.StandardWatchEventKinds.ENTRY_CREATE;
+import static java.nio.file.StandardWatchEventKinds.ENTRY_DELETE;
+import static java.nio.file.StandardWatchEventKinds.ENTRY_MODIFY;
+import static java.nio.file.StandardWatchEventKinds.OVERFLOW;
+
+/** Service which is able to watch local filesystem directories. */
+public class FsWatchService extends Thread {
```
morhidi, line 36: nit: FileSystemWatchService

morhidi:
> Thanks, @gaborgsomogyi this is much needed.
> 
> The code LGTM, I would suggest to add a note to the documentation stating that this functionality exists. 😏 I do not see a clear cut best place for it, it might make sense to add an `Admission Control` subsection to the [architecture](https://nightlies.apache.org/flink/flink-kubernetes-operator-docs-main/docs/concepts/architecture/) page with a paragraph on the webhook functionality anyway and also include that the certificate is dynamically watched. What do you think @morhidi, @gaborgsomogyi?

Nice Job @gaborgsomogyi! Verified it locally, works as expected. If you prefer you can open a separate Jira and add some proper documentation for webhook.

flink-kubernetes-webhook/src/main/java/org/apache/flink/kubernetes/operator/admission/FlinkOperatorWebhook.java
gaborgsomogyi: Please have a look at the first call of `createSslContext` and you see why it can be null.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/fs/FsWatchService.java
gaborgsomogyi: Renamed.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/fs/FsWatchService.java
gaborgsomogyi: I think it's internal detail that the internal implementation uses `WatchService` which is sending `WatchEvent.Kind` type which shouldn't appear in the API of the class. Simply loose couple.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/fs/FsWatchService.java
gaborgsomogyi: It blows up in a noisy way which is good because one can see that something is really bad:
```
java.nio.file.NoSuchFileException: /intentionally/missing/directory
```
Not starting it silently would give the user the feeling that everything is fine but it would be a lie, right?

gaborgsomogyi: What I can imagine is to add a check in the constructor and throw an exception. That way the user thread will be notified. Added a new test.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/fs/FileSystemWatchService.java
```diff
@@ -0,0 +1,94 @@
...
+                    } else if (watchEvent.kind() == ENTRY_MODIFY) {
+                        onFileOrDirectoryModified((Path) watchEvent.context());
+                    } else {
+                        throw new IllegalStateException("Invalid event kind: " + watchEvent.kind());
+                    }
+                }
+                watchKey.reset();
+            }
+        } catch (InterruptedException e) {
+            LOG.info("Filesystem watcher interrupted");
```
SteNicholas, line 80:
```suggestion
            LOG.warn("Filesystem watcher interrupted");
```

gaborgsomogyi: Why warning? Normal operation uses `Thread.interrupt` to notify the thread to exit from `watcher.take()`.

gaborgsomogyi:
@mbalassi @morhidi I've considered the doc in-depth and I think better to do it in separate jira since there is no doc for the webhook in general.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/fs/FileSystemWatchService.java
SteNicholas: Just directly ignore the `InterruptedException`?

SteNicholas (approved):
LGTM.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/fs/FileSystemWatchService.java
gaborgsomogyi: It would be better to have a private function which is called from 2 different tests in order not to mix-up left and right functionality but I'm personally fine w/ the actual code stand. It's a little bit harder to understand what's going on inside this single test but not horror.

gaborgsomogyi: The last message not here :)

gaborgsomogyi:
Thanks guys for the help!

PR #365, opened 2022-09-12
Flink 29257

mxm:

Users are confused how the SAVEPOINT ugprade mode works.

1. It's not clear to users that the savepoint is drawn during the upgrade process.
2. It's not clear to users what happens when the process fails.

## Brief change log

- modify docs

### DIFF
--- a/docs/content/docs/custom-resource/job-management.md
+++ b/docs/content/docs/custom-resource/job-management.md
@@ -98,8 +98,8 @@ The `upgradeMode` setting controls both the stop and restore mechanisms as detai
 The three  upgrade modes are intended to support different scenarios:
 
  1. **stateless**: Stateless application upgrades from empty state
- 2. **last-state**: Quick upgrades in any application state (even for failing jobs), does not require a healthy job as it always uses last checkpoint information. Manual recovery may be necessary if HA metadata is lost.
- 3. **savepoint**: Use savepoint (when possible) for upgrade, providing maximal safety and possibility to serve as backup/fork point.
+ 2. **last-state**: Quick upgrades in any application state (even for failing jobs), does not require a healthy job as it always uses the latest checkpoint information. Manual recovery may be necessary if HA metadata is lost.
+ 3. **savepoint**: Use savepoint for upgrade, providing maximal safety and possibility to serve as backup/fork point. The savepoint will be created during the upgrade process. Note that the Flink job needs to be running to allow the savepoint to get created. If the job is in an unhealthy state, the last checkpoint will be used. If the last checkpoint is not available, the job upgrade will fail.
 
 During stateful upgrades there are always cases which might require user intervention to preserve the consistency of the application. Please see the [manual Recovery section](#manual-recovery) for details.
 


morhidi:
+1 LGTM

morhidi:
cc @mbalassi

docs/content/docs/custom-resource/job-management.md
```diff
@@ -98,8 +98,8 @@ The `upgradeMode` setting controls both the stop and restore mechanisms as detai
 The three  upgrade modes are intended to support different scenarios:
 
  1. **stateless**: Stateless application upgrades from empty state
- 2. **last-state**: Quick upgrades in any application state (even for failing jobs), does not require a healthy job as it always uses last checkpoint information. Manual recovery may be necessary if HA metadata is lost.
- 3. **savepoint**: Use savepoint (when possible) for upgrade, providing maximal safety and possibility to serve as backup/fork point.
+ 2. **last-state**: Quick upgrades in any application state (even for failing jobs), does not require a healthy job as it always uses the latest checkpoint information. Manual recovery may be necessary if HA metadata is lost.
+ 3. **savepoint**: Use savepoint for upgrade, providing maximal safety and possibility to serve as backup/fork point. The savepoint will be created during the upgrade process. Note that the Flink job needs to be running to allow the savepoint to get created. If the job is in an unhealthy state, the last checkpoint will be used. If the last checkpoint is not available, the job upgrade will fail.
```
tweise, line 102: This isn't entirely correct either. Fallback to last checkpoint is a configurable option.

morhidi: Yup, forgot to comment it here: https://issues.apache.org/jira/browse/FLINK-29258

mbalassi:
Thanks, @mxm. Happy to merge this after you had the chance to address @tweise's comment.

docs/content/docs/custom-resource/job-management.md
mxm: I didn't know about the configuration option. I opened the issue because it's impossible for users to know what fallback behavior is configured. This should really be a separate upgrade mode.

mxm:
Updated.

docs/content/docs/custom-resource/job-management.md
gyfora: Originally we had no fallback behavior but in almost all cases users would prefer to upgrade the job even if it’s not running for some reason. There is always a good reason for the upgrade if there is a spec change and savepoint is kind of optional. 

For forking jobs the user can trigger manual savepoints and if they really need the savepoint upgrade they can set the config . Although they will be in big trouble if they must take savepoint and the job is failing for any reason :)

docs/content/docs/custom-resource/job-management.md
```diff
@@ -98,8 +98,8 @@ The `upgradeMode` setting controls both the stop and restore mechanisms as detai
...
 
  1. **stateless**: Stateless application upgrades from empty state
- 2. **last-state**: Quick upgrades in any application state (even for failing jobs), does not require a healthy job as it always uses last checkpoint information. Manual recovery may be necessary if HA metadata is lost.
- 3. **savepoint**: Use savepoint (when possible) for upgrade, providing maximal safety and possibility to serve as backup/fork point.
+ 2. **last-state**: Quick upgrades in any application state (even for failing jobs), does not require a healthy job as it always uses the latest checkpoint information. Manual recovery may be necessary if HA metadata is lost.
+ 3. **savepoint**: Use savepoint for upgrade, providing maximal safety and possibility to serve as backup/fork point. The savepoint will be created during the upgrade process. Note that the Flink job needs to be running to allow the savepoint to get created. If the job is in an unhealthy state, the last checkpoint will be used (if `kubernetes.operator.job.upgrade.last-state-fallback.enabled` is set to true). If the last checkpoint is not available, the job upgrade will fail.
```
gyfora, line 102: Technically it would be more precise to say:
"unless `kubernetes.operator.job.upgrade.last-state-fallback.enabled` is set to false" 

because the default is true (you don't have to set it)

PR #371, opened 2022-09-19
[FLINK-29313] Observe spec should include latest operator configs + improved test coverage

gyfora:

This PR improves the logic of the observe configuration to avoid problems with juggling operator specific confg between the new and deployed specs. We now ensure that operator related configs from the latest spec are always accounted for in the observe config.

The PR includes additional test coverage to ensure that the correct config is used during JM deployment recovery + improves the savepoint cleanup logic.

## Brief change log

  - *Add latest operator related configs to observeconfig*
  - *Improve savepoint cleanup*
  - *Add tests to ensure that new operator configs are picked up in reconcile/observe even if they didnt lead to a job upgrade*
  - *Add config test for JM deployment recovery*

## Verifying this change

Partly covered by existing tests and new tests have been added for the cases outlined in the ticket

## Documentation


gyfora:
cc @morhidi

gyfora:
In the long run we probably need a small overhaul of the config handling to eliminate all weird corner cases and special handling. I opened a jira ticket and will work on this for 1.3.0:  https://issues.apache.org/jira/browse/FLINK-29338

morhidi:
Thanks, will test this tomorrow.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/spec/AbstractFlinkSpec.java
```diff
@@ -51,8 +53,7 @@ public abstract class AbstractFlinkSpec implements Diffable<AbstractFlinkSpec> {
     /** Flink configuration overrides for the Flink deployment or Flink session job. */
     @SpecDiff.Config({
         @SpecDiff.Entry(prefix = "parallelism.default", type = DiffType.IGNORE),
-        @SpecDiff.Entry(prefix = "kubernetes.operator", type = DiffType.IGNORE),
-        @SpecDiff.Entry(prefix = "metrics.scope.k8soperator", type = DiffType.IGNORE)
+        @SpecDiff.Entry(prefix = K8S_OP_CONF_PREFIX, type = DiffType.IGNORE),
```
morhidi, line 56: Shell we keep the the deprecated configs e.g. `metrics.scope.k8soperator` here for a while, or we simply say that deprecated operator configs will trigger an upgrade?

morhidi:
Went through the PR, the situation with the configs actually not that bad. Just by looking at the code I was not able to figure out what happens with the operator configs in case of a rollback. I guess we can potentially loose some in case those are not kept in the new spec too.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/RollbackTest.java
```diff
@@ -179,6 +181,17 @@ public void testRollbackFailureWithLastState(FlinkVersion flinkVersion) throws E
...
+                    // Trigger deployment recovery
+                    flinkService.clear();
+                    flinkService.setPortReady(false);
+
+                    testController.reconcile(dep, context);
+                    flinkService.setPortReady(true);
+                    testController.reconcile(dep, context);
+                    var jobs = flinkService.listJobs();
+                    // Make sure deployment was recovered with correct spec/config
+                    assertTrue(jobs.get(jobs.size() - 1).f2.containsKey("t"));
```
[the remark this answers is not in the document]
gyfora: Not really, this change make it so that you pick up `kubernetes.operator.*` configs from the latest spec so even if you roll back you still pick the same.

gyfora:
> Went through the PR, the situation with the configs actually not that bad. Just by looking at the code I was not able to figure out what happens with the operator configs in case of a rollback. I guess we can potentially loose some in case those are not kept in the new spec too.

I don't think we have any problems with rollbacks because now we always use the latest operator related configs.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/spec/AbstractFlinkSpec.java
gyfora: I think we can say that they trigger an upgrade to keep the code clean, I doubt that anyone would set/change these very often

PR #372, opened 2022-09-19
[FLINK-29261] Use FAIL_ON_UNKNOWN_PROPERTIES=false in ReconciliationUtils

morhidi:
The operator cannot be downgraded with new CRD, once the CR containing a new field is written to the last reconciled/stable spec.

## Brief change log
The `ObjectMapper` is modified in `ReconciliationUtils`
```
private static final ObjectMapper objectMapper =
            new ObjectMapper().configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```

## Verifying this change

Added new `ReconciliationUtilsTest.testSpecDeserializationWithUnknownField()`

## Documentation

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
@@ -37,6 +37,7 @@ import org.apache.flink.kubernetes.operator.utils.StatusRecorder;
 import org.apache.flink.util.Preconditions;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import io.fabric8.kubernetes.client.CustomResource;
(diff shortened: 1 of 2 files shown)


morhidi:
> There are other components that use `ObjectMapper` the `FlinkValidator` for instance should probably also need this config otherwise the webhook might not be compatible

I've been thinking about this and I guess this is the only place we want to deserialize an object with unknown fields, otherwise there's no other way for rolling back to an older operator version. I don't see a reason yet we should allow this in webhook. I'm convincible though :)

gyfora:
> > There are other components that use `ObjectMapper` the `FlinkValidator` for instance should probably also need this config otherwise the webhook might not be compatible
> 
> I've been thinking about this and I guess this is the only place we want to deserialize an object with unknown fields, otherwise there's no other way for rolling back to an older operator version. I don't see a reason yet we should allow this in webhook. I'm convincible though :)

 1. You deploy a new operator version that has a new flinkSpec field.  
 2. You set that field in a CR
 3. You roll back the operator
 4. Any change to the CR will call the webhook with the CR containing the now unknown field you set in 2.

SteNicholas:
@morhidi @gyfora when using `FAIL_ON_UNKNOWN_PROPERTIES=false`, is it necessary to tell the users to make sure that the correct json properties have already verified and guaranteed by Kubernetes itself based on the installed CRD on the document?

gyfora:
> @morhidi @gyfora when using `FAIL_ON_UNKNOWN_PROPERTIES=false`, is it necessary to tell the users to make sure that the correct json properties have already verified and guaranteed by Kubernetes itself based on the installed CRD on the document?

Not sure if I understand the question correctly but Kubernetes always validates the json against the CRD and even if the user could submit extra fields they would be ignored and dropped without the user noticing. So I don't think it has any user impact this way.

PR #375, opened 2022-09-20
[FLINK-29327][operator] remove operator config from job runtime config before d…

HuangZhenQiu:
Remove operator related configs from flink runtime config, so that users will not see any operator related config in web ui.

## Brief change log
  - remove operator configs before deployments in Flink services.

## Verifying this change

This change is a trivial rework / code cleanup with unit test coverage.

## Documentation

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
@@ -184,13 +184,7 @@ public abstract class AbstractFlinkService implements FlinkService {
             throws Exception {
         // we generate jobID in advance to help deduplicate job submission.
         var jobID = FlinkUtils.generateSessionJobFixedJobID(meta);
-        Configuration runtimeConfig = removeOperatorConfigs(conf);
-        runJar(
-                spec.getJob(),
-                jobID,
-                uploadJar(meta, spec, runtimeConfig),
-                runtimeConfig,
-                savepoint);
+        runJar(spec.getJob(), jobID, uploadJar(meta, spec, conf), conf, savepoint);
         LOG.info("Submitted job: {} to session cluster.", jobID);
         return jobID;
     }
@@ -632,7 +626,7 @@ public abstract class AbstractFlinkService implements FlinkService {
                 conf, clusterId, (c, e) -> new StandaloneClientHAServices(restServerAddress));
     }
 
-    protected JarRunResponseBody runJar(
+    private JarRunResponseBody runJar(
             JobSpec job,
             JobID jobID,
             JarUploadResponseBody response,
@@ -674,7 +668,7 @@ public abstract class AbstractFlinkService implements FlinkService {
         }
     }
 
-    protected JarUploadResponseBody uploadJar(
+    private JarUploadResponseBody uploadJar(
             ObjectMeta objectMeta, FlinkSessionJobSpec spec, Configuration conf) throws Exception {
         String targetDir = artifactManager.generateJarDir(objectMeta, spec);
         File jarFile = artifactManager.fetch(spec.getJob().getJarURI(), conf, targetDir);
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
@@ -165,7 +165,8 @@ public abstract class AbstractFlinkService implements FlinkService {
         if (requireHaMetadata) {
             validateHaMetadataExists(conf);
         }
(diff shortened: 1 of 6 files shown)


morhidi:
@mbalassi can you kick off the workflow on this pls?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
```diff
@@ -165,7 +165,8 @@ public void submitApplicationCluster(
         if (requireHaMetadata) {
             validateHaMetadataExists(conf);
         }
-        deployApplicationCluster(jobSpec, conf);
+
+        deployApplicationCluster(jobSpec, removeOperatorConfigs(conf));
```
gyfora, line 169: I think you might have forgotten adding it for session cluster deployments.

gyfora (changes requested):

gyfora:
Btw in this outstanding PR (https://github.com/apache/flink-kubernetes-operator/pull/371/files) I have added some convenient utilities to test the configuration that was submitted to the cluster. Maybe you could simply rebase on it to leverage it and we can make sure to merge that first.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
HuangZhenQiu: Do you mean submitSessionCluster in NativeFlinkService? kubernetesClusterDescriptor.deploySessionCluster takes cluster specification as input. It already helped to filter unnecessary configs?

gyfora: Yes I think this needs to be added to the `submitSessionCluster` implementations for both the native and standalone service.

A proper test would actually validate all deployment modes and make sure the config is filtered :
  Native:
   - Application
   - Session
  Standalone
   - Application
   - Session
  
  + SessionJob

HuangZhenQiu: Thanks for the suggestions. Test cases are added in the latest revision.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
```diff
@@ -182,7 +184,13 @@ public JobID submitJobToSessionCluster(
...
         // we generate jobID in advance to help deduplicate job submission.
         var jobID = FlinkUtils.generateSessionJobFixedJobID(meta);
-        runJar(spec.getJob(), jobID, uploadJar(meta, spec, conf), conf, savepoint);
+        Configuration runtimeConfig = removeOperatorConfigs(conf);
+        runJar(
+                spec.getJob(),
+                jobID,
+                uploadJar(meta, spec, runtimeConfig),
+                runtimeConfig,
+                savepoint);
```
gyfora, line 193: I looked at the code and I think for sessionJobs we should not remove any configs. When we submit the jar through the rest api the configuration is not passed. 

Removing operator configs here can actually break some artifact fetcher mechanisms within the operator here.

We don't have very good test coverage it seems that would catch this problem. I will open some tickets

gyfora (changes requested):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
HuangZhenQiu: Thanks for the suggestion. Revised acccordingly.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
```diff
@@ -624,7 +626,7 @@ protected ClusterClient<String> getClusterClient(Configuration conf) throws Exce
                 conf, clusterId, (c, e) -> new StandaloneClientHAServices(restServerAddress));
     }
 
-    private JarRunResponseBody runJar(
+    protected JarRunResponseBody runJar(
```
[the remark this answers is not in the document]
HuangZhenQiu: Done

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
```diff
@@ -666,7 +668,7 @@ private JarRunResponseBody runJar(
         }
     }
 
-    private JarUploadResponseBody uploadJar(
+    protected JarUploadResponseBody uploadJar(
```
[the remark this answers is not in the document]
HuangZhenQiu: Done

PR #376, opened 2022-09-20
[FLINK-28755] Fix error when switching from stateless to savepoint upgrade mode

gaborgsomogyi:

Stateless to savepoint upgrade mode change blown up with exception. The reason was that stateless configuration doesn't require `state.savepoints.dir` config and `cancel` operation of the old job wanted to use that to make a savepoint. In this PR I've copied the necessary config from deploy config into observe config to make them available.

## Brief change log

* Copied `state.savepoints.dir` from deploy config to observe config
* Moved all upgrade mode tests to `ApplicationReconcilerUpgradeModeTest`
* Split a single upgrade test into 9 which represent all the transition combinations
* Enforce state pre-requisites in transitions which is described [here](https://nightlies.apache.org/flink/flink-kubernetes-operator-docs-main/docs/custom-resource/job-management/#stateful-and-stateless-application-upgrades)
* Minor beautifications here and there

## Verifying this change

Existing + additional unit tests

## Documentation


gaborgsomogyi:
cc @gyfora @mbalassi

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigManager.java
```diff
@@ -157,7 +158,14 @@ public Configuration getObserveConfig(FlinkDeployment deployment) {
             throw new RuntimeException(
                     "Cannot create observe config before first deployment, this indicates a bug.");
         }
-        return getConfig(deployment.getMetadata(), deployedSpec);
+        var conf = getConfig(deployment.getMetadata(), deployedSpec);
+        if (deployment.getSpec().getFlinkConfiguration() != null) {
+            var deployConfig = Configuration.fromMap(deployment.getSpec().getFlinkConfiguration());
+            deployConfig
+                    .getOptional(SAVEPOINT_DIRECTORY)
+                    .ifPresent(dir -> conf.set(SAVEPOINT_DIRECTORY, dir));
```
gyfora, line 166: Maybe we could factor this out to a static utility like:
```
applyConfigsFromCurrentSpec(ConfigOption... configs)
```

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/TestUtils.java
```diff
@@ -137,6 +137,36 @@ public static FlinkDeployment buildApplicationCluster(FlinkVersion version) {
         return buildApplicationCluster(TEST_DEPLOYMENT_NAME, TEST_NAMESPACE, version);
     }
 
+    public static FlinkDeployment buildApplicationCluster(
```
gyfora, line 140: If this is only used in one test class let's keep it there otherwise it gets a bit confusing

gyfora (changes requested):

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/TestUtils.java
gaborgsomogyi: Moved back to test.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigManager.java
gaborgsomogyi: Added.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigManager.java
```diff
@@ -177,6 +180,17 @@ private void addOperatorConfigsFromSpec(AbstractFlinkSpec spec, Configuration co
         }
     }
 
+    @SuppressWarnings("unchecked")
+    private void applyConfigsFromCurrentSpec(
```
gyfora, line 184: This method should internally call the `addOperatorConfigsFromSpec` so we keep things tidy and logically together.

gyfora (approved):
Looks good, one minor comment :)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigManager.java
gaborgsomogyi: Moved.

PR #377, opened 2022-09-21
[FLINK-28979] Add owner reference to flink deployment object

zezaeoh:

Add owner reference to `FlinkDeployment` object when create `Deployment`

## Brief change log

  - add owner reference to `FlinkDeployment` object when create `Deployment`

## Verifying this change

## Documentation

---
**NOTE:** 
With this change, we can see relation of `FlinkDeployment` with other resources in ArgoCD Web UI

**AS-IS:**
image

**TO-BE:**
image

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
@@ -18,6 +18,7 @@
 package org.apache.flink.kubernetes.operator.reconciler.deployment;
 
 import org.apache.flink.configuration.Configuration;
+import org.apache.flink.kubernetes.configuration.KubernetesConfigOptions;
 import org.apache.flink.kubernetes.operator.config.FlinkConfigManager;
 import org.apache.flink.kubernetes.operator.config.KubernetesOperatorConfigOptions;
 import org.apache.flink.kubernetes.operator.crd.AbstractFlinkResource;
@@ -433,13 +434,13 @@ public abstract class AbstractFlinkResourceReconciler<
 
     private void setOwnerReference(CR owner, Configuration deployConfig) {
         final Map<String, String> ownerReference =
-            Map.of(
-                    "apiVersion", owner.getApiVersion(),
-                    "kind", owner.getKind(),
-                    "name", owner.getMetadata().getName(),
-                    "uid", owner.getMetadata().getUid(),
-                    "blockOwnerDeletion", "false",
-                    "controller", "false");
+                Map.of(
+                        "apiVersion", owner.getApiVersion(),
+                        "kind", owner.getKind(),
+                        "name", owner.getMetadata().getName(),
+                        "uid", owner.getMetadata().getUid(),
+                        "blockOwnerDeletion", "false",
+                        "controller", "false");
         deployConfig.set(
                 KubernetesConfigOptions.JOB_MANAGER_OWNER_REFERENCE, List.of(ownerReference));
     }
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
@@ -100,6 +100,8 @@ public abstract class AbstractFlinkResourceReconciler<
             return;
         }
 
+        setOwnerReference(cr, deployConfig);
+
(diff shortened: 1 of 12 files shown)


morhidi:
> Hi! I understand that this improves how the job is displayed in ArgoCD but I don't see what real benefit it brings.
> 
> I wonder if on the other hand this might have some unintended negative consequences?
> 
> cc @morhidi @wangyang0918

Hi! @zezaeoh Thanks for your contribution, I share @gyfora's concern here, since we're finalizing the 1.2 release. But based on your screen shot this can greatly improve the user experience and visibility of operator managed resources in ArgoCD. If there are no obvious gotchas, I would give this a shot in 1.3 release.

zezaeoh:
Hi there, thanks for comments!

We can find out what an owner reference is, and what it can do in kubernetes in the following docs:
* https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/
* https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#OwnerReference

What i understand is there are no other effects when `blockOwnerDeletion` is set to `false`

gyfora:
What I don't understand is how this would interact with the CR deletion process. Currently when the user deletes a Flink CR the operator itself cleans up the dependent resources (deployment, HA configmaps).

What would happen if the owner reference is set?

gyfora:
Also since now we support standalone integration we should probably add the owner reference for both JM and TM deployments there. That should be part of the PR

+ some tests :)

zezaeoh:
It looks like e2e test failed because of connection time out from maven repo 😢 
Could you give me a one more try? @gyfora 

image

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
```diff
@@ -426,4 +431,17 @@ protected boolean flinkVersionChanged(SPEC oldSpec, SPEC newSpec) {
...
+                Map.of(
+                        "apiVersion", owner.getApiVersion(),
+                        "kind", owner.getKind(),
+                        "name", owner.getMetadata().getName(),
+                        "uid", owner.getMetadata().getUid(),
+                        "blockOwnerDeletion", "false",
+                        "controller", "false");
+        deployConfig.set(
+                KubernetesConfigOptions.JOB_MANAGER_OWNER_REFERENCE, List.of(ownerReference));
+    }
```
gyfora, line 446: Feels like the `FlinkConfigBuilder` might be a better place for this, then we can be sure that it's universally set in a single place.

gyfora: Or the `getDeployConfig` method of the `FlinkConfigManager` it might be easier to add there :)

zezaeoh: @gyfora 
I tried it but, the problem is with the apiVersion and kind

Is there a good way to add the apiVersion and kind of a CustomResource, if not in the `AbstractFlinkResourceReconciler`?

gyfora: We could probably change the information passed to getDeployConfig if necessary. Pass the whole CR + spec instead of meta + spec

gyfora: but I dont have a better idea

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
```diff
@@ -97,6 +100,8 @@ public final void reconcile(CR cr, Context<?> ctx) throws Exception {
             return;
         }
 
+        setOwnerReference(cr, deployConfig);
+
```
gyfora, line 104: I think we should call this method from inside the `deploy` method otherwise the ownerreference might not be set correctly during rollbacks, restarts etc

gyfora (changes requested):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
zezaeoh: You are right, it might be have too broad impact to call this method from there :)

PR #378, opened 2022-09-21
[FLINK-29383] Add PrinterColumn annotation for status fields

haoxins:

Add `PrinterColumn` annotation to the state fields, so that we can use `k get flinksession -o wide` to see session job list with status fields.

## Brief change log

- Add `@PrinterColumn` annotation to the state fields

## Verifying this change

I think the generated CRDs have covered the change 

## Documentation

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/status/JobStatus.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/status/JobStatus.java
@@ -19,6 +19,7 @@ package org.apache.flink.kubernetes.operator.crd.status;
 
 import org.apache.flink.annotation.Experimental;
 
+import io.fabric8.kubernetes.model.annotation.PrinterColumn;
 import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.Data;
@@ -38,6 +39,7 @@ public class JobStatus {
     private String jobId;
 
     /** Last observed state of the job. */
+    @PrinterColumn(name = "Job Status")
     private String state;
 
     /** Start time of the job. */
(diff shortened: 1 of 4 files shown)


gyfora:
What do you think @tweise @morhidi , to me this feels a little arbitrary why expose those specific fields. Might mislead the users.

gyfora:
I think if users care about certain specific status fields (which are anyways internal to the operator logic) they could simply modify the CRD that they install with these properties. Adding a doc guide instead of hardcoding it might be more helpful, not sure...

gyfora:
I agree with @tweise , I think if you want adding some documentation on how to this would be ok. Or we could simply close this ticket, up to you

morhidi:
I've no objection against this, it'd be great to see the lifecycle state there as well:)

haoxins:
When I have a lot of session jobs, this will be useful to figure out which one meets error.
This is what I'm facing when I submit hundreds of session jobs.

I can use `k get flinksessionjobs -o wide | grep xxx` to filter the jobs with problems,
and then debug what is wrong.

haoxins:
The other popular apps such as Argo CD also includes special status fields. such as
```
k get app -o wide -n argocd
NAME                   SYNC STATUS   HEALTH STATUS   REVISION
my-app-1             Synced               Healthy                   b3288befa7789047ff06c7e8631cbb65185ea3fe
```
I think the users of the Flink operator should also get familiar with these status fields' meanings.

haoxins:
1. the use case is to filter out the jobs with problems
2. and these fields are the most helpful ones

This is the reason why we need these additional printer columns

haoxins:
I just read another Jira issue about this https://issues.apache.org/jira/browse/FLINK-27893, it is worth discussing which fields should be exposed.

But the status fields are beneficial for deployment and debugging, and we can change the print fields without compatibility concerns.

I mean we can expose these now and change in the future base on the users feedback?

gyfora:
Since it doesn’t introduce new things to the status I think this is ok on a second thought :)

gyfora:
@tweise any objections against merging this based on the additional input?

tweise:
No objections. How would we accommodate another request that prefers some different columns though :)

gyfora:
> No objections. How would we accommodate another request that prefers some different columns though :)

We could then start to prioritize some fields and limit it to a fairly small number of total extra fields. But so far people did not seem to be too interested in these print columns :)

PR #382, opened 2022-09-24
[FLINK-29392] Trigger error when session job is lost without HA

gyfora:

Fix a bug where without HA a deployed SessionJob could be lost without any errors triggered.

## Brief change log

  - *Detect when a session job is completely lost*
  - *Trigger error event and set error status*
  - *Add tests*

## Verifying this change

SessionJobObserverTest have been extended to cover the expected cases for HA/non-HA configs.

## Documentation

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
@@ -19,11 +19,14 @@ package org.apache.flink.kubernetes.operator.observer;
 
 import org.apache.flink.configuration.Configuration;
 import org.apache.flink.kubernetes.operator.crd.AbstractFlinkResource;
+import org.apache.flink.kubernetes.operator.crd.FlinkDeployment;
+import org.apache.flink.kubernetes.operator.crd.FlinkSessionJob;
 import org.apache.flink.kubernetes.operator.crd.status.JobStatus;
 import org.apache.flink.kubernetes.operator.reconciler.ReconciliationUtils;
 import org.apache.flink.kubernetes.operator.service.FlinkService;
 import org.apache.flink.kubernetes.operator.utils.EventRecorder;
 import org.apache.flink.runtime.client.JobStatusMessage;
+import org.apache.flink.runtime.jobmanager.HighAvailabilityMode;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -37,7 +40,10 @@ import java.util.concurrent.TimeoutException;
 public abstract class JobStatusObserver<CTX> {
 
     private static final Logger LOG = LoggerFactory.getLogger(JobStatusObserver.class);
+
     private static final int MAX_ERROR_STRING_LENGTH = 512;
+    public static final String MISSING_SESSION_JOB_ERR = "Missing Session Job";
+
     private final FlinkService flinkService;
     private final EventRecorder eventRecorder;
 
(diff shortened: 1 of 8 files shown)


gyfora:
cc @morhidi @Aitozi @pvary

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
```diff
@@ -72,22 +81,92 @@ public boolean observe(
...
                     filterTargetJob(jobStatus, clusterJobStatuses);
+
             if (targetJobStatusMessage.isEmpty()) {
-                jobStatus.setState(org.apache.flink.api.common.JobStatus.RECONCILING.name());
+                LOG.warn("No matching jobs found on the cluster");
+                ifRunningMoveToReconciling(jobStatus, previousJobStatus);
+                // We could list the jobs but cannot find the one for this resource
+                if (resource instanceof FlinkDeployment) {
+                    // This should never happen for application clusters, there is something wrong
+                    setUnknownJobError((FlinkDeployment) resource);
```
pvary, line 94: Could this be normal if we concurrently removed a job as well with a complex configuration change?

gyfora: Not really, this means that there was a job on the cluster that we cannot match to the application deployment. this should never happen.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
```diff
@@ -72,22 +81,91 @@ public boolean observe(
...
                 updateJobStatus(resource, targetJobStatusMessage.get(), deployedConfig);
             }
             ReconciliationUtils.checkAndUpdateStableSpec(resource.getStatus());
             return true;
         } else {
+            LOG.debug("No jobs found on the cluster");
+            // No jobs found on the cluster, it is possible that the jobmanager is still starting up
             ifRunningMoveToReconciling(jobStatus, previousJobStatus);
+
+            if (resource instanceof FlinkSessionJob) {
```
gaborgsomogyi, line 109: Seems like this is not common logic, maybe this can be extracted from here.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
```diff
@@ -72,22 +81,91 @@ public boolean observe(
...
+            // There are jobs on the cluster, we filter the ones for this resource
             Optional<JobStatusMessage> targetJobStatusMessage =
                     filterTargetJob(jobStatus, clusterJobStatuses);
+
             if (targetJobStatusMessage.isEmpty()) {
-                jobStatus.setState(org.apache.flink.api.common.JobStatus.RECONCILING.name());
+                LOG.warn("No matching jobs found on the cluster");
+                ifRunningMoveToReconciling(jobStatus, previousJobStatus);
+                // We could list the jobs but cannot find the one for this resource
+                if (resource instanceof FlinkDeployment) {
```
gaborgsomogyi, line 92: Seems like this is not common logic, maybe this can be extracted from here.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
```diff
@@ -72,22 +81,91 @@ public boolean observe(
...
     }
 
+    /**
+     * When HA is disabled the session job will not recover on JM restarts. If the JM goes down /
+     * restarted the session job should be marked missing.
+     *
+     * @param sessionJob Flink session job.
+     * @param conf Flink config.
+     */
+    private void ifHaDisabledMarkSessionJobMissing(FlinkSessionJob sessionJob, Configuration conf) {
```
gaborgsomogyi, line 123: This is more like a taste question so we may or may not do it. I prefer `get/is` and `set` functions and build upper level functionality like
```
if (something) {
  react
}
```
More re-usable if somebody wants to implement `ifTheMoonIsGreenMarkSessionJobMissing` functionality.
But again, this is taste question so we can leave it like it if you don't agree.

gaborgsomogyi (commented):
Basically looks good. Left some comments, big +1 on adding some comments to edge case if conditions.

gyfora:
Addressed your comments @gaborgsomogyi please take a look

gaborgsomogyi (approved):
LGTM.

PR #383, opened 2022-09-24
[FLINK-29384] Bump snakeyaml version to 1.32

mbalassi:
Also bumped the jackson version as snakeyaml is only transitively pulled by jackson to keep the versions closer and more up to date.

### DIFF
--- a/flink-kubernetes-operator/pom.xml
+++ b/flink-kubernetes-operator/pom.xml
@@ -47,6 +47,20 @@ under the License.
             <groupId>io.fabric8</groupId>
             <artifactId>kubernetes-client</artifactId>
             <version>${fabric8.version}</version>
+            <!-- FLINK-29384, can be removed when fabric8 pulls snakeyaml 1.32+ -->
+            <exclusions>
+                <exclusion>
+                    <groupId>org.yaml</groupId>
+                    <artifactId>snakeyaml</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+
+        <!-- FLINK-29384, can be removed when fabric8 pulls snakeyaml 1.32+ -->
+        <dependency>
+            <groupId>org.yaml</groupId>
+            <artifactId>snakeyaml</artifactId>
+            <version>${snakeyaml.version}</version>
         </dependency>
 
         <dependency>
--- a/flink-kubernetes-shaded/pom.xml
+++ b/flink-kubernetes-shaded/pom.xml
@@ -54,6 +54,19 @@ under the License.
                 </exclusion>
             </exclusions>
         </dependency>
+        <!-- FLINK-29384, Since the transitive snakeyaml that we are filtering out from
+            flink-kubernetes was relocated we have to bundle a relocated new version -->
+        <dependency>
+            <groupId>org.yaml</groupId>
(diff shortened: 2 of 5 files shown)


gyfora (commented):
I think we need to manually exclude this from the shaded jar where we import the Flink-kubernetes fatjar .

that won’t expose it as a dependency so we can’t exclude it but the classes might still be there ?

flink-kubernetes-shaded/pom.xml
```diff
@@ -88,6 +103,7 @@ under the License.
                                         <exclude>META-INF/DEPENDENCIES</exclude>
                                         <exclude>META-INF/LICENSE</exclude>
                                         <exclude>META-INF/MANIFEST.MF</exclude>
+                                        <exclude>org/apache/flink/kubernetes/shaded/org/yaml/snakeyaml/**</exclude>
```
gyfora, line 106: Wouldn't this line be the only change we need for `flink-kubernetes-shaded`?

flink-kubernetes-operator/pom.xml
```diff
@@ -47,6 +47,20 @@ under the License.
...
+            <exclusions>
+                <exclusion>
+                    <groupId>org.yaml</groupId>
+                    <artifactId>snakeyaml</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+
+        <!-- FLINK-29384, can be removed when fabric8 pulls snakeyaml 1.32+ -->
+        <dependency>
```
wangyang0918, line 60: Could you please share me why we include the dependency explicitly and not use the `dependencyManagement`?

flink-kubernetes-shaded/pom.xml
wangyang0918: I am not sure whether it is possible to update the bundled `NOTICE` file to also remove the `org.yaml:snakeyaml:1.27`.

mbalassi: @wangyang0918 
The NOTICE has this and only this relevant entry:
```
  - SnakeYAML (https://bitbucket.org/snakeyaml/snakeyaml) org.yaml:snakeyaml:bundle:1.32
    License: Apache License, Version 2.0  (http://www.apache.org/licenses/LICENSE-2.0.txt)
```

@gyfora Because of the relocation, I added comments to make it more clear.

flink-kubernetes-operator/pom.xml
mbalassi: I tried the dependency management route in `flink-kubernetes-parent`, it did not seems to work. I see your point, and can give it a try if you prefer. Given that it is hopefully just a temporary fix I would suggest to leave it as is.

flink-kubernetes-shaded/pom.xml
wangyang0918: What I mean is we still have `org.yaml:snakeyaml:1.27` in the NOTICE file of `flink-kubernetes-shaded-1.2-SNAPSHOT.jar` even after this PR. Right?

gyfora: Yes that is correct, @mbalassi didnt find a straightforward way to do this but this should have no release implications as we correctly indicate the new version in the other notice file.

PR #384, opened 2022-09-25
[FLINK-29159] Harden initial deployment logic

gyfora:

Fixes a small flaw in the initialDeployment logic that made it useless after the job was actually submitted. This also simplifies code in other places.

## Brief change log

  - *Rename ReconStatus#isFirstDeployment to isBeforeFirstDeployment to better reflect logic*
  - *Fix reconciliation metadata isFirstDeployment computation*
  - *Add some tests*

## Verifying this change

Partly covered by existing tests + new tests added for the fixed logic

## Documentation

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/status/CommonStatus.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/status/CommonStatus.java
@@ -54,7 +54,7 @@ public abstract class CommonStatus<SPEC extends AbstractFlinkSpec> {
     public ResourceLifecycleState getLifecycleState() {
         var reconciliationStatus = getReconciliationStatus();
 
-        if (reconciliationStatus.isFirstDeployment()) {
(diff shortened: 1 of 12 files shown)


gyfora:
cc @morhidi

tweise (approved):

morhidi:
+1 Thanks, Gyula. I have the feeling though that on long term we should rely on a more consistent state-machine model, instead of introducing new states/substates in utility methods. WDYT?

gyfora:
On a very high level I agree with you @morhidi and the `ResourceLifecycleState` helper serves a related purpose. Unfortunately I feel that there are so many corner cases and slightly different deployment scenarios that a purely state machine based approach would have waaaay too many states and many of these would be hard to name or abstract properly. Even in the best case I think we would end up with a mix of a better state machine represenation + extra status info to capture some cornercases.

While the status at the moment contains a quite a lot of different types of information, at least it is somewhat normalized.

It would be great to see what kind of state machine you have in mind, maybe there is a nice way to simplify this.

PR #385, opened 2022-09-25
[FLINK-29109] Avoid checkpoint path conflicts (Flink < 1.16)

tweise:

Automatically set jobId for Flink version < 1.16 to avoid conflicts in checkpoint path (and other jobId based effects).

## Brief change log

* Set a jobId when no jobID is defined by user and Flink version is < 1.16.
* Rotate jobId when upgrade mode is stateless

## Verifying this change

Manual verification:
* No jobId set in Flink config
  * Set upgrade mode to stateless. Trigger CR changes and check that every redeploy uses a new ID.
  * Set upgrade mode to last_state. Trigger CR changes and check that ID does not change

## Documentation

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
@@ -17,13 +17,16 @@
 
 package org.apache.flink.kubernetes.operator.reconciler.deployment;
 
+import org.apache.flink.api.common.JobID;
 import org.apache.flink.configuration.Configuration;
 import org.apache.flink.configuration.HighAvailabilityOptions;
+import org.apache.flink.configuration.PipelineOptionsInternal;
 import org.apache.flink.kubernetes.configuration.KubernetesConfigOptions;
 import org.apache.flink.kubernetes.operator.config.FlinkConfigManager;
 import org.apache.flink.kubernetes.operator.config.KubernetesOperatorConfigOptions;
 import org.apache.flink.kubernetes.operator.crd.FlinkDeployment;
 import org.apache.flink.kubernetes.operator.crd.spec.FlinkDeploymentSpec;
+import org.apache.flink.kubernetes.operator.crd.spec.FlinkVersion;
 import org.apache.flink.kubernetes.operator.crd.spec.UpgradeMode;
 import org.apache.flink.kubernetes.operator.crd.status.FlinkDeploymentStatus;
 import org.apache.flink.kubernetes.operator.crd.status.JobManagerDeploymentStatus;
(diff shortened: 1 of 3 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/status/FlinkDeploymentStatus.java
```diff
@@ -53,4 +53,7 @@ public class FlinkDeploymentStatus extends CommonStatus<FlinkDeploymentSpec> {
 
     /** Information about the TaskManagers for the scale subresource. */
     private TaskManagerInfo taskManager;
+
+    /** The jobId, if it was set by the operator. */
+    private String overrideJobId;
```
gyfora, line 58: I think we should simply set it into the jobStatus.jobId and avoid introducing an extra field for it. 

As long as the job is running we should assume that the jobStatus reflects the correct jobid

gyfora: All the current logic uses the jobStatus.jobId for savepointing, cancelling, observing etc, it would avoid a lot of complications :)

tweise: Reusing jobStatus.jobId is not perfect as we are using the same field for two different purposes. But it works pretty well and is better than adding another field to status for something that is going to be phased out when >=1.16 becomes the minimum version. PTAL

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
```diff
@@ -195,6 +201,34 @@ protected void deploy(
...
+        }
+
+        final JobID jobId;
+        // generate jobId initially or rotate on every deployment when mode is stateless
+        if (status.getJobStatus().getJobId() == null
+                || spec.getJob().getUpgradeMode() == UpgradeMode.STATELESS) {
+            if (deployConfig.get(PipelineOptionsInternal.PIPELINE_FIXED_JOB_ID) != null) {
+                // user managed, don't touch
+                return;
+            }
```
gyfora, line 219: We could move this before the `if` clause to simplify

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
```diff
@@ -195,6 +201,34 @@ protected void deploy(
...
 
+    private void setJobIdIfNecessary(
+            FlinkDeploymentSpec spec, FlinkDeploymentStatus status, Configuration deployConfig) {
+        // https://issues.apache.org/jira/browse/FLINK-19358
+        // https://issues.apache.org/jira/browse/FLINK-29109
+        if (spec.getFlinkVersion().isNewerVersionThan(FlinkVersion.v1_15)) {
+            return;
+        }
+
+        final JobID jobId;
```
gyfora, line 212: We could use the String representation and set this directly from `status.getJobStatus().getJobId()` to get rid of the `else` branch later.

gyfora (approved):
Added some minor comments, otherwise looks good!

gyfora (approved):

PR #386, opened 2022-09-26
[FLINK-29376] Set SHUTDOWN_ON_APPLICATION_FINISH to false only on 1.15 and above

gaborgsomogyi:

If the user incorrectly sets a Flink version lower than 1.15 when using 1.15 or above:
```
image: flink:1.15
  flinkVersion: v1_14
```
then a spec change ends-up in exception and restart loop. In this PR we set `SHUTDOWN_ON_APPLICATION_FINISH` to `false` when Flink version is less than 1.15.

## Brief change log

* Set `SHUTDOWN_ON_APPLICATION_FINISH` to false only on 1.15 and above
* Added unit test for all available Flink versions

## Verifying this change

* New unit test
* Manually
  * Deploy app where
  `image: flink:1.15 flinkVersion: v1_14`
  * Wait for RUNNING state
  * Change Flink config
  * Double check that Flink is re-deployed and in RUNNING state again

## Documentation


gaborgsomogyi:
cc @gyfora

PR #394, opened 2022-10-06
[FLINK-29394] Add observe Flink job health

gaborgsomogyi:

Flink has its own restart strategies which are working fine. But there are certain circumstances when Flink can stuck in a restart loop. A good example is when checkpointing is activated and the restart strategy has not been configured, then the fixed-delay strategy is used with `Integer.MAX_VALUE` restart attempts. When the JobManager (JM from now on) is able to solve its temporary issue, it can be that a permanent issue appears on TaskManager (TM from now on) side. A good example is that TM has a memory leak and just crashes. Such case the Flink job requires a restart from the outside, which can be done by the Flink k8s operator.

In this PR I've added job health check feature. Please be aware that the implementation is simple and has the following caveats:
* Restart count is watched in a normal non-sliding window
* When the last valid observed health info timestamp is outside of the watched window then the algorithm assumes even restart count distribution

## Brief change log

* Added `kubernetes.operator.cluster.health-check.enabled` config (default: false)
* Added `kubernetes.operator.cluster.health-check.restarts.window` config (default: 2 minutes)
* Added `kubernetes.operator.cluster.health-check.restarts.threshold` config (default: 64)
* Added `ClusterHealthInfo` field to `clusterInfo` as string field (internal structure may change suddenly at this stage)
* Added `ClusterHealthObserver` which is responsible to fetch job health information from the submitted job
* Added `ClusterHealthEvaluator` which is responsible to decide whether the job is healthy or not
* Added job restart functionality (with the same spec) when job considered unhealthy
* Added several unit tests
* Some simplifications/refactors

## Verifying this change

* Existing unit tests
* Additional unit tests
* Manually with my newly created [chaos monkey](https://github.com/gaborgsomogyi/flink-chaos-monkey-java-job) job
  * Job submitted with health check
  * Executed in the TM shell: `touch /tmp/throwExceptionInUDF`
  * Waited for job recovery

  - The public API, i.e., is any changes to the `CustomResourceDescriptors`: yes, new configs added

## Documentation

  - If yes, how is the feature documented? newly added configs + in docs documented

### DIFF
--- a/docs/content/docs/custom-resource/reference.md
+++ b/docs/content/docs/custom-resource/reference.md
@@ -197,7 +197,7 @@ This page serves as a full reference for FlinkDeployment custom resource definit
 | ----------| ---- | ---- |
 | jobStatus | org.apache.flink.kubernetes.operator.crd.status.JobStatus | Last observed status of the Flink job on Application/Session cluster. |
 | error | java.lang.String | Error information about the FlinkDeployment/FlinkSessionJob. |
-| clusterInfo | java.util.Map<java.lang.String,java.lang.String> | Config information from running clusters. |
+| clusterInfo | java.util.Map<java.lang.String,java.lang.String> | Information from running clusters. |
 | jobManagerDeploymentStatus | org.apache.flink.kubernetes.operator.crd.status.JobManagerDeploymentStatus | Last observed status of the JobManager deployment. |
 | reconciliationStatus | org.apache.flink.kubernetes.operator.crd.status.FlinkDeploymentReconciliationStatus | Status of the last reconcile operation. |
 | taskManager | org.apache.flink.kubernetes.operator.crd.status.TaskManagerInfo | Information about the TaskManagers for the scale subresource. |
--- a/docs/layouts/shortcodes/generated/kubernetes_operator_config_configuration.html
+++ b/docs/layouts/shortcodes/generated/kubernetes_operator_config_configuration.html
@@ -9,6 +9,24 @@
     </thead>
     <tbody>
         <tr>
+            <td><h5>kubernetes.operator.cluster.health-check.enabled</h5></td>
+            <td style="word-wrap: break-word;">false</td>
+            <td>Boolean</td>
+            <td>Whether to enable health check for clusters.</td>
+        </tr>
+        <tr>
+            <td><h5>kubernetes.operator.cluster.health-check.restarts.threshold</h5></td>
+            <td style="word-wrap: break-word;">64</td>
+            <td>Integer</td>
+            <td>The threshold which is checked against job restart count within a configured window. If the restart count is reaching the threshold then full cluster restart is initiated.</td>
+        </tr>
+        <tr>
+            <td><h5>kubernetes.operator.cluster.health-check.restarts.window</h5></td>
+            <td style="word-wrap: break-word;">2 min</td>
+            <td>Duration</td>
+            <td>The duration of the time window where job restart count measured.</td>
+        </tr>
+        <tr>
             <td><h5>kubernetes.operator.config.cache.size</h5></td>
             <td style="word-wrap: break-word;">1000</td>
             <td>Integer</td>
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
@@ -315,4 +315,28 @@ public class KubernetesOperatorConfigOptions {
                     .intType()
                     .defaultValue(8085)
                     .withDescription("The port the health probe will use to expose the status.");
+
+    @Documentation.Section(SECTION_DYNAMIC)
+    public static final ConfigOption<Boolean> OPERATOR_CLUSTER_HEALTH_CHECK_ENABLED =
+            operatorConfig("cluster.health-check.enabled")
+                    .booleanType()
+                    .defaultValue(false)
+                    .withDescription("Whether to enable health check for clusters.");
+
+    @Documentation.Section(SECTION_DYNAMIC)
+    public static final ConfigOption<Duration> OPERATOR_CLUSTER_HEALTH_CHECK_RESTARTS_WINDOW =
+            operatorConfig("cluster.health-check.restarts.window")
+                    .durationType()
+                    .defaultValue(Duration.ofMinutes(2))
+                    .withDescription(
+                            "The duration of the time window where job restart count measured.");
+
+    @Documentation.Section(SECTION_DYNAMIC)
+    public static final ConfigOption<Integer> OPERATOR_CLUSTER_HEALTH_CHECK_RESTARTS_THRESHOLD =
+            operatorConfig("cluster.health-check.restarts.threshold")
+                    .intType()
+                    .defaultValue(64)
+                    .withDescription(
+                            "The threshold which is checked against job restart count within a configured window. "
+                                    + "If the restart count is reaching the threshold then full cluster restart is initiated.");
 }
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/health/ClusterHealthInfo.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/health/ClusterHealthInfo.java
@@ -1,0 +1,76 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.kubernetes.operator.health;
+
+import org.apache.flink.annotation.Experimental;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+import java.time.Clock;
+
+/** Represents information about job health. */
+@Experimental
+@Data
+@AllArgsConstructor
+@NoArgsConstructor
+public class ClusterHealthInfo {
+    /** Millisecond timestamp of the last observed health information. */
+    private long timeStamp;
+
+    /** Number of restarts. */
+    private int numRestarts;
+
+    /** Calculated field whether the cluster is healthy or not. */
+    private boolean healthy;
+
+    public static ClusterHealthInfo of(int numRestarts) {
+        return of(Clock.systemDefaultZone(), numRestarts);
+    }
+
+    public static ClusterHealthInfo of(Clock clock, int numRestarts) {
+        return new ClusterHealthInfo(clock.millis(), numRestarts, true);
+    }
+
+    public static boolean isValid(ClusterHealthInfo clusterHealthInfo) {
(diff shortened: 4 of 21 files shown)


gaborgsomogyi:
cc @gyfora @morhidi @mbalassi

docs/content/docs/custom-resource/reference.md
```diff
@@ -225,6 +225,16 @@ This page serves as a full reference for FlinkDeployment custom resource definit
 | error | java.lang.String | Error information about the FlinkDeployment/FlinkSessionJob. |
 | reconciliationStatus | org.apache.flink.kubernetes.operator.crd.status.FlinkSessionJobReconciliationStatus | Status of the last reconcile operation. |
 
+### JobHealthInfo
+**Class**: org.apache.flink.kubernetes.operator.crd.status.JobHealthInfo
```
gyfora, line 229: We should move this class to another package, so that the reference generator does not pick it up. This object is not directly part of the status (only in a serialized form)

docs/content/docs/custom-resource/reference.md
```diff
@@ -251,6 +261,7 @@ This page serves as a full reference for FlinkDeployment custom resource definit
 | startTime | java.lang.String | Start time of the job. |
 | updateTime | java.lang.String | Update time of the job. |
 | savepointInfo | org.apache.flink.kubernetes.operator.crd.status.SavepointInfo | Information about pending and last savepoint for the job. |
+| jobHealthInfo | java.lang.String | Information about job health. |
```
gyfora, line 264: I have been thinking a bit about this, we currently have a `clusterInfo` field in the `FlinkDeploymentStatus` we could simply use that map to store this information. 

I think originally it was designed with some flexibility in mind so we can store different metadata. cc @morhidi

docs/layouts/shortcodes/generated/kubernetes_operator_config_configuration.html
```diff
@@ -80,6 +80,24 @@
...
+            <td>Whether to enable health check for jobs.</td>
+        </tr>
+        <tr>
+            <td><h5>kubernetes.operator.job.restart-check.duration-window</h5></td>
+            <td style="word-wrap: break-word;">2 min</td>
+            <td>Duration</td>
+            <td>The duration of the time window where job restart count measured.</td>
+        </tr>
+        <tr>
+            <td><h5>kubernetes.operator.job.restart-check.threshold</h5></td>
```
gyfora, line 96: There is some inconsistency in the naming here, `health-check` vs `restart-check`. 
Maybe it could be:
```
kubernetes.operator.job.health-check.enabled
kubernetes.operator.job.health-check.window
kubernetes.operator.job.health-check.restarts.threshold / kubernetes.operator.job.health-check.num- restarts.threshold

```

gyfora (changes requested):

docs/content/docs/custom-resource/reference.md
gaborgsomogyi: Moved.

docs/layouts/shortcodes/generated/kubernetes_operator_config_configuration.html
gaborgsomogyi: Good point, renamed to:
```
kubernetes.operator.job.health-check.enabled
kubernetes.operator.job.health-check.restarts.window
kubernetes.operator.job.health-check.restarts.threshold
```

docs/content/docs/custom-resource/reference.md
gaborgsomogyi: That's a key-value store which is designed to store config info:
```
    /** Config information from running clusters. */
    private Map<String, String> clusterInfo = new HashMap<>();
```
For example one key is `DashboardConfiguration.FIELD_NAME_FLINK_VERSION`.
I can put it there but at the first glance it's not intended to store structures like this.
Are we sure that we want to do that?

gaborgsomogyi: OK, moved the health info to the `clusterInfo` map.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/AbstractDeploymentObserver.java
```diff
@@ -116,13 +116,10 @@ public void observe(FlinkDeployment flinkApp, Context<?> context) {
...
 
     private void observeClusterInfo(FlinkDeployment flinkApp, Configuration configuration) {
-        if (!flinkApp.getStatus().getClusterInfo().isEmpty()) {
-            return;
-        }
         try {
             Map<String, String> clusterInfo = flinkService.getClusterInfo(configuration);
-            flinkApp.getStatus().setClusterInfo(clusterInfo);
-            logger.debug("ClusterInfo: {}", clusterInfo);
+            flinkApp.getStatus().getClusterInfo().putAll(clusterInfo);
```
gaborgsomogyi, line 121: Important change that now we merge fetched info in order to keep previously set health info or other upcoming things.

gaborgsomogyi:
From high level perspective now we fill `clusterInfo` from multiple places but if we want to use it as custom and flexible storage then we need to go this direction instead of having a single `flinkService.getClusterInfo`.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/AbstractDeploymentObserver.java
```diff
@@ -139,8 +135,6 @@ protected void observeJmDeployment(
             return;
         }
 
-        flinkApp.getStatus().setClusterInfo(new HashMap<>());
```
gaborgsomogyi, line 142: This was blocking the reconciler to get valid health info in case of the cluster was in restarting state.

gaborgsomogyi:
Just renamed all params from `...job...` to `...cluster...`

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/JobHealthChecker.java
```diff
@@ -0,0 +1,130 @@
...
+    @VisibleForTesting @Nullable JobHealthInfo lastValidJobHealthInfo;
+
+    public JobHealthChecker(Clock clock) {
+        this.clock = clock;
+    }
+
+    @VisibleForTesting
+    void setLastValidJobHealthInfo(JobHealthInfo lastValidJobHealthInfo) {
+        LOG.debug("Setting last valid health check info");
+        this.lastValidJobHealthInfo = lastValidJobHealthInfo;
```
gyfora, line 50: If the isHealthy logic depends only on the last health info, it might be best to compute it at the time when you observe the new one but before you put it in the status. At that point you have the previous and new health info

gyfora: This way you would also put the isHealthy info in the status and the reconciler would not need to compute anything

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
```diff
@@ -315,4 +315,28 @@ public static String operatorConfigKey(String key) {
                     .intType()
                     .defaultValue(8085)
                     .withDescription("The port the health probe will use to expose the status.");
+
+    @Documentation.Section(SECTION_SYSTEM)
```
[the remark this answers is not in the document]
gaborgsomogyi: Fixed.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/JobHealthChecker.java
gaborgsomogyi: Refactored fully.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
```diff
@@ -272,6 +286,46 @@ private void recoverJmDeployment(
...
+            }
+        }
+
+        boolean result = false;
+
+        if (jobHealthChecker != null) {
+            final String clusterInfoKey = JobHealthInfo.class.getSimpleName();
+            if (deployment.getStatus().getClusterInfo().containsKey(clusterInfoKey)) {
+                LOG.debug("Cluster info contains job health info");
+                if (!jobHealthChecker.isJobHealthy(
```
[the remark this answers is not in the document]
gaborgsomogyi: Refactored fully.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
```diff
@@ -272,6 +286,46 @@ private void recoverJmDeployment(
...
+            FlinkDeployment deployment, Configuration observeConfig) {
+        if (observeConfig.getBoolean(OPERATOR_CLUSTER_HEALTH_CHECK_ENABLED)) {
+            if (jobHealthChecker == null) {
+                jobHealthChecker = new JobHealthChecker(Clock.systemDefaultZone());
+            }
+        } else {
+            if (jobHealthChecker != null) {
+                jobHealthChecker = null;
+            }
+        }
```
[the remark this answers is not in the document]
gaborgsomogyi: Refactored fully.

gaborgsomogyi:
Updated PR description to reflect latest code.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/health/ClusterHealthInfo.java
```diff
@@ -0,0 +1,78 @@
...
+
+import java.time.Clock;
+
+/** Represents information about job health. */
+@Experimental
+@Data
+@AllArgsConstructor
+@NoArgsConstructor
+@ToString
+public class ClusterHealthInfo {
```
gyfora, line 37:
```
@Data
@NoArgsConstructor
```
Should be enough, the rest comes from Data

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/ApplicationObserver.java
```diff
@@ -32,15 +33,21 @@
...
 
+import static org.apache.flink.kubernetes.operator.config.KubernetesOperatorConfigOptions.OPERATOR_CLUSTER_HEALTH_CHECK_ENABLED;
+
 /** The observer of {@link org.apache.flink.kubernetes.operator.config.Mode#APPLICATION} cluster. */
 public class ApplicationObserver extends AbstractDeploymentObserver {
 
     private final SavepointObserver<FlinkDeployment, FlinkDeploymentStatus> savepointObserver;
     private final JobStatusObserver<FlinkDeployment, ApplicationObserverContext> jobStatusObserver;
 
+    @Nullable private ClusterHealthObserver clusterHealthObserver;
```
gyfora, line 49: Not `@Nullable`

gyfora (changes requested):
Added some minor comments otherwise looks good.

I think we are still missing a controller test that would validate the entire restart logic in case of an unhealthy cluster 

 + some documentation would be great. Probably a short section on the Job Management page (https://nightlies.apache.org/flink/flink-kubernetes-operator-docs-main/docs/custom-resource/job-management/) is enough

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/ApplicationObserver.java
gaborgsomogyi: Fixed.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/health/ClusterHealthInfo.java
gaborgsomogyi: `@ToString` dropped, the printout is the same.
`@AllArgsConstructor` is must because otherwise `new ClusterHealthInfo(clock.millis(), numRestarts, true);` fails to compile

gaborgsomogyi:
Added doc + controller test
Screen Shot 2022-10-13 at 13 40 57

PR #395, opened 2022-10-07
Add CRD Documentation about the Flink Deployment Modes

usamj:
* Adding documentation about Native and Standalone deployment modes

### DIFF
--- a/docs/content/docs/custom-resource/overview.md
+++ b/docs/content/docs/custom-resource/overview.md
@@ -149,6 +149,30 @@ For Session clusters the operator only provides very basic management and monito
  - Monitor overall cluster health
  - Stop / Delete Session cluster
 
+### Cluster Deployment Modes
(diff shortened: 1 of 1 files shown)


docs/content/docs/custom-resource/overview.md
```diff
@@ -149,6 +149,26 @@ For Session clusters the operator only provides very basic management and monito
  - Monitor overall cluster health
  - Stop / Delete Session cluster
 
+### Cluster Deployment Modes
```
dannycranmer, line 152: This reads ok, however as a consumer of this documentation I might want to know when to pick Native vs Standalone or vice versa. Can we add some basic use-case examples where each mode is useful for?

dannycranmer (changes requested):

gyfora (approved):
I think this looks very nice for this doc section and gives a good idea to the user.

Maybe we could later add a more detailed architecture comparison to some other pages where we show what it means in practice -> that we create Deployments for TMs and JMs from the operator etc.

PR #400, opened 2022-10-12
[FLINK-29401] Refactor observer structure

gyfora:

The goal is to create a shared observer hierarchy for all Flink resources and factor out the common bits to base classes in a similar way how it is already done for Reconcilers.

## Brief change log

 - Create AbstractFlinkResourceObserver for shared logic
 - Other refactorings in sessionjob and application observer contexts

## Verifying this change

This change is already covered by existing observer/controller tests.

## Documentation


gyfora:
cc @morhidi @gaborgsomogyi

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/AbstractFlinkResourceObserver.java
```diff
@@ -0,0 +1,129 @@
...
+            return false;
+        }
+
+        if (reconciliationStatus.getState() == ReconciliationState.ROLLING_BACK) {
+            logger.debug("Skipping observe during rollback operation");
+            return false;
+        }
+
+        // We are in the middle or possibly right after an upgrade
+        if (reconciliationStatus.getState() == ReconciliationState.UPGRADING) {
```
gaborgsomogyi, line 94: Nit: `reconciliationStatus.getState()` can be put to local var.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/AbstractFlinkResourceObserver.java
```diff
@@ -0,0 +1,129 @@
...
+     * Checks a resource that is currently in the UPGRADING state whether it was already deployed
+     * but we simply miss the status information. After comparing the target resource generation
+     * with the one from the possible deployment if they match we update the status to the already
+     * DEPLOYED state.
+     *
+     * @param resource Flink resource.
+     * @param resourceContext Context for resource.
+     * @param observerContext Context for observer.
+     */
+    protected abstract void checkIfAlreadyUpgraded(
```
gaborgsomogyi, line 127: Maybe this should be called `updateStatusToDeployedIfAlreadyUpgraded` or something more meaningful. W/o checking the function description I can't really tell what is the outcome of this.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
```diff
@@ -132,9 +133,9 @@ private void ifRunningMoveToReconciling(JobStatus jobStatus, String previousJobS
     /**
      * Callback when list jobs timeout.
      *
-     * @param ctx Observe context.
+     * @param observerContext Observe context.
      */
-    protected abstract void onTimeout(CTX ctx);
+    protected abstract void onTimeout(R resource, Context<?> resourceContext, CTX observerContext);
```
gaborgsomogyi, line 138: Nit: `resource` is not documented.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/ObserverContext.java
```diff
@@ -15,9 +15,12 @@
  * limitations under the License.
  */
 
-package org.apache.flink.kubernetes.operator.observer.context;
+package org.apache.flink.kubernetes.operator.observer;
 
-/** An empty observer context. */
-public class VoidObserverContext {
```
gaborgsomogyi, line 21: Just for my own understanding why we've had `VoidObserverContext`? Looks like it doesn't do any meaningful task.

The other thing what I don't understand why we introduce another layer to pass `DeployedConfig`? Having tons of function params would justify the need but seems like it's not the case.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/ApplicationObserver.java
```diff
@@ -48,66 +48,72 @@ public ApplicationObserver(
         super(flinkService, configManager, eventRecorder);
         this.savepointObserver =
                 new SavepointObserver<>(flinkService, configManager, eventRecorder);
-        this.jobStatusObserver =
```
gaborgsomogyi, line 51: Big +1 to collapse this huge inline implementation.

gaborgsomogyi (commented):
Looks basically good, some minors left.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/ObserverContext.java
morhidi: This is weird indeed.

gyfora: It's not that weird to have some context implementations empty in cases where no extra contextual information is necessary. But in any case this is removed now :)

PR #403, opened 2022-10-13
[FLINK-29633] Pass fromSavepoint argument

sap1ens:

Passing `initialSavepointPath` from the JobSpec as a `fromSavepoint` argument in the standalone mode.

## Brief change log

- Updated `StandaloneKubernetesJobManagerParameters` to include savepoint path.
- Updated `CmdStandaloneJobManagerDecorator` to pass that as `fromSavepoint`.
- Also dropping the boolean parameter for the `allowNonRestoredState` flag: Flink may not parse it properly.

## Verifying this change

- Modified `CmdStandaloneJobManagerDecoratorTest` to confirm the argument is passed properly.
- Confirmed it's working after deploying to an internal staging environment.

## Documentation

### DIFF
--- a/flink-kubernetes-standalone/src/main/java/org/apache/flink/kubernetes/operator/kubeclient/decorators/CmdStandaloneJobManagerDecorator.java
+++ b/flink-kubernetes-standalone/src/main/java/org/apache/flink/kubernetes/operator/kubeclient/decorators/CmdStandaloneJobManagerDecorator.java
@@ -80,9 +80,8 @@ public class CmdStandaloneJobManagerDecorator extends AbstractKubernetesStepDeco
         }
 
         Boolean allowNonRestoredState = kubernetesJobManagerParameters.getAllowNonRestoredState();
-        if (allowNonRestoredState != null) {
+        if (allowNonRestoredState != null && allowNonRestoredState) {
             args.add("--allowNonRestoredState");
-            args.add(allowNonRestoredState.toString());
         }
 
         String savepointPath = kubernetesJobManagerParameters.getSavepointPath();
--- a/flink-kubernetes-standalone/src/main/java/org/apache/flink/kubernetes/operator/kubeclient/decorators/CmdStandaloneJobManagerDecorator.java
+++ b/flink-kubernetes-standalone/src/main/java/org/apache/flink/kubernetes/operator/kubeclient/decorators/CmdStandaloneJobManagerDecorator.java
@@ -85,6 +85,12 @@ public class CmdStandaloneJobManagerDecorator extends AbstractKubernetesStepDeco
             args.add(allowNonRestoredState.toString());
         }
 
+        String savepointPath = kubernetesJobManagerParameters.getSavepointPath();
+        if (savepointPath != null) {
+            args.add("--fromSavepoint");
+            args.add(savepointPath);
+        }
+
         List<String> jobSpecArgs = kubernetesJobManagerParameters.getJobSpecArgs();
         if (jobSpecArgs != null) {
             args.addAll(kubernetesJobManagerParameters.getJobSpecArgs());
--- a/flink-kubernetes-standalone/src/test/java/org/apache/flink/kubernetes/operator/kubeclient/decorators/CmdStandaloneJobManagerDecoratorTest.java
+++ b/flink-kubernetes-standalone/src/test/java/org/apache/flink/kubernetes/operator/kubeclient/decorators/CmdStandaloneJobManagerDecoratorTest.java
@@ -75,7 +75,7 @@ public class CmdStandaloneJobManagerDecoratorTest {
                 StandaloneKubernetesConfigOptionsInternal.CLUSTER_MODE,
                 StandaloneKubernetesConfigOptionsInternal.ClusterMode.APPLICATION);
         configuration.set(ApplicationConfiguration.APPLICATION_MAIN_CLASS, testMainClass);
-        configuration.set(SavepointConfigOptions.SAVEPOINT_IGNORE_UNCLAIMED_STATE, false);
+        configuration.set(SavepointConfigOptions.SAVEPOINT_IGNORE_UNCLAIMED_STATE, true);
         configuration.set(SavepointConfigOptions.SAVEPOINT_PATH, "/tmp/savepoint/path");
 
         FlinkPod decoratedPod = decorator.decorateFlinkPod(new FlinkPod.Builder().build());
@@ -86,7 +86,6 @@ public class CmdStandaloneJobManagerDecoratorTest {
                 containsInAnyOrder(
                         CmdStandaloneJobManagerDecorator.APPLICATION_MODE_ARG,
                         "--allowNonRestoredState",
-                        "false",
                         "--fromSavepoint",
                         "/tmp/savepoint/path",
                         "--job-classname",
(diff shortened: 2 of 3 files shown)


gyfora:
cc @usamj

sap1ens:
I was surprised to see that this change didn't fix the issue on staging... After spending more time on debugging I found that the `allowNonRestoredState` flag actually doesn't require a boolean! And Flink's parser can parse it incorrectly when it's supplied.

With `--allowNonRestoredState true`:

Screen Shot 2022-10-13 at 4 19 04 PM

With `--allowNonRestoredState`:

Screen Shot 2022-10-13 at 4 19 17 PM

Pushed https://github.com/apache/flink-kubernetes-operator/pull/403/commits/add45fc400511280c0cf8cdd97d4514e2375fa57 as a fix.

sap1ens:
Sorry, GitHub decided to close it automatically after I referenced it in an internal PR that was merged :)

gyfora:
We should think of a way to test these in a reliable way. Testing the parameters is not enough as might not be understood by the JM as you pointed out.

Maybe we need some e2e tests for standalone mode that test these params properly

sap1ens:
> We should think of a way to test these in a reliable way. Testing the parameters is not enough as might not be understood by the JM as you pointed out.
> 
> Maybe we need some e2e tests for standalone mode that test these params properly

I agree, e2e testing seems to be the only reliable way to verify that it works.

I've checked the `e2e-tests` folder and it looks like it contains only fairly basic tests. To properly test this change we need to:

1. Start a stateful application
2. Take a savepoint
3. Stop it
4. Start another stateful application with a slightly different topology so `allowNonRestoredState` is required
5. Confirm it's restored properly 

Perhaps we can skip steps 1-3 if we're ok with keeping a savepoint file as a test resource. Also, step 4 is particularly tricky: is there a standard Flink example that we can use? Otherwise, we'd need to build a new app just for testing.

This feel like a new test file.

@gyfora thoughts?

gyfora:
> > We should think of a way to test these in a reliable way. Testing the parameters is not enough as might not be understood by the JM as you pointed out.
> > Maybe we need some e2e tests for standalone mode that test these params properly
> 
> I agree, e2e testing seems to be the only reliable way to verify that it works.
> 
> I've checked the `e2e-tests` folder and it looks like it contains only fairly basic tests. To properly test this change we need to:
> 
> 1. Start a stateful application
> 2. Take a savepoint
> 3. Stop it
> 4. Start another stateful application with a slightly different topology so `allowNonRestoredState` is required
> 5. Confirm it's restored properly
> 
> Perhaps we can skip steps 1-3 if we're ok with keeping a savepoint file as a test resource. Also, step 4 is particularly tricky: is there a standard Flink example that we can use? Otherwise, we'd need to build a new app just for testing.
> 
> This feel like a new test file.
> 
> @gyfora thoughts?

I agree, it would be nice to have comprehensive tests that actually test the allowNonRestoredState logic. On the other hand if we simply enable that flag on existing tests at least we provide a regression test for this particular issue.

Because current test  would have still failed before this fix if that flag is enabled for standalone.

I think we should do something simple for this ticket and open another jira for a complete test case.

flink-kubernetes-standalone/src/test/java/org/apache/flink/kubernetes/operator/kubeclient/decorators/CmdStandaloneJobManagerDecoratorTest.java
```diff
@@ -85,7 +86,8 @@ public void testApplicationCommandAndArgsAdded() {
                 containsInAnyOrder(
                         CmdStandaloneJobManagerDecorator.APPLICATION_MODE_ARG,
                         "--allowNonRestoredState",
-                        "false",
+                        "--fromSavepoint",
+                        "/tmp/savepoint/path",
```
gyfora, line 90: Why do we need to add this here? Seems like it worked before without it

sap1ens:
> I agree, it would be nice to have comprehensive tests that actually test the allowNonRestoredState logic. On the other hand if we simply enable that flag on existing tests at least we provide a regression test for this particular issue.
> 
> Because current test would have still failed before this fix if that flag is enabled for standalone.

Not quite: `allowNonRestoredState` is pretty much ignored unless `fromSavepoint` is also provided. WIth or without the boolean: it doesn't matter. That's why I think we need a real savepoint recovery test, probably as a separate issue / PR.

flink-kubernetes-standalone/src/test/java/org/apache/flink/kubernetes/operator/kubeclient/decorators/CmdStandaloneJobManagerDecoratorTest.java
sap1ens: Because I'm testing the logic I added in the `CmdStandaloneJobManagerDecorator` file.

flink-kubernetes-standalone/src/main/java/org/apache/flink/kubernetes/operator/kubeclient/decorators/CmdStandaloneJobManagerDecorator.java
```diff
@@ -80,9 +80,14 @@ private List<String> getApplicationClusterArgs() {
...
-        if (allowNonRestoredState != null) {
+        if (allowNonRestoredState != null && allowNonRestoredState) {
             args.add("--allowNonRestoredState");
-            args.add(allowNonRestoredState.toString());
+        }
+
+        String savepointPath = kubernetesJobManagerParameters.getSavepointPath();
+        if (savepointPath != null) {
+            args.add("--fromSavepoint");
+            args.add(savepointPath);
```
gyfora, line 90: What I meant with my other question is actually why we need this part here :) 
Seems like before this change the savepoint restore logic worked correctly so the config is probably enough.

Wouldn't be enough to simply fix the allowNonRestoredState logic to not put true/false there?

gyfora: As we discussed on Slack, before this fix. Both `initialSavepoint` and savepoint upgrade logic worked correctly so there must not really be a need for the `--fromSavepoint` argument

sap1ens: Yep, I think you're right: it seems like just fixing the `allowNonRestoredState` flag would be enough. 

However, I think it's logical to provide the two options together via the command-line. They always come together (check `SavepointConfigOptions`, `SavepointRestoreSettings`, etc.) and I was really surprised when I didn't see the `fromSavepoint` argument when debugging. 

But if you insist on making the smallest change in this PR then I'll revert the `fromSavepoint` part. Let me know!

gyfora: I don't insist I think it's good like this for more visibility, I just wanted to make sure I fully understand the reason behind this particular change. :)

gyfora: actually I think it might be enough to remove both `fromSavepoint` and `allowNonRestoredState` command line arguments and rely purely on the config. If that works I would actually lean toward that. It's much simpler and more bulletproof to just use the conf

sap1ens: I thought about it too, but based on these lines:

- https://github.com/apache/flink/blob/master/flink-container/src/main/java/org/apache/flink/container/entrypoint/StandaloneApplicationClusterEntryPoint.java#L95
- https://github.com/apache/flink/blob/master/flink-runtime/src/main/java/org/apache/flink/runtime/jobgraph/SavepointRestoreSettings.java#L166-L168

it looks like if the flag is not present on the command-line, Flink will override the value from the config file to `false`.

gyfora: You are right, let’s keep it as is

gyfora:
@sap1ens could you please open a backport PR against the `release-1.2` branch based on the merged commit?

sap1ens:
Done https://github.com/apache/flink-kubernetes-operator/pull/404

PR #407, opened 2022-10-22
[FLINK-29713] Kubernetes operator should restart failed jobs

pvary:

It would be good to have the possibility to restart the Flink Application if it goes to `FAILED` state.
This could be used to restart, and reconfigure the job dynamically in the application `main` method if the current application can not handle the incoming data.

## Brief change log

  - Creates a new configuration `cluster.restart.failed` with the default `false`
  - If the Application state is `FAILED` and the configuration is set to `true` then the application is removed and redeployed

## Verifying this change
  - Added tests where the configuration is turned on, and the application is redeployed

## Documentation

### DIFF
--- a/docs/content/docs/custom-resource/job-management.md
+++ b/docs/content/docs/custom-resource/job-management.md
@@ -241,6 +241,21 @@ In order this feature to work one must enable [recovery of missing job deploymen
 At the moment deployment is considered unhealthy when Flink's restarts count reaches `kubernetes.operator.cluster.health-check.restarts.threshold` (default: `64`)
 within time window of `kubernetes.operator.cluster.health-check.restarts.window` (default: 2 minutes).
 
+## Restart failed job deployments
+
+The operator can restart a failed Flink cluster deployment. This could be useful in cases when the job main task is
+able to reconfigure the job to handle these failures.
+
+For example a job could dynamically create the DAG based on some job configuration which job configuration could
+change over time. When a task detects a record which could not be handled with the current configuration then the task
+should throw a `SuppressRestartsException` to fail the job. If `kubernetes.operator.cluster.restart.failed` is set to 
+`true` (default: `false`) then the operator detects the failed job and restarts it. When the job restarts then it reads
+the new job configuration and creates the new DAG based on this new configuration. The new deployment could handle the
+incoming records and no manual intervention is needed.
+
+When this configuration is set to `true` then at the moment when the deployment status is set to `FAILED` the kubernetes
+operator will delete the current deployment and redeploys the application using the latest successful checkpoint.
+
 ## Application upgrade rollbacks (Experimental)
 
 The operator supports upgrade rollbacks as an experimental feature.
--- a/docs/content/docs/custom-resource/job-management.md
+++ b/docs/content/docs/custom-resource/job-management.md
@@ -243,18 +243,12 @@ within time window of `kubernetes.operator.cluster.health-check.restarts.window`
 
 ## Restart failed job deployments
 
-The operator can restart a failed Flink cluster deployment. This could be useful in cases when the job main task is
+The operator can restart a failed Flink job. This could be useful in cases when the job main task is
 able to reconfigure the job to handle these failures.
 
-For example a job could dynamically create the DAG based on some job configuration which job configuration could
-change over time. When a task detects a record which could not be handled with the current configuration then the task
-should throw a `SuppressRestartsException` to fail the job. If `kubernetes.operator.cluster.restart.failed` is set to 
-`true` (default: `false`) then the operator detects the failed job and restarts it. When the job restarts then it reads
-the new job configuration and creates the new DAG based on this new configuration. The new deployment could handle the
-incoming records and no manual intervention is needed.
-
-When this configuration is set to `true` then at the moment when the deployment status is set to `FAILED` the kubernetes
-operator will delete the current deployment and redeploys the application using the latest successful checkpoint.
+When `kubernetes.operator.job.restart.failed` is set to `true` then at the moment when the job status is
+set to `FAILED` the kubernetes operator will delete the current job and redeploy the job using the
+latest successful checkpoint.
 
 ## Application upgrade rollbacks (Experimental)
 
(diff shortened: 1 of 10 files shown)


docs/content/docs/custom-resource/job-management.md
```diff
@@ -241,6 +241,21 @@ In order this feature to work one must enable [recovery of missing job deploymen
...
+
+The operator can restart a failed Flink cluster deployment. This could be useful in cases when the job main task is
+able to reconfigure the job to handle these failures.
+
+For example a job could dynamically create the DAG based on some job configuration which job configuration could
+change over time. When a task detects a record which could not be handled with the current configuration then the task
+should throw a `SuppressRestartsException` to fail the job. If `kubernetes.operator.cluster.restart.failed` is set to 
+`true` (default: `false`) then the operator detects the failed job and restarts it. When the job restarts then it reads
+the new job configuration and creates the new DAG based on this new configuration. The new deployment could handle the
+incoming records and no manual intervention is needed.
```
gyfora, line 254: Let's remove this paragraph, this is very use-case specific.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
```diff
@@ -339,4 +339,11 @@ public static String operatorConfigKey(String key) {
                     .withDescription(
                             "The threshold which is checked against job restart count within a configured window. "
                                     + "If the restart count is reaching the threshold then full cluster restart is initiated.");
+
+    @Documentation.Section(SECTION_DYNAMIC)
+    public static final ConfigOption<Boolean> OPERATOR_CLUSTER_RESTART_FAILED =
+            operatorConfig("cluster.restart.failed")
```
gyfora, line 345: this should be:

```
job.restart.failed
```

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
```diff
@@ -262,26 +264,44 @@ public boolean reconcileOtherChanges(
             return true;
         }
 
+        if (JobStatus.valueOf(deployment.getStatus().getJobStatus().getState()) == JobStatus.FAILED
```
gyfora, line 267: Maybe we could move this logic up one level to the `AbstractJobReconciler` this logic and config makes sense for both Applications and SessionJobs

gyfora (changes requested):

docs/content/docs/custom-resource/job-management.md
pvary: Done

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
pvary: Done

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
pvary: Done

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractJobReconciler.java
```diff
@@ -238,8 +240,34 @@ protected void rollback(CR resource, Context<?> ctx, Configuration observeConfig
...
+            throws Exception {
+        LOG.info("Resubmitting Flink Cluster deployment...");
+        SPEC specToRecover = ReconciliationUtils.getDeployedSpec(deployment);
+        restoreJob(
+                deployment,
+                specToRecover,
+                deployment.getStatus(),
+                ctx,
+                observeConfig,
+                requireHaMetadata);
```
gyfora, line 270: The `resubmitJmDeployment` name is not correct here. Maybe it's better to simply inline the `restoreJob` call in the if branch if it's only called in one place

gyfora: Or just rename it to `resubmintJob` and modify the logging accordingly

pvary: I like this. Renamed to `resubmitJob`, and modified the log message accordingly

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractJobReconciler.java
```diff
@@ -253,4 +281,14 @@ public boolean reconcileOtherChanges(
...
+
+    /**
+     * Removes a failed job.
+     *
+     * @param resource The failed job.
+     * @param observeConfig Observe configuration.
+     * @throws Exception Error during cancellation.
+     */
+    protected abstract void removeFailedJob(
+            CR resource, Context<?> ctx, Configuration observeConfig) throws Exception;
```
gyfora, line 293: I suggest we rename this to `cleanupAfterFailedJob`

gyfora (approved):
One minor comment about method naming, otherwise looks good :)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractJobReconciler.java
pvary: Done

gyfora (approved):
Very nice job @pvary!
I will merge this later today after we fixed the flaky test for the CI

jeesmon:
@pvary Can this be configured per FlinkDeployment or is it global config for all Deployments managed by the operator? Thanks!

gyfora:
> @pvary Can this be configured per FlinkDeployment or is it global config for all Deployments managed by the operator? Thanks!

@jeesmon this can be configured on a per CR basis (it is in dynamic section, thats the hint)

jeesmon:
@gyfora Just to confirm, so just need to add it to the `spec.flinkConfiguration` if I want to enable it just for a specific `FlinkDeployment`?

gyfora:
@jeesmon yes, same as with any other config

PR #412, opened 2022-10-25
[hotfix] Set the default helm image repo

xccui:

Not sure if this is intended, but the old default value `flink-kubernetes-operator` wasn't a valid image repository. Users have to override it when using the helm in the git repo.

## Brief change log

Replace the default value with a valid helm image repository `apache/flink-kubernetes-operator`

## Verifying this change

Verified with a local Argo CD setup (`apache/flink-kubernetes-operator:latest`).

## Documentation

### DIFF
--- a/helm/flink-kubernetes-operator/values.yaml
+++ b/helm/flink-kubernetes-operator/values.yaml
@@ -23,7 +23,7 @@
 # watchNamespaces: ["flink"]
 
 image:
-  repository: flink-kubernetes-operator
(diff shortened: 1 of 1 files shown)


gyfora:
The content of the Helm Values.yaml in the working branch is intended to use it with your local operator image.

I don't think we should set this to the apache repo...

xccui:
Thanks for the explanation @gyfora. That makes sense. Will close this.

PR #414, opened 2022-10-26
[FLINK-29771] Fix flaky rollback test

darenwkt:

This PR is to fix a flaky Unit Test in Rollback Test that has been failing Github Actions in the past.

## Brief change log

Increase `deployment.readiness.timeout` that is used in rollback test to simulate a rollback. This config is increased from 100ms to 400ms. 

When it's set to 100ms, it takes Github action env longer than 100ms to run line 334 to 339, therefore triggering an undesired auto-rollback and failing Unit Test. This test may not break in local dev as local env may be able to execute the code faster, however it can be easily replicated locally as well by setting the config to 10ms.

## Verifying this change

This change is already covered by existing tests, such as RollbackTest.

It has been tested in local and fork's github action.

## Documentation

  - If yes, how is the feature documented? (not applicable )


mbalassi:
Thanks @darenwkt, makes sense. Could you please extract the sleep times to a single `ROLLBACK_DELAY` or similarly named constant please?

gyfora (approved):
Very nice catch :)

darenwkt:
> Thanks @darenwkt, makes sense. Could you please extract the sleep times to a single `ROLLBACK_DELAY` or similarly named constant please?

Thanks for the feedback, I have extracted them into `ROLLBACK_DELAY`

PR #417, opened 2022-10-28
[FLINK-29655] Split Flink CRD from flink-kubernates-operator module

mbalassi:
This refactor introduces a new module called `flink-kubernetes-operator-api` and moves the CRD in said module with the aim of making the transitive dependencies of it as slim as possible via removing flink from the compile time dependency tree.


flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/crd/status/ReconciliationMetadata.java
```diff
@@ -15,10 +15,9 @@
  * limitations under the License.
  */
 
-package org.apache.flink.kubernetes.operator.reconciler;
+package org.apache.flink.kubernetes.operator.crd.status;
```
gyfora, line 18: I suggest keeping this out of the `crd` package. It is internal to the operator logic and moving it under the CRD pollutes the CRD ref docs

flink-kubernetes-operator/pom.xml
```diff
@@ -195,32 +202,6 @@ under the License.
         </resources>
 
         <plugins>
-            <plugin>
```
gyfora, line 198: We should also move the `maven-antrun-plugin` usages for the backward compatibility check to the api plugin

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/crd/status/ReconciliationMetadata.java
mbalassi: Good catch, done.

flink-kubernetes-operator/pom.xml
mbalassi: Done.

mbalassi:
@tweise has pointed out to me that we actually do not guarantee Java API compatibility at this point, and consequently it is best to make the changes now necessary to be able to provide said guarantee later.
https://nightlies.apache.org/flink/flink-kubernetes-operator-docs-main/docs/operations/compatibility/#java-api-compatibility

gyfora (commented):
Having both `api` and `crd` in the package name feels a bit redundant for me.

CRD does not really mean anything in the java world, and means api in the kubernetes world. I   don't feel strongly about renaming the packages but if we are going to do it I think we should simply change crd -> api

mbalassi:
Thanks, @gyfora. Moved the packages as you suggested.

PR #420, opened 2022-10-31
FLINK-29536 - Add WATCH_NAMESPACE env var to operator

tagarr:

OLM currently provides the ability to install an operator either into a single namespace or all namespaces (though multiple namespaces may come later). The only way to determine what namespace the operator is targeted to watch for CR's is by looking at the annotations that olm injects into the operator deployment i.e. `olm.targetNamespaces`. By adding the WATCH_NAMESPACE env var that references this annotation we will be able to get the operator to correctly monitor the targeted namespace. N.b. if the operator is installed to watch all namespaces then the annotation is just an empty string

## Brief change log

- Add the WATCH_NAMESPACE env var to EnvUtils
- Override Flink config to use this env var if it is defined i.e. not empty
- Modified helm charts so that the env vars for the pod use the deployments annotations
- Added tests to FlinkConfigManagerTest to test the env var changes

## Verifying this change

- Manually installed the helm charts to ensure the default behaviour was not modified
- Generated an OLM bundle with the changes and tested in both single and all namespaces to ensure that the operator was set to watch the correct namespace

  - Dependencies (does it add or upgrade a dependency): (yes, added moquito to test env var changes)

## Documentation

  - If yes, how is the feature documented?  (not documented). Though once we have automated olm it will become part of that documentation work

### DIFF
--- a/helm/flink-kubernetes-operator/templates/flink-operator.yaml
+++ b/helm/flink-kubernetes-operator/templates/flink-operator.yaml
@@ -82,10 +82,6 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.namespace
-            - name: WATCH_NAMESPACES
-              valueFrom:
-                fieldRef:
-                  fieldPath: metadata.annotations['olm.targetNamespaces']
             - name: OPERATOR_NAME
               value: {{ include "flink-operator.name" . }}
             - name: FLINK_CONF_DIR
@@ -164,10 +160,6 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.namespace
-            - name: WATCH_NAMESPACES
-              valueFrom:
-                fieldRef:
-                  fieldPath: metadata.annotations['olm.targetNamespaces']
           securityContext:
             {{- toYaml .Values.webhookSecurityContext | nindent 12 }}
           volumeMounts:
--- a/helm/flink-kubernetes-operator/templates/flink-operator.yaml
+++ b/helm/flink-kubernetes-operator/templates/flink-operator.yaml
@@ -79,7 +79,13 @@ spec:
           {{- end }}
           env:
             - name: OPERATOR_NAMESPACE
-              value: {{ .Release.Namespace }}
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
+            - name: WATCH_NAMESPACES
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.annotations['olm.targetNamespaces']
             - name: OPERATOR_NAME
               value: {{ include "flink-operator.name" . }}
             - name: FLINK_CONF_DIR
@@ -155,7 +161,13 @@ spec:
             - name: FLINK_PLUGINS_DIR
               value: /opt/flink/plugins
             - name: OPERATOR_NAMESPACE
-              value: {{ .Release.Namespace }}
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
+            - name: WATCH_NAMESPACES
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.annotations['olm.targetNamespaces']
           securityContext:
             {{- toYaml .Values.webhookSecurityContext | nindent 12 }}
           volumeMounts:
--- a/helm/flink-kubernetes-operator/templates/webhook.yaml
+++ b/helm/flink-kubernetes-operator/templates/webhook.yaml
@@ -85,7 +85,7 @@ metadata:
     cert-manager.io/inject-ca-from: {{ .Release.Namespace }}/flink-operator-serving-cert
   name: flink-operator-{{ .Release.Namespace }}-webhook-configuration
 webhooks:
-- name: flinkoperator.flink.apache.org
+- name: validationwebhook.flink.apache.org
   admissionReviewVersions: ["v1"]
   clientConfig:
     service:
@@ -94,7 +94,7 @@ webhooks:
       path: /validate
   failurePolicy: Fail
   rules:
-  - apiGroups: ["*"]
+  - apiGroups: ["flink.apache.org"]
     apiVersions: ["*"]
     scope: "Namespaced"
     operations:
@@ -121,7 +121,7 @@ metadata:
     cert-manager.io/inject-ca-from: {{ .Release.Namespace }}/flink-operator-serving-cert
   name: flink-operator-{{ .Release.Namespace }}-webhook-configuration
 webhooks:
-  - name: flinkoperator.flink.apache.org
+  - name: mutationwebhook.flink.apache.org
     admissionReviewVersions: ["v1"]
     clientConfig:
       service:
@@ -130,7 +130,7 @@ webhooks:
         path: /mutate
     failurePolicy: Fail
     rules:
-      - apiGroups: ["*"]
+      - apiGroups: ["flink.apache.org"]
         apiVersions: ["*"]
         scope: "Namespaced"
         operations:
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkOperatorConfiguration.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkOperatorConfiguration.java
@@ -21,6 +21,7 @@ package org.apache.flink.kubernetes.operator.config;
 import org.apache.flink.configuration.Configuration;
 import org.apache.flink.configuration.IllegalConfigurationException;
 import org.apache.flink.kubernetes.operator.metrics.KubernetesOperatorMetricOptions;
+import org.apache.flink.kubernetes.operator.utils.EnvUtils;
 
 import io.javaoperatorsdk.operator.api.config.LeaderElectionConfiguration;
 import io.javaoperatorsdk.operator.api.config.RetryConfiguration;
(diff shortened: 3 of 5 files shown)


gyfora:
cc @morhidi

morhidi:
Hi @tagarr could you please update the PR description using the PR template? I'd like to understand also why the current CM based configuration is not sufficient?

morhidi:
The main idea of having the watch namespaces in the configs is to be able to reload it dynamically without an operator restart. This would break that obviously and pin the versions to the one in the env variable.

tagarr:
If we don't use this method the only way I can see this working is by creating an init container that has the WATCH_NAMESPACE env var and then finding a way to modify the config map to set the target namespaces from the env var so that it is correctly configured. Helm installs are not affected by these changes and will always use the configmap and reload it dynamically when any changes have been made. It is highly unlikely that users using OLM would modify this env var post install, though it would be possible

tagarr:
@morhidi is the information I've suppled above adequate ?

morhidi:
I'm unfortunately not familiar with OLM capabilities, it seems it can inject certain things into the K8S yamls during installation, but cannot change a config in a  config map for example, hence the PR. We already allow configs to be overridden from a mounted config map, controlled by `CONF_OVERRIDE_DIR` have you considered this solution?

tagarr:
@morhidi I don't think providing an additional config map to use as an override is a solution. OLM already creates a configmap for the config, but there is no way to get this dynamic variable into it, other than an init container to modify post creation or user modifications post install. The easiest way, which is done for pretty much all olm'd operators is by using an Env ver injected by these annotations. If you don't want it in the helm chart I can remove it, we should be able to add it to the csv when we generate it, though it does mean a bit more work. But we do need to env var override

flink-kubernetes-webhook/pom.xml
```diff
@@ -96,6 +96,12 @@ under the License.
             <version>${okhttp.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
```
morhidi, line 100: We're trying to avoid using mockito in the project. We write test classes instead.

helm/flink-kubernetes-operator/templates/webhook.yaml
```diff
@@ -85,7 +85,7 @@ metadata:
     cert-manager.io/inject-ca-from: {{ .Release.Namespace }}/flink-operator-serving-cert
   name: flink-operator-{{ .Release.Namespace }}-webhook-configuration
 webhooks:
-- name: flinkoperator.flink.apache.org
+- name: validationwebhook.flink.apache.org
```
morhidi, line 88: Why the change?

pom.xml
```diff
@@ -89,6 +89,8 @@ under the License.
         <okhttp.version>4.10.0</okhttp.version>
 
         <snakeyaml.version>1.32</snakeyaml.version>
+
+        <mockito.version>4.8.1</mockito.version>
```
morhidi, line 93: Let's remove mockito from the tests.

morhidi (changes requested):

morhidi:
@gyfora @mbalassi can you kick off the workflow on this pls?

helm/flink-kubernetes-operator/templates/webhook.yaml
tagarr: the webhook names need to be unique

gyfora: Do you mean the validating and the mutating webhook name clashes?

gyfora: If so, why not change both?

tagarr: will do

helm/flink-kubernetes-operator/templates/webhook.yaml
```diff
@@ -121,7 +121,7 @@ metadata:
     cert-manager.io/inject-ca-from: {{ .Release.Namespace }}/flink-operator-serving-cert
   name: flink-operator-{{ .Release.Namespace }}-webhook-configuration
 webhooks:
-  - name: flinkoperator.flink.apache.org
+  - name: mutationwebhook.flink.apache.org
```
tagarr, line 124: it has been changed

flink-kubernetes-webhook/pom.xml
tagarr: @morhidi Understand, I looked at using https://junit-pioneer.org/docs/environment-variables/ which has some nice capabilities to change system env vars, but noticed when running it I get `This extension uses reflection to mutate JDK-internal state, which is fragile` warnings when running in intellij. Do you think it would be ok to add a method to EnvUtils e.g. setEnvironment(Map env) and use this to override System env vars ? Or can you think of something better ?

tagarr: Actually noticed there is a method in TestUtils, am I ok to use that ?

morhidi: @gyfora what do you think? Shall we lift the 'mockito' restriction? It's quite convenient to use the EnvUtil how it is implemented today.

helm/flink-kubernetes-operator/templates/flink-operator.yaml
```diff
@@ -79,7 +79,13 @@ spec:
...
           env:
             - name: OPERATOR_NAMESPACE
-              value: {{ .Release.Namespace }}
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
+            - name: WATCH_NAMESPACES
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.annotations['olm.targetNamespaces']
```
morhidi, line 88: We should try to generalise this somehow. Having `WATCH_NAMESPACES` as overrides are ok, ingesting them from the 'olm.targetNamespaces' is not.

tagarr: I can remove this and work out how to inject it into the csv if you prefer ?

flink-kubernetes-webhook/pom.xml
morhidi: Ah ok, I didn't realize there's `TestUtils.setEnv(Map<String, String> newEnv)`, yes you can use that.

helm/flink-kubernetes-operator/templates/flink-operator.yaml
morhidi: I don't mind if there's no Helm support for this, if you can injest it using olm.

jbusche:
I built a new flink-operator image based on your changes
- Twistlock scan shows no known fixable vulnerabilities
- helm install looks as before, able to deploy the basic.yaml.  `WATCH_NAMESPACES=`
- Built an OLM bundle using the same operator image, together with your branch.   When installed in default it properly shows: `WATCH_NAMESPACES=default`

jbusche:
I just tried an install in all namespaces - that looked good too:
```oc logs -f flink-kubernetes-operator-5b457fb9b7-znhfw |grep watch
2022-11-16 21:53:07,843 o.a.f.k.o.FlinkOperator        [INFO ] Configuring operator to watch the following namespaces: [JOSDK_ALL_NAMESPACES].
2022-11-16 21:53:07,876 o.a.f.k.o.FlinkOperator        [INFO ] Configuring operator to watch the following namespaces: [JOSDK_ALL_NAMESPACES].```

File: helm/flink-kubernetes-operator/templates/flink-operator.yaml (hunk 5 above)
Comment (morhidi) [reply helm/flink-kubernetes-operator/templates/flink-operator.yaml:88]: Can we remove this part of the logic? 

```
- name: WATCH_NAMESPACES
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.annotations['olm.targetNamespaces']
```

Comment (tagarr) [reply helm/flink-kubernetes-operator/templates/flink-operator.yaml]: done

**Comment by jbusche** [comment]
Hi @tagarr @morhidi @tedhtchang 
I've tested Tony's PR.  With a helm install, it acts as before, no change in behavior.
```
oc logs -f flink-kubernetes-operator-fdc8c59b9-dzpzr |grep watch
2022-12-01 18:31:07,701 o.a.f.k.o.FlinkOperator        [INFO ] Configuring operator to watch the following namespaces: [JOSDK_ALL_NAMESPACES].
```
But when I create an OLM bundle and install it in a particular namespace, for example default, then it properly only watches flinkdeps in that particular namespace:
```
oc rsh flink-kubernetes-operator-58f4967899-jz4w2 env |grep -i watch
WATCH_NAMESPACE=default

oc logs -f flink-kubernetes-operator-55585bc6bc-6lltj |grep watch
2022-12-01 18:12:54,931 o.a.f.k.o.FlinkOperator        [INFO ] Configuring operator to watch the following namespaces: [default].
```
Is it possible to get this merged in prior to the release 1.3.0 cutoff?  Thanks!

File: helm/flink-kubernetes-operator/templates/flink-operator.yaml [hunk 6]
```diff
@@ -79,7 +79,9 @@ spec:
           {{- end }}
           env:
             - name: OPERATOR_NAMESPACE
```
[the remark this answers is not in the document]
morhidi: @jbusche @tagarr ??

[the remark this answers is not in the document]
tagarr: So it is a more consistent way of defining the operator namespace and works for all install methods. That's why I made the change.

PR #424, opened 2022-11-02
[FLINK-29655] Move FlinkResourceListener to api

mbalassi:
I expect that this is finally the last commit for the API refactor change.

This is a potentially controversial one, because for this I needed to introduce an intentionally slim dependency on `flink-core`, this is why I ruled this out in the original PR #417. However on more thorough testing I realized that the listener interface is best exposed here.

I intentionally left the tests for the listener in the operator package, they use too much operator utilities to feasibly refactor them at this point.

The `flink-core` dependency let me revert some ugly minor changes I made in #417.


mbalassi:
cc @morhidi

morhidi:
LGTM, it seems we only needed the plugin interface from flink-core. 

Nit: The STATUS generic type is not used. `public <STATUS extends CommonStatus<?>> ResourceLifecycleState getLifecycleState() {`

mbalassi:
Thanks @morhidi, good catch. The beauty of refactoring.

PR #427, opened 2022-11-07
[FLINK-29871] Add 1.16 to CI tests

gyfora:

Enable 1.16 CI tests as a first step to support Flink 1.16. As agreed on JIRA we will move to using the 1.16 flink dependency once 1.16.1 is out.

## Brief change log

Added new e2e test case for 1.16

### DIFF
--- a/e2e-tests/data/multi-sessionjob.yaml
+++ b/e2e-tests/data/multi-sessionjob.yaml
@@ -22,8 +22,8 @@ metadata:
   namespace: default
   name: session-cluster-1
 spec:
-  image: flink:1.15
-  flinkVersion: v1_15
+  image: flink:1.16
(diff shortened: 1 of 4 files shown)


e2e-tests/data/multi-sessionjob.yaml
```diff
@@ -74,7 +74,7 @@ metadata:
   namespace: flink
   name: session-cluster-1
 spec:
-  image: flink:1.15
+  image: flink:1.16
   flinkVersion: v1_15
```
wangyang0918, line 78: Do we forget to update this line?

gyfora: Seems like it. :) will fix

PR #434, opened 2022-11-09
[FLINK-29919] Support operator leader election

gyfora:

Support enabling and configuring leader election for the kubernetes operator to make it possible to run standby operator instances.

## Brief change log

  - *Add configs to enable and configure the leader election service*
  - *Add tests*
  - *Update helm chart*

## Verifying this change

1. Extended the `FlinkOperatorTest` to cover the leader election config.
2. Manually tested on minikube

## Documentation

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkOperatorConfiguration.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkOperatorConfiguration.java
@@ -19,8 +19,10 @@
 package org.apache.flink.kubernetes.operator.config;
 
 import org.apache.flink.configuration.Configuration;
+import org.apache.flink.configuration.IllegalConfigurationException;
 import org.apache.flink.kubernetes.operator.metrics.KubernetesOperatorMetricOptions;
 
+import io.javaoperatorsdk.operator.api.config.LeaderElectionConfiguration;
 import io.javaoperatorsdk.operator.api.config.RetryConfiguration;
 import lombok.Value;
 import org.apache.commons.lang3.StringUtils;
@@ -60,6 +62,7 @@ public class FlinkOperatorConfiguration {
     int exceptionFieldLengthThreshold;
     int exceptionThrowableCountThreshold;
     String labelSelector;
+    LeaderElectionConfiguration leaderElectionConfiguration;
 
     public static FlinkOperatorConfiguration fromConfiguration(Configuration operatorConfig) {
         Duration reconcileInterval =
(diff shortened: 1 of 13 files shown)


gyfora:
cc @morhidi

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkOperatorConfiguration.java
```diff
@@ -60,6 +62,7 @@ public class FlinkOperatorConfiguration {
     int exceptionFieldLengthThreshold;
     int exceptionThrowableCountThreshold;
     String labelSelector;
+    LeaderElectionConfiguration leaderElectionConfiguration;
```
gaborgsomogyi, line 65: Nit: any reason why these are not private?

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/FlinkOperatorTest.java
```diff
@@ -74,4 +93,16 @@ public void testConfigurationPassedToJOSDK() {
...
+
+    @Test
+    public void testLeaderElectionConfig() {
+        var operatorConfig = new Configuration();
+        operatorConfig.set(KubernetesOperatorConfigOptions.OPERATOR_LEADER_ELECTION_ENABLED, true);
+
+        try {
+            new FlinkOperator(operatorConfig);
+        } catch (IllegalConfigurationException ice) {
+            assertTrue(ice.getMessage().startsWith("Lease name must be defined"));
```
gaborgsomogyi, line 105: Nit: Maybe we can mention either the feature (leader election) or config name. Such case one don't need to dig for conf name just copy-paste.

pom.xml
```diff
@@ -69,7 +69,7 @@ under the License.
         <maven-resources-plugin.version>3.2.0</maven-resources-plugin.version>
         <git-commit-id-maven-plugin.version>5.0.0</git-commit-id-maven-plugin.version>
 
-        <operator.sdk.version>4.1.0</operator.sdk.version>
+        <operator.sdk.version>4.1.1</operator.sdk.version>
```
gaborgsomogyi, line 72: Just for my own understanding, is this needed for `KUBERNETES_NAMESPACE_SYSTEM_PROPERTY`?

gaborgsomogyi (commented):
Basically looks good, only questions/nits.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/FlinkOperatorTest.java
gaborgsomogyi: Oh, now see that the feature later mentioned. Maybe adding the key would provide more convenience.

morhidi:
+1 LGTM overall. Some e2e would be nice. Is this going to be the recommended prod setup or we're not sure yet?

gyfora:
> +1 LGTM overall. Some e2e would be nice. Is this going to be the recommended prod setup or we're not sure yet?

I think the recommended setup is without leader election as it's a very optional feature. I think most prod settings do not require this and it only adds complication

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkOperatorConfiguration.java
gyfora: Lombok @Value makes them private

pom.xml
gyfora: No this is just so we have the latest josdk version and some fixes

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/FlinkOperatorTest.java
gyfora: Good idea, I will add this

PR #438, opened 2022-11-14
[FLINK-29974] Not allowing the cancelling the which are already in the completed state.

rgsriram:

*Allowing the only required job can be cancelled.*

## Brief change log

  - *Changed the `AbstractFlinkService` class `CancelSessionJob` method to allow only the necessary jobs can be cancelled*

## Verifying this change

  - *Extended the `SessionJobReconcilerTest` for testing this scenario.*
  
## Documentation

  - Does this pull request introduces a new feature? no
  - If yes, how is the feature documented? docs / JavaDocs

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
@@ -369,22 +369,28 @@ public abstract class AbstractFlinkService implements FlinkService {
         var jobId = JobID.fromHexString(jobIdString);
         Optional<String> savepointOpt = Optional.empty();
 
+        if (ReconciliationUtils.isJobInTerminalState(sessionJobStatus)) {
+            LOG.info("Job is already in terminal state.");
+        } else if (!ReconciliationUtils.isJobRunning(sessionJobStatus)) {
+            throw new RuntimeException(
+                    "Unexpected non-terminal status: "
+                            + sessionJobStatus.getJobStatus().getState());
+        }
+
         try (ClusterClient<String> clusterClient = getClusterClient(conf)) {
             final String clusterId = clusterClient.getClusterId();
             switch (upgradeMode) {
                 case STATELESS:
-                    if (ReconciliationUtils.isJobRunning(sessionJobStatus)) {
-                        LOG.info("Cancelling job.");
-                        clusterClient
-                                .cancel(jobId)
-                                .get(
-                                        configManager
-                                                .getOperatorConfiguration()
-                                                .getFlinkCancelJobTimeout()
-                                                .toSeconds(),
-                                        TimeUnit.SECONDS);
-                        LOG.info("Job successfully cancelled.");
-                    }
+                    LOG.info("Cancelling job.");
+                    clusterClient
+                            .cancel(jobId)
+                            .get(
+                                    configManager
+                                            .getOperatorConfiguration()
+                                            .getFlinkCancelJobTimeout()
+                                            .toSeconds(),
+                                    TimeUnit.SECONDS);
+                    LOG.info("Job successfully cancelled.");
                     break;
                 case SAVEPOINT:
                     LOG.info("Suspending job with savepoint.");
@@ -395,30 +401,22 @@ public abstract class AbstractFlinkService implements FlinkService {
                             conf.get(ExecutionCheckpointingOptions.CHECKPOINTING_TIMEOUT)
                                     .getSeconds();
                     try {
-                        if (ReconciliationUtils.isJobRunning(sessionJobStatus)) {
-                            String savepoint =
-                                    clusterClient
-                                            .stopWithSavepoint(
-                                                    jobId,
-                                                    false,
-                                                    savepointDirectory,
-                                                    conf.get(FLINK_VERSION)
-                                                                    .isNewerVersionThan(
-                                                                            FlinkVersion.v1_14)
-                                                            ? conf.get(
-                                                                    KubernetesOperatorConfigOptions
-                                                                            .OPERATOR_SAVEPOINT_FORMAT_TYPE)
-                                                            : null)
-                                            .get(timeout, TimeUnit.SECONDS);
-                            savepointOpt = Optional.of(savepoint);
-                            LOG.info("Job successfully suspended with savepoint {}.", savepoint);
-                        } else if (ReconciliationUtils.isJobInTerminalState(sessionJobStatus)) {
-                            LOG.info(
-                                    "Job is already in terminal state skipping cancel-with-savepoint operation.");
-                        } else {
-                            throw new RuntimeException(
-                                    "Unexpected non-terminal status: " + sessionJobStatus);
-                        }
+                        String savepoint =
+                                clusterClient
+                                        .stopWithSavepoint(
+                                                jobId,
+                                                false,
+                                                savepointDirectory,
+                                                conf.get(FLINK_VERSION)
+                                                                .isNewerVersionThan(
+                                                                        FlinkVersion.v1_14)
+                                                        ? conf.get(
+                                                                KubernetesOperatorConfigOptions
+                                                                        .OPERATOR_SAVEPOINT_FORMAT_TYPE)
+                                                        : null)
+                                        .get(timeout, TimeUnit.SECONDS);
+                        savepointOpt = Optional.of(savepoint);
+                        LOG.info("Job successfully suspended with savepoint {}.", savepoint);
                     } catch (TimeoutException exception) {
                         throw new FlinkException(
                                 String.format(
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
@@ -362,11 +362,27 @@ public abstract class AbstractFlinkService implements FlinkService {
             FlinkSessionJob sessionJob, UpgradeMode upgradeMode, Configuration conf)
             throws Exception {
 
+        // Not allowing the jobs which are already in the completed state.
+        String[] jobStateCannotBeCancelled =
+                new String[] {
+                    JobStatus.CANCELLING.name(),
+                    JobStatus.CANCELED.name(),
+                    JobStatus.FAILING.name(),
+                    JobStatus.FAILED.name(),
+                    JobStatus.FINISHED.name()
+                };
+
         var jobStatus = sessionJob.getStatus().getJobStatus();
+
+        if (Arrays.stream(jobStateCannotBeCancelled).anyMatch(jobStatus.getState()::contains)) {
+            throw new RuntimeException("Job is Already in " + jobStatus.getState() + " state");
+        }
+
         var jobIdString = jobStatus.getJobId();
         Preconditions.checkNotNull(jobIdString, "The job to be suspend should not be null");
         var jobId = JobID.fromHexString(jobIdString);
         Optional<String> savepointOpt = Optional.empty();
+
         try (ClusterClient<String> clusterClient = getClusterClient(conf)) {
             final String clusterId = clusterClient.getClusterId();
             switch (upgradeMode) {
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
@@ -362,22 +362,8 @@ public abstract class AbstractFlinkService implements FlinkService {
             FlinkSessionJob sessionJob, UpgradeMode upgradeMode, Configuration conf)
             throws Exception {
 
-        // Not allowing the jobs which are already in the completed state.
-        String[] jobStateCannotBeCancelled =
-                new String[] {
-                    JobStatus.CANCELLING.name(),
-                    JobStatus.CANCELED.name(),
-                    JobStatus.FAILING.name(),
-                    JobStatus.FAILED.name(),
-                    JobStatus.FINISHED.name()
-                };
-
-        var jobStatus = sessionJob.getStatus().getJobStatus();
-
-        if (Arrays.stream(jobStateCannotBeCancelled).anyMatch(jobStatus.getState()::contains)) {
-            throw new RuntimeException("Job is Already in " + jobStatus.getState() + " state");
-        }
-
+        var sessionJobStatus = sessionJob.getStatus();
+        var jobStatus = sessionJobStatus.getJobStatus();
         var jobIdString = jobStatus.getJobId();
         Preconditions.checkNotNull(jobIdString, "The job to be suspend should not be null");
         var jobId = JobID.fromHexString(jobIdString);
@@ -387,16 +373,18 @@ public abstract class AbstractFlinkService implements FlinkService {
             final String clusterId = clusterClient.getClusterId();
             switch (upgradeMode) {
                 case STATELESS:
-                    LOG.info("Cancelling job.");
-                    clusterClient
-                            .cancel(jobId)
-                            .get(
-                                    configManager
-                                            .getOperatorConfiguration()
-                                            .getFlinkCancelJobTimeout()
-                                            .toSeconds(),
-                                    TimeUnit.SECONDS);
-                    LOG.info("Job successfully cancelled.");
+                    if (ReconciliationUtils.isJobRunning(sessionJobStatus)) {
+                        LOG.info("Cancelling job.");
+                        clusterClient
+                                .cancel(jobId)
+                                .get(
+                                        configManager
+                                                .getOperatorConfiguration()
+                                                .getFlinkCancelJobTimeout()
+                                                .toSeconds(),
+                                        TimeUnit.SECONDS);
+                        LOG.info("Job successfully cancelled.");
+                    }
                     break;
                 case SAVEPOINT:
                     LOG.info("Suspending job with savepoint.");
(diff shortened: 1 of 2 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
```diff
@@ -362,11 +362,27 @@ public void cancelSessionJob(
...
+                    JobStatus.FAILED.name(),
+                    JobStatus.FINISHED.name()
+                };
+
         var jobStatus = sessionJob.getStatus().getJobStatus();
+
+        if (Arrays.stream(jobStateCannotBeCancelled).anyMatch(jobStatus.getState()::contains)) {
+            throw new RuntimeException("Job is Already in " + jobStatus.getState() + " state");
+        }
+
```
gyfora, line 380: This logic is not correct. We should not throw an error, here that would only make the current situation worse.

please look at the logic for application clusters:
```
if (ReconciliationUtils.isJobRunning(deploymentStatus)) {
                            LOG.info("Suspending job with savepoint.");
                            String savepoint =
                                    clusterClient
                                            .stopWithSavepoint(
                                                    Preconditions.checkNotNull(jobId),
                                                    false,
                                                    savepointDirectory,
                                                    conf.get(FLINK_VERSION)
                                                                    .isNewerVersionThan(
                                                                            FlinkVersion.v1_14)
                                                            ? savepointFormatType
                                                            : null)
                                            .get(timeout, TimeUnit.SECONDS);
                            savepointOpt = Optional.of(savepoint);
                            LOG.info("Job successfully suspended with savepoint {}.", savepoint);
                        } else if (ReconciliationUtils.isJobInTerminalState(deploymentStatus)) {
                            LOG.info(
                                    "Job is already in terminal state skipping cancel-with-savepoint operation.");
                        } else {
                            throw new RuntimeException(
                                    "Unexpected non-terminal status: " + deploymentStatus);
                        }
```
if the job is already in terminal state we **do not** throw error, because the job is already in the state where we want to get to.

gyfora (changes requested):
The point of this fix is to not get an error if a session job is finished/failed when upgrading or suspending.

Your change actually does the opposite and extends the cases where an exception would be encountered by the user

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
rgsriram: Got your point. I thought of another way to handle it. I will fix this.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
```diff
@@ -391,22 +395,30 @@ public void cancelSessionJob(
...
+                                            .get(timeout, TimeUnit.SECONDS);
+                            savepointOpt = Optional.of(savepoint);
+                            LOG.info("Job successfully suspended with savepoint {}.", savepoint);
+                        } else if (ReconciliationUtils.isJobInTerminalState(sessionJobStatus)) {
+                            LOG.info(
+                                    "Job is already in terminal state skipping cancel-with-savepoint operation.");
+                        } else {
+                            throw new RuntimeException(
+                                    "Unexpected non-terminal status: " + sessionJobStatus);
+                        }
```
gyfora, line 421: any reason why the logging and exception is not present for the STATELESS branch?

gyfora: I think it could be moved simply before the switch statement to apply uniformly

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
```diff
@@ -362,11 +362,21 @@ public void cancelSessionJob(
...
-        var jobStatus = sessionJob.getStatus().getJobStatus();
+        var sessionJobStatus = sessionJob.getStatus();
+        var jobStatus = sessionJobStatus.getJobStatus();
         var jobIdString = jobStatus.getJobId();
         Preconditions.checkNotNull(jobIdString, "The job to be suspend should not be null");
         var jobId = JobID.fromHexString(jobIdString);
         Optional<String> savepointOpt = Optional.empty();
+
+        if (ReconciliationUtils.isJobInTerminalState(sessionJobStatus)) {
+            LOG.info("Job is already in terminal state.");
```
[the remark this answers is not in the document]
rgsriram: I found the following behaviour. When the job is in Terminal State (COMPLETED/FINISHED/FAILED). There is no exception. It goes to a terminal state and continues to cancel the job. Ideally which is not expected. I checked behaviour with both STATELESS and STATEFULL jobs. I attached only the STATELESS job behaviour in the below image for reference.

Screenshot 2022-11-16 at 10 09 08 AM

Coming to the actual error reported in the ticket "Job not found". It happens only in case the job is not present in the session jobs list. I tested by the following test case.

```
@Test
public void testCancelTheDeletedSessionJob() throws Exception {
    FlinkSessionJob sessionJob = TestUtils.buildSessionJob();
    var readyContext = TestUtils.createContextWithReadyFlinkDeployment();
    reconciler.reconcile(sessionJob, readyContext);
    assertEquals(1, flinkService.listJobs().size());
    verifyAndSetRunningJobsToStatus(
            sessionJob,
            JobState.RUNNING,
            org.apache.flink.api.common.JobStatus.RECONCILING.name(),
            null,
            flinkService.listJobs());

    var statelessSessionJob = ReconciliationUtils.clone(sessionJob);
    var jobConfig = flinkService.listJobs().get(0).f2;

    // JobID must be equal.
    assertEquals(
            statelessSessionJob.getStatus().getJobStatus().getJobId(),
            flinkService.listJobs().get(0).f1.getJobId().toHexString());

    statelessSessionJob.getSpec().getJob().setUpgradeMode(UpgradeMode.STATELESS);
    statelessSessionJob.getSpec().getJob().setParallelism(2);
    // job suspended first
    reconciler.reconcile(statelessSessionJob, readyContext);
    assertEquals(
            org.apache.flink.api.common.JobStatus.FINISHED,
            flinkService.listJobs().get(0).f1.getJobState());
    verifyJobState(statelessSessionJob, JobState.SUSPENDED, "FINISHED");

    // Test - cancel the deleted session job
    flinkService.listJobs().remove(0);

    flinkService.cancelSessionJob(
                            statelessSessionJob, UpgradeMode.STATELESS, jobConfig);
}
```

I got the below exception. 

 
Screenshot 2022-11-16 at 9 48 17 AM

IMO - The exception is correct because the job itself is not present. But I think we can find the job if it exists before cancelling it. If not exists we need to throw an error but it gets cut before the execution.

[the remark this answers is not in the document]
rgsriram: I have validated the error through the minikube. Following are my observations.
1. Suspending the job using Custom Resource is working fine. 
2. If we cancel the job via flink cli, we are getting the error mentioned in the ticket. That is coming from the flink code. one example,

- [Dispatcher.java#stopWithSavepointAndGetLocation](https://github.com/apache/flink/blob/5d66e82915eace9342c175163b17f610bfbf7fa4/flink-runtime/src/main/java/org/apache/flink/runtime/dispatcher/Dispatcher.java#L976)
- [FlinkJobNotExceptionNotFound](https://github.com/apache/flink/blob/99c2a415e9eeefafacf70762b6f54070f7911ceb/flink-runtime/src/main/java/org/apache/flink/runtime/messages/FlinkJobNotFoundException.java#L29).

Correct me If I am wrong, this is something we can't control via operator code right?. We need to modify the cli and flink session cluster code in flink.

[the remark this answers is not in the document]
gyfora: If the job is cancelled via the CLI why isn't it observed in a terminal state? I don't really think that should give us any error if the operator works correctly.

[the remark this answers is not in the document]
rgsriram: Yes, You're right. 

Let me put it clearly, there are the following scenarios.
1. Cancel the same job via cli more than 1 time-> Throws an error which is out of the scope of the operator. (IMO)
2. Cancelled the job via cli and suspended the same job via operator -> No error.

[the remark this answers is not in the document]
gyfora: Ah ok, then I misunderstood what you wrote earlier a little :)

yea that sounds good!

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
```diff
@@ -362,11 +362,22 @@ public void cancelSessionJob(
...
+        var jobStatus = sessionJobStatus.getJobStatus();
         var jobIdString = jobStatus.getJobId();
         Preconditions.checkNotNull(jobIdString, "The job to be suspend should not be null");
         var jobId = JobID.fromHexString(jobIdString);
         Optional<String> savepointOpt = Optional.empty();
+
+        if (ReconciliationUtils.isJobInTerminalState(sessionJobStatus)) {
+            LOG.info("Job is already in terminal state. JobID {}", jobId.toHexString());
+            return;
+        } else if (!ReconciliationUtils.isJobRunning(sessionJobStatus)) {
```
gyfora, line 375: I think we don’t need this branch here, this would prevent us from suspending failing jobs

gyfora (changes requested):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
rgsriram: I have re-written the logic and added test cases accordingly. Please check.

rgsriram: @gaborgsomogyi - Please kindly check.

rgsriram: @gyfora - The build pipeline is failing with the below error. At this place: https://github.com/apache/flink-kubernetes-operator/actions/runs/3574448062/jobs/6009735670. 

Could you please help me here? Not sure if re-triggering solves it. 

```
flinkdeployment.flink.apache.org "session-cluster-1" deleted
flinksessionjob.flink.apache.org "flink-example-statemachine" deleted
persistentvolumeclaim "session-cluster-1-pvc" deleted
ingressclass.networking.k8s.io "nginx" deleted
No resources found
Error: Process completed with exit code 1.
```

rgsriram: above issue is resolved now.

gaborgsomogyi:
From directional perspective it looks good but I need to go through all the details.
In the meantime I would like to ask you 2 things:
* Please rebase the code to the latest master to have important changes like [this](https://github.com/apache/flink-kubernetes-operator/blob/80e5f3f4add50355de6964c1466ba558d46bc79d/e2e-tests/test_multi_sessionjob.sh#L41)
* Could you please squash all the commits into a single one to ease life (force pushing this branch is good in general)

rgsriram:
Sure, Gabor. I will complete it sometime.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/reconciler/sessionjob/SessionJobReconcilerTest.java
```diff
@@ -232,6 +232,218 @@ public void testStatelessUpgrade() throws Exception {
                 flinkService.listJobs());
     }
 
+    @Test
```
gaborgsomogyi, line 235: I think this can be compacted such a way but that said in my previous comment these are passing all the time no matter if your fix is there or not:
```
    @ParameterizedTest
    @EnumSource(org.apache.flink.api.common.JobStatus.class)
    public void testCancelStatelessSessionJob(org.apache.flink.api.common.JobStatus fromJobStatus) throws Exception {
        FlinkSessionJob sessionJob = TestUtils.buildSessionJob();

        var readyContext = TestUtils.createContextWithReadyFlinkDeployment();

        reconciler.reconcile(sessionJob, readyContext);
        assertEquals(1, flinkService.listJobs().size());
        verifyAndSetRunningJobsToStatus(
                sessionJob,
                JobState.RUNNING,
                org.apache.flink.api.common.JobStatus.RECONCILING.name(),
                null,
                flinkService.listJobs());

        var statelessSessionJob = ReconciliationUtils.clone(sessionJob);
        var jobConfig = flinkService.listJobs().get(0).f2;

        // JobID must be equal.
        assertEquals(
                statelessSessionJob.getStatus().getJobStatus().getJobId(),
                flinkService.listJobs().get(0).f1.getJobId().toHexString());

        statelessSessionJob.getSpec().getJob().setUpgradeMode(UpgradeMode.STATELESS);
        statelessSessionJob.getSpec().getJob().setParallelism(2);
        reconciler.reconcile(statelessSessionJob, readyContext);

        statelessSessionJob
                .getStatus()
                .getJobStatus()
                .setState(fromJobStatus.name());
        flinkService.cancelSessionJob(statelessSessionJob, UpgradeMode.STATELESS, jobConfig);
        assertEquals(
                org.apache.flink.api.common.JobStatus.FINISHED,
                flinkService.listJobs().get(0).f1.getJobState());
        verifyJobState(statelessSessionJob, JobState.SUSPENDED, "FINISHED");
    }
```

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
```diff
@@ -362,25 +362,31 @@ public void cancelSessionJob(
...
-                    clusterClient
-                            .cancel(jobId)
-                            .get(
-                                    configManager
-                                            .getOperatorConfiguration()
-                                            .getFlinkCancelJobTimeout()
-                                            .toSeconds(),
-                                    TimeUnit.SECONDS);
-                    LOG.info("Job successfully cancelled.");
+                    if (ReconciliationUtils.isJobRunning(sessionJobStatus)) {
```
gaborgsomogyi, line 378: I've just removed this if condition and the tests are still passing. Something is wrong.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/reconciler/sessionjob/SessionJobReconcilerTest.java
```diff
@@ -232,6 +232,218 @@ public void testStatelessUpgrade() throws Exception {
...
+        statelessSessionJob
+                .getStatus()
+                .getJobStatus()
+                .setState(org.apache.flink.api.common.JobStatus.CANCELED.name());
+        flinkService.cancelSessionJob(statelessSessionJob, UpgradeMode.STATELESS, jobConfig);
+        verifyJobState(statelessSessionJob, JobState.SUSPENDED, "FINISHED");
+    }
+
+    @Test
+    public void testCancelStatefulSessionJob() throws Exception {
```
gaborgsomogyi, line 321: * Here the same, test parts which are not expecting an exception are passing w/o the fix.
* Please split the test per source state and don't create 10+ assertions within a single test.
Either one can do parameterized test like I've suggested before or one can extract the common functionality into a function. An example can be found [here](https://github.com/apache/flink-kubernetes-operator/blob/80e5f3f4add50355de6964c1466ba558d46bc79d/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconcilerUpgradeModeTest.java#L86-L102).

gaborgsomogyi (changes requested):
In general I suggest to write tests which are failing first which doesn't contain too many copy-paste things. Then add the fixes and realize that they're passing now.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/reconciler/sessionjob/SessionJobReconcilerTest.java
rgsriram: Hi Gabor, 

This is expected for a stateless session job like in the application mode. In application mode, it will throw an error if it can not gracefully shut down the stateless job. By any chance did you mean that?  I also thought about it but it wasn't there initially. I think it should be added.

```
          case STATELESS:
                    if (ReconciliationUtils.isJobRunning(deployment.getStatus())) {
                        LOG.info("Job is running, cancelling job.");
                        try {
                            clusterClient
                                    .cancel(Preconditions.checkNotNull(jobId))
                                    .get(
                                            configManager
                                                    .getOperatorConfiguration()
                                                    .getFlinkCancelJobTimeout()
                                                    .toSeconds(),
                                            TimeUnit.SECONDS);
                            LOG.info("Job successfully cancelled.");
                        } catch (Exception e) {
                            LOG.error("Could not shut down cluster gracefully, deleting...", e);
                        }
                    }
                    deleteClusterDeployment(deployment.getMetadata(), deploymentStatus, true);
                    break;
```

Stateful job is failing as expected.

Screenshot 2022-11-30 at 6 39 54 PM

And I agree with you on splitting the test cases into multiple small functions.

```
RUNNING JOB -> Savepoint Cancel -> OK
TERMINAL JOB -> Savepoint Cancel -> OK
NOT RUNNING or TERMINAL -> Savepoint Cancel -> EXCEPTION

RUNNING JOB -> Stateless Cancel -> OK
TERMINAL JOB -> Stateless Cancel -> OK
NOT RUNNING or TERMINAL -> Stateless Cancel -> OK
```
 Above is the core logic for the test cases. Please feel free to correct me if I am wrong in any place.

rgsriram: Thanks for your valuable time in reviewing :)

mbalassi:
Closing in favor of #459

PR #439, opened 2022-11-14
[FLINK-29959] Use optimistic locking when updating the status

gyfora:

Currently the operator uses `patchStatus()` without any resource version locking to update the Flink resources. This means that the operator overwrites whatever status was there previously. This was convenient so far but with leader election it is much more likely to get into a situation where this can be problematic.

This change introduces optimistic locking for the status which detects if someone externally modified the status without the reconciliation logic being aware of it.

There are 2 main problems that we target here:
 1. Status updates by zombie operators who has last lost leadership but not realized/dead yet.
 2. Stale status received when a new leader starts
 
**Why would these happen?**

Zombie operator: 
It could in theory happen that an operator loses leadership in a middle of reconciliation due to a very long GC pause (or some network issue or whatever) and the current CR reconcile loop continues while the new leader already started to reconcile this resource. This is very unlikely but can happen with leader election and a standby operator. In these cases we don't want to allow the old operator who lost leadership to be able to make any status updates. The new logic guarantees that if the new leader made any status update the old would never be able to do so again.

Stale status:
When the new leader starts processing (if it was on standby) there is no guarantee that the status/spec reconciled at the first time is up to date. This can happen because due to some unlucky cache update timing or even a zombie operator submitting late status updates. The current operator logic very much relies on seeing the last status otherwise we can have some very weird cornercases that would definitely cause problems for the resources.

**How the new logic tackles this in a safe way**
 
What the new logic does is that it basically only allows status updates to go through when the operator has the latest status information. So it's sort of a locking on the current status. If anyone else changed the status in the meantime, we simply throw an error and retrigger the reconciliation. This is actually safe to do as the operator reconcile logic already runs with the assumption that the operator can fail at any time before status update, and we always use the status as a "write-ahead-log" of the actions we are taking. In these cases zombie operators who have already lost leadership would never reconcile again (the leader election guarantees that), and in other cases this would give us the latest version of the resource.

*Note:*
It's not easy to write unit tests that validate the logic as the fabric8 mockserver/utilities do not work well with optimistic locking. See https://github.com/fabric8io/kubernetes-client/issues/4573 for details.

## Brief change log

  - *Add optimistic locking for status updates*
  - *Remove empty strings from status*
  - *Added new integration test*

## Verifying this change

 - Extended the operator IT case
 - Manually verified on minikube
 - e2es

## Documentation

### DIFF
--- a/flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/SavepointInfo.java
+++ b/flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/SavepointInfo.java
@@ -38,15 +38,16 @@ public class SavepointInfo {
     private Savepoint lastSavepoint;
 
     /** Trigger id of a pending savepoint operation. */
-    private String triggerId = "";
+    private String triggerId;
 
     /** Trigger timestamp of a pending savepoint operation. */
-    private long triggerTimestamp = 0L;
+    private Long triggerTimestamp;
 
     /** Savepoint trigger mechanism. */
-    private SavepointTriggerType triggerType = SavepointTriggerType.UNKNOWN;
+    private SavepointTriggerType triggerType;
 
-    private SavepointFormatType formatType = SavepointFormatType.UNKNOWN;
+    /** Savepoint format. */
+    private SavepointFormatType formatType;
 
     /** List of recent savepoints. */
     private List<Savepoint> savepointHistory = new ArrayList<>();
@@ -63,8 +64,10 @@ public class SavepointInfo {
     }
 
     public void resetTrigger() {
-        this.triggerId = "";
-        this.triggerTimestamp = 0L;
+        this.triggerId = null;
+        this.triggerTimestamp = null;
+        this.triggerType = null;
+        this.formatType = null;
     }
 
     /**
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/StatusRecorder.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/StatusRecorder.java
@@ -18,7 +18,6 @@
 
 package org.apache.flink.kubernetes.operator.utils;
 
-import org.apache.flink.api.java.tuple.Tuple2;
 import org.apache.flink.kubernetes.operator.api.AbstractFlinkResource;
 import org.apache.flink.kubernetes.operator.api.FlinkDeployment;
 import org.apache.flink.kubernetes.operator.api.lifecycle.ResourceLifecycleState;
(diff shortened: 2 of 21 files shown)


csviri (commented):
Makes sense, note that this just needed in case a persistent state is managed in status. If that state would be in a separate ConfigMap/Secret might feel less "unnatural" to update it with optimistic locking. 
Status in Kubernetes by definition is "best effort", basically product of latest observation, so storing state little bit out of this definition. 

Also note that it is not guaranteed that in next reconciliation you will receive the latest object unless you call one of these:
https://github.com/java-operator-sdk/java-operator-sdk/blob/80e0696e4f3af371be21c15ced805ba4c2e9a1b6/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/ManagedInformerEventSource.java#L80-L91
For `ControllerResourceEventSource` in this case.

gyfora:
cc @tweise @morhidi

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/SavepointInfo.java
```diff
@@ -63,7 +63,7 @@ public void setTrigger(
     }
 
     public void resetTrigger() {
-        this.triggerId = "";
+        this.triggerId = null;
         this.triggerTimestamp = 0L;
```
morhidi, line 67: Can't we use `Long` and `null`s as default for the timestamps as well (lastPeriodicSavepointTimestamp,triggerTimestamp) or it would break backward compatibility?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/StatusRecorder.java
```diff
@@ -97,20 +96,79 @@ public void patchAndCacheStatus(CR resource) {
...
+                replaceStatus(resource, prevStatus);
+                err = null;
+            } catch (KubernetesClientException e) {
                 LOG.error("Error while patching status, retrying {}/3...", (i + 1), e);
                 Thread.sleep(1000);
                 err = e;
             }
         }
-        throw err;
+
```
morhidi, line 109: I trust you with this @gyfora. Do you think we should add some multithreaded unit tests for this or you verified it locally.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/SavepointInfo.java
gaborgsomogyi: Adding this would end-up in additional code like null checks, etc..., but what would be the benefit?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/StatusRecorder.java
gyfora: This is not really a question of multithreading, the main problem with unit tests is that the fabric8 mockservers cannot really handle the resource locking.

I have added a case to the Operator integration test which validates the core conflict detection. Otherwise the e2e tests and manual testing should be enough to cover the basic behaviour (we don't expect any errors in the logs normally related to this part)

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/SavepointInfo.java
gyfora: The benefit is that there would not be `triggerTimestamp` in the status when there is nothing triggered.

I agree with @morhidi I will change this, originally I thought this would be a Java API breaking change but I believe Kubernetes never sends `null` values so it's not going to be a problem.

gaborgsomogyi: Sounds fair.

gaborgsomogyi (approved):
LGTM from my point.

gaborgsomogyi:
Not sure why couple of CI things cancelled :/

gyfora:
@morhidi I updated the PR and I made some more savepoint triggering related fields nullable.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/SavepointInfo.java
morhidi: It'll keep already stored `0` values however, so no business logic should check `(triggerTimestamp == null)`, it'll be `0`

PR #446, opened 2022-11-21
[FLINK-29475] Add error checker for the operator in e2e tests

gaborgsomogyi:

At the moment we're not checking that the operator is not having an error in its log. In this PR I've added an assertion when error exists in the log. There is an exclusion list which is easy to extend.

## Brief change log

Add error checker for the operator in e2e tests

## Verifying this change

Existing e2e tests.

## Documentation

### DIFF
--- a/e2e-tests/utils.sh
+++ b/e2e-tests/utils.sh
@@ -83,6 +83,11 @@ function wait_for_jobmanager_running() {
     wait_for_logs $jm_pod_name "Rest endpoint listening at" ${TIMEOUT} || exit 1
 }
 
+function get_operator_pod_name() {
+   operator_pod_name=$(kubectl get pods --selector="app.kubernetes.io/name=flink-kubernetes-operator" -o jsonpath='{..metadata.name}')
+   echo "${operator_pod_name}"
+}
+
 function get_jm_pod_name() {
    CLUSTER_ID=$1
    jm_pod_name=$(kubectl get pods --selector="app=${CLUSTER_ID},component=jobmanager" -o jsonpath='{..metadata.name}')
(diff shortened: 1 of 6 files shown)


gaborgsomogyi:
Excluded all the errors for now for testing purposes. Each and every error needs evaluation one by one.

gaborgsomogyi:
I've gone through the whitelisted messages and I think there is no issue w/ the actual feature set.
Newly coming error messages will be breaking the build when this PR is merged.

gaborgsomogyi:
cc @gyfora @morhidi

morhidi:
Hi @gaborgsomogyi thanks for opening the PR. The main motivator behind this Jira is to eliminate ERRORS/WARNINGS from the operator logs that are harmless but generating unnecessary  noise. E.g. cannot list jobs on a cluster that is coming up. Similarly how you would eliminate compile WARNINGs from the code. Feel free to turn this JIRA into an epic and open subtasks for eliminating the errors currently added as an exception. @gyfora WDYT??

e2e-tests/utils.sh
```diff
@@ -108,12 +113,37 @@ function retry_times() {
     return 1
 }
 
+function check_operator_log_for_errors {
+  echo "Checking for operator log errors..."
+  operator_pod_name=$(get_operator_pod_name)
+  errors=$(kubectl logs "${operator_pod_name}" \
+      | grep -v "JmDeploymentStatus.ERROR.Count" \
```
morhidi, line 120: Please add a JIRA ticket for all these exceptions we are planning to eliminate. cc @gyfora

PR #449, opened 2022-11-23
[FLINK-30176] Fix assert_available_slots when curl is returning empty response

gaborgsomogyi:

The condition is simply wrong, please see the following case:
```
$ expected=0
$ actual=""

$ if [[ expected -ne actual ]] then echo "Not same"; else echo "Same"; fi
Same

$ if [[ $expected != $actual ]] then echo "Not same"; else echo "Same"; fi
Not same
```
In this PR I've fixed this edge case.

## Brief change log

Fixed the bash logic.

## Verifying this change

Existing e2e tests.

## Documentation

### DIFF
--- a/e2e-tests/utils.sh
+++ b/e2e-tests/utils.sh
@@ -61,9 +61,9 @@ function assert_available_slots() {
   expected=$1
   CLUSTER_ID=$2
   ip=$(minikube ip)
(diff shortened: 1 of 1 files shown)


gaborgsomogyi:
cc @gyfora

mbalassi (approved):
Good catch

PR #462, opened 2022-12-01
[FLINK-30266] Roll back change to HA metadata check logic

gyfora:

Revert an "improvement" to the HA metadata checking logic that actually only causes problems. In this case it did not allow jobs to recover without checkpoints in last-state mode.

## Brief change log

revert change

## Verifying this change

manually verified 

## Documentation

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkUtils.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkUtils.java
@@ -46,6 +46,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
+import java.util.function.Predicate;
 
 /** Flink Utility methods used by the operator. */
 public class FlinkUtils {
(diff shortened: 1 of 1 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkUtils.java
```diff
@@ -147,12 +148,7 @@ public static boolean isHaMetadataAvailable(
                         .list()
                         .getItems();
 
-        return configMaps.stream()
-                .anyMatch(
-                        map ->
-                                !map.isMarkedForDeletion()
-                                        && map.getData() != null
-                                        && !map.getData().isEmpty());
+        return configMaps.stream().anyMatch(Predicate.not(ConfigMap::isMarkedForDeletion));
```
mxm, line 151: In case somebody wonders, this change was included in 45a280aa2d6824ded3a03b503dc33208f159f5d9.

mxm (commented):
LGTM but I think we should cover this by a test of some sort.

mbalassi:
Closing in favor of #465

PR #463, opened 2022-12-01
[FLINK-30119] Breaking change: Flink Kubernetes Operator should store…

clarax:
… last savepoint in the SavepointInfo.lastSavepoint field whether it is completed or pending

End user experience proposal:

User can see the properties of last savepoint pending or completed and can get status in one of three states for the status: PENDING, SUCCEEDED and ABANDONED. If there is never savepoint taken or attempted, it is empty. Completed savepoints (manual, periodic and upgrade) are included Savepoint history, merged with savepoints form Flink job.

User can manually trigger a savepoint. If there is another savepoint pending, triggering is disabled. User can see this savepoint with PENDING status. Once completed, user can see last savepoint status changed to SUCCEEDED and included in savepoint history or ABANDONED and not in savepoint history. If there is other savepoint triggered after completion before user checks, user cannot see the status of the one they triggered but they can check if the savepoint is in the history.

Currently lastSavepoint only stores the last completed one, duplicate with savepoint history.

To expose the properties of the currently pending savepoint or last savepoint that failed, we need to expose those info in separate fields in SavepointInfo. The internal logic of Operator uses those fields for triggering and retries and creates compatibility issues with client.

Code change proposal:

Use lastSavepoint to store the last completed/attempted one and remove SavepointInfo.triggerTimstamp, SavepointInfo.triggerType and SavepointInfo.formatType.

Add SavepointInfo::retrieveLastSavepoint method to return the last succeeded one.

Update getLastSavepointStatus to simplify the logic and change SavepointStatus.ABANDONED state to FAILED.

This is a breaking change:
 - Existing application using lastSavepoint or trigger related fields in Savepoint info except triggerId may be broken. 
 - For triggerId dependence, better use SavepointUtils.savepointInProgress which is guaranteed to work.
 - savepointHistory is not changed and still keeps the completed savepoints only. Unit tests for savepoint history also cover a few scenarios.

## Brief change log

  - Use Savepoint.lastSavepoint to store last savepoint, completed or attempted .
  - Added SavepointInfo::retrieveLastCompletedSavepoint() method to find the last completed one.
  - Updated related logic.
  - Added migration logic to handle lingering trigger information on old CR

## Verifying this change

*(example:)*
  - Updated existing tests to check for regression.
  - Added tests for migration and backward compatibility with old CR
  - Updated e2e testes to remove tests that are no longer needed

## Documentation

  - If yes, how is the feature documented?  JavaDocs and will roll into public doc

### DIFF
--- a/docs/content/docs/custom-resource/reference.md
+++ b/docs/content/docs/custom-resource/reference.md
@@ -267,7 +267,7 @@ This page serves as a full reference for FlinkDeployment custom resource definit
 ### Savepoint
 **Class**: org.apache.flink.kubernetes.operator.api.status.Savepoint
 
-**Description**: Represents information about a finished savepoint.
+**Description**: Represents information about an attempted or finished savepoint.
 
 | Parameter | Type | Docs |
 | ----------| ---- | ---- |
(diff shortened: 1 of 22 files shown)


morhidi:
After discussing it with Clara, I'm colsing this for the time being.

PR #465, opened 2022-12-02
[FLINK-30266] Roll back change to HA metadata check logic

morhidi:

Revert an "improvement" to the HA metadata checking logic that actually only causes problems. In this case it did not allow jobs to recover without checkpoints in last-state mode.

## Brief change log

revert change

## Verifying this change

manually verified 

## Documentation


morhidi:
@mbalassi I've rebased this revert on the reverted e2es :) , please merge it once the CI is finished and green.

mbalassi (approved):
ACK @morhidi

PR #467, opened 2022-12-02
[FLINK-30281] Parsing for log4j error entry instead of plain error string in e2e tests

gaborgsomogyi:

The operator log error parser identified the metric as error which is definitely false positive:
```
Found error in log files.

flink-kubernetes-operator-7456697448-mgw42.k8soperator.flink.flink-kubernetes-operator.namespace.default.FlinkDeployment.JmDeploymentStatus.ERROR.Count: 0
```
In this PR I'm parsing for log4j error entry instead of plain error string in e2e tests.

## Brief change log

* Parsing for log4j error entry instead of plain error string in e2e tests.
* Added whitelist entry: `Error while patching status`

## Verifying this change

Existing e2e tests.

## Documentation


gaborgsomogyi:
cc @morhidi

morhidi:
There can be all sort of `ERROR.Count` like metrics in the log, not just `JmDeploymentStatus.ERROR.Count`. I simple don't understand why don't we check only the ERROR level entries in the logs and check any string containing the error word?

gaborgsomogyi:
Makes sense, changing the approach...

PR #468, opened 2022-12-02
[FLINK-30287] Delete job manager deployment first when cleaning up standalone cluster

stevenpyzhang:

Delete the Job manager resources first when deleting a standalone Flink cluster as part of an upgrade in order to prevent HA configmaps from getting cleaned up. As described in the ticket, I was seeing job HA data getting deleted between cluster upgrades.

## Brief change log

Changed the resource clean up order from TM then JM to JM then TM for standalone clusters.

## Verifying this change

Manually verified when upgrading a standalone session cluster with a SessionJob running on it, configmaps didn't get cleaned up after switching operator to deleting JM resources first.

## Documentation

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/StandaloneFlinkService.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/StandaloneFlinkService.java
@@ -134,20 +134,20 @@ public class StandaloneFlinkService extends AbstractFlinkService {
         final String clusterId = meta.getName();
         final String namespace = meta.getNamespace();
 
-        LOG.info("Deleting Flink Standalone cluster TM resources");
-        kubernetesClient
-                .apps()
-                .deployments()
-                .inNamespace(namespace)
-                .withName(StandaloneKubernetesUtils.getTaskManagerDeploymentName(clusterId))
-                .delete();
-
         LOG.info("Deleting Flink Standalone cluster JM resources");
         kubernetesClient
                 .apps()
                 .deployments()
                 .inNamespace(namespace)
                 .withName(StandaloneKubernetesUtils.getJobManagerDeploymentName(clusterId))
+                .delete();
+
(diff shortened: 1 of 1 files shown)


morhidi:
cc @usamj WDYT?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/StandaloneFlinkService.java
```diff
@@ -134,20 +134,20 @@ protected void deleteClusterInternal(ObjectMeta meta, boolean deleteHaConfigmaps
...
         kubernetesClient
                 .apps()
                 .deployments()
                 .inNamespace(namespace)
-                .withName(StandaloneKubernetesUtils.getTaskManagerDeploymentName(clusterId))
+                .withName(StandaloneKubernetesUtils.getJobManagerDeploymentName(clusterId))
                 .delete();
 
-        LOG.info("Deleting Flink Standalone cluster JM resources");
+        LOG.info("Deleting Flink Standalone cluster TM resources");
```
gyfora, line 145: Do we need to wait until the JM is deleted before starting the TM deletion?

usamj (approved):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/StandaloneFlinkService.java
usamj: I don't think that is needed

stevenpyzhang:
@usamj @morhidi @gyfora thanks for reviewing the pr and helping merge it in. Were there any thoughts on including a more advanced testing framework into Operator such as Mockito? I was thinking about adding a unit test to enforce the new clean up order, but I didn't see a way of doing it with just Junit

stevenpyzhang:
Thanks for the clarification, I've opened another pr to add a unit test.

PR #469, opened 2022-12-03
[FLINK-29109] Generate random jobId for stateless upgrade mode irrespective of Flink version

tweise:

* Previously the logic to assign jobId to avoid checkpoint path collision was applied for Flink versions >= 1.16. Since even with the newer version the jobId is constant, albeit derived from clusterId, we need to overwrite it regardless of Flink version.

## Verifying this change

Modified existing tests.

## Documentation


morhidi (approved):
+1 LGTM

gyfora (approved):

PR #473, opened 2022-12-06
[FLINK-30154] Allow to set flinkConfiguration in SessionJob

gaborgsomogyi:

At the moment it's not possible to add any flinkConfiguration in SessionJob which is blocking to submit session jobs. I can imagine configs which can be harmful in case of session jobs but we must have a blacklist like all other cases.

## Brief change log

Allow to set flinkConfiguration in SessionJob.

## Verifying this change

Existing unit tests.

## Documentation


gaborgsomogyi:
cc @mbalassi @morhidi

gyfora (commented):
Could you please add one positive test case that shows that adding any config won’t cause a validation error?

gaborgsomogyi:
Sure, added `kubernetes.operator.periodic.savepoint.interval` which is mentioned first in the jira.

gyfora (approved):

PR #476, opened 2022-12-06
[FLINK-30315] Add more information about image pull failures to the operator log

pvary:

Throw DeploymentFailedException on ErrImagePull.
This would add more info for the user to understand why the issue happened.

## Brief change log

- Introduce a new value to check in the `ContainerStateWaiting.reason`
- Updated the relevant tests
- Removed a method which was accidentally added during a previous change

## Verifying this change

This change is already covered by the existing `FlinkDeploymentControllerTest.verifyInProgressDeploymentWithBackoff` test, just added a new reason to check 

## Documentation

### DIFF
--- a/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentControllerTest.java
+++ b/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentControllerTest.java
@@ -297,88 +297,10 @@ public class FlinkDeploymentControllerTest {
     @ValueSource(
             strings = {
                 DeploymentFailedException.REASON_CRASH_LOOP_BACKOFF,
-                DeploymentFailedException.REASON_IMAGE_PULL_BACKOFF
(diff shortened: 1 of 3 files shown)


flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentControllerTest.java
```diff
@@ -297,88 +297,10 @@ public void verifyFailedDeployment() throws Exception {
     @ValueSource(
             strings = {
                 DeploymentFailedException.REASON_CRASH_LOOP_BACKOFF,
-                DeploymentFailedException.REASON_IMAGE_PULL_BACKOFF
+                DeploymentFailedException.REASON_IMAGE_PULL_BACKOFF,
+                DeploymentFailedException.REASON_ERR_IMAGE_PULL
             })
-    public void verifyInProgressDeploymentWithBackoff(String reason) throws Exception {
```
pvary, line 302: This was a duplicated code, but without the `Test` annotation

PR #481, opened 2022-12-12
[FLINK-30361] Cluster deleted and created back while updating replicas

swathic95:

Whenever we scale the standalone session cluster, currently the existing job managers and task managers are deleted and added back. During scale, we need not delete the cluster and only delete/add the new task managers depending on the replica count.

## Brief change log

  - *Changed the reconcilation logic of the SessionCluster when there is a task manager replica change*

## Verifying this change

*Unit test cases are added with only scale change and upgrade change ( along with scale change )*
*Integration testing done E2E manually*

## Documentation


swathic95:
Hi @gyfora , @morhidi , can you please review this PR

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/SessionReconciler.java
```diff
@@ -96,6 +96,19 @@ protected void reconcileSpecChange(
             Configuration deployConfig,
             DiffType type)
             throws Exception {
+        if (type == DiffType.SCALE) {
```
gyfora, line 99: If this logic is the same as the one in the AbstracJobReconciler we should move it instead to the shared parent class (AbstractFlinkResourceRecnciler)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/StandaloneFlinkService.java
```diff
@@ -171,11 +170,6 @@ protected void deleteClusterInternal(ObjectMeta meta, boolean deleteHaConfigmaps
 
     @Override
     public boolean scale(ObjectMeta meta, JobSpec jobSpec, Configuration conf) {
-        if (conf.get(JobManagerOptions.SCHEDULER_MODE) == null) {
```
gyfora, line 174: This check was supposed to check whether reactive scheduler is enabled for applications, we can’t simply remove it.

Maybe we are lacking test coverage for this…

In any case it checks not null which already looks incorrect. @morhidi you worked on this in the past what do you think?

gyfora (changes requested):
We need to add tests for the change in the reconciler logic not only the diff type changes to avoid future regressions

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/SessionReconciler.java
swathic95: Since this case was more applicable for SessionReconciler, added the changes there. But I agree with your  point as both JobReconciler and SessionReconciler is re-using it, we can add it in the parent.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/StandaloneFlinkService.java
swathic95: Even if there is no scheduler mode enabled , i.e if the scheduler mode is null, still we should be able to scale the cluster i.e even if the adaptive or reactive mode is not enabled and only scaling change should allow to only scale. But let me know your thoughts @gyfora , @morhidi

swathic95: In this case, we are calculating the desired replicas as 
```
var desiredReplicas =
                conf.get(StandaloneKubernetesConfigOptionsInternal.KUBERNETES_TASKMANAGER_REPLICAS);
```
Which is actually ```kubernetes.internal.taskmanager.replicas``` which is created whenever there is a CR update while building the task manager. Here, this value would be equal to the following:
1) If taskmanager replicas are present, KUBERNETES_TASKMANAGER_REPLICAS = task_manager_replicas configured. [Code Reference to 1](https://github.com/apache/flink-kubernetes-operator/blob/dc173cbccd5f9a90fe573ce01790684253dcc0a9/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java#L233) 
2) Else, if a job is running, it calculates the parallelism = task_slots * task_manager_replicas or parallelism configured ( if no TM replicas ) and [Code reference to 2 ](https://github.com/apache/flink-kubernetes-operator/blob/dc173cbccd5f9a90fe573ce01790684253dcc0a9/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java#L247)
```KUBERNETES_TASKMANAGER_REPLICAS =  (parallelism + taskSlots - 1) / taskSlots;```
[Internal task manager calc for 2 ]( https://github.com/apache/flink-kubernetes-operator/blob/dc173cbccd5f9a90fe573ce01790684253dcc0a9/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkUtils.java#L174)

We'll always fall to 1st category as the user has updated the task manager replicas manually. 


@gyfora , @morhidi  if you think, we have other scenarios which might affect it,let me know, I'll create another diffType ( SCALE_REPLICAS ) which will be used only by task manager replicas change and add a separate logic to scale based on the desired TM replicas and not ```kubernetes.internal.taskmanager.replicas```

Comment (gyfora) [reply flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/StandaloneFlinkService.java:174]: That's fine but what I am saying is that the StandaloneFlinkService#scale method is used by both Application and Session cluster deployments. You cannot simply scale an application cluster by adding more replicas unless reactive scheduler is configured for the job. 

So the check that you removed is neeeded for the application logic to work. 
And it should be improved to:
```
if (conf.get(JobManagerOptions.SCHEDULER_MODE) == REACTIVE) {
            return false;
        }
```

Comment (gyfora) [reply flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/StandaloneFlinkService.java:174]: As you can see your change actually breaks the test case for this:
```
StandaloneFlinkServiceTest.testReactiveScale:180 expected: <false> but was: <true>
```

You need to add new tests in this class to make sure this works well for sessions too and still preserve the existing behaviour for Application deployments

Comment (swathic95) [reply flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/StandaloneFlinkService.java:174]: Understood your concern, will make changes to make it work for both session and application mode

File: flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/service/StandaloneFlinkServiceTest.java [hunk 3]
```diff
@@ -177,11 +177,123 @@ public void testReactiveScale() throws Exception {
...
+        assertTrue(
+                flinkStandaloneService.scale(
+                        flinkDeployment.getMetadata(),
+                        flinkDeployment.getSpec().getJob(),
+                        buildConfig(flinkDeployment, configuration)));
+    }
+
+    @Test
+    public void testTMReplicaScale() throws Exception {
+        var flinkDeployment = TestUtils.buildApplicationCluster();
```
gyfora, line 189: would be nice to test this with a session cluster resource where `job = null`.
You could use `TestUtils.buildSessionCluster()` instead

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
```diff
@@ -316,6 +320,35 @@ private boolean checkNewSpecAlreadyDeployed(CR resource, Configuration deployCon
...
+            throws Exception {
+        boolean scaled =
+                getFlinkService(cr, ctx)
+                        .scale(cr.getMetadata(), cr.getSpec().getJob(), deployConfig);
+        if (scaled) {
+            LOG.info("Scaling succeeded");
+            ReconciliationUtils.updateStatusForDeployedSpec(cr, deployConfig);
+            return;
+        }
+        reconcileSpecChange(cr, ctx, observeConfig, deployConfig, diffType);
```
gyfora, line 349: it might be better to return a boolean from this method so we don't need to call `reconcileSpecChange` from two places.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/StandaloneFlinkService.java
swathic95: Addressed the comment where it scales in appmode only if parallelism changes, and if there is a replica change when the reactive scheduler is not enabled, the scaling is not successful. Added test cases as well in both session and app mode for different scenarios. 
Currently scaling can happen due to the following in standalone clusters :
1) Session mode - replica change
2) App mode enabled with scheduler configured

swathic95:
Thanks @gyfora for reviewing the PR. I have addressed all the comments, at your time can you please review it

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
```diff
@@ -316,6 +319,38 @@ private boolean checkNewSpecAlreadyDeployed(CR resource, Configuration deployCon
...
+    /**
+     * Scale the cluster whenever there is a scaling change, based on the task manager replica
+     * update or the parallelism in case of scheduler mode.
+     *
+     * @param cr Resource being reconciled.
+     * @param ctx Reconciliation context.
+     * @param observeConfig Observe configuration.
+     * @param deployConfig Configuration to be deployed.
+     * @param diffType Spec change type.
+     * @throws Exception
```
gyfora, line 331: missing `@return` javadoc

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/StandaloneFlinkService.java
```diff
@@ -171,7 +171,11 @@ protected void deleteClusterInternal(ObjectMeta meta, boolean deleteHaConfigmaps
...
+        if (conf.get(JobManagerOptions.SCHEDULER_MODE) == null
+                && conf.get(StandaloneKubernetesConfigOptionsInternal.CLUSTER_MODE)
+                        .equals(
+                                StandaloneKubernetesConfigOptionsInternal.ClusterMode
+                                        .APPLICATION)) {
```
gyfora, line 178: Instead of checking the ClusterMode, it's simpler to check `jobSpec != null`, and also please explicitly check the `SCHEDULER_MODE != REACTIVE` instead of just checking null

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/service/StandaloneFlinkServiceTest.java
```diff
@@ -184,6 +184,125 @@ public void testReactiveScale() throws Exception {
                         buildConfig(flinkDeployment, configuration)));
     }
 
+    @Test
+    public void testTMReplicaScaleApplication() throws Exception {
```
gyfora, line 188: I think the new tests are good enough and you can simply remove the old unit test. That is covered in the one you added

gyfora (approved):

gyfora:
Thank you for the work @swathic95 :) Looks great!

PR #483, opened 2022-12-13
[FLINK-30381] Bump minikube kubernetes version for operator

gaborgsomogyi:

The operator uses a quite old version of kubernetes for minikube. The latest supported version is v1.25.3. I've bumped it.

## Brief change log

Bump minikube kubernetes version for operator.

## Verifying this change

Executed `basic.yaml` on local minikube.

## Documentation


gaborgsomogyi:
cc @mbalassi @gyfora

mbalassi (commented):
Thanks. It would make sense to use the same version for testing [during the CI](https://github.com/apache/flink-kubernetes-operator/blob/main/e2e-tests/utils.sh#L197), just to make sure that this will continue to work.
If you have the bandwidth I would appreciate if you could add it there too with a comment that when that version is changed this doc should be changed too. If it becomes tedious then we can scope that part out of this ticket.

gaborgsomogyi:
Added.

PR #487, opened 2022-12-18
[FLINK-30411] Validate empty JmSpec and TmSpec

nowke:

This PR validates `FlinkDeployment` without `jobManager` or `taskManager` spec. Currently, `FlinkDeployment` gets stuck in `UPGRADING` state if you don't specify `jobManager` or `taskManager` spec. 

By adding validation, `helm install` directly fails.

**Example: `basic.yaml`**

```yaml
kind: FlinkDeployment
metadata:
  name: basic-example
spec:
  image: flink:1.15
  flinkVersion: v1_15
  flinkConfiguration:
  serviceAccount: flink
  job:
    parallelism: 2
    upgradeMode: stateless
```

## Brief change log

- Validate `FlinkDeployment` without `jobManager` or `taskManager` spec.

## Verifying this change

- Extended existing unit test for the change as part of the PR
- Manually verified by removing `jobManager` and `taskManager` from [`examples/basic.yaml`](https://github.com/apache/flink-kubernetes-operator/blob/c8ed11dfde1e0fc57d8fa45b0121ecec362bd03d/examples/basic.yaml#L29-L36)
  - *Before change* - ensured `FlinkDeployment` gets stuck in `UPGRADING` state.
  - *After change* - `helm install example/basic.yaml` fails with error message. Reverting back the file deploys the job successfully.

## Documentation

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultValidator.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultValidator.java
@@ -19,6 +19,8 @@ package org.apache.flink.kubernetes.operator.validation;
 
 import org.apache.flink.configuration.CheckpointingOptions;
 import org.apache.flink.configuration.Configuration;
+import org.apache.flink.configuration.IllegalConfigurationException;
+import org.apache.flink.configuration.JobManagerOptions;
 import org.apache.flink.configuration.MemorySize;
 import org.apache.flink.kubernetes.configuration.KubernetesConfigOptions;
 import org.apache.flink.kubernetes.operator.api.FlinkDeployment;
@@ -42,6 +44,8 @@ import org.apache.flink.kubernetes.operator.reconciler.ReconciliationUtils;
 import org.apache.flink.kubernetes.operator.utils.FlinkUtils;
 import org.apache.flink.kubernetes.operator.utils.IngressUtils;
 import org.apache.flink.kubernetes.utils.Constants;
+import org.apache.flink.runtime.clusterframework.TaskExecutorProcessUtils;
+import org.apache.flink.runtime.jobmanager.JobManagerProcessUtils;
 import org.apache.flink.util.StringUtils;
 
 import javax.annotation.Nullable;
@@ -88,7 +92,7 @@ public class DefaultValidator implements FlinkResourceValidator {
                 validateLogConfig(spec.getLogConfiguration()),
                 validateJobSpec(spec.getJob(), spec.getTaskManager(), effectiveConfig),
                 validateJmSpec(spec.getJobManager(), effectiveConfig),
-                validateTmSpec(spec.getTaskManager()),
+                validateTmSpec(spec.getTaskManager(), effectiveConfig),
                 validateSpecChange(deployment, effectiveConfig),
                 validateServiceAccount(spec.getServiceAccount()));
     }
@@ -241,13 +245,33 @@ public class DefaultValidator implements FlinkResourceValidator {
     }
 
     private Optional<String> validateJmSpec(JobManagerSpec jmSpec, Map<String, String> confMap) {
+        Configuration conf = Configuration.fromMap(confMap);
+        var jmMemoryDefined =
+                jmSpec != null
+                        && jmSpec.getResource() != null
+                        && !StringUtils.isNullOrWhitespaceOnly(jmSpec.getResource().getMemory());
+        Optional<String> jmMemoryValidation =
+                jmMemoryDefined ? Optional.empty() : validateJmMemoryConfig(conf);
+
         if (jmSpec == null) {
-            return Optional.of("JobManager spec must be specified.");
+            return jmMemoryValidation;
         }
 
         return firstPresent(
+                jmMemoryValidation,
                 validateResources("JobManager", jmSpec.getResource()),
                 validateJmReplicas(jmSpec.getReplicas(), confMap));
+    }
+
+    private Optional<String> validateJmMemoryConfig(Configuration conf) {
+        try {
+            JobManagerProcessUtils.processSpecFromConfigWithNewOptionToInterpretLegacyHeap(
+                    conf, JobManagerOptions.JVM_HEAP_MEMORY);
+        } catch (IllegalConfigurationException e) {
+            return Optional.of(e.getMessage());
+        }
+
+        return Optional.empty();
     }
 
     private Optional<String> validateJmReplicas(int replicas, Map<String, String> confMap) {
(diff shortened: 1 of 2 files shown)


gyfora (changes requested):
There are two problems with the PR:

1. It doesn't validate that the memort is set in the TM/JM spec. The spec can be non-null and the memory config still be null that would lead to the same error.

2. This is a breaking change to people who have not set the spec/memory but instead configured memory directly through flinkConfiguration

morhidi:
@gyfora how do you think we should fix this?

nowke:
> @gyfora how do you think we should fix this?

@morhidi @gyfora I added further validations with configs set through `flinkConfiguration`. It does make the validation logic a little complicated. Do you suggest any alternative approach?

Besides helm validations, the deployment gets stuck in `UPGRADING` state due to `ClusterDeploymentException`. Should it fail the deployment instead upon encountering  `ClusterDeploymentException` a fixed number of times? Even after the validations in this PR, the deployment can still go into this state by setting insufficient memory.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultValidator.java
```diff
@@ -241,15 +245,35 @@ private Optional<String> validateJobSpec(
...
 
         return firstPresent(
+                jmMemoryValidation,
                 validateResources("JobManager", jmSpec.getResource()),
                 validateJmReplicas(jmSpec.getReplicas(), confMap));
     }
 
+    private Optional<String> validateJmMemoryConfig(Configuration conf) {
+        try {
+            JobManagerProcessUtils.processSpecFromConfigWithNewOptionToInterpretLegacyHeap(
```
gyfora, line 268: Why do we use this method vs using `processSpecFromConfig`?

nowke: - `processSpecFromConfigWithNewOptionToInterpretLegacyHeap` internally calls `processSpecFromConfig` - [JobManagerProcessUtils.java#L73-L74](https://github.com/apache/flink/blob/75a92efd7b35501698e5de253e5231d680830c16/flink-runtime/src/main/java/org/apache/flink/runtime/jobmanager/JobManagerProcessUtils.java#L73-L74)
- Also, `processSpecFromConfig` is not `public`, but package private - [JobManagerProcessUtils.java#L82](https://github.com/apache/flink/blob/master/flink-runtime/src/main/java/org/apache/flink/runtime/jobmanager/JobManagerProcessUtils.java#L82)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultValidator.java
```diff
@@ -241,15 +245,35 @@ private Optional<String> validateJobSpec(
...
+                    conf, JobManagerOptions.JVM_HEAP_MEMORY);
+        } catch (IllegalConfigurationException e) {
```
gyfora, line 270: I think we should catch any exception here in case Flink throws something else.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultValidator.java
```diff
@@ -261,16 +285,40 @@ private Optional<String> validateJmReplicas(int replicas, Map<String, String> co
...
+        return firstPresent(
+                tmMemoryConfigValidation,
+                validateResources("TaskManager", tmSpec.getResource()),
+                validateTmReplicas(tmSpec));
+    }
+
+    private Optional<String> validateTmMemoryConfig(Configuration conf) {
+        try {
+            TaskExecutorProcessUtils.processSpecFromConfig(conf);
+        } catch (IllegalConfigurationException e) {
```
gyfora, line 311: I think we should catch any exception here in case Flink throws something else.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/validation/DefaultValidatorTest.java
```diff
@@ -248,13 +252,40 @@ public void testValidationWithoutDefaultConfig() {
         testError(
                 dep -> dep.getSpec().getJobManager().getResource().setMemory("invalid"),
                 "JobManager resource memory parse error");
-
         testError(
                 dep -> dep.getSpec().getTaskManager().getResource().setMemory(null),
-                "TaskManager resource memory must be defined");
+                "TaskManager memory configuration failed");
```
gyfora, line 257: I think the previous error message was much more informative for the user. 
"TaskManager memory configuration failed" is pretty hard to understand.

gyfora (changes requested):
I am still a bit torn on this PR. This adds quite a lot of complexity to the validation with some risks involved. Maybe we should just leave it as it is?

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/validation/DefaultValidatorTest.java
nowke: It is thrown by `IllegalConfigurationException` - [JobManagerProcessUtils.java#L7](https://github.com/apache/flink/blob/5c3658aa06b79e8039043145560a1ad2bcce68b0/flink-runtime/src/main/java/org/apache/flink/runtime/jobmanager/JobManagerProcessUtils.java#L78)

Full error message is very descriptive, example:

```
TaskManager memory configuration failed: Either required fine-grained memory 
(taskmanager.memory.task.heap.size and taskmanager.memory.managed.size), 
or Total Flink Memory size (Key: 'taskmanager.memory.flink.size' , default: null (fallback keys: [])), 
or Total Process Memory size (Key: 'taskmanager.memory.process.size' , 
default: null (fallback keys: [])) need to be configured explicitly.
```

nowke:
> I am still a bit torn on this PR. This adds quite a lot of complexity to the validation with some risks involved. Maybe we should just leave it as it is?

Agreed on the added complexity, given the scope of the issue. Only additional benefit we get out of `processSpecFromConfig` from TM/JM utils is validating different combinations of `flinkConfiguration`, example: [https://github.com/apache/flink-kubernetes-operator/pull/487#discussion_r1059562798](https://github.com/apache/flink-kubernetes-operator/pull/487#discussion_r1059562798).

I am totally fine closing the PR unless you have any alternative approach in mind.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/validation/DefaultValidatorTest.java
gyfora: In this case I would expect an error instructing the user to set `spect.taskmanager.resource.memory` instead of the flink configs directly.

nowke: Changed it to `TaskManager resource memory must be defined using 'spec.taskManager.resource.memory'`.

PR #489, opened 2022-12-19
[FLINK-30406] Detect when jobmanager never started

gyfora:

The purpose of this PR is to fix the long standing annoying case where the job was stuck after a non-upgradable state after starting/upgrading from a savepoint but the JobManager never starts.

In these cases previously we only supported last-state (HA based) upgrade which was impossible to do if the JM never started and never created the HA metadata configmaps.

The PR introduces a check whether the JobManager pods ever started by checking the Availability conditions on the JM deployment and comparing condition times with the deployment creation timestamp.

If availability is False and the Deployment never transitioned out of this state after creation, we can then assume that the JM never started and we can perform the upgrade using the last recorded savepoint.

This also removes the slightly adhoc logic we had in place for upgrades on initial deployments before stable state (that basically intended to work around this limitaiton).

## Verifying this change

Unit tests + manual testing on minikube

## Documentation

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
@@ -27,9 +27,12 @@ import org.apache.flink.kubernetes.operator.api.spec.AbstractFlinkSpec;
 import org.apache.flink.kubernetes.operator.api.spec.FlinkDeploymentSpec;
 import org.apache.flink.kubernetes.operator.api.spec.JobSpec;
 import org.apache.flink.kubernetes.operator.api.spec.JobState;
+import org.apache.flink.kubernetes.operator.api.spec.UpgradeMode;
 import org.apache.flink.kubernetes.operator.api.status.CommonStatus;
 import org.apache.flink.kubernetes.operator.api.status.JobManagerDeploymentStatus;
 import org.apache.flink.kubernetes.operator.api.status.ReconciliationState;
+import org.apache.flink.kubernetes.operator.api.status.Savepoint;
+import org.apache.flink.kubernetes.operator.api.status.SavepointTriggerType;
 import org.apache.flink.kubernetes.operator.autoscaler.JobAutoScaler;
 import org.apache.flink.kubernetes.operator.config.FlinkConfigManager;
 import org.apache.flink.kubernetes.operator.config.KubernetesOperatorConfigOptions;
@@ -99,7 +102,7 @@ public abstract class AbstractFlinkResourceReconciler<
     }
 
     @Override
-    public final void reconcile(CR cr, Context<?> ctx) throws Exception {
+    public void reconcile(CR cr, Context<?> ctx) throws Exception {
         var spec = cr.getSpec();
         var deployConfig = getDeployConfig(cr.getMetadata(), spec, ctx);
         var status = cr.getStatus();
@@ -115,12 +118,7 @@ public abstract class AbstractFlinkResourceReconciler<
         // No further logic is required at this point.
         if (reconciliationStatus.isBeforeFirstDeployment()) {
             LOG.info("Deploying for the first time");
-
-            // Before we try to submit the job we record the current spec in the status so we can
-            // handle subsequent deployment and status update errors
-            ReconciliationUtils.updateStatusBeforeDeploymentAttempt(cr, deployConfig);
-            statusRecorder.patchAndCacheStatus(cr);
-
+            updateStatusBeforeFirstDeployment(cr, spec, deployConfig, status);
             deploy(
                     cr,
                     spec,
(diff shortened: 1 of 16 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
```diff
@@ -116,9 +119,23 @@ public final void reconcile(CR cr, Context<?> ctx) throws Exception {
...
+
+                if (initialSp != null) {
+                    status.getJobStatus()
+                            .getSavepointInfo()
+                            .setLastSavepoint(
+                                    Savepoint.of(initialSp, SavepointTriggerType.UNKNOWN));
+                    initialUpgradeMode = UpgradeMode.SAVEPOINT;
+                }
+
+                spec.getJob().setUpgradeMode(initialUpgradeMode);
```
tweise, line 134: @gyula-fora this wasn't resolved. I would expect the "reconciled" (and stable) spec to match the spec that was supplied. AFAIK, comparing desired spec with reconciled/stable spec is currently the only way to determine if all changes have been applied. If the specs don't match, how can that be achieved?

gyfora: We already store modified versions of the spec as the operator reconciles the resource. We had similar logic already during suspend/ upgrade/savepoints etc.

this is not a problem because we do not use a naive equality check to test whether user requested changes have been applied. But instead we determine the diff and ignore changes that do not result in an upgrade but only affect reconciliation (such as upgrade mode and operator configs)

Since this is already done and works without any issues and is actually a very clean solution I would suggest we not change it.

gyfora: Also I would note that this only happens if what the user requested in the spec cannot be applied directly but we apply an allowed but different upgrade mode instead . It would be strange if this was not reflected in the status and would make it difficult to build some feature that require knowing what actually was done during last upgrade

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconcilerUpgradeModeTest.java
```diff
@@ -260,11 +265,96 @@ private FlinkDeployment cloneDeploymentWithUpgradeMode(
     }
 
     @ParameterizedTest
-    @EnumSource(UpgradeMode.class)
-    public void testUpgradeBeforeReachingStableSpec(UpgradeMode upgradeMode) throws Exception {
+    @MethodSource("testUpgradeJmDeployCannotStartParams")
+    public void testUpgradeJmDeployCannotStart(UpgradeMode fromMode, UpgradeMode toMode)
+            throws Exception {
```
morhidi, line 270: Does it makes sense Gyula to break up this test into multiple methods? It's really hard to decode what are the internal status changes / checks mean actually.
- succesfulJobSubmit
- suspendWithSavepoint
- etc..

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
tweise: Makes sense. Agreed that it must be possible to find the actually used upgrade mode. And since upgrade mode change does not cause reconciliation there is no need to consider it when comparing the specs from user side either. That a new mode was applied is evident by its presence in the latest spec.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconcilerUpgradeModeTest.java
gyfora: This test is indeed a bit too complex , I will try to make some extra methods comments to make it clearer

gyfora: I have added some more comments to the tests @morhidi . I could not find a good way to break it up to methods, I feel that would just add more complexity.

These tests are pretty generic in the sense that they test different upgradeMode/savepoint setting combinations. While that helps us to cover all cases in a robust way it makes it a little difficult to simplify beyond a point.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
```diff
@@ -112,18 +116,23 @@ protected Optional<UpgradeMode> getAvailableUpgradeMode(
...
         }
 
+        if (status.getReconciliationStatus()
+                                .deserializeLastReconciledSpec()
+                                .getJob()
+                                .getUpgradeMode()
+                        != UpgradeMode.LAST_STATE
+                && FlinkUtils.jmPodNeverStarted(ctx)) {
+            deleteJmThatNeverStarted(deployment, deployConfig);
+            return getAvailableUpgradeMode(deployment, ctx, deployConfig, observeConfig);
```
morhidi, line 133: Can the process stuck here in a recursive loop?

morhidi:
+1 nice job @gyfora !

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
gyfora: I imporved the check to make this more clear, but essentially after we delete the JobManager deployment, the job will be put in a terminal state and the JmDeployment in the MISSING state. Therefore this branch would not be hit again.

PR #492, opened 2022-12-20
[FLINK-30463] Start stabilization period only after job goes into RUNNING

mxm:
We were tracking the stabilization period from the job start time which can greatly vary from the time the job goes into RUNNING state. This can result in no stabilization period at all leading to poor scaling decisions.

Thus, we change the logic to use the update time in combination with a RUNNING job status.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/ScalingExecutor.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/ScalingExecutor.java
@@ -18,6 +18,7 @@
 package org.apache.flink.kubernetes.operator.autoscaler;
 
 import org.apache.flink.annotation.VisibleForTesting;
+import org.apache.flink.api.common.JobStatus;
 import org.apache.flink.configuration.ConfigOption;
 import org.apache.flink.configuration.ConfigOptions;
 import org.apache.flink.configuration.Configuration;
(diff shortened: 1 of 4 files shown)


mxm:
Note: I need to tag the commit with the correct JIRA on merge.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/ScalingExecutor.java
```diff
@@ -121,13 +122,22 @@ public boolean scaleResource(
 
     private boolean stabilizationPeriodPassed(
             AbstractFlinkResource<?, ?> resource, Configuration conf) {
-        var now = clock.instant();
+        var jobStatus = resource.getStatus().getJobStatus();
+
+        if (!JobStatus.RUNNING.name().equals(jobStatus.getState())) {
```
gyfora, line 127: This might be a redundant check because the `JobAutoscaler` class already runs this check before calling any collection, evaluation etc.

mxm: I noticed while implementing that there are various invariants enforced in different layers. This per se is not a problem if they weren't leaky, i.e. we wouldn't pass the same data structures with different invariants through the layers. This becomes apparent in unit tests where implied invariants have to be modeled. One way to solve this would be to pass a different data structure, e.g. not pass JobStatus to lower levels but only the update time / running time.

I think the check here doesn't hurt as it asserts an important invariant.

mxm:
Do you think there are any blockers merging this?

gyfora:
Looks good +1

mxm:
Thanks for the review!

PR #493, opened 2022-12-23
[FLINK-30464] Exclude metrics during stabilization and ensure a full metric window

mxm:
So far, the code collects metrics as soon as the job goes into RUNNING mode, regardless of whether we are still in the stabilization period. This means the first metric collection window after rescaling will include metrics from the stabilization period. This is not desirable. The stabilization period is supposed to allow the job to stabilize. Metrics during stabilization shouldn't be used to make new scaling decisions.

Further, the collected metrics for the scaling decider initially do not span a full metric window. Only after the first metric window is complete, we are guaranteed to have a full windows available. This leads to short-sighted scaling decisions.

### DIFF
--- a/docs/layouts/shortcodes/generated/auto_scaler_configuration.html
+++ b/docs/layouts/shortcodes/generated/auto_scaler_configuration.html
@@ -21,16 +21,10 @@
             <td>Enable job autoscaler module.</td>
         </tr>
         <tr>
-            <td><h5>kubernetes.operator.job.autoscaler.metrics.window.size.max</h5></td>
+            <td><h5>kubernetes.operator.job.autoscaler.metrics.window</h5></td>
             <td style="word-wrap: break-word;">5 min</td>
             <td>Duration</td>
-            <td>Scaling metrics aggregation maximum window size.</td>
-        </tr>
-        <tr>
-            <td><h5>kubernetes.operator.job.autoscaler.metrics.window.size.min</h5></td>
-            <td style="word-wrap: break-word;">5 min</td>
-            <td>Duration</td>
-            <td>Scaling metrics aggregation minimum window size.</td>
+            <td>Scaling metrics aggregation window size.</td>
         </tr>
         <tr>
             <td><h5>kubernetes.operator.job.autoscaler.restart.time</h5></td>
--- a/docs/layouts/shortcodes/generated/auto_scaler_configuration.html
+++ b/docs/layouts/shortcodes/generated/auto_scaler_configuration.html
@@ -21,10 +21,16 @@
             <td>Enable job autoscaler module.</td>
         </tr>
         <tr>
-            <td><h5>kubernetes.operator.job.autoscaler.metrics.window</h5></td>
+            <td><h5>kubernetes.operator.job.autoscaler.metrics.window.size.max</h5></td>
             <td style="word-wrap: break-word;">5 min</td>
             <td>Duration</td>
-            <td>Scaling metrics aggregation window size.</td>
+            <td>Scaling metrics aggregation maximum window size.</td>
+        </tr>
+        <tr>
+            <td><h5>kubernetes.operator.job.autoscaler.metrics.window.size.min</h5></td>
+            <td style="word-wrap: break-word;">5 min</td>
+            <td>Duration</td>
+            <td>Scaling metrics aggregation minimum window size.</td>
         </tr>
         <tr>
             <td><h5>kubernetes.operator.job.autoscaler.restart.time</h5></td>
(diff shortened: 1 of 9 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/ScalingMetricCollector.java
```diff
@@ -122,7 +130,15 @@ public CollectedMetrics getMetricsHistory(
 
         // Add scaling metrics to history if they were computed successfully
         scalingMetricHistory.put(clock.instant(), scalingMetrics);
-        scalingInformation.updateMetricHistory(currentJobStartTs, scalingMetricHistory);
+        scalingInformation.updateMetricHistory(currentJobUpdateTs, scalingMetricHistory);
+
+        if (currentJobUpdateTs
+                .plus(stabilizationDuration)
+                .isAfter(clock.instant().minus(metricsWindowDuration))) {
```
gyfora, line 137: I would prefer to have a config setting for a minimum metrics window (that should default to the metrics window), so that we can have quick scaling decisions after stabilization.

Ideally users would use longer metrics windows (15-30 minutes) and a minimum window of something much shorter like 5 mins.

This is very beneficial also for testing new changes to the algorithm, or when tuning the paramaters.

mxm: That's a great idea. I've added another commit which does that. I might still create an extra JIRA for it to document the new feature set.

mxm: Created [FLINK-30510](https://issues.apache.org/jira/browse/FLINK-30510) and tagged the commit.

gyfora: Looks good! Please also update the autoscaler doc page section about the improved configuration (https://nightlies.apache.org/flink/flink-kubernetes-operator-docs-main/docs/custom-resource/autoscaler/#stabilization-and-metrics-collection-intervals) and the example configs in docs and yamls.

mxm: After thinking about this change more, it might actually be tricky for users to set the minimum metric window size correctly. There could be unexpected side effects if the delta between the two is too high, e.g. oscillation where the minimum window scales up, the maximum window scales down, and so on. It seems we only want to permit the minimum window for the initial scaling but not for subsequent scaling decisions. We could keep `metrics.window` and introduce `metrics.window.initial` which is only respected as long as the last scaling decision didn't come from the regular window.

gyfora: I think the min/max is actually simpler to understand and configure . If oscillation occurs then your stabilization period may be too small?

the default for min/max could be the same duration but I would not make this more complicated

gyfora: At any point the data rate might suddenly increase (during any scale operation) the minimum window allows us to react to it. This is not only about the first time we start the job

gyfora: I think in this early state it’s better to have flexibility with the configuration and document recommendations than to restrict the behavior. We still have time to simplify this in the future as we gather working knowledge

docs/layouts/shortcodes/generated/auto_scaler_configuration.html
```diff
@@ -21,10 +21,16 @@
             <td>Enable job autoscaler module.</td>
         </tr>
         <tr>
-            <td><h5>kubernetes.operator.job.autoscaler.metrics.window</h5></td>
+            <td><h5>kubernetes.operator.job.autoscaler.metrics.window.size.max</h5></td>
```
gyfora, line 24: could also be simply `...metrics.window.max` and `...metrics.window.min` to simplify naming

gyfora:
After some offline discussions, we have agreed with @mxm to keep the original version of this PR and not introduce an extra config for minimum metric window now to keep this simple and stable.

gyfora (approved):
Looks good!

PR #496, opened 2022-12-29
[FLINK-30527] Validate flinkVersion change for suspended jobs

gyfora:

The PR adds validation for flinkVersion changes on SUSPENDED application. If the flinkVersion is changed for an application that was suspended with last-state upgrade mode (relying on the HA metadata) this may lead to state loss depending on the flink version.

This validation is necessary to avoid unintentional state loss. 

If the user wants to change the flink version in these cases, they first have to restart the job with the old version then perform the upgrade from that state or use a savepoint.

## Brief change log

  - *Added required validation & tests*
  
## Verifying this change

New validator unit tests added to check the version change scenarios.

## Documentation

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultValidator.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultValidator.java
@@ -25,7 +25,6 @@ import org.apache.flink.kubernetes.operator.api.FlinkDeployment;
 import org.apache.flink.kubernetes.operator.api.FlinkSessionJob;
 import org.apache.flink.kubernetes.operator.api.spec.FlinkDeploymentSpec;
 import org.apache.flink.kubernetes.operator.api.spec.FlinkSessionJobSpec;
-import org.apache.flink.kubernetes.operator.api.spec.FlinkVersion;
 import org.apache.flink.kubernetes.operator.api.spec.IngressSpec;
 import org.apache.flink.kubernetes.operator.api.spec.JobManagerSpec;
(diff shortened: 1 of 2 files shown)


morhidi:
Thanks for the PR @gyfora. Regarding:
> If the user wants to change the flink version in these cases, they first have to restart the job with the old version then perform the upgrade from that state or use a savepoint.

We should advise the user to follow these instructions. The validate error could contain this IMO. WDYT?

morhidi:
Slightly related: How do you feel @gyfora about adding instructions / a link to instructions for manual restore steps, etc that users are supposed to perform.

gyfora:
@morhidi i think improving the error message is a good idea and I will do that, but I wouldn’t add detailed instructions / links. This is not common practice I feel and this also this is a manual recovery scenario but simply a spec validation error .

morhidi:
> @morhidi i think improving the error message is a good idea and I will do that, but I wouldn’t add detailed instructions / links. This is not common practice I feel and this also isn’t a manual recovery scenario but simply a spec validation error .

But the instructions are so specific ` they first have to restart the job with the old version then perform the upgrade from that state or use a savepoint`. We should do something with it. The users should know how to address the validation error.

gyfora:
The validation is about the version only, the user can restart the job with the old version but change the spec otherwise. This is not something we should do automatically for an already suspended job.

gyfora:
Maybe I just don’t understand what your concrete suggestion is :)

morhidi:
> Maybe I just don’t understand what your concrete suggestion is :)

Just append this to the validation error you already added: `they first have to restart the job with the old version then perform the upgrade from that state or use a savepoint.` Otherwise the users wont know what to do.

gyfora:
Makes sense, I will do this tomorrow

PR #504, opened 2023-01-13
[FLINK-30669] update recent job status in flinkdeployment object

kmozaid:
JIRA: https://issues.apache.org/jira/browse/FLINK-30669

This pull request fix an issue where incorrect job status is updated in FlinkDeployment resource.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
@@ -158,6 +158,7 @@ public abstract class JobStatusObserver<R extends AbstractFlinkResource<?, ?>> {
     private void updateJobStatus(FlinkResourceContext<R> ctx, JobStatusMessage clusterJobStatus) {
         var resource = ctx.getResource();
         var jobStatus = resource.getStatus().getJobStatus();
+        var previousJobId = jobStatus.getJobId();
         var previousJobStatus = jobStatus.getState();
 
         jobStatus.setState(clusterJobStatus.getJobState().name());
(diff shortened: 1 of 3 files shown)


gyfora:
I don't really see the value of this fix as the core operator logic such as observe/upgrades etc won't work when there are multiple jobs in a single application.

litiliu:
> I don't really see the value of this fix as the core operator logic such as observe/upgrades etc won't work when there are multiple jobs in a single application.

Are we going to support running multiple jobs in a single application?   There are some use case to submitting multiple pipelines in the loop within the same batch job.  Then using the `state` field in `jobStatus` to query whether all the pipeline in that batch job has finished.

gyfora:
@litiliu multiple job support is not on our roadmap. In general we are not focusing on batch processing specific features currently.

kmozaid:
Hi @gyfora IMHO, this is very minimalistic change which does not alter any existing flow. As you can see, in [ApplicationObserver.java](https://github.com/apache/flink-kubernetes-operator/pull/504/files#diff-7bd417adb2e95b5e8b81371d3022fa637a74a0a1a96403fbcc5cc5049a4df379) , just added sorting but still picks only one element from list and in `JobStatusObserver` adding if condition on jobid equality also does not change anything existing.

I have tested these changes. you could initiate workflows to see that it works fine.

kmozaid:
> I don't really see the value of this fix as the core operator logic such as observe/upgrades etc won't work when there are multiple jobs in a single application.

When there are multiple jobs running in sequence, everything works fine.

gyfora:
I agree @kmozaid if this doesn't affect existing jobs and improves some others then the change makes sense :) 

Could you please add a unit test to guard this change against possible future regressions?

kmozaid:
> I agree @kmozaid if this doesn't affect existing jobs and improves some others then the change makes sense :)
> 
> Could you please add a unit test to guard this change against possible future regressions?

sure, let me add a test.

litiliu:
@gyfora 2 workflows awaiting approval. Could you help to approve it?

PR #507, opened 2023-01-18
[FLINK-30705] Fix CR ref gen to use JsonProperty annotation values

zch93:

*This PR will resolve the CRD reference generation process. Currently the CRD reference generator does not respect JsonProperty annotation in case of enums (eg.: `public enum JobState`) which means the procedure generates the name of the ENUM instead of the value of the particular json property (eg.: `SUSPENDED` instead of suspended from `@JsonProperty("suspended")`). With these modifications generator will return with the values of the json properties in case of enums.*

#### In `CrdReferenceDoclet` class:
  - A new method added: `getNameOrJsonPropValue(Element e)`
  - Call the `getNameOrJsonPropValue(Element e)` inside the `public Void scan(Element e, Integer depth)` method

## Verifying this change

  - First I suggest to check the current version here: [/custom-resource/reference/](https://nightlies.apache.org/flink/flink-kubernetes-operator-docs-main/docs/custom-resource/reference/)
  - Then checkout this PR and
  - Run the `mvn clean install -Pgenerate-docs -DskipTests` command within the `/flink-kubernetes-operator` (so from the top level within the project)
  - After it returned with **success**, then you can go to the `/flink-kubernetes-operator/docs` folder and
  - Run the `hugo -b "" serve` command (for more information about this and about the pre-reqs, please check this [READ ME](https://github.com/apache/flink-kubernetes-operator/tree/main/docs#build-the-site-locally))
  - After you built the site locally, you can check the changes at `http://localhost:1313/`

*Example:*

#### 1. Before changes:

Screenshot 2023-01-18 at 15 17 13

#### 2. After changes:

Screenshot 2023-01-18 at 15 17 24

## Documentation

  - Does this pull request introduce a new feature? yes (new method within `CrdReferenceDoclet` class)

### DIFF
--- a/flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/docs/CrdReferenceDoclet.java
+++ b/flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/docs/CrdReferenceDoclet.java
@@ -55,7 +55,7 @@ public class CrdReferenceDoclet implements Doclet {
     private String outputFile;
     private Map<Element, Element> child2ParentElements;
 
-    private String getJsonPropValueOfEnum(Element e) {
+    private String getNameOrJsonPropValue(Element e) {
         return e.getAnnotationMirrors().stream()
                 .filter(
                         am ->
@@ -213,7 +213,7 @@ public class CrdReferenceDoclet implements Doclet {
                 case FIELD:
                     out.println(
                             "| "
-                                    + e
+                                    + getNameOrJsonPropValue(e)
                                     + " | "
                                     + e.asType().toString()
                                     + " | "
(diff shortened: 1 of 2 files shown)


flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/docs/CrdReferenceDoclet.java
```diff
@@ -210,7 +233,7 @@ public Void scan(Element e, Integer depth) {
                 case ENUM_CONSTANT:
                     out.println(
                             "| "
-                                    + e
+                                    + getJsonPropValueOfEnum(e)
```
gyfora, line 236: I think this logic should be applied to all fields, not only enums. Currently only enums are affected but we could define @JsonProperty annotations on other fields in the future

gyfora (changes requested):
Looks good! Please also include the changed regenerated documentation in the commit.

PR #510, opened 2023-01-24
FLINK-30577 unable to access flink artifacts dir

tagarr:

In openshift when the operator is installed into a restricted namespace the operator is unable to create the /opt/flink/artifacts dir in order to download remote jars for jobs

## Brief change log

 - Modified the flink-operator.yaml to add an emptyDir volume mount for /opt/flink/artifacts

## Verifying this change

This change can be verified as follows:

  - On an opensshift cluster create a new namespace
  - Install the helm chart
  - Apply the examples/basic-session-deployment-and-job.yaml
  - Verify the jobs are running correctly
 
## Documentation

### DIFF
--- a/helm/flink-kubernetes-operator/templates/flink-operator.yaml
+++ b/helm/flink-kubernetes-operator/templates/flink-operator.yaml
@@ -101,6 +101,8 @@ spec:
           volumeMounts:
             - name: flink-operator-config-volume
               mountPath: /opt/flink/conf
+            - name: flink-artifacts-volume
+              mountPath: /opt/flink/artifacts
             {{- if .Values.operatorVolumeMounts.create }}
                 {{- toYaml .Values.operatorVolumeMounts.data | nindent 12 }}
             {{- end }}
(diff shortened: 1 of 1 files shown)


jbusche:
I tested on an OpenShift 4.10.46 Cluster, both with the helm install method, as well as building a bundle and installing in multiple namespaces.  It looked good in all cases.  The operator logs looked like this:
```
oc logs -f flink-kubernetes-operator-5b6d584c7c-jgttf |grep artifacts
2023-01-24 23:21:01,713 o.a.f.k.o.a.ArtifactManager    [INFO ][flink/basic-session-job-only-example] Created dir: /opt/flink/artifacts/flink/basic-session-deployment-only-example/basic-session-job-only-example
2023-01-24 23:21:24,427 o.a.f.k.o.a.ArtifactManager    [INFO ][jim/basic-session-job-only-example] Created dir: /opt/flink/artifacts/jim/basic-session-deployment-only-example/basic-session-job-only-example
```
Did a twistlock scan of the image created from the Dockerfile, and no known vulnerabilities.  

I think it looks good, the artifacts folders can be written now both in `default` and non-default namespaces on OpenShift.

PR #513, opened 2023-01-26
[FLINK-30781] Add completed checkpoint check to cluster health check

gaborgsomogyi:

There are workloads which stuck in such a way that they're in RUNNING state most of the time but not able to proceed and make checkpoints. Such cases must be detected by the operator. In this PR I've added the possibility to ask the operator to watch the number of successful checkpoints. If the feature is enabled by `cluster.health-check.completed-checkpoints.enabled` and there are no successful checkpoint within the defined window in `cluster.health-check.completed-checkpoints.window` then the operator considers it as unhealthy deployment and re-creates it.

## Brief change log

* Added config `cluster.health-check.completed-checkpoints.enabled`
* Added config `cluster.health-check.completed-checkpoints.window`
* Added number of successful checkpoints watching

## Verifying this change

Changed/added automated tests + manually on Minikube (stateless job w/o checkpoint restarted all the time).

## Documentation

### DIFF
--- a/docs/layouts/shortcodes/generated/dynamic_section.html
+++ b/docs/layouts/shortcodes/generated/dynamic_section.html
@@ -9,6 +9,18 @@
     </thead>
     <tbody>
         <tr>
+            <td><h5>kubernetes.operator.cluster.health-check.checkpoint-progress.enabled</h5></td>
+            <td style="word-wrap: break-word;">false</td>
+            <td>Boolean</td>
+            <td>Whether to enable checkpoint progress health check for clusters.</td>
+        </tr>
+        <tr>
+            <td><h5>kubernetes.operator.cluster.health-check.checkpoint-progress.window</h5></td>
+            <td style="word-wrap: break-word;">5 min</td>
+            <td>Duration</td>
+            <td>If no checkpoints are completed within the defined time window, the job is considered unhealthy. This must be bigger than checkpointing interval.</td>
+        </tr>
+        <tr>
             <td><h5>kubernetes.operator.cluster.health-check.enabled</h5></td>
             <td style="word-wrap: break-word;">false</td>
             <td>Boolean</td>
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
@@ -380,6 +380,24 @@ public class KubernetesOperatorConfigOptions {
                                     + "If the restart count is reaching the threshold then full cluster restart is initiated.");
 
     @Documentation.Section(SECTION_DYNAMIC)
+    public static final ConfigOption<Boolean>
+            OPERATOR_CLUSTER_HEALTH_CHECK_CHECKPOINT_PROGRESS_ENABLED =
+                    operatorConfig("cluster.health-check.checkpoint-progress.enabled")
+                            .booleanType()
+                            .defaultValue(false)
+                            .withDescription(
+                                    "Whether to enable checkpoint progress health check for clusters.");
+
+    @Documentation.Section(SECTION_DYNAMIC)
+    public static final ConfigOption<Duration>
+            OPERATOR_CLUSTER_HEALTH_CHECK_CHECKPOINT_PROGRESS_WINDOW =
+                    operatorConfig("cluster.health-check.checkpoint-progress.window")
+                            .durationType()
+                            .defaultValue(Duration.ofMinutes(5))
+                            .withDescription(
+                                    "If no checkpoints are completed within the defined time window, the job is considered unhealthy. This must be bigger than checkpointing interval.");
+
+    @Documentation.Section(SECTION_DYNAMIC)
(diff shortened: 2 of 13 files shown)


gaborgsomogyi:
cc @gyfora

docs/layouts/shortcodes/generated/dynamic_section.html
```diff
@@ -8,6 +8,18 @@
...
     </thead>
     <tbody>
+        <tr>
+            <td><h5>kubernetes.operator.cluster.health-check.completed-checkpoints.enabled</h5></td>
+            <td style="word-wrap: break-word;">false</td>
+            <td>Boolean</td>
+            <td>Whether to enable completed checkpoint health check for clusters.</td>
+        </tr>
+        <tr>
+            <td><h5>kubernetes.operator.cluster.health-check.completed-checkpoints.window</h5></td>
```
gyfora, line 18: maybe it would be better to call it `checkpoint-progress`, and make it more explicit in the description that we record unhealthy state if no checkpoints were taken at all in the check window

gaborgsomogyi: Changed + fixed other issues in the doc.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
```diff
@@ -379,6 +379,24 @@ public static String operatorConfigKey(String key) {
...
+                                    "Whether to enable checkpoint progress health check for clusters.");
+
+    @Documentation.Section(SECTION_DYNAMIC)
+    public static final ConfigOption<Duration>
+            OPERATOR_CLUSTER_HEALTH_CHECK_CHECKPOINT_PROGRESS_WINDOW =
+                    operatorConfig("cluster.health-check.checkpoint-progress.window")
+                            .durationType()
+                            .defaultValue(Duration.ofMinutes(5))
+                            .withDescription(
+                                    "The duration of the time window where job completed checkpoint count measured. This must be bigger than checkpointing interval.");
```
gyfora, line 398: This description should be something like:
```
If no checkpoints are completed within the defined time window, the job is considered unhealthy. This must be bigger than checkpointing interval.
```

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/ClusterHealthEvaluator.java
```diff
@@ -75,72 +77,139 @@ public void evaluate(
...
-
-                // Update the health flag
-                lastValidClusterHealthInfo = getLastValidClusterHealthInfo(clusterInfo);
+                boolean isHealthy =
+                        evaluateRestarts(
+                                configuration,
+                                clusterInfo,
+                                lastValidClusterHealthInfo,
+                                observedClusterHealthInfo);
+                isHealthy &=
```
gyfora, line 102: let's use
```
isHealthy = evaluteRestarts(...) && evaluateCheckpoints(...)
```
for clarity

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/ClusterHealthEvaluator.java
```diff
@@ -75,72 +77,139 @@ public void evaluate(
             var lastValidClusterHealthInfo = getLastValidClusterHealthInfo(clusterInfo);
             if (lastValidClusterHealthInfo == null) {
                 LOG.debug("No last valid health info, skipping health check");
```
gyfora, line 79: I think we should clear all health info (restarts, checkpoint progress) during an upgrade/suspend, so we guarantee that it is reinitialized after the jm starts after the upgrade.

gyfora (changes requested):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/ClusterHealthEvaluator.java
gaborgsomogyi: Added.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
```diff
@@ -145,6 +147,15 @@ public void deploy(
         var relatedResource = ctx.getResource();
         var status = relatedResource.getStatus();
         var flinkService = ctx.getFlinkService();
+
+        if (spec.getJob().getState().equals(JobState.SUSPENDED)
+                || status.getReconciliationStatus()
+                        .getState()
+                        .equals(ReconciliationState.UPGRADING)) {
```
gyfora, line 154: Do we need any of these checks? If we are deploying a job then the previous health info should probably be cleared always

gaborgsomogyi: Well, such case we're more on the safe side so not needed. Removed.

PR #514, opened 2023-01-26
[FLINK-27273] Zookeeper HA support

antonipp:
https://issues.apache.org/jira/browse/FLINK-27273
This PR makes sure that Flink applications which use Zookeeper-based HA can be deployed via the Kubernetes Operator.
It wasn't previously possible mainly due to the lack of proper clean-up of Zookeeper data which could cause unexpected behaviour in Flink applications, as mentioned in the ticket.

## Brief change log
- JobGraph-related data is now properly cleaned in Zookeeper (same behaviour as when using Kubernetes HA)
- All HA metadata is now properly cleaned when application is deleted or suspended in some cases (same behaviour as when using Kubernetes HA)
- Code has been updated to support both Kubernetes and Zookeper-based HA settings. All features which worked with Kubernetes HA (such as JM deployment recovery, unhealthy job restarts or rollbacks) should also work with Zookeeper HA as well.
- Made sure https://github.com/apache/flink-kubernetes-operator/pull/518 is compatible with ZK as well

## Verifying this change
- Wrote unit tests
- Verified all existing unit tests succeed
- Tried some scenarios in our Kubernetes environment:
  - `upgradeMode: savepoint`:
    - [x] Successfully deployed a Flink application with `upgradeMode: savepoint` and Zookeeper HA enabled
    - [x] Updated the `FlinkDeployment` object, verified that the JobGraph data was successfully deleted in Zookeeper and that the new version of the application was successfully rolled out
    - [x] Manually deleted the JobManager Kubernetes Deployment, verified that the Operator was able to recover it (with `kubernetes.operator.jm-deployment-recovery.enabled` and `kubernetes.operator.job.upgrade.last-state-fallback.enabled` set to `true`) and that the application restarted successfully (this process relies on HA metadata too)
    - [x] Suspended a Standalone Flink application, verified that after the Savepoint was taken, the Zookeeper HA metadata was cleaned-up 
    - [x] Deleted the Flink Deployment, verified that the HA metadata was cleaned-up 
  - `upgradeMode: last-state`:
    - [x] Successfully deployed a Flink application with `upgradeMode: last-state` and Zookeeper HA enabled
    - [x] Updated the `FlinkDeployment` object, verified that the JobGraph data was successfully deleted in Zookeeper and that the new version of the application was successfully rolled out
    - [x] Manually deleted the JobManager Kubernetes Deployment, verified that the Operator was able to recover it with `kubernetes.operator.jm-deployment-recovery.enabled` set to `true` and that the application restarted successfully (this process relies on HA metadata too) 
  - [x] Set `kubernetes.operator.cluster.health-check.enabled` to `true` and deployed a Flink application which was continuously failing and restarting. Once the restart threshold (`kubernetes.operator.cluster.health-check.restarts.threshold`) was hit, the Operator was able to successfully validate that the Zookeeper HA metadata exists and restarted the job  
  - [x] Set `kubernetes.operator.deployment.rollback.enabled` to `true` and deployed a Flink application which was in `CrashLoopBackOff`. Once the `kubernetes.operator.deployment.readiness.timeout` passed, the Operator was able to successfully validate that the Zookeeper HA metadata exists and rolled back the job to the previous version. 
  - `upgradeMode: stateless`:
    - [x] Suspended a Flink application, verified that the Zookeeper HA metadata was cleaned-up (it's always deleted regardless of Native vs Standalone mode)

## Documentation
  - Does this pull request introduce a new feature?: yes
  - If yes, how is the feature documented?: Updated the documentation where applicable

### DIFF
--- a/docs/content/docs/concepts/overview.md
+++ b/docs/content/docs/concepts/overview.md
@@ -88,7 +88,7 @@ The examples are maintained as part of the operator repo and can be found [here]
 ## Known Issues & Limitations
 
 ### JobManager High-availability
-The Operator leverages [Kubernetes HA Services](https://nightlies.apache.org/flink/flink-docs-master/docs/deployment/ha/kubernetes_ha/) for providing High-availability for Flink jobs. The HA solution can benefit form using additional [Standby replicas](https://nightlies.apache.org/flink/flink-docs-master/docs/deployment/ha/overview/), it will result in a faster recovery time, but Flink jobs will still restart when the Leader JobManager goes down.
+The Operator supports both [Kubernetes HA Services](https://nightlies.apache.org/flink/flink-docs-master/docs/deployment/ha/kubernetes_ha/) and [Zookeeper HA Services](https://nightlies.apache.org/flink/flink-docs-master/docs/deployment/ha/zookeeper_ha/) for providing High-availability for Flink jobs. The HA solution can benefit form using additional [Standby replicas](https://nightlies.apache.org/flink/flink-docs-master/docs/deployment/ha/overview/), it will result in a faster recovery time, but Flink jobs will still restart when the Leader JobManager goes down.
 
 ### JobResultStore Resource Leak
 To mitigate the impact of [FLINK-27569](https://issues.apache.org/jira/browse/FLINK-27569) the operator introduced a workaround [FLINK-27573](https://issues.apache.org/jira/browse/FLINK-27573) by setting `job-result-store.delete-on-commit=false` and a unique value for `job-result-store.storage-path` for every cluster launch. The storage path for older runs must be cleaned up manually, keeping the latest directory always:
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
@@ -24,6 +24,7 @@ import org.apache.flink.configuration.Configuration;
 import org.apache.flink.configuration.CoreOptions;
 import org.apache.flink.configuration.DeploymentOptions;
 import org.apache.flink.configuration.DeploymentOptionsInternal;
+import org.apache.flink.configuration.HighAvailabilityOptions;
 import org.apache.flink.configuration.JobManagerOptions;
 import org.apache.flink.configuration.MemorySize;
 import org.apache.flink.configuration.PipelineOptions;
@@ -340,6 +341,9 @@ public class FlinkConfigBuilder {
         // Set cluster config
         effectiveConfig.setString(KubernetesConfigOptions.NAMESPACE, namespace);
         effectiveConfig.setString(KubernetesConfigOptions.CLUSTER_ID, clusterId);
+        if (HighAvailabilityMode.isHighAvailabilityModeActivated(effectiveConfig)) {
+            effectiveConfig.setString(HighAvailabilityOptions.HA_CLUSTER_ID, clusterId);
+        }
         return effectiveConfig;
     }
 
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
@@ -166,13 +166,17 @@ public abstract class AbstractFlinkService implements FlinkService {
         LOG.info(
                 "Deploying application cluster{}",
                 requireHaMetadata ? " requiring last-state from HA metadata" : "");
+
+        // If Kubernetes or Zookeeper HA are activated, delete the job graph in HA storage so that
+        // the newly changed job config (e.g. parallelism) could take effect
         if (FlinkUtils.isKubernetesHAActivated(conf)) {
             final String clusterId = conf.get(KubernetesConfigOptions.CLUSTER_ID);
             final String namespace = conf.get(KubernetesConfigOptions.NAMESPACE);
-            // Delete the job graph in the HA ConfigMaps so that the newly changed job config(e.g.
-            // parallelism) could take effect
             FlinkUtils.deleteJobGraphInKubernetesHA(clusterId, namespace, kubernetesClient);
+        } else if (FlinkUtils.isZookeeperHAActivated(conf)) {
+            FlinkUtils.deleteJobGraphInZookeeperHA(conf);
         }
+
         if (requireHaMetadata) {
             validateHaMetadataExists(conf);
         }
@@ -182,7 +186,13 @@ public abstract class AbstractFlinkService implements FlinkService {
 
     @Override
     public boolean isHaMetadataAvailable(Configuration conf) {
-        return FlinkUtils.isHaMetadataAvailable(conf, kubernetesClient);
+        if (FlinkUtils.isKubernetesHAActivated(conf)) {
+            return FlinkUtils.isKubernetesHaMetadataAvailable(conf, kubernetesClient);
+        } else if (FlinkUtils.isZookeeperHAActivated(conf)) {
+            return FlinkUtils.isZookeeperHaMetadataAvailable(conf);
+        }
+
+        return false;
     }
 
     @Override
@@ -283,7 +293,7 @@ public abstract class AbstractFlinkService implements FlinkService {
                             LOG.error("Could not shut down cluster gracefully, deleting...", e);
                         }
                     }
-                    deleteClusterDeployment(deployment.getMetadata(), deploymentStatus, true);
+                    deleteClusterDeployment(deployment.getMetadata(), deploymentStatus, conf, true);
                     break;
                 case SAVEPOINT:
                     final String savepointDirectory =
@@ -329,11 +339,14 @@ public abstract class AbstractFlinkService implements FlinkService {
                     }
                     if (deleteClusterAfterSavepoint) {
                         LOG.info("Cleaning up deployment after stop-with-savepoint");
-                        deleteClusterDeployment(deployment.getMetadata(), deploymentStatus, true);
+
+                        deleteClusterDeployment(
+                                deployment.getMetadata(), deploymentStatus, conf, true);
                     }
                     break;
                 case LAST_STATE:
-                    deleteClusterDeployment(deployment.getMetadata(), deploymentStatus, false);
+                    deleteClusterDeployment(
+                            deployment.getMetadata(), deploymentStatus, conf, false);
                     break;
                 default:
                     throw new RuntimeException("Unsupported upgrade mode " + upgradeMode);
@@ -907,8 +920,11 @@ public abstract class AbstractFlinkService implements FlinkService {
 
     @Override
     public final void deleteClusterDeployment(
-            ObjectMeta meta, FlinkDeploymentStatus status, boolean deleteHaData) {
-        deleteClusterInternal(meta, deleteHaData);
+            ObjectMeta meta,
+            FlinkDeploymentStatus status,
+            Configuration conf,
+            boolean deleteHaData) {
+        deleteClusterInternal(meta, conf, deleteHaData);
         updateStatusAfterClusterDeletion(status);
     }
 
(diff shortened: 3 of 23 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
```diff
@@ -340,6 +341,11 @@ protected Configuration build() {
         // Set cluster config
         effectiveConfig.setString(KubernetesConfigOptions.NAMESPACE, namespace);
         effectiveConfig.setString(KubernetesConfigOptions.CLUSTER_ID, clusterId);
+
+        if (FlinkUtils.isZookeeperHAActivated(effectiveConfig)) {
+            effectiveConfig.setString(HighAvailabilityOptions.HA_CLUSTER_ID, clusterId);
```
antonipp, line 346: By default, `clusterId` is generated [based on the FlinkDeployment resource name](https://github.com/antonipp/flink-kubernetes-operator/blob/6b0d91fe2c1dc41eeba655337c1d99fedabe3ef8/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java#L96).

The value of `HighAvailabilityOptions.HA_CLUSTER_ID` is later [used](https://github.com/apache/flink/blob/c57b84921447bb0ade5e1ff77a05ebd8bbbe71b7/flink-runtime/src/main/java/org/apache/flink/runtime/util/ZooKeeperUtils.java#L226) as the namespace for storing Zookeeper ZNodes for each Flink application.

This means that if someone creates two `FlinkDeployment`s with the same name in two different Kubernetes namespaces and these two `FlinkDeployment`s use the same ZK cluster for HA metadata storage, there might be a collision. So I think we should append a random UUID to `HighAvailabilityOptions.HA_CLUSTER_ID` :
```java
effectiveConfig.setString(HighAvailabilityOptions.HA_CLUSTER_ID, clusterId + "-RANDOM-UUID-HERE");
```
Not 100% sure of all the implications of doing that but I think it should be safe enough?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
```diff
@@ -160,19 +163,31 @@ public KubernetesClient getKubernetesClient() {
         return kubernetesClient;
     }
 
+    public CuratorFrameworkWithUnhandledErrorListener getCurator(Configuration conf) {
```
antonipp, line 166: Wrapped ZK client creation in a separate method so that it's easier to mock once I get to unit tests

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
```diff
@@ -160,19 +163,31 @@ public KubernetesClient getKubernetesClient() {
...
+        // If Kubernetes or Zookeeper HA are activated, delete the job graph in HA storage so that
+        // the newly changed job config (e.g. parallelism) could take effect
         if (FlinkUtils.isKubernetesHAActivated(conf)) {
             final String clusterId = conf.get(KubernetesConfigOptions.CLUSTER_ID);
             final String namespace = conf.get(KubernetesConfigOptions.NAMESPACE);
-            // Delete the job graph in the HA ConfigMaps so that the newly changed job config(e.g.
-            // parallelism) could take effect
             FlinkUtils.deleteJobGraphInKubernetesHA(clusterId, namespace, kubernetesClient);
+        } else if (FlinkUtils.isZookeeperHAActivated(conf)) {
+            try (var curator = getCurator(conf)) {
```
antonipp, line 184: Here, the ZK client is created only for one request and then torn down (in `FlinkUtils ::isZookeeperHaMetadataAvailable` it's the same thing). We could somehow cache these clients but I think it's not worth it. 

First of all, these ZK requests happen quite rarely and the overhead of starting one client per request is negligible IMO.  

Moreover, if we decide to cache the client, we'll need to cache one per Flink application. Since we use the application `conf` when instantiating the client, we would also need to re-create the client every time ZK configuration values change in the application `conf`, which adds even more unnecessary complexity.

gyfora:
I think from an initial review this looks pretty good. We just have to ensure that we handle the Zookeeper dependency as optional as this is the less frequent integration :)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
```diff
@@ -160,19 +163,31 @@ public KubernetesClient getKubernetesClient() {
         return kubernetesClient;
     }
 
+    public CuratorFrameworkWithUnhandledErrorListener getCurator(Configuration conf) {
+        return ZooKeeperUtils.startCuratorFramework(conf, exception -> {});
```
antonipp, line 167: Note that by leveraging the [ZooKeeperUtils::startCuratorFramework](https://github.com/apache/flink/blob/c57b84921447bb0ade5e1ff77a05ebd8bbbe71b7/flink-runtime/src/main/java/org/apache/flink/runtime/util/ZooKeeperUtils.java#L200) function and by taking the application `conf`, we create the client in exactly the same way as the Flink application itself. This means that we should be able to support all ZK configuration settings that are supported by Flink applications themselves (ex: ACLs).

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
```diff
@@ -160,19 +163,31 @@ public KubernetesClient getKubernetesClient() {
...
+        // the newly changed job config (e.g. parallelism) could take effect
         if (FlinkUtils.isKubernetesHAActivated(conf)) {
             final String clusterId = conf.get(KubernetesConfigOptions.CLUSTER_ID);
             final String namespace = conf.get(KubernetesConfigOptions.NAMESPACE);
-            // Delete the job graph in the HA ConfigMaps so that the newly changed job config(e.g.
-            // parallelism) could take effect
             FlinkUtils.deleteJobGraphInKubernetesHA(clusterId, namespace, kubernetesClient);
+        } else if (FlinkUtils.isZookeeperHAActivated(conf)) {
+            try (var curator = getCurator(conf)) {
+                ZooKeeperUtils.deleteZNode(
```
antonipp, line 185: [This comment](https://issues.apache.org/jira/browse/FLINK-27273?focusedCommentId=17552240&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17552240) on the Jira ticket suggested using `HighAvailabilityServicesUtils::createAvailableOrEmbeddedServices`.

I gave it a try with this piece of code (obviously not production ready but good enough for a POC):
```java
try (var availableOrEmbeddedServices =
                    HighAvailabilityServicesUtils.createAvailableOrEmbeddedServices(
                            conf, UnsupportedOperationExecutor.INSTANCE, exception -> {})) {
  availableOrEmbeddedServices
          .getJobGraphStore()
          .globalCleanupAsync(
                  availableOrEmbeddedServices.getJobGraphStore().getJobIds().stream()
                          .findFirst()
                          .get(),
                  Executors.newFixedThreadPool(1));
}
```

The problem is that under the hood this will call `HighAvailabilityServicesUtils::createZooKeeperHaServices` which in turn will [instantiate](https://github.com/apache/flink/blob/fb05a7be9b828b7e582e75e4832443806fa4ff17/flink-runtime/src/main/java/org/apache/flink/runtime/highavailability/HighAvailabilityServicesUtils.java#L89) a `BlobStoreService` object. I immediately ran into issues with S3 FileSystem initialization since this is the file system we use for `high-availability.storageDir`. This is why I switched to `ZooKeeperUtils::startCuratorFramework` instead.

As a side note, we could also authorize the Operator to access `high-availability.storageDir` and clean-up the data there as well (right now, both for Kubernetes HA and ZK HA, we only delete the pointer to the JobGraph from the HA metadata but we don't delete the JobGraph object itself). This adds a lot of complexity though, since we'll need a mechanism for adding relevant FS client dependencies and a mechanism for configuring these clients as well (i.e. specify credentials, retry options, etc).

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkUtils.java
```diff
@@ -133,7 +136,23 @@ public static void deleteJobGraphInKubernetesHA(
         }
     }
 
-    public static boolean isHaMetadataAvailable(
+    public static boolean isZookeeperHaMetadataAvailable(
+            Configuration conf, CuratorFrameworkWithUnhandledErrorListener curator) {
+        try (curator) {
+            return curator.asCuratorFramework().checkExists().forPath("/") != null;
```
antonipp, line 142: The ZK client is instantiated with `HighAvailabilityOptions.HA_ZOOKEEPER_ROOT / HighAvailabilityOptions.HA_CLUSTER_ID` as the namespace for all operations (set [here](https://github.com/apache/flink/blob/c57b84921447bb0ade5e1ff77a05ebd8bbbe71b7/flink-runtime/src/main/java/org/apache/flink/runtime/util/ZooKeeperUtils.java#L267)).
This code basically tests whether the root ZNode exists.

Note that `checkExists` will return `null` if the path doesn't exist (http://www.mtitek.com/tutorials/zookeeper/cf_check_exists.php)

docs/content/docs/concepts/overview.md
```diff
@@ -88,7 +88,7 @@ The examples are maintained as part of the operator repo and can be found [here]
 ## Known Issues & Limitations
 
 ### JobManager High-availability
```
antonipp, line 90: TBH I don't know why this was in the "Known Issues & Limitations" section? Maybe we can move it somewhere else?

antonipp:
Thanks a lot @gyfora!

Regarding:
> We just have to ensure that we handle the Zookeeper dependency as optional

The dependency was actually already there! I didn't have to add anything.

I looked into it and all Curator classes are coming from the `flink-clients` package which is listed as a dependency in `pom.xml` [here](https://github.com/antonipp/flink-kubernetes-operator/blob/57b2832d97a4d52d7fcd814d50b94f4e745ce027/flink-kubernetes-operator/pom.xml#L75-L79).
<pre>
$ mvn dependency:tree -pl 'flink-kubernetes-operator'
[...]
[INFO] +- org.apache.flink:flink-clients:jar:1.15.3:compile
[INFO] |  +- org.apache.flink:flink-runtime:jar:1.15.3:compile
[INFO] |  |  +- org.apache.flink:flink-rpc-core:jar:1.15.3:compile
[INFO] |  |  +- [...]
<b>[INFO] |  |  +- org.apache.flink:flink-shaded-zookeeper-3:jar:3.5.9-15.0:compile</b>
</pre>

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
gyfora: In that case I would suggest using the namespace + name combination itself to get a unique resource identifier. That is basically the same logic for Kubernetes HA

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
gyfora: I think it's fine to create the client on demand

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
gyfora: 👍

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
gyfora: The problem with HA storage access is both dependencies and credentials. Of which credentials is more problematic, it would open a big attack surface if the operator held storage creds of users somehow.

gyfora:
> Thanks a lot @gyfora!
> 
> Regarding:
> 
> > We just have to ensure that we handle the Zookeeper dependency as optional
> 
> The dependency was actually already there! I didn't have to add anything.
> 
> I looked into it and all Curator classes are a transitive dependency of the `flink-clients` package which is listed as a dependency in `pom.xml` [here](https://github.com/antonipp/flink-kubernetes-operator/blob/57b2832d97a4d52d7fcd814d50b94f4e745ce027/flink-kubernetes-operator/pom.xml#L75-L79).
> 
> $ mvn dependency:tree -pl 'flink-kubernetes-operator'
> [...]
> [INFO] +- org.apache.flink:flink-clients:jar:1.15.3:compile
> [INFO] |  +- org.apache.flink:flink-runtime:jar:1.15.3:compile
> [INFO] |  |  +- org.apache.flink:flink-rpc-core:jar:1.15.3:compile
> [INFO] |  |  +- [...]
> [INFO] |  |  +- org.apache.flink:flink-shaded-zookeeper-3:jar:3.5.9-15.0:compile

Hm, thats convenient :)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
antonipp: Yeah makes sense, then even more arguments for not doing this.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
antonipp: Unfortunately I couldn't implement this logic because the value of `HA_CLUSTER_ID` is overridden way deep in Flink's code 😞 

For Standalone applications it's overridden here in the Operator's code, so it's easy to change:
https://github.com/apache/flink-kubernetes-operator/blob/bf3e1d44506809fda9ec1ef75308496c7202bbff/flink-kubernetes-standalone/src/main/java/org/apache/flink/kubernetes/operator/standalone/KubernetesStandaloneClusterDescriptor.java#L140

However, for Native applications, it's overridden in Flink's code:
https://github.com/apache/flink/blob/ab3eb40d920fa609f49164a0bbb5fcbb3004a808/flink-kubernetes/src/main/java/org/apache/flink/kubernetes/KubernetesClusterDescriptor.java#L255

Since this requires changing logic in the main Flink project, I didn't touch the `HA_CLUSTER_ID` value in this Operator PR. It's probably worth doing as a separate ticket later.

antonipp:
I made some progress on the PR but still need to implement this HA data deletion logic for Zookeeper: 
https://github.com/apache/flink-kubernetes-operator/blob/d0b769c0d80fee526f5abda42673ae4c856fee3c/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java#L931-L938

PR #524, opened 2023-02-03
[FLINK-30405] Add ResourceLifecycleStatus to CommonStatus and printer column

zch93:

*I had to remove the JsonIgnore annotation above the ResourceLifecycleState getter in CommonStatus and also I had to expose this as a printer column instead of the reconciliation status that is used currently.*

## Brief change log

  - *Remove the `@JsonIgnore` annotation above the `ResourceLifecycleState` getter*
  - *Added a new private field to `CommonStatus` with `@PrinterColumn` annotation*
  - *Remove the `@PrinterColumn` annotation above the `state` field within the `ReconciliationStatus` class*

## Verifying this change

1. Checkout this PR
2. And go through the quick-start guide
3. Finally use the `kubectl get flinkdeployments` command to check the results

*(OR)*

1. (After you checked out this PR) Build the flink-kubernetes-operator with `mvn clean install -Pgenerate-docs -DskipTests` command
2. Upgrade the `flinkdeployments` and `flinksessionjobs` CRDs
4. Build a new docker image
5. Load that image into minikube
6. Install the operator locally
7. Submit the basic-example Flink job
8. Finally use the `kubectl get flinkdeployments` command to check the results

You should see something like this as result:
image

## Documentation

### DIFF
--- a/helm/flink-kubernetes-operator/crds/flinkdeployments.flink.apache.org-v1.yml
+++ b/helm/flink-kubernetes-operator/crds/flinkdeployments.flink.apache.org-v1.yml
@@ -18,8 +18,10 @@ spec:
       jsonPath: .status.jobStatus.state
       name: Job Status
       type: string
-    - jsonPath: .status.lifecycleState
-      name: Resource Lifecycle State
+    - description: "Lifecycle state of the Flink resource (including being rolled\
+        \ back, failed etc.)."
+      jsonPath: .status.lifecycleState
+      name: Lifecycle State
       type: string
     name: v1beta1
     schema:
@@ -9252,8 +9254,6 @@ spec:
                 - ROLLING_BACK
                 - ROLLED_BACK
                 - FAILED
-                - terminal
-                - description
                 type: string
             type: object
         type: object
--- a/helm/flink-kubernetes-operator/crds/flinkdeployments.flink.apache.org-v1.yml
+++ b/helm/flink-kubernetes-operator/crds/flinkdeployments.flink.apache.org-v1.yml
@@ -18,9 +18,8 @@ spec:
       jsonPath: .status.jobStatus.state
       name: Job Status
       type: string
-    - description: Deployment state of the last reconciled spec.
-      jsonPath: .status.reconciliationStatus.state
-      name: Reconciliation Status
+    - jsonPath: .status.lifecycleState
+      name: Resource Lifecycle State
       type: string
     name: v1beta1
     schema:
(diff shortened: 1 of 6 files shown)


helm/flink-kubernetes-operator/crds/flinkdeployments.flink.apache.org-v1.yml
```diff
@@ -9243,6 +9242,19 @@ spec:
...
+                enum:
+                - CREATED
+                - SUSPENDED
+                - UPGRADING
+                - DEPLOYED
+                - STABLE
+                - ROLLING_BACK
+                - ROLLED_BACK
+                - FAILED
+                - terminal
```
morhidi, line 9255: What is this exactly? :)

morhidi (changes requested):
I don't see in the code where we set this field?

zch93:
> I don't see in the code where we set this field?

My feeling is that, there are some implicit setter in the code. For example if I tried to add the `@PrinterColumn` annotation to fields from the `JobStatus` like `jobId` or `startTime`, or to any field from `ReconciliationStatus`, then despite there are not any explicit setter for those as well, the values appear when I run the `kubectl get flinkdeployments` command.

I think since the `StatusRecorder ` extends the `CommonStatus` that has an affect for the new field as well with an implicit setter. There is a connection between different layers like the following:
`ResourceLifecycleMetricTracker.onUpdate()` --> `LifecycleMetrics.onUpdate()`  --> `MetricManager.onUpdate()` --> StatusRecorder.patchAndCacheStatus()`  and `StatusRecorder` extends `CommonStatus`

helm/flink-kubernetes-operator/crds/flinkdeployments.flink.apache.org-v1.yml
zch93: For some reason the fabric8 crd gen includes all the fields from the `ResourceLifecycleState`, and this is why `terminal` and `description` also appear. This is not good I agree, maybe some annotation needed above those fields to ignore those by the crd generator. I can take a look at those tomorrow afternoon (CET). However, maybe that is not required to commit the automated generated yaml files, am I right?

mbalassi:
I will take a look during CET work hours to help wrap this up for 1.4.0.

mbalassi (commented):
I could verify locally that the change works from a functional perspective. I added I suggestion and one fix, pushing the code change myself so that we can unblock the upcoming release.

mbalassi (approved):
After my changes I am happy with this 😄

PR #525, opened 2023-02-03
[FLINK-27491] Support env replacement in flink-kubernetes-operator

morhidi:
Flink deployment resources support env interpolation natively using $() syntax. Users expected this to "just work" like other resources when using the operator, but it does not. It would be a great addition, simplifying job startup decision-making while following existing conventions.

```
job:  
  args:
    - "--sleep"
    - "${SLEEP}"
    - "--error-rate"
    - "${ERROR_RATE}"
```

Note: The feature is considered experymental and controlled by a flag `argument.variables.enabled=false`

## Brief change log

The logic uses the idea to append the `-D$internal.application.program-args=--sleep;${SLEEP};--error-rate;${ERROR_RATE}` as a dynamic property to `kubernetes.jobmanager.entrypoint.args` which results in the following container command:
```
 - args:
    - bash
    - -c
    - kubernetes-jobmanager.sh kubernetes-application -D\$internal.application.program-args=--sleep\;${SLEEP}\;--error-rate\;${ERROR_RATE}
```
## Verifying this change

- Unit tests where added to cover the functionality
- Also verified manually by running:
```
kind: FlinkDeployment
metadata:
  name: basic-example
spec:
  image: flink:1.16
  flinkVersion: v1_16
  flinkConfiguration:
  serviceAccount: flink
  podTemplate:
    apiVersion: v1
    kind: Pod
    metadata:
      name: pod-template
    spec:
      containers:        
        - name: flink-main-container
          env:
            - name: SLEEP
              value: "10"
            - name: ERROR_RATE
              value: "0.1"
  jobManager:
    resource:
      memory: "2048m"
      cpu: 1
  taskManager:
    resource:
      memory: "2048m"
      cpu: 1
  job:
    args: ["--sleep","${SLEEP}","--error-rate", "${ERROR_RATE}"]
    parallelism: 2
    upgradeMode: stateless
```

## Documentation

  - If yes, how is the feature documented? (autoganarated)

### DIFF
--- a/docs/layouts/shortcodes/generated/dynamic_section.html
+++ b/docs/layouts/shortcodes/generated/dynamic_section.html
@@ -9,6 +9,12 @@
     </thead>
     <tbody>
         <tr>
+            <td><h5>kubernetes.operator.argument.variables.enabled</h5></td>
+            <td style="word-wrap: break-word;">false</td>
+            <td>Boolean</td>
+            <td>Whether to resolve job args containing environment variables, e.g. '--args ${ENV}'</td>
+        </tr>
+        <tr>
             <td><h5>kubernetes.operator.cluster.health-check.checkpoint-progress.enabled</h5></td>
             <td style="word-wrap: break-word;">false</td>
             <td>Boolean</td>
--- a/docs/layouts/shortcodes/generated/kubernetes_operator_config_configuration.html
+++ b/docs/layouts/shortcodes/generated/kubernetes_operator_config_configuration.html
@@ -9,6 +9,12 @@
     </thead>
     <tbody>
         <tr>
+            <td><h5>kubernetes.operator.argument.variables.enabled</h5></td>
+            <td style="word-wrap: break-word;">false</td>
+            <td>Boolean</td>
+            <td>Whether to resolve job args containing environment variables, e.g. '--args ${ENV}'</td>
+        </tr>
+        <tr>
             <td><h5>kubernetes.operator.cluster.health-check.checkpoint-progress.enabled</h5></td>
             <td style="word-wrap: break-word;">false</td>
             <td>Boolean</td>
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
@@ -20,6 +20,7 @@ package org.apache.flink.kubernetes.operator.config;
 import org.apache.flink.client.deployment.application.ApplicationConfiguration;
 import org.apache.flink.configuration.ConfigOption;
 import org.apache.flink.configuration.ConfigOptions;
+import org.apache.flink.configuration.ConfigUtils;
 import org.apache.flink.configuration.Configuration;
 import org.apache.flink.configuration.CoreOptions;
 import org.apache.flink.configuration.DeploymentOptions;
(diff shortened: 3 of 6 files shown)


morhidi:
cc @gaborgsomogyi @pvary

docs/layouts/shortcodes/generated/dynamic_section.html
```diff
@@ -8,6 +8,12 @@
         </tr>
     </thead>
     <tbody>
+        <tr>
+            <td><h5>kubernetes.operator.argument.variables.enabled</h5></td>
+            <td style="word-wrap: break-word;">false</td>
+            <td>Boolean</td>
```
gaborgsomogyi, line 14: Not sure what would be a good name but the feature is job args env var resolving, right? If one takes a look purely at the conf name then not able to tell the main intention. It would be good to give a name which reflects resolving. Maybe 'kubernetes.operator.job.arguments.env-var.resolve.enabled' or something.

gaborgsomogyi (commented):
Looks good, only a minor suggestion.

docs/layouts/shortcodes/generated/dynamic_section.html
gyfora: `kubernetes.operator.flink-config.env-variables.enabled` or something like that

gyfora: ah okay, I was under the impression that this PR worked for all flink configs

morhidi: This is already given by `kubernetes.jobmanager.entrypoint.args`, the extra that we're adding here is to add the internal args automatically, avoid users manually dealing with the property`-D\$internal.application.program-args=--sleep\;${SLEEP}\;--error-rate\;${ERROR_RATE}`

gyfora: It's somewhat of a hidden feature technically. I thought instead of just applying the current logic to job args we could apply it to all Flink Config key/value pairs.

That would cover both args and all else too.

morhidi: With `kubernetes.jobmanager.entrypoint.args` you already getting this functionality you're requesting. If you think it's sufficient we can simply close out this one.

gyfora: I think it's nice if the operator provides these convenience features, so the user doesn't need to figure it out how to do this.

So I completely agree with this PR, all I am trying to say is that we could extend the scope to the entire spec.flinkConfiguration not only to spec.job.args

gyfora: but we don't have to do this right now, we can also add this later

morhidi: Ah ok I guess I see your point. So you want automatic substitution of:
```
security.ssl.internal.truststore: ${TRUSTSTORE}
security.ssl.internal.truststore-password: ${TRUSTSTORE_PASSWROD}
```

Instead of:

```
kubernetes.jobmanager.entrypoint.args: -Dsecurity.ssl.internal.truststore: ${TRUSTSTORE} -Dsecurity.ssl.internal.truststore-password: ${TRUSTSTORE_PASSWROD}
```

I'd prefer to do this in a separate jira.

gyfora: The only problem then is that we will have to create a new config and immediately deprecate the one you just created now, because the other will completely subsume it. 

We could also just take a little more time and do this properly instead of rushing this in for the release. Users already have a workaround and I would prefer to get this right at first.

morhidi: ok, closing the pr for the time being

seano288:
I was wondering if this is still in the works? If not, what is the recommended way to get variable replacement in the job args?

vbabenkoru:
Yeah, is there a proper way of doing this?

PR #529, opened 2023-02-09
[FLINK-30776] Move autoscaler code into a separate Maven module

mxm:
The autoscaling code is tightly coupled with the main operator code. In an
effort to make it pluggable or run independently, we should move the code into a
dedicated Maven module.


flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/JobAutoscalerFactoryImpl.java
```diff
@@ -0,0 +1,41 @@
...
+import org.apache.flink.kubernetes.operator.utils.EventRecorder;
+
+import com.google.auto.service.AutoService;
+import io.fabric8.kubernetes.client.KubernetesClient;
+
+/**
+ * Factory for loading JobAutoScalerImpl included in this module. This class will be dynamically
+ * instantiated by the main operator module.
+ */
+@AutoService(JobAutoScalerFactory.class)
```
gyfora, line 31: I think we should follow the same Plugin mechanism that is used by other components (metrics reporters, validators, resource listeners).

I am not against innovating in the plugin loading space but we should not introduce a different mechanism for a single module

mxm: This is still required. All plugins have to declare a META-INF/service file which this annotation does automatically.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
```diff
@@ -468,6 +468,16 @@ protected void setOwnerReference(CR owner, Configuration deployConfig) {
                 KubernetesConfigOptions.JOB_MANAGER_OWNER_REFERENCE, List.of(ownerReference));
     }
 
+    private JobAutoScaler loadJobAutoscaler() {
+        for (JobAutoScalerFactory factory : ServiceLoader.load(JobAutoScalerFactory.class)) {
+            LOG.info("Loading JobAutoScaler implementation: {}", factory.getClass().getName());
+            return factory.create(kubernetesClient, eventRecorder);
+        }
+        LOG.info("No JobAutoscaler implementation found. Autoscaling is disabled.");
+        return new NoopJobAutoscaler();
```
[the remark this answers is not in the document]
mxm: See the latest commit.

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/JobAutoscalerFactoryImpl.java
gyfora: I think we should not introduce this as part of the PR. We should have a separate jira/commit and then apply for all the plugins within the operator project. It doesn't look too good to introduce a new dependency and only apply it for one out of 3 plugins

mxm: The other built-in plugins do not require this process because they are bundled together with the operator and never dynamically created. Any custom plugin would need this. This dependency only applies for the new autoscaler module.

gyfora: makes sense

gyfora (approved):

PR #536, opened 2023-02-23
[FLINK-31140] - Load additional dependencies in operator classpath

tamirsagi:

Allow users to add additional dependencies onto operator classpath

## Brief change log

  - Create a target folder with a dedicated environment variable for user dependencies in root dockerFile

  - Dependencies (does it add or upgrade a dependency): (yes / no) no
  - The public API, i.e., is any changes to the `CustomResourceDescriptors`: (yes / no)  no
  - Core observer or reconciler logic that is regularly executed: (yes / no)  no

## Documentation

  - Does this pull request introduce a new feature? (yes / no) yes
  - If yes, how is the feature documented? (not applicable / docs / JavaDocs / not documented) docs

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,12 +36,15 @@ RUN cd /app/tools/license; mkdir jars; cd jars; \
 # stage
 FROM eclipse-temurin:11-jre-jammy
 ENV FLINK_HOME=/opt/flink
-ENV FLINK_PLUGINS_DIR=/opt/flink/plugins
+ENV FLINK_PLUGINS_DIR=$FLINK_HOME/plugins
 ENV OPERATOR_VERSION=1.5-SNAPSHOT
 ENV OPERATOR_JAR=flink-kubernetes-operator-$OPERATOR_VERSION-shaded.jar
 ENV AUTOSCALER_JAR=flink-kubernetes-operator-autoscaler-$OPERATOR_VERSION.jar
 ENV WEBHOOK_JAR=flink-kubernetes-webhook-$OPERATOR_VERSION-shaded.jar
 ENV FLINK_KUBERNETES_SHADED_JAR=flink-kubernetes-standalone-$OPERATOR_VERSION-shaded.jar
+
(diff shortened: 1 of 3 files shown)


tamirsagi:
@gyfora  Would you please  review?

Dockerfile
```diff
@@ -36,13 +36,16 @@ RUN cd /app/tools/license; mkdir jars; cd jars; \
...
-ENV FLINK_PLUGINS_DIR=/opt/flink/plugins
+ENV FLINK_PLUGINS_DIR=$FLINK_HOME/plugins
 ENV OPERATOR_VERSION=1.5-SNAPSHOT
 ENV OPERATOR_JAR=flink-kubernetes-operator-$OPERATOR_VERSION-shaded.jar
 ENV AUTOSCALER_JAR=flink-kubernetes-operator-autoscaler-$OPERATOR_VERSION.jar
 ENV WEBHOOK_JAR=flink-kubernetes-webhook-$OPERATOR_VERSION-shaded.jar
 ENV FLINK_KUBERNETES_SHADED_JAR=flink-kubernetes-standalone-$OPERATOR_VERSION-shaded.jar
 
+ENV USER_DEPENDENCIES_DIR=$FLINK_HOME/dependencies
+RUN mkdir -p $USER_DEPENDENCIES_DIR
```
asardaes, line 47: Why not use the same convention as Flink's application mode and name the directory `usrlib`?

tamirsagi: I did not want to mix up user jar located under 'userlib' in Flink application and user dependencies on Flink operator.

asardaes: Mm I don't have a strong opinion either way, if you think `usrlib` should have different semantics, I think that's fine.

tamirsagi: @gyfora  what do you think about this?

gyfora: I think this is not usrlib, the operator is controlled by the platform not the user. We could even call it `$FLINK_HOME/operator-lib`

tamirsagi: agree. changed it.

PR #538, opened 2023-02-27
[hotfix] Fix autoscaler module mockwebserver tests

gyfora:
Currently the mockwebserver tests for the autoscaler module don't run.

There is still one issue I need to fix before merging this.


mxm (approved):

mateczagany:
@gyfora  it seems the autoscaler tests are still not included in the e2e CI actions, is that intentional?
https://github.com/apache/flink-kubernetes-operator/blob/8a30e9bd770b32e28d2a7c0fe1830f5f6d9ab090/.github/workflows/ci.yml#L82

gyfora:
@mateczagany This is unrelated to what was fixed in this PR but you are compeltely right, the CI conf is now wrong. cc @mxm

PR #539, opened 2023-02-28
[FLINK-31246] Beautify the SpecChange message

pvary:

Currently the Spec Change message contains the full PodTemplate twice.
This makes the message seriously big and also contains very little useful information.

We should abbreviate the message

## Brief change log

- When we found different pod templates, we just print `before` and `after` instead of printing out the pod template itself

## Verifying this change

*(example:)*
  - *Added unit test verifying the text*

## Documentation

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/diff/DiffResult.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/diff/DiffResult.java
@@ -21,10 +21,10 @@ import org.apache.flink.annotation.Experimental;
 import org.apache.flink.kubernetes.operator.api.diff.DiffType;
 import org.apache.flink.kubernetes.operator.api.diff.Diffable;
 
-import io.fabric8.zjsonpatch.JsonDiff;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import io.fabric8.zjsonpatch.JsonDiff;
 import lombok.Getter;
 import lombok.NonNull;
 
@@ -98,9 +98,11 @@ public class DiffResult<T> {
                                                 .append(", ");
                                     }
                                 });
+                        builder.setLength(builder.length() - 2);
                     } catch (JsonProcessingException je) {
                         builder.append(diff.getLeft()).append(" -> ").append(diff.getRight());
                     }
+                    builder.append(", ");
                 });
         builder.setLength(builder.length() - 2);
         builder.append("]");
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/diff/DiffResult.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/diff/DiffResult.java
@@ -100,7 +100,11 @@ public class DiffResult<T> {
                                 });
                         builder.setLength(builder.length() - 2);
                     } catch (JsonProcessingException je) {
-                        builder.append(diff.getLeft()).append(" -> ").append(diff.getRight());
+                        builder.append(diff.getFieldName())
+                                .append(" : ")
+                                .append(diff.getLeft())
+                                .append(" -> ")
+                                .append(diff.getRight());
                     }
                     builder.append(", ");
                 });
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/diff/DiffResult.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/diff/DiffResult.java
@@ -21,7 +21,6 @@ import org.apache.flink.annotation.Experimental;
 import org.apache.flink.kubernetes.operator.api.diff.DiffType;
 import org.apache.flink.kubernetes.operator.api.diff.Diffable;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.fabric8.zjsonpatch.JsonDiff;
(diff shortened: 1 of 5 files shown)


pvary:
@tweise, @gyfora: Could you please take a look.

This is the proposed result after the change:
```
Diff: FlinkDeploymentSpec[image : "img1" -> "img2", ingress : {"template":"temp","className":null,"annotations":null} -> null, podTemplate.apiVersion : "v1" -> "v2", podTemplate.spec.hostname : "localhost1" -> "localhost2", restartNonce : null -> 1]
```

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/diff/DiffResult.java
```diff
@@ -59,18 +63,48 @@ public String toString() {
...
-        final ToStringBuilder rhsBuilder =
-                new ToStringBuilder(right, ToStringStyle.SHORT_PREFIX_STYLE);
+        final StringBuilder builder = new StringBuilder();
+        builder.append(left.getClass().getSimpleName()).append("[");
 
         diffList.forEach(
                 diff -> {
-                    lhsBuilder.append(diff.getFieldName(), diff.getLeft());
-                    rhsBuilder.append(diff.getFieldName(), diff.getRight());
+                    try {
```
tamirsagi, line 71: This part should be extracted to a method

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/diff/DiffResult.java
```diff
@@ -59,18 +63,48 @@ public String toString() {
...
+                        JsonNode before =
+                                objectMapper.readTree(
+                                        objectMapper.writeValueAsString(diff.getLeft()));
+                        JsonNode after =
+                                objectMapper.readTree(
+                                        objectMapper.writeValueAsString(diff.getRight()));
+                        JsonNode jsonDiff = JsonDiff.asJson(before, after);
+                        jsonDiff.forEach(
+                                row -> {
```
tamirsagi, line 80: that clause should be extracted to another method

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/diff/DiffResult.java
pvary: Which part do you think should be extracted to a method, and why?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/diff/DiffResult.java
pvary: Again, I am not sure what part do you think would merit its own method?
The block should be quite easy to understand, just an if statement with a bunch of append statements.
I am not sure that putting it into a method would help with readability

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/diff/DiffResult.java
```diff
@@ -59,18 +63,48 @@ public String toString() {
...
+                                                .append(", ");
+                                    }
+                                });
+                    } catch (JsonProcessingException je) {
+                        builder.append(diff.getLeft()).append(" -> ").append(diff.getRight());
+                    }
                 });
-
-        return String.format("%s differs from %s", lhsBuilder.build(), rhsBuilder.build());
+        builder.setLength(builder.length() - 2);
```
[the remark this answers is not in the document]
pvary: Actually this was in a wrong place, I have moved it.
OTOH again this is something which is done quite often in my experience - removing the last separator between the fields after writing out things in a loop.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/reconciler/diff/SpecDiffTest.java
```diff
@@ -181,4 +181,28 @@ public void testFlinkSessionJobSpecChanges() {
...
+        var right = BaseTestUtils.buildApplicationCluster().getSpec();
+        right.setPodTemplate(BaseTestUtils.getTestPod("localhost2", "v2", List.of()));
+        right.setImage("img2");
+        right.setRestartNonce(1L);
+
+        var diff = new ReflectiveDiffBuilder<>(left, right).build();
+        assertEquals(DiffType.UPGRADE, diff.getType());
+        assertEquals(
+                "Diff: FlinkDeploymentSpec[image : \"img1\" -> \"img2\", "
+                        + "ingress : {\"template\":\"temp\",\"className\":null,\"annotations\":null} -> null, "
```
gyfora, line 202: We could potentiall improve the behaviour when nulling out objects. It should not be necessary to show the entire left side object . Maybe something like `ingress : {...} -> null`

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/diff/DiffResult.java
tamirsagi: IMHO,  Extracting multiple lines within a loop to a well named method help others to understand the intention behind that piece of code. it makes the code smaller, help with debugging and error handling when needed.

i.e , obj.forEach(x -> createDiffObject(x))

I see that in that case there is another loop for every row, it would be nicer to have another small method which indicates the operation per row.(IMO)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/diff/DiffResult.java
```diff
@@ -59,18 +63,54 @@ public String toString() {
...
+                                                                .replaceAll("/", "."))
+                                                .append(" : ")
+                                                .append(before.at(row.get("path").asText()))
+                                                .append(" -> ")
+                                                .append(after.at(row.get("path").asText()))
+                                                .append(", ");
+                                    }
+                                });
+                        builder.setLength(builder.length() - 2);
+                    } catch (JsonProcessingException je) {
```
gyfora, line 102: Maybe we can be a bit more careful here and catch all kinds of exceptions, to be on the safe side :)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/diff/DiffResult.java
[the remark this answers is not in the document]
tamirsagi: yes that's true, my point was to add a small comment next to it. "Removing last 2 .."
Alternatively,  wrap it with another method called "excludingLastTwoSeparators".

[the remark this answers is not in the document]
gyfora: I think this would be an overkill @tamirsagi , I think this is readable and we should also give some freedom to @pvary :)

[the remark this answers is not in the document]
tamirsagi: fair enough :)

pvary:
After some discussion with @gyfora, the message looks like this:

> Diff: FlinkDeploymentSpec[image : img1 -> img2, ingress : {..} -> null, podTemplate.apiVersion : v1 -> v2, podTemplate.spec.hostAliases.0.hostnames.1 : host2 -> null, podTemplate.spec.hostname : localhost1 -> localhost2, restartNonce : null -> 1]

PR #540, opened 2023-02-28
Flink 29566

haoxins:
https://issues.apache.org/jira/browse/FLINK-29566

Reschedule the cleanup logic if cancel job failed

## Brief change log

- Reschedule the cleanup logic if cancel job failed

## Verifying this change

This change added tests

## Documentation


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/sessionjob/SessionJobReconciler.java
```diff
@@ -100,6 +106,26 @@ public DeleteControl cleanupInternal(FlinkResourceContext<FlinkSessionJob> ctx)
...
+                        LOG.error("Job {} not found in the Flink cluster.", jobID, e);
+                        return DeleteControl.defaultDelete();
+                    }
+
+                    if (cause instanceof FlinkJobTerminatedWithoutCancellationException) {
+                        LOG.error("Job {} already terminated without cancellation.", jobID, e);
+                        return DeleteControl.defaultDelete();
+                    }
+
+                    final var delay = 10_000L;
```
tamirsagi, line 122: What about allow users to define that value via Flink configurations?

haoxins: good point ~

gyfora: We could also use the KubernetesOperatorConfigOptions.OPERATOR_OBSERVER_PROGRESS_CHECK_INTERVAL that we use for similar purposes. That also defaults to 10s

gyfora (commented):
Looks pretty good!
Added a minor comment on the delay, after that it should be good to go if the tests pass :)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/sessionjob/SessionJobReconciler.java
```diff
@@ -24,30 +24,41 @@
...
 import java.util.Optional;
+import java.util.concurrent.ExecutionException;
 
 /** The reconciler for the {@link FlinkSessionJob}. */
 public class SessionJobReconciler
         extends AbstractJobReconciler<FlinkSessionJob, FlinkSessionJobSpec, FlinkSessionJobStatus> {
 
     private static final Logger LOG = LoggerFactory.getLogger(SessionJobReconciler.class);
 
+    private UpdateControl updateControl;
```
gyfora, line 51: Can you please remove this field? I think it's unused

gyfora (changes requested):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/sessionjob/SessionJobReconciler.java
haoxins: done

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/sessionjob/SessionJobReconciler.java
haoxins: done

PR #542, opened 2023-03-02
[FLINK-31299] PendingRecords metric might not be available

mxm:
The Kafka pendingRecords metric is only initialized on receiving the first record. For empty topics or checkpointed topics without any incoming data, the metric won't appear.

We need to handle this case in the autoscaler and allow downscaling.

### DIFF
--- a/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/metrics/ScalingMetrics.java
+++ b/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/metrics/ScalingMetrics.java
@@ -156,6 +156,8 @@ public class ScalingMetrics {
         var pendingRecords = flinkMetrics.get(FlinkMetric.PENDING_RECORDS);
         if (pendingRecords != null) {
             scalingMetrics.put(ScalingMetric.LAG, pendingRecords.getSum());
+        } else {
+            scalingMetrics.put(ScalingMetric.LAG, 0.);
         }
     }
 }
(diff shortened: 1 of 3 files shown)


flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/metrics/ScalingMetrics.java
```diff
@@ -156,6 +156,8 @@ public static void computeLagMetrics(
         var pendingRecords = flinkMetrics.get(FlinkMetric.PENDING_RECORDS);
         if (pendingRecords != null) {
             scalingMetrics.put(ScalingMetric.LAG, pendingRecords.getSum());
+        } else {
+            scalingMetrics.put(ScalingMetric.LAG, 0.);
```
gyfora, line 160: Wouldn't this logic affect sources that do not report the pending records metric at all ?
Those currently would go through a different codepath in: https://github.com/apache/flink-kubernetes-operator/blob/main/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/metrics/ScalingMetrics.java#L111

mxm: I think both can share the same logic now. The linked code may be dead code now.

gyfora: Ok, in that case can you please remove the parts that are not necessary anymore?

gyfora (approved):

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/metrics/ScalingMetrics.java
mxm: I'll follow up with a cleanup PR.

PR #546, opened 2023-03-07
[FLINK-31326] Consolidate source scaling logic

mxm:
Previous changes removed the requirement for sources to have the `pendingRecords` metric. This allows to refactor and consolidate the source scaling code. We remove the option to disable scaling sources because the primary motivation was to skip scaling legacy sources which we can achieve without an extra configuration option. 

The refactored source scaling logic works like this:

* Modern sources (busy time available)

Scale source based on the busy time up. Add extra capacity if the lag
information, provided by the pending records metric, is available. Otherwise
assume 0 lag. Cap by the maximum number of partitions, if available.

* Legacy sources (busy time unavailable)

Leave the source parallelism unchanged, use the current processing rate as the
target rate downstream.

### DIFF
--- a/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/metrics/MetricAggregator.java
+++ b/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/metrics/MetricAggregator.java
@@ -19,7 +19,6 @@ package org.apache.flink.kubernetes.operator.autoscaler.metrics;
 
 import org.apache.flink.runtime.rest.messages.job.metrics.AggregatedMetric;
 
-import java.util.Optional;
 import java.util.function.Function;
 
 /** Enum specifying which aggregator to use when getting a metric value. */
@@ -34,7 +33,11 @@ public enum MetricAggregator {
         this.getter = getter;
     }
 
-    public Optional<Double> get(AggregatedMetric metric) {
-        return Optional.ofNullable(metric).map(getter).filter(d -> !d.isNaN());
+    public double get(AggregatedMetric metric) {
+        if (metric != null) {
+            return getter.apply(metric);
+        } else {
+            return Double.NaN;
+        }
     }
 }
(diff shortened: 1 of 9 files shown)


gyfora (approved):
Looks great!

mxm:
Thanks for testing @mateczagany! Good catch. I've removed the obsolete config entry in the example yml.

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/metrics/MetricAggregator.java
```diff
@@ -34,7 +33,11 @@ public enum MetricAggregator {
...
     }
 
-    public Optional<Double> get(AggregatedMetric metric) {
-        return Optional.ofNullable(metric).map(getter).filter(d -> !d.isNaN());
+    public double get(AggregatedMetric metric) {
+        if (metric != null) {
+            return getter.apply(metric);
+        } else {
+            return Double.NaN;
+        }
```
gyfora, line 41: This change here could be the reason why the lag growth rate may be nan after this.
Before this commit after your pending record related change the lag would be 0 , and now this may return Double.nan.

mxm: We discussed that this may delay decisions in case of NaNs but should not cause issues.

gyfora (approved):
🚢

gyfora (changes requested):
I found a bug here while testing this further. The logic for faking the busyTime for sources where this is not available:
```
return conf.get(AutoScalerOptions.TARGET_UTILIZATION) * 1000;
```
Is not sufficient to prevent scaling of these vertices. This is because lag, restart-time, and catchup duration all factor in the decision.

I noticed that these sources would be scaled up under some circumstances from their original parallelism.

The fix would be to mark these jobvertices explicitly excluded from scaling through the JobTopology. Then we can pass the jobtopology object to the ScalingExecutor and where we exclude the configured exluded vertices we can also exclude the ones that were marked through the topology.

mxm:
Good point. This would only affect legacy sources which do not expose their busy time. I've pushed a fix.

gyfora (approved):

PR #547, opened 2023-03-07
[FLINK-31345] Reduce AutoScalerInfo size by rounding metrics and compression

gyfora:

Compresses (GZip) the autoscaler info (metrics and scaling history) before storing in the configmap to greately reduce the size and allow keeping longer histories without reaching the 1MB limit.

We also introduce a 3 decimal place rounding for metrics to further reduce size and make the reported metrics nicer.

## Verifying this change

Change is covered by existing tests.

## Documentation

### DIFF
--- a/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/AutoScalerInfo.java
+++ b/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/AutoScalerInfo.java
@@ -35,16 +35,24 @@ import io.fabric8.kubernetes.api.model.ConfigMap;
 import io.fabric8.kubernetes.api.model.ObjectMeta;
 import io.fabric8.kubernetes.client.KubernetesClient;
 import lombok.SneakyThrows;
+import org.apache.commons.io.IOUtils;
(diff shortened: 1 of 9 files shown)


flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/AutoScalerInfo.java
```diff
@@ -119,7 +127,7 @@ public Map<JobVertexID, SortedMap<Instant, ScalingSummary>> getScalingHistory()
         if (scalingHistory != null) {
             return scalingHistory;
         }
-        var yaml = configMap.getData().get(SCALING_HISTORY_KEY);
+        var yaml = decompress(configMap.getData().get(SCALING_HISTORY_KEY));
```
mxm, line 130: Does that mean all existing ConfigMaps have to be deleted? Could we somehow achieve a graceful migration?

gyfora: Migration should fall back to uncompressed data so it should require no intervention. I just added a test to verify this

mxm (approved):

PR #548, opened 2023-03-13
[FLINK-31407] Bump fabric8 version to 6.5.0

mbalassi:
This new version brings critical stability fixes for the informers. I fixed the newly introduced deprecation warnings.

### DIFF
--- a/flink-kubernetes-standalone/src/main/java/org/apache/flink/kubernetes/operator/kubeclient/Fabric8FlinkStandaloneKubeClient.java
+++ b/flink-kubernetes-standalone/src/main/java/org/apache/flink/kubernetes/operator/kubeclient/Fabric8FlinkStandaloneKubeClient.java
@@ -23,6 +23,7 @@ import org.apache.flink.kubernetes.configuration.KubernetesConfigOptions;
 import org.apache.flink.kubernetes.kubeclient.Fabric8FlinkKubeClient;
 import org.apache.flink.kubernetes.operator.utils.StandaloneKubernetesUtils;
 
+import io.fabric8.kubernetes.api.model.DeletionPropagation;
 import io.fabric8.kubernetes.api.model.apps.Deployment;
 import io.fabric8.kubernetes.client.DefaultKubernetesClient;
 import io.fabric8.kubernetes.client.NamespacedKubernetesClient;
@@ -57,14 +58,14 @@ public class Fabric8FlinkStandaloneKubeClient extends Fabric8FlinkKubeClient
                 .apps()
                 .deployments()
                 .withName(StandaloneKubernetesUtils.getJobManagerDeploymentName(clusterId))
-                .cascading(true)
+                .withPropagationPolicy(DeletionPropagation.FOREGROUND)
                 .delete();
 
         this.internalClient
                 .apps()
                 .deployments()
                 .withName(StandaloneKubernetesUtils.getTaskManagerDeploymentName(clusterId))
-                .cascading(true)
+                .withPropagationPolicy(DeletionPropagation.FOREGROUND)
                 .delete();
     }
 
--- a/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/AutoScalerInfo.java
+++ b/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/AutoScalerInfo.java
@@ -181,7 +181,7 @@ public class AutoScalerInfo {
 
     public void replaceInKubernetes(KubernetesClient client) throws Exception {
         trimHistoryToMaxCmSize();
-        client.resource(configMap).replace();
+        client.resource(configMap).update();
     }
 
     @VisibleForTesting
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
@@ -808,7 +808,6 @@ public abstract class AbstractFlinkService implements FlinkService {
                                 .inNamespace(namespace)
                                 .withName(
                                         ExternalServiceDecorator.getExternalServiceName(clusterId))
-                                .fromServer()
                                 .get();
                 if (service == null) {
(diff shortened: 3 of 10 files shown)


flink-kubernetes-standalone/src/main/java/org/apache/flink/kubernetes/operator/kubeclient/Fabric8FlinkStandaloneKubeClient.java
```diff
@@ -57,14 +58,14 @@ public void stopAndCleanupCluster(String clusterId) {
                 .apps()
                 .deployments()
                 .withName(StandaloneKubernetesUtils.getJobManagerDeploymentName(clusterId))
-                .cascading(true)
+                .withPropagationPolicy(DeletionPropagation.ORPHAN)
```
mbalassi, line 61: @usamj could you please confirm that this was the intention here.

mbalassi:
The changes broke our tests that rely on the `KubernetesMockClient`, e.g.:

```
Caused by: io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: PATCH at: https://localhost:54301/apis/flink.apache.org/v1beta1/namespaces/flink-operator-test/flinkdeployments/test-cluster?fieldManager=fabric8. Message: Not Found.
	at io.fabric8.kubernetes.client.dsl.internal.OperationSupport.requestFailure(OperationSupport.java:701)
```

I will come back to this a bit later, any pointers are welcome. 😏

flink-kubernetes-standalone/src/main/java/org/apache/flink/kubernetes/operator/kubeclient/Fabric8FlinkStandaloneKubeClient.java
gyfora: Why note foreground? Orphan would not delete pods, configmaps etc right?

gaborgsomogyi:
I've just had a deeper look at the test issues and here are my findings:
* The originally used `createOrReplace` has the following semantics
  * Create the resource with `POST` operation
  * If response is HTTP_CONFLICT (409) then replace the resource with `PUT` operation (replace operation is effectively a forced update with retries)
* The `createOrReplace` doc says the following:
  ```
  /**
   * Creates a provided resource in a Kubernetes Cluster. If creation
   * fails with a HTTP_CONFLICT, it tries to replace resource.
   *
   * @return created item returned in kubernetes api response
   *
   * @deprecated please user {@link ServerSideApplicable#serverSideApply()} or attempt a create and edit/patch operation.
   */
  @Deprecated
  T createOrReplace();
  ```
* The doc is cloudy what `serverSideApply` really does so had an in-depth look of the source code.
* In case of custom resource the code is unconditionally patch the resource with `PATCH` operation

My general conclusion is that `createOrReplace` and `serverSideApply` are having totally different semantics if we consider the source code. My proposal is the following:
* Use `create` when new resource is needed
* Use `update` when existing resource need to be upgraded (since `replace` is deprecated too)
* If it's super hard to decide whether create or update is needed then one can re-implement to original `createOrReplace` functionality but I discourage to do it

gyfora:
> I've just had a deeper look at the test issues and here are my findings:
> 
> * The originally used `createOrReplace` has the following semantics
>   
>   * Create the resource with `POST` operation
>   * If response is HTTP_CONFLICT (409) then replace the resource with `PUT` operation (replace operation is effectively a forced update with retries)
> * The `createOrReplace` doc says the following:
>   ```
>   /**
>    * Creates a provided resource in a Kubernetes Cluster. If creation
>    * fails with a HTTP_CONFLICT, it tries to replace resource.
>    *
>    * @return created item returned in kubernetes api response
>    *
>    * @deprecated please user {@link ServerSideApplicable#serverSideApply()} or attempt a create and edit/patch operation.
>    */
>   @Deprecated
>   T createOrReplace();
>   ```
> * The doc is cloudy what `serverSideApply` really does so had an in-depth look of the source code.
> * In case of custom resource the code is unconditionally patch the resource with `PATCH` operation
> 
> My general conclusion is that `createOrReplace` and `serverSideApply` are having totally different semantics if we consider the source code. My proposal is the following:
> 
> * Use `create` when new resource is needed
> * Use `update` when existing resource need to be upgraded (since `replace` is deprecated too)
> * If it's super hard to decide whether create or update is needed then one can re-implement to original `createOrReplace` functionality but I discourage to do it

I agree with @gaborgsomogyi 's assessment. If we need we can keep the serverSideApply in some tests but we should not use it in the core logic.

flink-kubernetes-standalone/src/main/java/org/apache/flink/kubernetes/operator/kubeclient/Fabric8FlinkStandaloneKubeClient.java
mbalassi: That makes more sense, to me - the javadoc was ambiguous tbh. I prefer foreground unless @usamj or @dannycranmer disagrees.

flink-kubernetes-standalone/src/main/java/org/apache/flink/kubernetes/operator/kubeclient/Fabric8FlinkStandaloneKubeClient.java
```diff
@@ -57,14 +58,14 @@ public void stopAndCleanupCluster(String clusterId) {
...
                 .delete();
 
         this.internalClient
                 .apps()
                 .deployments()
                 .withName(StandaloneKubernetesUtils.getTaskManagerDeploymentName(clusterId))
-                .cascading(true)
+                .withPropagationPolicy(DeletionPropagation.ORPHAN)
```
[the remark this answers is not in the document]
mbalassi: Same as https://github.com/apache/flink-kubernetes-operator/pull/548#discussion_r1142261062.

flink-kubernetes-standalone/src/main/java/org/apache/flink/kubernetes/operator/kubeclient/Fabric8FlinkStandaloneKubeClient.java
usamj: I agree, foreground is more in-line with the behaviour that we want.

gyfora (approved):

PR #549, opened 2023-03-17
[FLINK-31502] Limit the number of scale operations to reduce cluster churn

mxm:
Until we move to using the upcoming Rescale API which recycles pods, we need to be mindful with how many deployments we scale at the same time because each of them is going to give up all its pods and require the new number of required pods.

This can cause churn in the cluster and temporary lead to "unallocatable" pods which triggers the k8s cluster autoscaler to add more cluster nodes. That is often not desirable because the actual required resources after the scalings have been settled, are lower.

### DIFF
--- a/docs/layouts/shortcodes/generated/dynamic_section.html
+++ b/docs/layouts/shortcodes/generated/dynamic_section.html
@@ -87,6 +87,12 @@
             <td>Interval at which periodic savepoints will be triggered. The triggering schedule is not guaranteed, savepoints will be triggered as part of the regular reconcile loop.</td>
         </tr>
         <tr>
+            <td><h5>kubernetes.operator.rescaling.cluster-cooldown</h5></td>
+            <td style="word-wrap: break-word;">1 min</td>
+            <td>Duration</td>
+            <td>Cooldown period after rescaling jobs where no further scaling will be performed to minimize cluster churn.</td>
+        </tr>
+        <tr>
             <td><h5>kubernetes.operator.savepoint.format.type</h5></td>
             <td style="word-wrap: break-word;">CANONICAL</td>
             <td><p>Enum</p></td>
--- a/docs/layouts/shortcodes/generated/kubernetes_operator_config_configuration.html
+++ b/docs/layouts/shortcodes/generated/kubernetes_operator_config_configuration.html
@@ -225,6 +225,12 @@
             <td>The maximum number of threads running the reconciliation loop. Use -1 for infinite.</td>
         </tr>
         <tr>
+            <td><h5>kubernetes.operator.rescaling.cluster-cooldown</h5></td>
+            <td style="word-wrap: break-word;">1 min</td>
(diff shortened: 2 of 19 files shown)


docs/layouts/shortcodes/generated/dynamic_section.html
```diff
@@ -86,6 +86,12 @@
             <td>Duration</td>
             <td>Interval at which periodic savepoints will be triggered. The triggering schedule is not guaranteed, savepoints will be triggered as part of the regular reconcile loop.</td>
         </tr>
+        <tr>
+            <td><h5>kubernetes.operator.rescaling.cluster-cooldown</h5></td>
+            <td style="word-wrap: break-word;">1 min</td>
+            <td>Duration</td>
```
gyfora, line 92: The default value should be 0 I think to preserve the current behavior as this does not affect all kinds of Kubernetes clusters .

many are actually fixed sized

mxm: This one is tricky. I would rather like to be conservative here because concurrent scaling operations can create a lot of unexpected churn in the cluster. Users have seen cost increase from resource spikes which can paralyze the entire cluster and lead to adding more k8s nodes (if the cluster autoscaler is active). Users may always change this default. 

Beyond this, we may add something like a concurrency parameter which allows to define how many pipelines are allowed to scale before the cool down. Additionally, checking the actual allocatable resources before scaling is something we need to look into.

gyfora (changes requested):
I think the approach here is a little bit too complex and we should simplify it.

Restrictions related to scaling operations (specifically to executing scaling) should go into the ScalingExecutor class. This way we also won't impact the logic if actual scaling is already disabled.

With adding the context based logic we again moved back part of the autoscaler logic into the operator core which should not be necessary. This also requires a changes to a lot of tests etc. With this in the scaling executor it would be a simple few lines of change.

docs/layouts/shortcodes/generated/dynamic_section.html
gyfora: I personally would not enforce any cross deployment timeout / cooldown by default. I don't think it's intuitive why something on one deployment should have an effect on others. Also this is specific to EKS not a general problem.

While the lack of this config may cause problems on EKS over time, it causes immediate weird behaviour on every operator that is using the autoscaler for multiple jobs. Imagine having 60 jobs, it will take an hour to scale once. Also some jobs may never be scaled because there is no faire queing etc.

mxm:
The logic was implemented in the global context because it handles logic across pipelines, something which the pipeline-specific ScalingExecutor cannot do. The logic can be used independently of the autoscaler implementation. I tried to implement the logic in the autoscaler directly but the the current design of operator / autoscaler doesn't easily allow this because the autoscaler logic is pipeline-specific without state across pipelines. If you have any ideas, please let me know.

mxm:
Closing for now.

PR #551, opened 2023-03-23
[FLINK-31592] Improve Flink dependent resource deletion

gyfora:

Currently we use the default (background) deletion propagation when we delete Kubernetes resources such as the JobManager deployment.

In most cases this is acceptable however this can lead to lingering resources when the garbage collection is slow on the k8s cluster. We therefore propose to make this configurable and also change the default to Foreground which is more predictable and works better for our purposes in the operator.

## Brief change log

  - *Change default resource deletion to foreground propagation*
  - *Make it configurable*
  - *Add tests*

## Verifying this change

Added unit test to NativeFlinkServiceTest

## Documentation

### DIFF
--- a/docs/layouts/shortcodes/generated/kubernetes_operator_config_configuration.html
+++ b/docs/layouts/shortcodes/generated/kubernetes_operator_config_configuration.html
@@ -243,6 +243,12 @@
             <td>The timeout for the resource clean up to wait for flink to shutdown cluster.</td>
         </tr>
         <tr>
+            <td><h5>kubernetes.operator.resource.deletion.propagation</h5></td>
+            <td style="word-wrap: break-word;">Foreground</td>
+            <td><p>Enum</p></td>
+            <td>JM/TM Deployment deletion propagation.<br /><br />Possible values:<ul><li>"Orphan"</li><li>"Background"</li><li>"Foreground"</li></ul></td>
+        </tr>
+        <tr>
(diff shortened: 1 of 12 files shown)


mxm (commented):
Does this mean the operator could get stuck if foreground deletion hangs?

gyfora:
@mxm do you have any further concerns?

docs/layouts/shortcodes/generated/kubernetes_operator_config_configuration.html
```diff
@@ -236,6 +236,12 @@
             <td>Duration</td>
             <td>The timeout for the resource clean up to wait for flink to shutdown cluster.</td>
         </tr>
+        <tr>
+            <td><h5>kubernetes.operator.resource.deletion.propagation</h5></td>
+            <td style="word-wrap: break-word;">Foreground</td>
+            <td><p>Enum</p></td>
+            <td>JM/TM Resource deletion propagation.<br /><br />Possible values:<ul><li>"Orphan"</li><li>"Background"</li><li>"Foreground"</li></ul></td>
```
mbalassi, line 243: nit: Technically not just the JM/TM resources, but all dependant resources including services, configmaps etc. It is still reasonably descriptive in this form. 😏

mbalassi (approved):
LGTM, thanks.

mxm (approved):

PR #556, opened 2023-03-28
[FLINK-30575] Set processing capacity to infinite if task is idle

mxm:
In a prior change we set the processing rates and capacity (true processing rate) to near zero values to allow scaling down in case of idle tasks. This has worked well but caused issues for tasks which are fed via upstream selective outputs (side outputs). Those tests usually received none or only a few records of the upstream outputted records. Setting low values for the incoming records rate will yield a relatively low true processing rate which quickly triggers upscaling even if no records are being processed at all.

We don't currently have per-output/per-input metrics, so the best we can do, is to assume infinite processing capacity if we don't receive any records or are completely idle. Once we received records, we can measure our true processing rate. In the case of selective outputs, we will still assume we receive all the data which is suboptimal but far better than scaling up all the way to the max parallelism as we have seen for the near zero true processing rates during idleness.

### DIFF
--- a/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/metrics/ScalingMetrics.java
+++ b/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/metrics/ScalingMetrics.java
@@ -38,10 +38,12 @@ public class ScalingMetrics {
     private static final Logger LOG = LoggerFactory.getLogger(ScalingMetrics.class);
 
     /**
-     * The minimum value to avoid using zero values which cause side effects like division by zero
-     * or out of bounds (infinitive) floats.
+     * A large, effectively infinite value which continues to allow arithmetics for the scaling
+     * algorithm. It is used in case the task is not processing data or is completely idle. We can't
+     * accurately sample its true processing rate in this case and assume a high capacity until we
+     * know any better. This is to allow scale down in case a task becomes completely idle.
      */
-    public static final double EFFECTIVELY_ZERO = 1e-10;
+    public static final double EFFECTIVELY_INFINITE = 1e100;
 
     public static void computeLoadMetrics(
             Map<FlinkMetric, AggregatedMetric> flinkMetrics,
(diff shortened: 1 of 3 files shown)


flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/metrics/ScalingMetrics.java
```diff
@@ -38,10 +38,12 @@ public class ScalingMetrics {
...
     /**
-     * The minimum value to avoid using zero values which cause side effects like division by zero
-     * or out of bounds (infinitive) floats.
+     * A large, effectively infinite value which continues to allow arithmetics for the scaling
+     * algorithm. It is used in case the task is not processing data or is completely idle. We can't
+     * accurately sample its true processing rate in this case and assume a high capacity until we
+     * know any better. This is to allow scale down in case a task becomes completely idle.
      */
-    public static final double EFFECTIVELY_ZERO = 1e-10;
+    public static final double EFFECTIVELY_INFINITE = 1e100;
```
gyfora, line 46: Any reason for not actually using `Double.POSITIVE_INFINITY`?

gyfora (commented):
Added a minor comment but otherwise I think the changes look good!

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/metrics/ScalingMetrics.java
mxm: Updated pushed.

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/ScalingMetricEvaluator.java
```diff
@@ -251,12 +251,20 @@ private static double getAverage(
...
                         .filter(m -> m.containsKey(metric))
                         .mapToDouble(m -> m.get(metric))
                         .filter(d -> !Double.isNaN(d))
-                        .toArray());
+                        .toArray();
+        for (double metricsValue : metricsValues) {
+            if (Double.isInfinite(metricsValue)) {
+                // As long as infinite values are present, we can't properly average. We need to
+                // wait until they are evicted.
+                return metricsValue;
```
gyfora, line 265: This confused me a little, should we simply return Double.positive_infinity here explicitly?

gyfora (approved):

PR #561, opened 2023-04-03
[FLINK-30609] Add ephemeral storage to CRD

HuangZhenQiu:
Add ephemeral storage to CRD, so that users can apply ephemeral storage as resource requirement to containers of JM and TM.

## Brief change log 

  - Add ephemeral storage to Resource of flink deployment spec

## Verifying this change

  - Changes are tested in unit test in FlinkConfigBuilderTest

  - The public API, i.e., is any changes to the `CustomResourceDescriptors`: (yes)

## Documentation

  - If yes, how is the feature documented? (not documented)

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
@@ -17,6 +17,7 @@
 
 package org.apache.flink.kubernetes.operator.config;
 
+import org.apache.flink.annotation.VisibleForTesting;
 import org.apache.flink.client.deployment.application.ApplicationConfiguration;
 import org.apache.flink.configuration.ConfigOption;
 import org.apache.flink.configuration.ConfigOptions;
@@ -31,6 +32,7 @@ import org.apache.flink.configuration.PipelineOptions;
 import org.apache.flink.configuration.TaskManagerOptions;
 import org.apache.flink.kubernetes.configuration.KubernetesConfigOptions;
 import org.apache.flink.kubernetes.configuration.KubernetesDeploymentTarget;
+import org.apache.flink.kubernetes.operator.api.CrdConstants;
 import org.apache.flink.kubernetes.operator.api.FlinkDeployment;
 import org.apache.flink.kubernetes.operator.api.spec.FlinkDeploymentSpec;
 import org.apache.flink.kubernetes.operator.api.spec.FlinkVersion;
@@ -40,6 +42,7 @@ import org.apache.flink.kubernetes.operator.api.spec.Resource;
 import org.apache.flink.kubernetes.operator.api.spec.UpgradeMode;
 import org.apache.flink.kubernetes.operator.standalone.StandaloneKubernetesConfigOptionsInternal;
 import org.apache.flink.kubernetes.operator.utils.FlinkUtils;
+import org.apache.flink.kubernetes.utils.Constants;
 import org.apache.flink.runtime.jobgraph.SavepointConfigOptions;
 import org.apache.flink.runtime.jobmanager.HighAvailabilityMode;
 import org.apache.flink.streaming.api.environment.CheckpointConfig;
@@ -47,7 +50,11 @@ import org.apache.flink.streaming.api.environment.ExecutionCheckpointingOptions;
 import org.apache.flink.util.FileUtils;
 import org.apache.flink.util.StringUtils;
 
+import io.fabric8.kubernetes.api.model.Container;
 import io.fabric8.kubernetes.api.model.Pod;
+import io.fabric8.kubernetes.api.model.PodSpec;
+import io.fabric8.kubernetes.api.model.Quantity;
+import io.fabric8.kubernetes.api.model.ResourceRequirements;
 import io.fabric8.kubernetes.client.utils.Serialization;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -211,6 +218,7 @@ public class FlinkConfigBuilder {
             setPodTemplate(
                     spec.getPodTemplate(),
                     spec.getJobManager().getPodTemplate(),
+                    spec.getJobManager().getResource(),
                     effectiveConfig,
                     true);
             if (spec.getJobManager().getReplicas() > 0) {
@@ -228,6 +236,7 @@ public class FlinkConfigBuilder {
             setPodTemplate(
                     spec.getPodTemplate(),
                     spec.getTaskManager().getPodTemplate(),
+                    spec.getTaskManager().getResource(),
                     effectiveConfig,
                     false);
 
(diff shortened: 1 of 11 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultValidator.java
```diff
@@ -366,6 +368,18 @@ private Optional<String> validateResources(String component, Resource resource)
             return Optional.of(component + " resource memory parse error: " + iae.getMessage());
         }
 
+        String storage = resource.getEphemeralStorage();
```
morhidi, line 371: This will not catch misconfigured storage resources if memory is not defined.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
```diff
@@ -409,15 +419,44 @@ private static void setPodTemplate(
...
                     podConfigOption,
                     createTempFile(
-                            mergePodTemplates(
-                                    basicPod,
-                                    appendPod,
-                                    effectiveConfig.get(
-                                            KubernetesOperatorConfigOptions
-                                                    .POD_TEMPLATE_MERGE_BY_NAME))));
+                            applyResourceToPodTemplate(
+                                    mergePodTemplates(
```
gyfora, line 423: I think we should not nest these calls but do one after the other for readability

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
```diff
@@ -409,15 +419,44 @@ private static void setPodTemplate(
...
+                                                            .POD_TEMPLATE_MERGE_BY_NAME)),
+                                    resource)));
         }
     }
 
+    private static Pod applyResourceToPodTemplate(Pod podTemplate, Resource resource) {
+        if (resource != null
+                && !StringUtils.isNullOrWhitespaceOnly(resource.getEphemeralStorage())
+                && podTemplate != null
+                && podTemplate.getSpec() != null) {
```
gyfora, line 437: This will ignore the ephemeralStorage setting if the user did not specify any podTemplate (or spec in the podTemplate), that is probably incorrect.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
```diff
@@ -409,15 +419,44 @@ private static void setPodTemplate(
...
+            for (Container container : podTemplate.getSpec().getContainers()) {
```
gyfora, line 438: Should this apply for all containers or only to the main jm/tm container? `Constants.MAIN_CONTAINER_NAME` 
I think only to the main one.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
gyfora: We need some more test coverage for different configs:
 1. Ephemeral storage settings without any podtemplate 
 2. Ephemeral storage without tm/jm podTemplate 
 3. Ephemeral storage without spec in podTemplate

gyfora (changes requested):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultValidator.java
HuangZhenQiu: Got catch. Thanks!

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
HuangZhenQiu: Updated

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
HuangZhenQiu: Added different scenarios in FlinkConfigBuilderTest

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
HuangZhenQiu: I was confused about the requirement.  It makes more sense to only apply it to flink main container.

ashangit:
I'm wondering if it won't be cleaner to let flink manage the resources field as it is done for cpu, memory and external resources (see https://github.com/apache/flink/blob/master/flink-kubernetes/src/main/java/org/apache/flink/kubernetes/kubeclient/decorators/InitTaskManagerDecorator.java#L168)
and having the operator only providing the ephemeral storage config inside the flink configuration?

gyfora:
> I'm wondering if it won't be cleaner to let flink manage the resources field as it is done for cpu, memory and external resources (see https://github.com/apache/flink/blob/master/flink-kubernetes/src/main/java/org/apache/flink/kubernetes/kubeclient/decorators/InitTaskManagerDecorator.java#L168)
> and having the operator only providing the ephemeral storage config inside the flink configuration?

That would be nice but even if we add this to a newer Flink version, this logic is required to support this to current / older versions .

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
```diff
@@ -391,31 +399,87 @@ private static void setResource(
...
+
+        if (podTemplate == null || podTemplate.getSpec() == null) {
+            PodSpec spec = new PodSpec();
+            spec.getContainers()
+                    .add(
+                            decorateContainerWithEphemeralStorage(
+                                    new Container(), resource.getEphemeralStorage()));
+            Pod newPodTemplate = new Pod();
+            newPodTemplate.setSpec(spec);
+            return newPodTemplate;
```
gyfora, line 446: If only the spec is null, this logic overwrites the entire pod, that sounds incorrect. Imagine the user only set the metadata. Please add a test to guard against this.

gyfora (changes requested):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
```diff
@@ -391,31 +399,87 @@ private static void setResource(
...
-
         // Avoid to create temporary pod template files for JobManager and TaskManager if it is not
         // configured explicitly via .spec.JobManagerSpec.podTemplate or
         // .spec.TaskManagerSpec.podTemplate.
+        final ConfigOption<String> podConfigOption =
+                isJM
+                        ? KubernetesConfigOptions.JOB_MANAGER_POD_TEMPLATE
+                        : KubernetesConfigOptions.TASK_MANAGER_POD_TEMPLATE;
+
+        Pod basicPodWithResource = applyResourceToPodTemplate(basicPod, resource);
```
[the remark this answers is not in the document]
HuangZhenQiu: Yes, you are right. I think the best way to add the feature without breaking feature is that:

Always merge first as long as not both basicPod and appendPod are null, then apply resource to pod.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
HuangZhenQiu: Thanks for pointing it out. Added the testApplyResourceToPodTemplate function in FlinkConfigBuilderTest to cover it.

PR #563, opened 2023-04-11
[FLINK-31716] Event UID field is missing the first time that an event…

rodmeneses:
What is the purpose of the change

Fixes a bug reported on https://issues.apache.org/jira/browse/FLINK-31716 where the event being consumed for the first time didn't have an UID field

This ticket was reopened for two reasons:
1. Before the check was only being done when the event was created for the first time, and not on updates
2. We are now accepting the actual return value of the `.createOrReplace` call. Before we were only using the `uid` field of it.

Brief change log

Fixes a bug on EventUtils where the event consumed was missing the UID field
Verifying this change

This change modified existing unit tests.

Does this pull request potentially affect one of the following parts:

Dependencies (does it add or upgrade a dependency): no
The public API, i.e., is any changes to the CustomResourceDescriptors: no
Core observer or reconciler logic that is regularly executed: no
Documentation

Does this pull request introduce a new feature? no
If yes, how is the feature documented? not applicable

### DIFF
--- a/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/utils/EventUtilsTest.java
+++ b/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/utils/EventUtilsTest.java
@@ -72,6 +72,7 @@ public class EventUtilsTest {
                         .withName(eventName)
                         .get();
         Assertions.assertEquals(event.getMetadata().getUid(), eventConsumed.getMetadata().getUid());
+        Assertions.assertEquals(eventConsumed, event);
         eventConsumed = null;
         Assertions.assertNotNull(event);
         Assertions.assertEquals(1, event.getCount());
@@ -93,7 +94,7 @@ public class EventUtilsTest {
                         .inNamespace(flinkApp.getMetadata().getNamespace())
                         .withName(eventName)
                         .get();
-        Assertions.assertEquals(event.getMetadata().getUid(), eventConsumed.getMetadata().getUid());
+        Assertions.assertEquals(eventConsumed, event);
         Assertions.assertEquals(2, event.getCount());
     }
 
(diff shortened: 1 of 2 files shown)


flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/utils/EventUtilsTest.java
```diff
@@ -72,6 +72,7 @@ public void accept(Event event) {
                         .withName(eventName)
                         .get();
         Assertions.assertEquals(event.getMetadata().getUid(), eventConsumed.getMetadata().getUid());
```
morhidi, line 74: This is redundant with the assertions below

morhidi:
+1 LGTM

mbalassi:
@rodmeneses there are multiple failed test scenarios in `FlinkDeploymentControllerTest` PTAL.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentControllerTest.java
```diff
@@ -217,9 +227,10 @@ public void verifyBasicReconcileLoop(FlinkVersion flinkVersion) throws Exception
 
     @Test
     public void verifyFailedDeployment() throws Exception {
+
         var submittedEventValidatingResponseProvider =
                 new TestUtils.ValidatingResponseProvider<>(
-                        new EventBuilder().withNewMetadata().endMetadata().build(),
+                        mockedEvent,
```
morhidi, line 233: good enough, but can't we send back the the event we received?

morhidi (changes requested):

morhidi (approved):

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentControllerTest.java
rodmeneses: I tried but the the `TestUtils.ValidatingResponseProvider` class requires the event to be returned at construction time, and I couldn't find a way to intercept the `createOrReplace` call inside `EventUtils.createOrUpdateEvent`

PR #568, opened 2023-04-18
[FLINK-31845] Make KubernetesStepDecorator Pluggable

darenwkt:

This PR extends KubernetesStepDecorator interface to make it pluggable for custom decorators to be loaded as JARs.

## Brief change log

- Added DecoratorUtils to discover CustomDecorators in flink plugin folder to be loaded as JAR.

## Verifying this change

- Added Unit Test.
- Manual tested by creating a custom Decorator plugin JAR that is copied over in DockerFile. Validated that the created JM container name is changed as per custom Decorator specification.

## Documentation

### DIFF
--- a/flink-kubernetes-standalone/pom.xml
+++ b/flink-kubernetes-standalone/pom.xml
@@ -128,6 +128,26 @@ under the License.
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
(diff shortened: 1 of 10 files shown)


gyfora:
I think this should be fixed first as part of https://issues.apache.org/jira/browse/FLINK-28831 and then we can adopt the solution here.

PR #570, opened 2023-04-19
[FLINK-31857] Observer plugin

darenwkt:

Support pluginable Observer to allow user to load a plugin JAR to run custom observer logic.

## Brief change log

- Added FlinkResourceObserverPlugin interface for users to implement in plugin JAR with custom observer logic.

## Verifying this change

- Added unit test to verify TestObserverPlugin is discoverable.
- Manual test by loading a sample JAR that changes the jobName in FlinkSessionJob status field and verified successfully.

## Documentation


gyfora:
I think we should design this together with the reconciler plugin and there should be a FLIP for it to make sure that other users can influence the design and we cover existing features like the autoscaler etc with the pluggable mechanism.

PR #571, opened 2023-04-20
[FLINK-31827] Compute output ratio per edge

gyfora:

This change introduces per-edge output ratio tracking for the JobGrapgh instead of the previous per-vertex tracking.
The previous approach assumed that all downstream vertices will receive the same output (all edges have the same num records out) which is not true when side output or more complex operator chains are present in the JobVertex.

While Flink does not directly expose per edge for the sent/received records, in many cases this can be computed.
 - If we have a single output then we use num records out
 - If the downstream vertex has a single input we use the downstream num records in
 - If the downstream vertex has only inputs with a single output then we subtrack the upstream numRecrods out from other inputs from the num records in
 
 As a related change we should also introduce per edge record count metrics in Flink which would allow us to use that in the autoscaler algorithm if enabled. 

## Brief change log

  - *Remove per-vertex OUTPUT_RATIO and TRUE_OUTPUT_RATE metrics*
  - *Rework the collected metrics / metric history to allow storing per-edge output ratio metrics*
  - *Compute per-edge output ratio depending on the topology*
  - *Use downstream num records in whenever possible instead of upstream num records out as the latter is very unreliable due to some Flink bugs (https://issues.apache.org/jira/browse/FLINK-18808 & https://issues.apache.org/jira/browse/FLINK-31752)*

## Verifying this change

New unit tests added for the output ratio computation and existing tests cover the current behaviour.

## Documentation

### DIFF
--- a/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/AutoScalerInfo.java
+++ b/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/AutoScalerInfo.java
@@ -21,12 +21,13 @@ import org.apache.flink.annotation.VisibleForTesting;
 import org.apache.flink.configuration.Configuration;
 import org.apache.flink.kubernetes.operator.api.AbstractFlinkResource;
 import org.apache.flink.kubernetes.operator.autoscaler.config.AutoScalerOptions;
-import org.apache.flink.kubernetes.operator.autoscaler.metrics.ScalingMetric;
-import org.apache.flink.kubernetes.operator.autoscaler.utils.JobVertexSerDeModule;
+import org.apache.flink.kubernetes.operator.autoscaler.metrics.CollectedMetrics;
+import org.apache.flink.kubernetes.operator.autoscaler.utils.AutoScalerSerDeModule;
 import org.apache.flink.kubernetes.utils.Constants;
 import org.apache.flink.runtime.jobgraph.JobVertexID;
 import org.apache.flink.util.Preconditions;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
@@ -71,7 +72,7 @@ public class AutoScalerInfo {
     protected static final ObjectMapper YAML_MAPPER =
             new ObjectMapper(yamlFactory())
                     .registerModule(new JavaTimeModule())
-                    .registerModule(new JobVertexSerDeModule());
+                    .registerModule(new AutoScalerSerDeModule());
 
     private final ConfigMap configMap;
     private Map<JobVertexID, SortedMap<Instant, ScalingSummary>> scalingHistory;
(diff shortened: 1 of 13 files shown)


flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/metrics/ScalingMetrics.java
```diff
@@ -153,37 +169,64 @@ private static double getNumRecordsInPerSecond(
...
+        // records in
+        if (toVertexInputs.size() == 1) {
+            LOG.debug(
+                    "Computing edge ({}, {}) data rate for single input downstream task", from, to);
+            return getNumRecordsInPerSecond(toMetrics, to, false);
+        }
+
+        // Case 2: Downstream vertex has only inputs from upstream vertices which don't have other
+        // outputs
+        double sumOtherUpstreamInputs = 0;
```
mxm, line 202:
```suggestion
        double numRecordsOutFromUpstreamInputs = 0;
```

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/metrics/ScalingMetrics.java
```diff
@@ -153,37 +169,64 @@ private static double getNumRecordsInPerSecond(
...
+        for (JobVertexID input : toVertexInputs) {
+            if (input.equals(from)) {
+                continue;
```
mxm, line 205:
```suggestion
                // Exclude source edge because we only want to consider other input edges
                continue;
```

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/AutoScalerInfo.java
```diff
@@ -86,20 +87,24 @@ public AutoScalerInfo(Map<String, String> data) {
...
             return new TreeMap<>();
         }
 
-        return YAML_MAPPER.readValue(decompress(historyYaml), new TypeReference<>() {});
+        try {
+            return YAML_MAPPER.readValue(decompress(historyYaml), new TypeReference<>() {});
+        } catch (JsonProcessingException e) {
+            LOG.error(
+                    "Could not deserialize metric history, possibly the format changed. Discarding...");
+            return new TreeMap<>();
```
mxm, line 101: This will clear all existing metrics when we deploy this new version.

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/metrics/CollectedMetrics.java
```diff
@@ -17,18 +17,19 @@
...
-@Value
+/** Collected scaling metrics. */
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
 public class CollectedMetrics {
-    JobTopology jobTopology;
-    SortedMap<Instant, Map<JobVertexID, Map<ScalingMetric, Double>>> metricHistory;
+    private Map<JobVertexID, Map<ScalingMetric, Double>> vertexMetrics;
+    private Map<Edge, Double> outputRatios;
```
mxm, line 34: This will increase the size of the metric history. We may have to pursue https://issues.apache.org/jira/browse/FLINK-31866 for future upgrades.

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/metrics/CollectedMetricHistory.java
```diff
@@ -0,0 +1,32 @@
...
+
+import lombok.Value;
+
+import java.time.Instant;
+import java.util.SortedMap;
+
+/** Topology and collected metric history. */
+@Value
+public class CollectedMetricHistory {
+    JobTopology jobTopology;
```
mxm, line 30: Why do we need to persist the JobTopology? Could we save some bytes here?

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/AutoScalerInfo.java
mxm: I think this will introduce a bug where only a single metric observation in a window will be considered: Fixed via https://issues.apache.org/jira/browse/FLINK-31866 or https://issues.apache.org/jira/browse/FLINK-31867.

mxm (commented):
Thanks @gyfora!

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/AutoScalerInfo.java
gyfora: Maybe a better solution would be to throw an exception and clear the autoscaler info completely. That will basically reset the history.

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/metrics/CollectedMetricHistory.java
gyfora: We don't persist the JobTopology in the autoscalerinfo, only the metric history

mxm: ok

mxm (approved):

PR #573, opened 2023-04-21
[FLINK-31866] Start metric window with timestamp of first observation

mxm:
The autoscaler uses a ConfigMap to store past metric observations which is used to re-initialize the autoscaler state in case of failures or upgrades.

Whenever trimming of the ConfigMap occurs, we need to make sure we also update the timestamp for the start of the metric collection, so any removed observations can be compensated with by collecting new ones. If we don't do this, the metric window will effectively shrink due to removing observations.

This can lead to triggering scaling decisions when the operator gets redeployed due to the removed items.

The solution we are opting here is to treat the first metric observation timestamp as the start of the metric collection.


mxm:
Rebased

gyfora (approved):

PR #574, opened 2023-04-24
[FLINK-31827] Discard old format scaling history

mxm:
We recently change the format for the metric history but this also affects the scaling history because it contains new / old scaling metrics.


flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/AutoScalerInfo.java
```diff
@@ -95,7 +95,7 @@ public SortedMap<Instant, CollectedMetrics> getMetricHistory() {
 
         try {
             return YAML_MAPPER.readValue(decompress(historyYaml), new TypeReference<>() {});
-        } catch (JsonProcessingException e) {
+        } catch (JacksonException e) {
```
gyfora, line 98: We could catch all `Exception`s maybe?

mxm: The reason I changed the exception type is that this is the base Jackson exception which catches all Jackson related errors which we can't do anything about. Leaving this as-is for now.

PR #575, opened 2023-04-26
[FLINK-31885] Trigger event on autoscaler error

gyfora:

Currently when there is an error in the atuoscaler logic due to some misconfiguration / invalid metrics etc we only log the problem in the operator side but don't show this to the user. From the user's perspective the autoscaler simply doesnt do anything in these cases if they dont have access to the operator logs.

This PR triggers a CR event in case an error happens in the autoscaler logic.

## Brief change log

 - Add error when exception happens
 - Add test

## Verifying this change

new unit test added

## Documentation

### DIFF
--- a/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/JobAutoScalerImpl.java
+++ b/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/JobAutoScalerImpl.java
@@ -24,6 +24,7 @@ import org.apache.flink.kubernetes.operator.autoscaler.metrics.ScalingMetric;
 import org.apache.flink.kubernetes.operator.controller.FlinkResourceContext;
 import org.apache.flink.kubernetes.operator.metrics.KubernetesResourceMetricGroup;
 import org.apache.flink.kubernetes.operator.reconciler.deployment.JobAutoScaler;
+import org.apache.flink.kubernetes.operator.utils.EventRecorder;
 import org.apache.flink.runtime.jobgraph.JobVertexID;
 
 import io.fabric8.kubernetes.client.KubernetesClient;
@@ -48,6 +49,7 @@ public class JobAutoScalerImpl implements JobAutoScaler {
     private final ScalingMetricCollector metricsCollector;
     private final ScalingMetricEvaluator evaluator;
     private final ScalingExecutor scalingExecutor;
+    private final EventRecorder eventRecorder;
 
     private final Map<ResourceID, Map<JobVertexID, Map<ScalingMetric, EvaluatedScalingMetric>>>
             lastEvaluatedMetrics = new ConcurrentHashMap<>();
@@ -57,13 +59,13 @@ public class JobAutoScalerImpl implements JobAutoScaler {
             KubernetesClient kubernetesClient,
             ScalingMetricCollector metricsCollector,
             ScalingMetricEvaluator evaluator,
-            ScalingExecutor scalingExecutor) {
-
+            ScalingExecutor scalingExecutor,
+            EventRecorder eventRecorder) {
         this.kubernetesClient = kubernetesClient;
-
         this.metricsCollector = metricsCollector;
         this.evaluator = evaluator;
         this.scalingExecutor = scalingExecutor;
+        this.eventRecorder = eventRecorder;
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/JobAutoScalerImpl.java
```diff
@@ -113,7 +115,13 @@ public boolean scale(FlinkResourceContext<? extends AbstractFlinkResource<?, ?>>
...
             return specAdjusted;
-        } catch (Exception e) {
+        } catch (Throwable e) {
+            eventRecorder.triggerEvent(
+                    resource,
+                    EventRecorder.Type.Warning,
+                    EventRecorder.Reason.AutoscalerError,
+                    EventRecorder.Component.Operator,
+                    e.getMessage());
             LOG.error("Error while scaling resource", e);
```
mxm, line 125: Can we log before triggering the event just in case...

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/metrics/ScalingMetrics.java
```diff
@@ -83,6 +83,8 @@ public static void computeDataRateMetrics(
             scalingMetrics.put(ScalingMetric.CURRENT_PROCESSING_RATE, numRecordsInPerSecond);
         } else {
             LOG.error("Cannot compute true processing rate without numRecordsInPerSecond");
+            scalingMetrics.put(ScalingMetric.TRUE_PROCESSING_RATE, Double.NaN);
+            scalingMetrics.put(ScalingMetric.CURRENT_PROCESSING_RATE, Double.NaN);
```
mxm, line 87: Unrelated change? Why is this necessary?

mxm (approved):

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/JobAutoScalerImpl.java
```diff
@@ -81,17 +83,17 @@ public boolean scale(FlinkResourceContext<? extends AbstractFlinkResource<?, ?>>
         var conf = ctx.getObserveConfig();
         var resource = ctx.getResource();
 
-        if (resource.getSpec().getJob() == null || !conf.getBoolean(AUTOSCALER_ENABLED)) {
-            LOG.info("Job autoscaler is disabled");
-            return false;
-        }
+        try {
+            if (resource.getSpec().getJob() == null || !conf.getBoolean(AUTOSCALER_ENABLED)) {
+                LOG.info("Job autoscaler is disabled");
```
rodmeneses, line 88: should we use `LOG.warn` instead?

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/JobAutoScalerImpl.java
```diff
@@ -113,7 +115,13 @@ public boolean scale(FlinkResourceContext<? extends AbstractFlinkResource<?, ?>>
...
             autoScalerInfo.replaceInKubernetes(kubernetesClient);
             return specAdjusted;
-        } catch (Exception e) {
+        } catch (Throwable e) {
+            eventRecorder.triggerEvent(
+                    resource,
+                    EventRecorder.Type.Warning,
+                    EventRecorder.Reason.AutoscalerError,
+                    EventRecorder.Component.Operator,
+                    e.getMessage());
```
rodmeneses, line 124: if an NPE occurs, `e.gestMessage()` will be null. We need to make sure that `triggerEvent` and the downstream caller are not throwing NPE because of this

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/metrics/ScalingMetrics.java
gyfora: I was testing the error scenarios and realized that if some metrics were missing from during collection, the scaling evaluator threw a nullpointer exception. This is to put nan instead of just not putting anything

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/JobAutoScalerImpl.java
gyfora: sure ^^

mxm (approved):
LGTM apart from any remaining comments.

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/JobAutoScalerImpl.java
gyfora: It's not really a warning, this is based on user config. Usually it's off

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/JobAutoScalerImpl.java
gyfora: Shouldn't be a problem and we do this in other places as well. I will add a simple test even if unrelated to the change.

PR #581, opened 2023-05-03
[FLINK-31936] Support setting scale up max factor

X-czh:

Support setting scale up max factor for autoscaler.

## Brief change log

- Add new option _kubernetes.operator.job.autoscaler.scale-up.max-factor_ for setting scale up max factor.
- Cap scale factor in `JobVertexScaler#computeScaleTargetParallelism` by max scale-up factor.
- Record capped target processing rate instead as the expected processing rate in `JobVertexScaler#computeScaleTargetParallelism`.
- Adapt UTs accordingly.

## Verifying this change

- Extended existing unit test

## Documentation

  - If yes, how is the feature documented? auto-generated doc

### DIFF
--- a/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/JobVertexScaler.java
+++ b/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/JobVertexScaler.java
@@ -38,6 +38,7 @@ import java.util.Map;
 import java.util.SortedMap;
 
 import static org.apache.flink.kubernetes.operator.autoscaler.config.AutoScalerOptions.MAX_SCALE_DOWN_FACTOR;
+import static org.apache.flink.kubernetes.operator.autoscaler.config.AutoScalerOptions.MAX_SCALE_UP_FACTOR;
 import static org.apache.flink.kubernetes.operator.autoscaler.config.AutoScalerOptions.SCALE_UP_GRACE_PERIOD;
 import static org.apache.flink.kubernetes.operator.autoscaler.config.AutoScalerOptions.TARGET_UTILIZATION;
 import static org.apache.flink.kubernetes.operator.autoscaler.config.AutoScalerOptions.VERTEX_MAX_PARALLELISM;
@@ -93,6 +94,7 @@ public class JobVertexScaler {
         LOG.debug("Target processing capacity for {} is {}", vertex, targetCapacity);
         double scaleFactor = targetCapacity / averageTrueProcessingRate;
         double minScaleFactor = 1 - conf.get(MAX_SCALE_DOWN_FACTOR);
+        double maxScaleFactor = 1 + conf.get(MAX_SCALE_UP_FACTOR);
         if (scaleFactor < minScaleFactor) {
             LOG.debug(
                     "Computed scale factor of {} for {} is capped by maximum scale down factor to {}",
@@ -100,7 +102,18 @@ public class JobVertexScaler {
                     vertex,
                     minScaleFactor);
             scaleFactor = minScaleFactor;
+        } else if (scaleFactor > maxScaleFactor) {
+            LOG.debug(
+                    "Computed scale factor of {} for {} is capped by maximum scale up factor to {}",
+                    scaleFactor,
+                    vertex,
+                    maxScaleFactor);
+            scaleFactor = maxScaleFactor;
         }
+
+        // Cap target capacity according to the capped scale factor
+        double cappedTargetCapacity = averageTrueProcessingRate * scaleFactor;
+        LOG.debug("Capped target processing capacity for {} is {}", vertex, cappedTargetCapacity);
 
         int newParallelism =
                 scale(
@@ -124,7 +137,8 @@ public class JobVertexScaler {
 
         // We record our expectations for this scaling operation
         evaluatedMetrics.put(
-                ScalingMetric.EXPECTED_PROCESSING_RATE, EvaluatedScalingMetric.of(targetCapacity));
+                ScalingMetric.EXPECTED_PROCESSING_RATE,
+                EvaluatedScalingMetric.of(cappedTargetCapacity));
         return newParallelism;
     }
 
--- a/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/config/AutoScalerOptions.java
+++ b/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/config/AutoScalerOptions.java
@@ -101,9 +101,9 @@ public class AutoScalerOptions {
     public static final ConfigOption<Double> MAX_SCALE_UP_FACTOR =
             autoScalerConfig("scale-up.max-factor")
                     .doubleType()
-                    .defaultValue(2.0)
+                    .defaultValue((double) Integer.MAX_VALUE)
                     .withDescription(
-                            "Max scale up factor. 2.0 means job can only be scaled up with 200% of the original parallelism.");
+                            "Max scale up factor. 2.0 means job can only be scaled up with 200% of the current parallelism.");
 
     public static final ConfigOption<Duration> CATCH_UP_DURATION =
             autoScalerConfig("catch-up.duration")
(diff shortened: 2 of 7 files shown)


flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/JobVertexScaler.java
```diff
@@ -93,15 +94,27 @@ public int computeScaleTargetParallelism(
...
+                    "Computed scale factor of {} for {} is capped by maximum scale up factor to {}",
+                    scaleFactor,
+                    vertex,
+                    maxScaleFactor);
+            scaleFactor = maxScaleFactor;
         }
 
+        // Cap target capacity according to the capped scale factor
+        double cappedTargetCapacity = averageTrueProcessingRate * scaleFactor;
+        LOG.debug("Capped target processing capacity for {} is {}", vertex, cappedTargetCapacity);
```
gyfora, line 116: Could you please explain why you introduced this change?

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/config/AutoScalerOptions.java
```diff
@@ -98,6 +98,13 @@ private static ConfigOptions.OptionBuilder autoScalerConfig(String key) {
                     .withDescription(
                             "Max scale down factor. 1 means no limit on scale down, 0.6 means job can only be scaled down with 60% of the original parallelism.");
 
+    public static final ConfigOption<Double> MAX_SCALE_UP_FACTOR =
+            autoScalerConfig("scale-up.max-factor")
+                    .doubleType()
+                    .defaultValue(2.0)
```
gyfora, line 104: I wonder if we should set this higher by default, @mxm what do you think?

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/config/AutoScalerOptions.java
```diff
@@ -98,6 +98,13 @@ private static ConfigOptions.OptionBuilder autoScalerConfig(String key) {
...
+                    .withDescription(
+                            "Max scale up factor. 2.0 means job can only be scaled up with 200% of the original parallelism.");
```
mxm, line 106:
```suggestion
                            "Max scale up factor. 2.0 means job can only be scaled up with 200% of the current parallelism.");
```

Just to make it clear that it can double every time.

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/config/AutoScalerOptions.java
mxm: In our production setup, we would definitely set this to a very high number, effectively disabling it. The reason is that we have pipelines which sit most idle (parallelism 1) but then receive huge traffic spikes which need quite high parallelisms (up to 500) to be able to keep up. The only way we limit the scale up is by setting a max parallelism.

Being able to scale up quickly is very important to keep up with the backlog size / growth. If you cap the scale up, you limit the effectiveness of the scaling. I get where it comes from but generally I think it is curing the symptoms and not the cause of these spikes.

mxm: I would suggest to set the default to the maximum parallelism, to be able to scale up from 1 to the max configured parallelism.

flink-kubernetes-operator-autoscaler/src/test/java/org/apache/flink/kubernetes/operator/autoscaler/BacklogBasedScalingTest.java
```diff
@@ -100,6 +100,7 @@ public void setup() {
         defaultConf.set(AutoScalerOptions.CATCH_UP_DURATION, Duration.ofSeconds(2));
         defaultConf.set(AutoScalerOptions.SCALING_ENABLED, true);
         defaultConf.set(AutoScalerOptions.MAX_SCALE_DOWN_FACTOR, 1.);
+        defaultConf.set(AutoScalerOptions.MAX_SCALE_UP_FACTOR, Double.MAX_VALUE);
```
mxm, line 103: In my tests, MAX_VALUE might be problematic because:

```
> Double.MAX_VALUE * 2
Infinity
```

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/config/AutoScalerOptions.java
X-czh: BTW, I'm quite interested in the scale, job type, business scenarios and of course all the problems encountered while adopting autoscaling in production. In my company, we encountered quite a few problems. It'll be good if we can have a discussion on flink-user mail-list sharing experience or even an online meetup would be great.

flink-kubernetes-operator-autoscaler/src/test/java/org/apache/flink/kubernetes/operator/autoscaler/BacklogBasedScalingTest.java
X-czh: I think we can set the default value to the same as that of vertex.max-parallelism: Integer.MAX_VALUE. This would avoid overflow issues while effectively allowing a job to scale up from 1 to the max configured parallelism.

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/config/AutoScalerOptions.java
X-czh: For example, in my company:
1. Job input usu. has periodic pattern, sometimes spikes but smoothly change in most cases
2. Underlying resources are not elastic, resource competition between jobs is not rare
3. Job types are diverse with heavy dependence on various external services, scaling Flink alone is often ineffective
4. The scale is gigantic (over 40k jobs), we have to carefully limit the autoscaler's behavior to prevent resonant scaling behaviors from causing massive failures

mxm: @X-czh Great comments! You are right that the autoscaler configuration is still quite important to handle load skew in big jobs. Concerning the constraints you listed, we definitely also fall under (1) although we are elastic to a large degree. It is definitely a good idea to configure the max parallelism even in an elastic cluster. 

Concerning (4) I'd be interested what you mean by massive failures? Jobs taking other jobs resources? One of the biggest concerns with the autoscaler currently is the lack of resource pre-allocation. Fortunately, this issue will soon be fixed once we start using Flink's Rescale API in the autoscaler instead of triggering a full redeploy every time.

X-czh: Hi @mxm, I'll expand a bit on (4). In case of a data center failure, we will go through a disaster recovery process, jobs will be migrated to another data center to recover. During the process,
- A huge backlog will be built up, and when these jobs recover, all jobs tend to scale up but the overall resource is limited.
- Many external services might not recover to in time or have to lower capacity for other higher priority services, this will cause Flink jobs replying on them process slower and tend to scale up (and very likely scale up a lot!) as well.
As a result, almost all the jobs will tend to scale up a lot during this process, and we need to limit the single-step scale-up behavior so that not too much resources are occupied by a small set of jobs before automatic scaling circuit breaker and human intervention to take action.

X-czh: But yes, our use case might be a bit extreme for most operator users. We tend to add tight constraints on scaling for the overall stability of our clusters while serving most "well-behaved" jobs well at the cost of limiting the elasticity of a single job. Here, "well-behaved" refer to the job input changing smoothly.

gyfora:
@mxm do you have any more concerns/comments? Or can we merge this?

mxm:
No more comments. Thanks @X-czh! 😃

PR #584, opened 2023-05-04
[FLINK-31867] Enforce a minimum number of observations within a metric window

mxm:
This makes sure that a minimum number of observations have been taken within the metric window. Otherwise the metrics won't be returned to prevent decision making on too few observations.

### DIFF
--- a/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/config/AutoScalerOptions.java
+++ b/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/config/AutoScalerOptions.java
@@ -52,6 +52,13 @@ public class AutoScalerOptions {
                     .defaultValue(Duration.ofMinutes(5))
                     .withDescription("Scaling metrics aggregation window size.");
 
+    public static final ConfigOption<Integer> METRICS_WINDOW_MIN_OBSERVATIONS =
+            autoScalerConfig("metrics.window.min-observations")
+                    .intType()
+                    .defaultValue(10)
+                    .withDescription(
+                            "The minimum number of observations to collect during a metric window");
+
     public static final ConfigOption<Duration> STABILIZATION_INTERVAL =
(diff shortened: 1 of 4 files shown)


flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/config/AutoScalerOptions.java
```diff
@@ -52,6 +52,13 @@ private static ConfigOptions.OptionBuilder autoScalerConfig(String key) {
                     .defaultValue(Duration.ofMinutes(5))
                     .withDescription("Scaling metrics aggregation window size.");
 
+    public static final ConfigOption<Integer> METRICS_WINDOW_MIN_OBSERVATIONS =
+            autoScalerConfig("metrics.window.min-observations")
+                    .intType()
+                    .defaultValue(10)
```
gyfora, line 58: I think the default value here is inconsistent with the default 5 minute metric window. I suggest setting the default here to 5

gyfora: Or we could compute a dynamic default here by computing `(metricWindow / reconcileInterval) - X`

mxm: True, the metric collection interval is coupled with the reconciliation interval (60 seconds by default). If we chose to calculate the min observations dynamically, I wonder whether we should even expose this as a config option then. It is more of a sanity check.

gyfora: We could also define the config in terms of % of the maximum observations possible (based on the other configs).

mxm:
Closing for now, this needs additional work.

PR #585, opened 2023-05-04
[FLINK-32005] Add a per-deployment error metric to signal about potential issues

mxm:
If any of the autoscaled deployment produce errors they are only visible in the logs or in the k8s events. Additionally, it would be good to have metrics to detect any potential issues.

### DIFF
--- a/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/JobAutoScalerImpl.java
+++ b/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/JobAutoScalerImpl.java
@@ -70,9 +70,7 @@ public class JobAutoScalerImpl implements JobAutoScaler {
         metricsCollector.cleanup(cr);
         var resourceId = ResourceID.fromResource(cr);
         lastEvaluatedMetrics.remove(resourceId);
-        // We are not removing the metrics registered with Flink (flinkMetrics)
-        // because Flink metrics can only be registered once! When the deployment
-        // comes back we would not be able to register and report metrics.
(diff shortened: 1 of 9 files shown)


gyfora (approved):

mxm:
Do you think having a "scaling" counter would be useful as well?

gyfora:
> Do you think having a "scaling" counter would be useful as well?

I think that could be useful also, because otherwise it's a bit tricky to see how many scalings actually happened. But not critical

mxm:
I'm going to follow up with the scaling counter which also includes a refactor.

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/JobAutoScalerImpl.java
```diff
@@ -82,8 +85,11 @@ public boolean scale(FlinkResourceContext<? extends AbstractFlinkResource<?, ?>>
 
         var conf = ctx.getObserveConfig();
         var resource = ctx.getResource();
+        var resouceId = ResourceID.fromResource(resource);
+        var autoscalerMetricGroup = ctx.getResourceMetricGroup().addGroup("AutoScaler");
```
mxm, line 89: Actually this one should probably just be added when the autoscaler is enabled.

mxm:
The latest commit contains a refactor and adds the successful scaling metric.

mxm:
@tamirsagi I think that is a great suggestion! We do not have good interfaces to add custom metrics. There is only the resource MetricGroup which one can add metrics to, but it is not directly exposed. I think we should create a JIRA to keep track of this work.

PR #586, opened 2023-05-05
[FLINK-32002] Adjust autoscaler defaults for release

mxm:
The goal is to make the make the out-of-the-box experience as smooth as possible. Hence, the defaults are adjusted to have a larger metric window, wider bounds for scaling up and down, less quick downscaling, and a maximum parallelism of 200.

### DIFF
--- a/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/config/AutoScalerOptions.java
+++ b/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/config/AutoScalerOptions.java
@@ -49,7 +49,7 @@ public class AutoScalerOptions {
     public static final ConfigOption<Duration> METRICS_WINDOW =
             autoScalerConfig("metrics.window")
                     .durationType()
-                    .defaultValue(Duration.ofMinutes(5))
+                    .defaultValue(Duration.ofMinutes(10))
                     .withDescription("Scaling metrics aggregation window size.");
 
     public static final ConfigOption<Duration> STABILIZATION_INTERVAL =
@@ -68,15 +68,16 @@ public class AutoScalerOptions {
     public static final ConfigOption<Double> TARGET_UTILIZATION_BOUNDARY =
             autoScalerConfig("target.utilization.boundary")
                     .doubleType()
-                    .defaultValue(0.1)
+                    .defaultValue(0.4)
                     .withDescription(
                             "Target vertex utilization boundary. Scaling won't be performed if utilization is within (target - boundary, target + boundary)");
 
     public static final ConfigOption<Duration> SCALE_UP_GRACE_PERIOD =
             autoScalerConfig("scale-up.grace-period")
                     .durationType()
-                    .defaultValue(Duration.ofMinutes(10))
-                    .withDescription("Period in which no scale down is allowed after a scale up");
+                    .defaultValue(Duration.ofHours(1))
+                    .withDescription(
+                            "Duration in which no scale down of a vertex is allowed after it has been scaled up.");
 
     public static final ConfigOption<Integer> VERTEX_MIN_PARALLELISM =
             autoScalerConfig("vertex.min-parallelism")
@@ -87,28 +88,28 @@ public class AutoScalerOptions {
     public static final ConfigOption<Integer> VERTEX_MAX_PARALLELISM =
             autoScalerConfig("vertex.max-parallelism")
                     .intType()
-                    .defaultValue(Integer.MAX_VALUE)
+                    .defaultValue(200)
                     .withDescription(
                             "The maximum parallelism the autoscaler can use. Note that this limit will be ignored if it is higher than the max parallelism configured in the Flink config or directly on each operator.");
 
     public static final ConfigOption<Double> MAX_SCALE_DOWN_FACTOR =
             autoScalerConfig("scale-down.max-factor")
                     .doubleType()
-                    .defaultValue(0.6)
+                    .defaultValue(1.0)
                     .withDescription(
                             "Max scale down factor. 1 means no limit on scale down, 0.6 means job can only be scaled down with 60% of the original parallelism.");
 
     public static final ConfigOption<Duration> CATCH_UP_DURATION =
             autoScalerConfig("catch-up.duration")
                     .durationType()
-                    .defaultValue(Duration.ofMinutes(10))
+                    .defaultValue(Duration.ofMinutes(5))
                     .withDescription(
                             "The target duration for fully processing any backlog after a scaling operation. Set to 0 to disable backlog based scaling.");
 
     public static final ConfigOption<Duration> RESTART_TIME =
             autoScalerConfig("restart.time")
                     .durationType()
-                    .defaultValue(Duration.ofMinutes(5))
+                    .defaultValue(Duration.ofMinutes(3))
                     .withDescription(
                             "Expected restart time to be used until the operator can determine it reliably from history.");
 
@@ -136,7 +137,7 @@ public class AutoScalerOptions {
     public static final ConfigOption<Integer> VERTEX_SCALING_HISTORY_COUNT =
             autoScalerConfig("history.max.count")
                     .intType()
-                    .defaultValue(1)
+                    .defaultValue(3)
                     .withDescription(
                             "Maximum number of past scaling decisions to retain per vertex.");
 
--- a/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/config/AutoScalerOptions.java
+++ b/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/config/AutoScalerOptions.java
@@ -70,7 +70,7 @@ public class AutoScalerOptions {
                     .doubleType()
                     .defaultValue(0.4)
                     .withDescription(
-                            "Target vertex utilization boundary. Scaling won't be performed if utilization is within (target - boundary, target + boundary)");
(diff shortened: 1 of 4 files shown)


gyfora (approved):
LGTM, please regenerate the docs

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/config/AutoScalerOptions.java
```diff
@@ -68,15 +68,16 @@ private static ConfigOptions.OptionBuilder autoScalerConfig(String key) {
     public static final ConfigOption<Double> TARGET_UTILIZATION_BOUNDARY =
             autoScalerConfig("target.utilization.boundary")
                     .doubleType()
-                    .defaultValue(0.1)
+                    .defaultValue(0.4)
```
X-czh, line 71: This makes the upper boundary to be 1.0, will this block all upscaling?

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/config/AutoScalerOptions.java
```diff
@@ -87,28 +88,28 @@ private static ConfigOptions.OptionBuilder autoScalerConfig(String key) {
...
-                    .defaultValue(Integer.MAX_VALUE)
+                    .defaultValue(200)
                     .withDescription(
                             "The maximum parallelism the autoscaler can use. Note that this limit will be ignored if it is higher than the max parallelism configured in the Flink config or directly on each operator.");
 
     public static final ConfigOption<Double> MAX_SCALE_DOWN_FACTOR =
             autoScalerConfig("scale-down.max-factor")
                     .doubleType()
-                    .defaultValue(0.6)
+                    .defaultValue(1.0)
```
X-czh, line 98: Curious why we choose to loose it to 1.0. We found in that TPR tends to be overestimated a lot and leading to overly aggressive downscaling when:

- the pipeline is underloaded and far from optimal.
- Avg CPU allocated per slot is < 1.

The reason is that linear scaling with busy time metrics assumes no resource competition between tasks when we pushing up the loads, however, when avg CPU allocated per slot is small, resource competition beween tasks will be more and more severve as we pushing up the overall loads of the pipeline.

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/config/AutoScalerOptions.java
mxm: Not at all. The boundary is used to calculate a scale down and a scale up rate. If the processing capacity falls below the scale up rate, we will scale up to reach the target capacity. If we exceeds the scale down rate, we will scale down to the target capacity. This is a bit counter-intuitive because the upper and lower bound are actually reversed.

A `1.0` utilization for the upscale threshold will lower the scale up rate which means that we delaying upscale to utilize 100% of our processing capacity based on the calculated rates. However, we will still scale up if our processing capacity is lower than the scale up rate. The scale up rate is always computed via the target rate, but the comparison is made using the actual processing capacity.

For example: Let's say we currently have a processing capacity of 100 records/sec. The processing capacity if always estimated at 100% utilization (we call this also *true rate*). At a target rate of 50 records/second (e.g. Kafka ingestion rate), the scale up bound will be 50 rec/s. That means we will only scale up once our processing capacity falls below 50 rec/s. So we will delay scaling as much as possible. If the target rate was to increase to 110 rec/s, we would scale up because our processing capacity of 100 rec/s is now lower.

Similarly, the downscale rate will be raised (instead of lowered) when we increase the utilization boundary. That means we won't scale down as quickly. The tradeoff here is slightly higher resource usage but the scaling becomes less aggressive because we will only scale down once our processing capacity exceeds the know increased "lower bound".

To illutrate this further here some sketches:

Balanced:
```
------ upscale target rate
------ processing capacity (true rate)
------ downscale target rate
```

We will scale up:
```
------ downscale rate
------ upscale rate
------ processing capacity (true rate)
```

We will scale down:

```
------ processing capacity (true rate)
------ downscale rate
------ upscale rate
```

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/config/AutoScalerOptions.java
gyfora: I tend to agree here, I think 0.6 is not a bad default. I could also see 0.7 to be good as well but I would not go higher than that because often a large scale down leads to an immediate scale up due to non-linearity

mxm: I've seen situations where vertices wouldn't be able to scale down all the way to 1 if they were idle. Maybe setting it to `0.5` would still allow that. From a theoretical perspective this makes sense because a halving (0.5) would ensure that you get from parallelism 2 to 1.

gyfora: What I don't understand is why `0.6` is worse than 0.5 :)

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/config/AutoScalerOptions.java
X-czh: I see your point. But I have to admit that it is counter-intuitive and the option description "Target vertex utilization boundary. Scaling won't be performed if utilization is within (target - boundary, target + boundary)" is rather misleading

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/config/AutoScalerOptions.java
mxm: The code looks like it works as expected. I've started a test pipeline to verify is this is still an issue.

@X-czh In my use case there is a CPU per slot, so I might be less affected by this. In any case, we probably keep 0.6 as the default here.

gyfora (approved):

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/config/AutoScalerOptions.java
mxm: True, the description suggests that the scaling is simply based on CPU utilization. I can try to rephrase.

mxm: Rephrased.

trystanj: @mxm your illustration was helpful. I'm having a hard time understanding why the autoscaler makes the decisions it makes. The logs, metrics, and definitions of terms are somewhat vague (e.g. I never knew what was meant by "true processing rate" until I saw you explain it here).

I can open a new ticket to document and perhaps work on it myself, because I think these are critical definitions 😄 

But is source backlog considered in this as well? I have a source which is lagging quite substantially, but never seems to scale out because the "true processing rate" suggests it should be keeping up, but it never does. The catch-up duration is having no effect, either.

At the root of what I am asking is: even if the "true" rate is _calculated_ to be sufficient to catch up, _but it isn't actually catching up_ (let's say because a source is doing something inefficient with its kafka poll settings)... is there any mechanism in place to detect this?

trystanj: Thanks, that makes a lot of sense! Is catchup data rate determined by literal timestamps compared against the catchup duration? eg if a record was placed in kafka 10m ago, and our expected catchup duration is 5m, then are we 5m behind, or are we still 10m behind? just trying to get a better sense of "catch up" statistics!

Perhaps our problem is that lag, for every single job tracked (operator 1.7, Flink 1.18.1, all using `KafkaSource`), is `N/A`. At least according to the exposed operator metrics themselves. If the operator can't see the lag then maybe it can't make an informed decision? I'm wondering if this is a bug on our configuration or maybe I'm just way off base. I should expect to see values for `LAG_Current`, right?

gyfora: lag time is defined as: number of outstanding records / current processing rate (so it's not based on timestamps)
I would double check the pendingRecords metrics from the Kafka source because those should be there

trystanj: Great, thank you!

Yes pendingRecords is present in all jobs and has a value (fluctuating, of course). I have never seen the operator's metric `jobVertexID_LAG_Current` be anything other than N/A, though. I'm observing it via prometheus, so maybe it's just a bug in the translation layer.

(also, if this conversation is out of scope I'd be happy to move it to somewhere that is less tangential!)

trystanj: Just checked in the operator logs as well, to potentially rule out a prom translation issue. I only see `NaN` for all three metrics in the logs.

PR #587, opened 2023-05-08
[FLINK-31717] Fix unit tests using local kube config

mateczagany:

Fix unit tests trying to access the Kubernetes cluster found in `~/.kube/config` as this might be dangerous if there's a production server set in this config.

## Brief change log

- Removed FlinkOperatorITCase
- Modified other tests so that they don't attempt to load `~/.kube/config` 
- Removed minikube-related commands from `test_ci` workflow job

## Verifying this change

- Ran all the non-e2e tests with a malformed `~/kube/config`
- Removed minikube from CI to ensure the unit tests can be ran without Kubernetes

## Documentation


gyfora:
@mateczagany seems like we are losing some test coverage by simply deleting the operator itcase. Why did you decide to delete it?

mateczagany:
I thought this test case was covered by the end-to-end tests.

I wanted to make sure that the user won't accidentally make changes to the Kubernetes cluster that is defined in `~/.kube/config` by simply running the tests, and this test case does exactly that. 

I know that it would require to specifically set `it.skip` to false, but removing this test allowed me to remove minikube from the CI step to make sure all new tests added in the future will not accidentally make changes to the users Kubernetes server.

I have some other solutions in mind if the current solution does not seem fine:
- Restore `FlinkOperatorITCase.java` and run this test in the e2e job of the CI workflow
- Restore `FlinkOperatorITCase.java` and revert changes to `ci.yml` but then we risk adding tests in the future that make changes to the users Kubernetes server

gyfora:
looking at the test now @mateczagany I agree that this is 99% covered by the existing e2e-s. I suggest we go ahead with your proposed changes :)

gyfora:
I will merge this once I have cut the 1.5 release branch hopefully tomorrow

PR #588, opened 2023-05-08
[hotfix] Update CRD compat check to 1.4.0

gyfora:

Update the CRD compatibility check to reference the most recent stable version and remove older versions from the check as the checks should be transitive.


mbalassi (approved):
LGTM

PR #590, opened 2023-05-10
[FLINK-32012] Rely on savepoint mechanism when performing rollback with saveoint upgrade mode

ashangit:

Operator failed to rollback due to missing HA data.
This is due to data being removed by flink job when relying on savepoint upgrade mode.

As performed by upgrade mechanism we should rely on savepoint to do rollback and not on HA data

## Brief change log

During savepoint rollback the HA data is not mandatory. Do not ensure it is available when rollbacking:
- during check if shouldRollback
- during restore of the job

Also enforce deletion of the HA data as performed during upgrade mechanism for savepoint upgrade mode

## Verifying this change

- This change is covered by change performed in RollbackTest.testStatefulRollback.SAVEPOINT
The flinkService now return false for haDataAvailable as if the HA data was deleted during upgrade phase and not recreated due to failure during upgrade phase.
Still the rollback reconcile work fine and the job has well taken in account the last savepoint.

- Also tested on one of our flinkdeployment
  - redeploy flinkdeployment with increase of parallelism work fine
  - then redeploy the flinkdeployment with a service account which didn't have access to s3 savepoint path. The operator successfully rollbacked to previous working state

  - Core observer or reconciler logic that is regularly executed: yes for rollback event with savepoint upgrade mode

## Documentation

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
@@ -384,6 +384,12 @@ public abstract class AbstractFlinkResourceReconciler<
             return false;
         }
 
+        if (resource.getSpec().getJob() != null
+                && resource.getSpec().getJob().getUpgradeMode() == UpgradeMode.SAVEPOINT) {
+            // HA data is not available during rollback for savepoint upgrade mode
+            return true;
+        }
+
         var haDataAvailable = flinkService.isHaMetadataAvailable(configuration);
         if (!haDataAvailable) {
             LOG.warn("Rollback is not possible due to missing HA metadata");
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractJobReconciler.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractJobReconciler.java
@@ -270,17 +270,13 @@ public abstract class AbstractJobReconciler<
 
         UpgradeMode upgradeMode = resource.getSpec().getJob().getUpgradeMode();
 
-        cancelJob(
-                ctx,
-                upgradeMode == UpgradeMode.STATELESS
-                        ? UpgradeMode.STATELESS
-                        : UpgradeMode.LAST_STATE);
+        cancelJob(ctx, upgradeMode);
 
         restoreJob(
                 ctx,
                 rollbackSpec,
                 ctx.getDeployConfig(rollbackSpec),
-                upgradeMode != UpgradeMode.STATELESS);
+                upgradeMode == UpgradeMode.LAST_STATE);
 
         reconciliationStatus.setState(ReconciliationState.ROLLED_BACK);
     }
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
@@ -36,6 +36,7 @@ import org.apache.flink.kubernetes.operator.api.spec.JobSpec;
 import org.apache.flink.kubernetes.operator.api.spec.UpgradeMode;
 import org.apache.flink.kubernetes.operator.api.status.FlinkDeploymentStatus;
 import org.apache.flink.kubernetes.operator.api.status.JobManagerDeploymentStatus;
+import org.apache.flink.kubernetes.operator.api.status.ReconciliationState;
 import org.apache.flink.kubernetes.operator.api.status.Savepoint;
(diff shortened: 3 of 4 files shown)


ashangit:
@gyfora please review

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
```diff
@@ -384,6 +384,12 @@ private boolean shouldRollBack(
             return false;
         }
 
+        if (resource.getSpec().getJob() != null
+                && resource.getSpec().getJob().getUpgradeMode() == UpgradeMode.SAVEPOINT) {
+            // HA data is not available during rollback for savepoint upgrade mode
+            return true;
+        }
```
gyfora, line 391: This check is not fully correct. Just because we started from a savepoint if the jobmanager started succesfully HA metadata would be available even if the job could not recover after the upgrade. 

Instead of simply checkping for savepoint upgrade mode we also have to check `FlinkUtils.jmPodNeverStarted(ctx)` to be completely safe.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractJobReconciler.java
```diff
@@ -270,17 +270,13 @@ protected void rollback(FlinkResourceContext<CR> ctx) throws Exception {
 
         UpgradeMode upgradeMode = resource.getSpec().getJob().getUpgradeMode();
 
-        cancelJob(
-                ctx,
-                upgradeMode == UpgradeMode.STATELESS
-                        ? UpgradeMode.STATELESS
-                        : UpgradeMode.LAST_STATE);
+        cancelJob(ctx, upgradeMode);
```
gyfora, line 273: Why did you change this? We don't want to use a savepoint cancel during rollback

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
```diff
@@ -338,6 +339,10 @@ protected void cancelJob(
                     } else if (ReconciliationUtils.isJobInTerminalState(deploymentStatus)) {
                         LOG.info(
                                 "Job is already in terminal state skipping cancel-with-savepoint operation.");
+                    } else if (deployment.getStatus().getReconciliationStatus().getState()
+                            == ReconciliationState.ROLLING_BACK) {
+                        LOG.info(
+                                "Job reconciliation status is in rolling back state. Job is expecting to not be in running or terminal state");
```
gyfora, line 345: We can't simply ignore errors here because the cluster may not have shut down. Please see my other comment, I dont think we should use savepoint cancel ever during rollbaack

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractJobReconciler.java
ashangit: I wanted to ensure that HA data is well deleted before to restoreJob for rollback also.

It looks to me that we should apply same pattern for upgrade and rollback relying only on the savepoint and ensuring HA data is here.
For ex. if the upgrade was linked to bump of flink version with some changes in HA data not compatible with rollbacked flink version

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
ashangit: Sorry but I'm not sure to see the issue if we just checkping for savepoint upgrade.
If the HA metadata has indeed been created by the jobmanger but the job has not recover we should not have any new checkpoint, am I wrong?
As in the rollback section we ensure that HA metadata is well deleted (in `cancelJob`) it should still be safe to 'not care' about this HA metadata, no?

gyfora: This is not that simple. The operator only periodically (lets say every 30sec) observes the job. It is possible that the job started, checkpointed and failed in that 30 sec period so we cannot simply ignore the HA metadata. If it's there we have to use it, we cannot delete it.

ashangit: Ok that was the thing I was missing

So we should indeed check that JM never start and ensure we rely on HA metadata if it exist otherwise rely on savepoint

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractJobReconciler.java
ashangit: Indeed it would be better
I will look into this and contact you through the original FLINK ticket

PR #596, opened 2023-05-14
fix: flink account cannot get resource "services" in API group

huaxk:

When I remotely connected to the Flink cluster using local SQL client, I encountered an error while submitting an SQL statement:
```
[ERROR] Could not execute SQL statement. Reason:
io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: GET at: https://10.53.0.1/api/v1/namespaces/flink-operator/services/basic-example-rest. Message: Forbidden!Configured service account doesn't have access. Service account may have been revoked. services "basic-example-rest" is forbidden: User "system:serviceaccount:flink-operator:flink" cannot get resource "services" in API group "" in the namespace "flink-operator".
```
this error indicating that Flink account lacks permission to operate services resource, missing permissions should be added to the Flink account in chart templates.

### DIFF
--- a/helm/flink-kubernetes-operator/templates/rbac.yaml
+++ b/helm/flink-kubernetes-operator/templates/rbac.yaml
@@ -87,6 +87,7 @@ rules:
     resources:
       - pods
       - configmaps
+      - services
     verbs:
       - '*'
   - apiGroups:


mbalassi:
Hi @huaxk! Connecting to a cluster managed by the Kubernetes Operator is an antipattern in its current form, because the operator will not be aware of the jobs you submit this way. Only jobs that are represented by the Kubernetes custom resource (as of today `FlinkDeployment` or `FlinkSessionJob`) can be managed by the operator. Adding an SQL gateway type functionality is interesting, but it would require a larger reconsideration. You can run SQL jobs directly via the operator as presented in this [example](https://github.com/apache/flink-kubernetes-operator/tree/main/examples/flink-sql-runner-example).

huaxk:
Yes, I see.

prakash-42:
Hi @mbalassi ! Is it possible to get this information added to the official documentation of the flink-kubernetes-operator?

I was stuck on the same issue for very long, and later updated the flink role to give access to "service" resource; only to find this discussion and realize that I created an antipattern.

It'd be very helpful if this was available in the main document itself. I can help in that documentation update if needed. Thank you!

mxm:
@prakash-42 You would greatly help us out here if you could open a PR against the this repo with the documentation changes 🙏

prakash-42:
Thanks @mxm  ! I have a few questions before I can start working on the PR, hopefully this is the right place to ask.

1. Where should I put this guide (new document) in the project structure? I suggest that we create a folder by the name "examples" (in the [docs](https://github.com/apache/flink-kubernetes-operator/tree/main/docs/content/docs) folder), and create this single guide in there. We can later add more examples to the folder. What do you think?

2. I will copy a lot of content from the readme of the [flink-sql-runner-example](https://github.com/apache/flink-kubernetes-operator/tree/main/examples/flink-sql-runner-example), with maybe a few additions. Is there a way to directly add that file as a page instead of duplicating content?

mxm:
I would add it on this page: https://nightlies.apache.org/flink/flink-kubernetes-operator-docs-main/docs/custom-resource/overview/ Actually, the examples linked above are already linked at the bottom of the page. Maybe we can add a section which expands a little bit on the concept.

prakash-42:
Sorry to bug you again @mxm , but where is the place to submit bugs/ask questions for the flink operator? (The JIRA project doesn't have public signup, are the flink mailing lists active?)

I tried to use the [flink-sql-runner-example](https://github.com/apache/flink-kubernetes-operator/tree/main/examples/flink-sql-runner-example) before writing documentation about it, and stumbled into two issues:

1. **The regex for SET statement is not flexible. (not a blocker)**
  [The regex for SET statement](https://github.com/apache/flink-kubernetes-operator/blob/release-1.7/examples/flink-sql-runner-example/src/main/java/org/apache/flink/examples/SqlRunner.java#L44) enforces that the user provides single quote (`'`) around the key and value. Also there must be space on both sides of the equals (`=`)

2. **Any text present between backticks(\`) is getting lost** (unable to debug the reason)
  When I try to submit a script that uses backticks around field names, the text between backticks isn't read by the code. ([FileUtils.readFileUtf8](https://github.com/apache/flink-kubernetes-operator/blob/release-1.7/examples/flink-sql-runner-example/src/main/java/org/apache/flink/examples/SqlRunner.java#L55) is responsible for this, I added a log). (I used `flink:1.17` as my base image). I'm not able to reproduce this behavior locally, only happens when I package the job and deploy it in our EKS cluster.

  Here's a portion of SQL script I used (I printed the full script to console, even the content b/w backticks from the comments was gone)
 ```
 SET 'execution.checkpointing.interval' = '30s';
SET 'pipeline.name' = 'MyFlinkJob';

-- some comment text in `back ticks`
-- some more text in ```back ticks x3```

CREATE TABLE Library_Book_Source (
   database_name STRING METADATA VIRTUAL,
   table_name STRING METADATA VIRTUAL,
   `ID` INT NOT NULL,
   Title VARCHAR(255),
   `Price` VARCHAR(50),
   PRIMARY KEY (`ID`) NOT ENFORCED
 ) WITH (
   'connector' = 'mysql-cdc',
   'hostname' = 'DB_HOST',
   'port' = '3306',
   'username' = 'username',
   'password' = 'password',
   'database-name' = 'Library',
   'table-name' = 'Book',
   'jdbc.properties.useSSL' = 'false'
 );
```

Due to these problems, I'm not sure if this example is production ready and whether or not it should be added in our documentation.

mxm:
@prakash-42 You can request a JIRA account here: https://selfserve.apache.org/jira-account.html Let me know the username once you have requested an account, I can then approve your request.

prakash-42:
Hi @mxm ! I have submitted the request for account with username `prakash.tiwari`

mxm:
Great! The request is approved 👍

PR #597, opened 2023-05-15
[FLINK-32100] Use the total number of Kafka partitions as the max source parallelism

mxm:
So far, we've taken the max number partitions we can find. However, the correct way to calculate the max source parallelism would be to sum the number of partitions of all topis.

### DIFF
--- a/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/ScalingMetricCollector.java
+++ b/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/ScalingMetricCollector.java
@@ -58,6 +58,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.SortedMap;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 /** Metric collector using flink rest api. */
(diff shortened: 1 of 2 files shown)


mxm:
We don't know how many divisors the number of partitions of a topic has. Similarly, we don't know how how many divisors the sum of the number of partitions of all topics has. What you say applies to any operator, not only to source operators. Source operators are limited by the max partitions which are usually smaller than the configured max vertex which will always be the max also for all vertices including sources.

From my experience the worst thing that can happen is that the source is underprovisioned which this PR tries to avoid. Underprovisioned sources suffer from watermark alignment issues and will build huge lag. We need to allow sources to reach the max scale up which is the sum of all partitions.

Let's say you have 73 (prime number) partitions because you read from topic1 with 60 partitions and topic2 with 13 partitions. The autoscaler can now decide to scale between 1 and 73. Any of these parallelisms will work. In case of a non-prime sum of partitions, the scale up will at worst case be `Math.min(sum_of_partitions / 2, max_parallelism)` in case of key-group alignment. Otherwise, we will choose the target parallelism like in the prime example.

I think we can think about adding a config option to disable keygroup alignment.

gyfora (approved):
Makes sense, I think we can think about this in the long run, and consider a flag for the keygroup alignment in general. Also keygroup alignment should not apply to stateless (no keyed state at least)operators such as sources

mxm:
I agree, probably makes sense to disable key group alignment for sources.

PR #598, opened 2023-05-15
[FLINK-32102] Aggregate multiple pendingRecords metric per source if present

mxm:
Some source expose multiple `.pendingRecords` metrics. If that is the case, we must sum up these records to yield the correct internal pending records count.


mxm:
This is a fix for https://cwiki.apache.org/confluence/display/FLINK/FLIP-246%3A+Multi+Cluster+Kafka+Source which we already use internally. It exposes multiple pending records metrics, one for each cluster.

PR #603, opened 2023-05-16
[FLINK-32111] - Add Null  check for checkpoints history obj

tamirsagi:

Fix NPE while getting in progress checkpoints.

## Brief change log

- Add null check CheckpointHistoryWrapper#getInProgressCheckpoint

## Verifying this change

- Added 2 Unitests to verify NPE is not thrown when calling getInProgressCheckpoint when history obj is null.

  - Dependencies (does it add or upgrade a dependency): (yes / no) no
  - The public API, i.e., is any changes to the `CustomResourceDescriptors`: (yes / no) no
  - Core observer or reconciler logic that is regularly executed: (yes / no) no

## Documentation

  - Does this pull request introduce a new feature? (yes / no) no

### DIFF
--- a/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/service/NativeFlinkServiceTest.java
+++ b/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/service/NativeFlinkServiceTest.java
@@ -66,11 +66,13 @@ import org.junit.jupiter.params.provider.MethodSource;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Map;
+import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 
 import static org.apache.flink.kubernetes.operator.config.FlinkConfigBuilder.FLINK_VERSION;
 import static org.apache.flink.kubernetes.operator.config.KubernetesOperatorConfigOptions.OPERATOR_HEALTH_PROBE_PORT;
 import static org.apache.flink.kubernetes.operator.config.KubernetesOperatorConfigOptions.OPERATOR_SAVEPOINT_FORMAT_TYPE;
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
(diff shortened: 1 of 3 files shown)


morhidi:
Thanks for taking a stab at this @tamirsagi. Could you please address the formatting issues? Hint: run `mvn spotless:apply` The CI/CD pipeline fails with this:
```
src/test/java/org/apache/flink/kubernetes/operator/service/NativeFlinkServiceTest.java:[27] (imports) ImportOrder: Import org.apache.flink.api.common.JobID appears after other imports that it should precede
[593](https://github.com/apache/flink-kubernetes-operator/actions/runs/4994518856/jobs/8946617823?pr=603#step:7:596)
#24 114.5 [ERROR] src/test/java/org/apache/flink/kubernetes/operator/service/NativeFlinkServiceTest.java:[58] (imports) ImportOrder: 'org.apache.flink.shaded.jackson2.com.fasterxml.jackson.core.JsonProcessingException' should be separated from previous imports.
[594](https://github.com/apache/flink-kubernetes-operator/actions/runs/4994518856/jobs/8946617823?pr=603#step:7:597)
#24 114.5 [ERROR] src/test/java/org/apache/flink/kubernetes/operator/service/NativeFlinkServiceTest.java:[60] (imports) ImportOrder: 'org.junit.jupiter.api.Assertions' should be separated from previous imports.
```

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/service/NativeFlinkServiceTest.java
```diff
@@ -307,6 +308,30 @@ public void testGetLastSavepointRestCompatibility() throws JsonProcessingExcepti
...
+
+        val checkpointHistoryWrapperV14 = objectMapper.readValue(flink14Response, CheckpointHistoryWrapper.class);
```
morhidi, line 319: Why using `val` here?

tamirsagi: I use `val` (which is `final var`) rather than just `var` . but to keep the tests aligned I changed it to` var` as used elsewhere.

gyfora:
Please run `mvn clean install` locally before you push to avoid the need to run the CI again and again. There is still a spotless formatting issue

tamirsagi:
> Please run `mvn clean install` locally before you push to avoid the need to run the CI again and again. There is still a spotless formatting issue

Hey Gyula,

I ran it locally prior MR, there were policy errors that are not related to my changes. for instance
com.diffplug.spotless.FormatExceptionPolicyLegacy error
SEVERE: Step 'removeUnusedImports' found problem in 'src\main\java\org\apache\flink\kubernetes\operator\kubeclient\decorators\CmdStandaloneJobManagerDecorator.java':null
Caused by: java.lang.IllegalAccessError: class com.google.googlejavaformat.java.RemoveUnusedImports (in unnamed module @0x70a984c5) cannot access class com.sun.tools.javac.util.Context (in module jdk.compiler) because module jdk.compiler does not export com.sun.tools.javac.util to unnamed module @0x70a984c5

I believe it's related to that fact that JDK 17 is being used.

I added a simple null check and 2 tests and got errors regarding format issues (First import order and now something else) . I checked the failure in CI but it was not clear what exactly the violation is. A little assistance with the format error would be greatly appreciated.

gyfora:
You should try using Java 11 and make sure the spotless plugin runs successfully. That will format the code for you.

we haven’t tested with Java 17

PR #604, opened 2023-05-19
[FLINK-32041] - Allow operator to manage leases when using watchNamespaces

tchin25:

The operator doesn't have the ability to manage leases in its own namespace when using `watchNamespaces`. This fixes that by adding a `Role` and `RoleBinding` for it when `watchNamespaces` is set.

## Brief change log

  - Add Role and RoleBinding for the operator in release namespace when using namespace-scoped RBAC
  - Remove the ability to create leases from `watchNamespaces` when using namespace-scoped RBAC

## Verifying this change

This change is a trivial rework / code cleanup without any test coverage. We do, however, currently have this change [deployed](https://github.com/wikimedia/operations-deployment-charts/commit/e06d57b5bdf06f827e4e1ca081320aee4cc82cb0) in our production setup

## Documentation

### DIFF
--- a/helm/flink-kubernetes-operator/templates/rbac.yaml
+++ b/helm/flink-kubernetes-operator/templates/rbac.yaml
@@ -71,17 +71,12 @@ rules:
       - ingresses
     verbs:
       - "*"
-{{/*
-watchNamespaces doesn't need this since the operator handles it
-*/}}
-{{- if not .Values.watchNamespaces }}
   - apiGroups:
       - coordination.k8s.io
     resources:
       - leases
     verbs:
       - "*"
-{{- end }}
 {{- end }}
 
 {{/*
@@ -179,7 +174,7 @@ subjects:
 {{/*
 Give operator the ability to operate on leases in the release namespace
 */}}
-{{- if .Values.rbac.operatorRole.create }}
+{{- if and .Values.rbac.operatorRole.create (not (has .Release.Namespace .Values.watchNamespaces)) }}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
@@ -196,7 +191,7 @@ rules:
       - "*"
 {{- end }}
 ---
-{{- if .Values.rbac.operatorRoleBinding.create }}
+{{- if and .Values.rbac.operatorRole.create (not (has .Release.Namespace .Values.watchNamespaces)) }}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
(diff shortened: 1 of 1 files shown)


helm/flink-kubernetes-operator/templates/rbac.yaml
```diff
@@ -71,13 +71,18 @@ rules:
...
+watchNamespaces doesn't need this since the operator handles it
+*/}}
+{{- if not .Values.watchNamespaces }}
   - apiGroups:
       - coordination.k8s.io
     resources:
       - leases
     verbs:
       - "*"
 {{- end }}
```
gyfora, line 84: I don't completely understand this. Could we simply remove this entire if logic? It seems that the required role is created later anyways

tchin25: I added the conditional since it doesn't seem like the namespaces in `watchNamespaces` uses leases, but I didn't want the behavior to change for cluster-scoped rules since on line 229 it uses the same rules template.

Unless I've overlooked something when you say that
> It seems that the required role is created later anyways

gyfora:
I see the following error on the e2es:
```
client.go:134: [debug] creating 21 resource(s)
Error: INSTALLATION FAILED: 2 errors occurred:
	* roles.rbac.authorization.k8s.io "flink-operator" already exists
	* rolebindings.rbac.authorization.k8s.io "flink-operator-role-binding" already exists


helm.go:84: [debug] 2 errors occurred:
	* roles.rbac.authorization.k8s.io "flink-operator" already exists
	* rolebindings.rbac.authorization.k8s.io "flink-operator-role-binding" already exists
```
@tchin25 can you please check? maybe some overlapping role names

tchin25:
Oh, I found the issue. The CI has `default` in `watchNamespaces` while the release namespace is also `default`, so it makes the role twice, one with the leases resource and one without it. Trying to think of the most elegant way to handle this

gyfora (approved):

ottomata:
Thanks @gyfora !  QQ: how long do you think it will take for this to make it into a release?  

I ask just for our own 'fork'. We'd prefer to apply these changes directly from your release rather than manually making them ourselves.

gyfora:
@ottomata we generally try to follow a 2 month release cycle. Details can be found here: https://cwiki.apache.org/confluence/display/FLINK/Release+Schedule+and+Planning

In your case I see no harm in using the helm chart from the `main` branch.

ottomata:
Okay, thank you!

PR #606, opened 2023-05-23
[FLINK-31684] Report autoscaling metrics before metric window is full

mxm:
The metrics get reported only after the metric window is full. This is not helpful for observability after rescaling. We need to make sure that metrics are reported even when the metric window is not yet full.


gyfora (approved):
Nice improvement!

PR #607, opened 2023-05-24
[FLINK-32148] Make operator logging less noisy

gyfora:

Reduce logging noise from the operator by changing the log level to debug in many cases. Currently the operator is very noisy and outputs a lot of redundant / usless logs during normal operations.

## Verifying this change

Tested locally.

## Documentation

### DIFF
--- a/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/JobVertexScaler.java
+++ b/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/JobVertexScaler.java
@@ -221,7 +221,7 @@ public class JobVertexScaler {
                 message);
 
         if (conf.get(AutoScalerOptions.SCALING_EFFECTIVENESS_DETECTION_ENABLED)) {
-            LOG.info(
+            LOG.warn(
                     "Ineffective scaling detected for {}, expected increase {}, actual {}",
                     vertex,
                     expectedIncrease,
--- a/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/metrics/ScalingMetrics.java
+++ b/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/metrics/ScalingMetrics.java
@@ -72,7 +72,7 @@ public class ScalingMetrics {
 
         if (isSource) {
             double sourceDataRate = Math.max(0, numRecordsInPerSecond + lagGrowthRate);
-            LOG.info("Using computed source data rate {} for {}", sourceDataRate, jobVertexID);
+            LOG.debug("Using computed source data rate {} for {}", sourceDataRate, jobVertexID);
             scalingMetrics.put(ScalingMetric.SOURCE_DATA_RATE, sourceDataRate);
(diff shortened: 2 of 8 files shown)


flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/JobVertexScaler.java
```diff
@@ -221,7 +221,7 @@ private boolean detectIneffectiveScaleUp(
                 message);
 
         if (conf.get(AutoScalerOptions.SCALING_EFFECTIVENESS_DETECTION_ENABLED)) {
-            LOG.info(
+            LOG.warn(
```
tweise, line 224: Why should this be warn? This looks more like something that should be reflected in the status?

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/metrics/ScalingMetrics.java
```diff
@@ -82,7 +82,7 @@ public static void computeDataRateMetrics(
             scalingMetrics.put(ScalingMetric.TRUE_PROCESSING_RATE, trueProcessingRate);
             scalingMetrics.put(ScalingMetric.CURRENT_PROCESSING_RATE, numRecordsInPerSecond);
         } else {
-            LOG.error("Cannot compute true processing rate without numRecordsInPerSecond");
+            LOG.warn("Cannot compute true processing rate without numRecordsInPerSecond");
```
tweise, line 85: Same as above, the logging is probably not the best way to bring this to the intended audience and warn should probably be mostly reserved for operator internal problems.

tweise (approved):
LGTM for the info => debug changes. The autoscaler observability probably needs more discussion, which can happen after this PR.

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/JobVertexScaler.java
gyfora: There is already an event triggered for this, that should be enough for most users. The log just outputs some additional info

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/metrics/ScalingMetrics.java
gyfora: We could trigger an event, but putting this into the status may be an overkill because this is not really actionable.

tweise: Not actionable is also why I'm not sure about logging at warn level. Who is going to read these logs? Will logs be flooded with repeating entries that are unlikely to be addressed? I thought a message in the status (or one time event) might be more useful, after all the application would need to be modified to remedy the condition.

mxm (approved):
@tweise I agree that we need to revisit some of the logging. I think this PR is a step in the right direction. Concerning observability in general, I think in addition to the metrics and events we already emit, we could add an autoscaler status field to the resource.

PR #609, opened 2023-05-25
[FLINK-32171] Add PostStart hook to flink k8s operator helm

xccui:

Introduce postStart hook to allow users to initialize the main container with some commands. For instance, users can download flink-fs-s3-hadoop plugin to enable using a JAR from s3 in session mode. There's no need to build their own docker images.

## Brief change log

* Added a helm value `postStart` that allows users to specify postStart hook of the main container.
* Some doc changes.

## Verifying this change

## Documentation


helm/flink-kubernetes-operator/values.yaml
```diff
@@ -183,3 +183,7 @@ operatorHealth:
   startupProbe:
     failureThreshold: 30
     periodSeconds: 10
+
+# Set exec command for the postStart hook of the main container
+postStart:
+  execCommand: []
```
gyfora, line 189: I think it would be more intuitive to provide the entire yaml content of the `postStart` section instead of `execCommand` which would require extra documentation and would limit the usage.

xccui: Sounds good. I'll make an update

PR #610, opened 2023-06-03
[FLINK-32243] Bump okhttp version to 4.11.0

czy006:
Bump okhttp version to 4.11.0


gyfora:
please update the NOTICE file in flink-kubernetes-operator accordingly

PR #613, opened 2023-06-07
[FLINK-32271] Report RECOMMENDED_PARALLELISM as an autoscaler metric

morhidi:

It is beneficial to report the recommended parallelism and overlay it with the current parallelism on the same chart when auto scaler is running in advisor mode.

## Brief change log
- Added a new `RECOMMENDED_PARALLELISM` entry to `ScalingMetric`
- The latest recommended parallelisms are being reported as metrics (gauges)
- Recommended parallelisms follow the logic below:
  - `RECOMMENDED_PARALLELISM` will be set to `PARALLELISM` while the metric window is filling up
  - `RECOMMENDED_PARALLELISM` may change according to the evaluated scaling metrics
  - `PARALLELISM` will be set to `RECOMMENDED_PARALLELISM` after scaling if scaling is enabled
  - `RECOMMENDED_PARALLELISM` will set to `null` during a scaling operation, while metric collection is not possible
   - `RECOMMENDED_PARALLELISM` will be set again to the changed `PARALLELISM` while the metric window is filling up

## Verifying this change

- Added unit test for recommended parallelisms
- Manual by checking the actual metrics being reported

```
jobVertexID.90bea66de1c231edf33913ecd54406c1.PARALLELISM.Current: 1.0
jobVertexID.90bea66de1c231edf33913ecd54406c1.PARALLELISM.Current: 1.0
jobVertexID.90bea66de1c231edf33913ecd54406c1.RECOMMENDED_PARALLELISM.Current: 1.0
jobVertexID.90bea66de1c231edf33913ecd54406c1.PARALLELISM.Current: 1.0
jobVertexID.90bea66de1c231edf33913ecd54406c1.RECOMMENDED_PARALLELISM.Current: 1.0
jobVertexID.90bea66de1c231edf33913ecd54406c1.PARALLELISM.Current: 1.0
jobVertexID.90bea66de1c231edf33913ecd54406c1.RECOMMENDED_PARALLELISM.Current: 6.0
jobVertexID.90bea66de1c231edf33913ecd54406c1.PARALLELISM.Current: 1.0
jobVertexID.90bea66de1c231edf33913ecd54406c1.RECOMMENDED_PARALLELISM.Current: null
jobVertexID.90bea66de1c231edf33913ecd54406c1.PARALLELISM.Current: 6.0
jobVertexID.90bea66de1c231edf33913ecd54406c1.RECOMMENDED_PARALLELISM.Current: 6.0
jobVertexID.90bea66de1c231edf33913ecd54406c1.PARALLELISM.Current: 6.0
jobVertexID.90bea66de1c231edf33913ecd54406c1.RECOMMENDED_PARALLELISM.Current: 6.0
jobVertexID.90bea66de1c231edf33913ecd54406c1.PARALLELISM.Current: 6.0
jobVertexID.90bea66de1c231edf33913ecd54406c1.RECOMMENDED_PARALLELISM.Current: 6.0
```

## Documentation


gyfora (changes requested):
Thanks @morhidi . I think this metric is useful for monitoring the autoscaler especially if automatic scaling is turned off.

However I think we need to add some logic to update the recommended parallelism when the job status / jobgraph changes as well. Whenever we reset the metrics we should also reset the recommended parallelism to the current job vertex parallelism. What do you think?

cc @mxm

morhidi:
> Thanks @morhidi . I think this metric is useful for monitoring the autoscaler especially if automatic scaling is turned off.
> 
> However I think we need to add some logic to update the recommended parallelism when the job status / jobgraph changes as well. Whenever we reset the metrics we should also reset the recommended parallelism to the current job vertex parallelism. What do you think?
> 
> cc @mxm

Makes sense to me, I'll sync-up with you offline and amend the PR with the recommended changes

morhidi:
> > Thanks @morhidi . I think this metric is useful for monitoring the autoscaler especially if automatic scaling is turned off.
> > However I think we need to add some logic to update the recommended parallelism when the job status / jobgraph changes as well. Whenever we reset the metrics we should also reset the recommended parallelism to the current job vertex parallelism. What do you think?
> > cc @mxm
> 
> Makes sense to me, I'll sync-up with you offline and amend the PR with the recommended changes

@gyfora I've updated the PR description with the logic we discussed offline, and implemented the changes we agreed, have another look pls

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/JobAutoScalerImpl.java
```diff
@@ -143,4 +163,18 @@ private AutoscalerFlinkMetrics getOrInitAutoscalerFlinkMetrics(
...
+            ResourceID resourceID) {
+        var parallelisms = new HashMap<JobVertexID, Integer>();
+        lastEvaluatedMetrics.forEach(
+                (jobVertexID, map) -> {
+                    parallelisms.put(
+                            jobVertexID, (int) map.get(ScalingMetric.PARALLELISM).getCurrent());
+                });
+        recommendedParallelisms.put(resourceID, parallelisms);
+        LOG.debug("Recommended parallelisms are reset to current {}", parallelisms);
+    }
```
gyfora, line 179: This could be removed and unified with the `updateRecommendedParallelisms` method (make it static utility maybe?) as the resetting is the same as updating with empty scaling summaries.

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/JobAutoScalerImpl.java
```diff
@@ -99,22 +104,37 @@ public boolean scale(FlinkResourceContext<? extends AbstractFlinkResource<?, ?>>
...
             LOG.debug("Evaluating scaling metrics for {}", collectedMetrics);
             var evaluatedMetrics = evaluator.evaluate(conf, collectedMetrics);
             LOG.debug("Scaling metrics evaluated: {}", evaluatedMetrics);
             lastEvaluatedMetrics.put(resourceId, evaluatedMetrics);
-            flinkMetrics.registerScalingMetrics(() -> lastEvaluatedMetrics.get(resourceId));
+
+            flinkMetrics.registerEvaluatedScalingMetrics(
+                    () -> lastEvaluatedMetrics.get(resourceId));
+            flinkMetrics.registerRecommendedParallelismMetrics(
+                    () -> recommendedParallelisms.get(resourceId));
```
gyfora, line 121: I think we should pass these together because they both need to be provided

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/JobAutoScalerImpl.java
```diff
@@ -99,22 +104,37 @@ public boolean scale(FlinkResourceContext<? extends AbstractFlinkResource<?, ?>>
...
             }
 
             var specAdjusted =
-                    scalingExecutor.scaleResource(resource, autoScalerInfo, conf, evaluatedMetrics);
+                    scalingExecutor.scaleResource(
+                            resource,
+                            autoScalerInfo,
+                            conf,
+                            evaluatedMetrics,
+                            recommendedParallelisms);
```
gyfora, line 136: Not something that should be fixed in this PR but we should consider a refactor by introducing some sort of autoscaler resource context to avoid so many params

gyfora (approved):
The changes look really good! 👍 

I had 2 minor cosmetic comments, please feel free to merge the PR after fixing those (and passing CI).

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/ScalingMetricCollector.java
```diff
@@ -100,7 +100,9 @@ public CollectedMetricHistory updateMetrics(
             cleanup(cr);
             metricHistory.clear();
             metricCollectionStartTs = now;
+            cleanupRecommendedParallelisms(recommendedParallelisms, resourceID);
         }
+        var topology = getJobTopology(flinkService, cr, conf, autoscalerInfo);
```
gyfora, line 105: Moving the topology assignment is not necessary right? (I made changes to this logic in my open PR so it would be better to not move this if not necessary)

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/AutoscalerFlinkMetrics.java
```diff
@@ -45,7 +43,13 @@ public class AutoscalerFlinkMetrics {
...
     private final MetricGroup metricGroup;
 
-    private final Set<JobVertexID> vertexMetrics = new HashSet<>();
+    private final Map<JobVertexID, MetricGroup> vertexMetricGroups = new ConcurrentHashMap<>();
+    private final Map<Tuple2<JobVertexID, ScalingMetric>, MetricGroup> scalingMetricGroups =
+            new ConcurrentHashMap<>();
+    private final Map<Tuple2<JobVertexID, ScalingMetric>, Gauge<Double>> currentScalingMetrics =
+            new ConcurrentHashMap<>();
+    private final Map<Tuple2<JobVertexID, ScalingMetric>, Gauge<Double>> averageScalingMetrics =
+            new ConcurrentHashMap<>();
```
mxm, line 52: Please revert back to using non-current hash map. This is not accessed concurrently.

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/AutoscalerFlinkMetrics.java
```diff
@@ -54,51 +58,122 @@ public AutoscalerFlinkMetrics(MetricGroup metricGroup) {
...
+                                    Tuple2.of(jobVertexID, RECOMMENDED_PARALLELISM),
+                                    key ->
+                                            smg.gauge(
+                                                    "Current",
+                                                    () ->
+                                                            getCurrentValue(
+                                                                    recommendedParallelisms,
+                                                                    jobVertexID)));
+                        });
+    }
```
mxm, line 108: Why does this require a separate method? The recommended parallelism should just be exposed like any other evaluated scaling metric.

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/AutoscalerFlinkMetrics.java
```diff
@@ -54,51 +58,122 @@ public AutoscalerFlinkMetrics(MetricGroup metricGroup) {
...
-                                                            Optional.ofNullable(
-                                                                            currentVertexMetrics
-                                                                                    .get())
-                                                                    .map(m -> m.get(jobVertexID))
-                                                                    .map(
-                                                                            metrics ->
-                                                                                    metrics.get(sm)
-                                                                                            .getAverage())
-                                                                    .orElse(null));
-                                        }
```
mxm, line 100: As far as I understand, this should have been a separate commit or PR. The refactoring wasn't necessary to make this work.

flink-kubernetes-operator-autoscaler/src/test/java/org/apache/flink/kubernetes/operator/autoscaler/BacklogBasedScalingTest.java
```diff
@@ -1,437 +0,0 @@
...
-
-    private void assertFlinkMetricsCount(
-            int scalingCount, int balancedCount, TestUtils.TestingContext<HasMetadata> ctx) {
-        AutoscalerFlinkMetrics autoscalerFlinkMetrics =
-                autoscaler.flinkMetrics.get(
-                        ResourceID.fromResource(getResourceContext(app, ctx).getResource()));
-        assertEquals(scalingCount, autoscalerFlinkMetrics.numScalings.getCount());
-        assertEquals(balancedCount, autoscalerFlinkMetrics.numBalanced.getCount());
-    }
-}
```
mxm, line 437: Can we keep this test class and keep the refactoring to a minimum in this feature PR?

flink-kubernetes-operator-autoscaler/src/test/java/org/apache/flink/kubernetes/operator/autoscaler/JobAutoScalerImplTest.java
```diff
@@ -18,59 +18,537 @@
...
+                        evaluator,
+                        scalingExecutor,
+                        new EventRecorder(kubernetesClient, eventCollector));
+
+        // Reset custom window size to default
+        metricsCollector.setTestMetricWindowSize(null);
+    }
+
+    @Test
+    public void testBacklogBasedScaling() throws Exception {
```
mxm, line 129: Is this a new test or an existing one? It's hard to review when there are many unrelated changes.

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/AutoscalerFlinkMetrics.java
```diff
@@ -54,51 +58,122 @@ public AutoscalerFlinkMetrics(MetricGroup metricGroup) {
...
+                                        registerScalingMetric(
+                                                currentVertexMetrics,
+                                                jobVertexID,
+                                                scalingMetric,
+                                                smg);
                                     });
                         });
     }
+
+    public void registerRecommendedParallelismMetrics(
```
mxm, line 83: (1) Special casing for the PR feature in new component

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/JobAutoScalerImpl.java
```diff
@@ -143,4 +163,18 @@ private AutoscalerFlinkMetrics getOrInitAutoscalerFlinkMetrics(
                         new AutoscalerFlinkMetrics(
                                 ctx.getResourceMetricGroup().addGroup("AutoScaler")));
     }
+
+    private void resetRecommendedParallelisms(
```
mxm, line 167: (2) Special casing for the PR feature in new component

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/ScalingExecutor.java
```diff
@@ -269,6 +273,25 @@ private void setVertexParallelismOverrides(
         resource.getSpec().setFlinkConfiguration(flinkConf.toMap());
     }
 
+    private void updateRecommendedParallelisms(
```
mxm, line 276: (3) Special casing for the PR feature in new component

flink-kubernetes-operator-autoscaler/src/test/java/org/apache/flink/kubernetes/operator/autoscaler/MetricsCollectionAndEvaluationTest.java
```diff
@@ -68,6 +69,12 @@ public class MetricsCollectionAndEvaluationTest {
 
     private final AutoScalerInfo scalingInfo = new AutoScalerInfo(new HashMap<>());
 
+    private final Map<ResourceID, Map<JobVertexID, Integer>> lastScalingSummaries =
+            new ConcurrentHashMap<>();
+
+    private final Map<ResourceID, Map<JobVertexID, Integer>> recommendedParallelisms =
+            new ConcurrentHashMap<>();
```
mxm, line 76: Test is not concurrent, hence normal HashMap would suffice and would be advisable to use here.

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/ScalingMetricCollector.java
```diff
@@ -441,6 +443,17 @@ public void cleanup(AbstractFlinkResource<?, ?> cr) {
         topologies.remove(resourceId);
     }
 
+    private void cleanupRecommendedParallelisms(
```
mxm, line 446: (4) Special casing for the PR feature in new component

mxm (commented):
Thanks @morhidi! That's a very usual addition. I'm just a bit concerned about the complexity here. I think this could have been implemented a bit simpler. If we were to expose the recommended parallelism as an evaluated scaling metric, the changes would be a few lines of code. We could reuse all existing logic to expose metrics.

I would also appreciate if we could keep the unrelated changes / refactoring to a bare minimum. This makes reviewing much easier and keeps the commit history clean.

All in all, this will be super useful.

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/ScalingMetricCollector.java
morhidi: It was necessary in certain cases it fetched the new job graph and then viped it out with the metric history.

morhidi:
> Thanks @morhidi! That's a very usual addition. I'm just a bit concerned about the complexity here. I think this could have been implemented a bit simpler. If we were to expose the recommended parallelism as an evaluated scaling metric, the changes would be a few lines of code. We could reuse all existing logic to expose metrics.
> 
> I would also appreciate if we could keep the unrelated changes / refactoring to a bare minimum. This makes reviewing much easier and keeps the commit history clean.
> 
> All in all, this will be super useful.

Thanks @mxm for the comprehensice review. My original thought was to implement this as an evaluated metric, but the idea was abandoned and went on this rout instead.  I'll give that idea another shot, also trying not to be so disruptive with refactoring on a single PR :)

PR #617, opened 2023-06-09
[FLINK-32271] Report RECOMMENDED_PARALLELISM as an autoscaler metric V2

morhidi:

It is beneficial to report the recommended parallelism and overlay it with the current parallelism on the same chart when auto scaler is running in advisor mode.

## Brief change log
- Added a new `RECOMMENDED_PARALLELISM` entry to `ScalingMetric`
- And the latest recommended parallelisms are now being reported as evaluated scaling metric
- Recommended parallelisms follow the logic below:
  - On job start all scaling metrics are registered to null until the job is in RUNNING state
  - `RECOMMENDED_PARALLELISM` will be kept `null` while the metric window is filling up
  - `RECOMMENDED_PARALLELISM` will be populated only after the metric window is full
  -  `RECOMMENDED_PARALLELISM` will pick up the PARALLELISM value by default
  -  `RECOMMENDED_PARALLELISM` on each vertex can be updated according to the calculated scaling logic

## Verifying this change

- Added unit test for recommended parallelisms
- Manual by checking the actual metrics being reported

```
# first deploy scaling disabled
jobVertexID.cbc357ccb763df2852fee8c4fc7d55f2.RECOMMENDED_PARALLELISM.Current: null
jobVertexID.cbc357ccb763df2852fee8c4fc7d55f2.PARALLELISM.Current: null
jobVertexID.90bea66de1c231edf33913ecd54406c1.RECOMMENDED_PARALLELISM.Current: null
jobVertexID.90bea66de1c231edf33913ecd54406c1.PARALLELISM.Current: null

# metric window filling up
jobVertexID.cbc357ccb763df2852fee8c4fc7d55f2.RECOMMENDED_PARALLELISM.Current: null
jobVertexID.cbc357ccb763df2852fee8c4fc7d55f2.PARALLELISM.Current: 1.0
jobVertexID.90bea66de1c231edf33913ecd54406c1.RECOMMENDED_PARALLELISM.Current: null
jobVertexID.90bea66de1c231edf33913ecd54406c1.PARALLELISM.Current: 1.0

# metric window full
jobVertexID.cbc357ccb763df2852fee8c4fc7d55f2.RECOMMENDED_PARALLELISM.Current: 1.0
jobVertexID.cbc357ccb763df2852fee8c4fc7d55f2.PARALLELISM.Current: 1.0
jobVertexID.90bea66de1c231edf33913ecd54406c1.RECOMMENDED_PARALLELISM.Current: 2.0
jobVertexID.90bea66de1c231edf33913ecd54406c1.PARALLELISM.Current: 1.0

# scaling enabled
jobVertexID.cbc357ccb763df2852fee8c4fc7d55f2.RECOMMENDED_PARALLELISM.Current: null
jobVertexID.cbc357ccb763df2852fee8c4fc7d55f2.PARALLELISM.Current: null
jobVertexID.90bea66de1c231edf33913ecd54406c1.RECOMMENDED_PARALLELISM.Current: null
jobVertexID.90bea66de1c231edf33913ecd54406c1.PARALLELISM.Current: null

# metric window filling up
jobVertexID.cbc357ccb763df2852fee8c4fc7d55f2.RECOMMENDED_PARALLELISM.Current: null
jobVertexID.cbc357ccb763df2852fee8c4fc7d55f2.PARALLELISM.Current: 1.0
jobVertexID.90bea66de1c231edf33913ecd54406c1.RECOMMENDED_PARALLELISM.Current: null
jobVertexID.90bea66de1c231edf33913ecd54406c1.PARALLELISM.Current: 2.0

# metric window full
jobVertexID.cbc357ccb763df2852fee8c4fc7d55f2.RECOMMENDED_PARALLELISM.Current: 1.0
jobVertexID.cbc357ccb763df2852fee8c4fc7d55f2.PARALLELISM.Current: 1.0
jobVertexID.90bea66de1c231edf33913ecd54406c1.RECOMMENDED_PARALLELISM.Current: 2.0
jobVertexID.90bea66de1c231edf33913ecd54406c1.PARALLELISM.Current: 2.0
```

## Documentation

### DIFF
--- a/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/JobAutoScalerImpl.java
+++ b/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/JobAutoScalerImpl.java
@@ -17,6 +17,7 @@
 
 package org.apache.flink.kubernetes.operator.autoscaler;
 
+import org.apache.flink.annotation.VisibleForTesting;
 import org.apache.flink.api.common.JobStatus;
 import org.apache.flink.kubernetes.operator.api.AbstractFlinkResource;
 import org.apache.flink.kubernetes.operator.autoscaler.metrics.EvaluatedScalingMetric;
@@ -35,6 +36,8 @@ import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
 import static org.apache.flink.kubernetes.operator.autoscaler.config.AutoScalerOptions.AUTOSCALER_ENABLED;
+import static org.apache.flink.kubernetes.operator.autoscaler.metrics.ScalingMetric.PARALLELISM;
+import static org.apache.flink.kubernetes.operator.autoscaler.metrics.ScalingMetric.RECOMMENDED_PARALLELISM;
 
 /** Application and SessionJob autoscaler. */
 public class JobAutoScalerImpl implements JobAutoScaler {
@@ -47,8 +50,11 @@ public class JobAutoScalerImpl implements JobAutoScaler {
     private final ScalingExecutor scalingExecutor;
     private final EventRecorder eventRecorder;
 
-    private final Map<ResourceID, Map<JobVertexID, Map<ScalingMetric, EvaluatedScalingMetric>>>
+    @VisibleForTesting
+    final Map<ResourceID, Map<JobVertexID, Map<ScalingMetric, EvaluatedScalingMetric>>>
             lastEvaluatedMetrics = new ConcurrentHashMap<>();
+
+    @VisibleForTesting
     final Map<ResourceID, AutoscalerFlinkMetrics> flinkMetrics = new ConcurrentHashMap<>();
 
     public JobAutoScalerImpl(
@@ -79,6 +85,8 @@ public class JobAutoScalerImpl implements JobAutoScaler {
         var conf = ctx.getObserveConfig();
         var resource = ctx.getResource();
         var resourceId = ResourceID.fromResource(resource);
+        var flinkMetrics = getOrInitAutoscalerFlinkMetrics(ctx, resourceId);
+        Map<JobVertexID, Map<ScalingMetric, EvaluatedScalingMetric>> evaluatedMetrics = null;
 
         try {
 
@@ -88,10 +96,10 @@ public class JobAutoScalerImpl implements JobAutoScaler {
             }
 
             // Initialize metrics only if autoscaler is enabled
-            var flinkMetrics = getOrInitAutoscalerFlinkMetrics(ctx, resourceId);
 
             if (!resource.getStatus().getJobStatus().getState().equals(JobStatus.RUNNING.name())) {
                 LOG.info("Job autoscaler is waiting for RUNNING job state");
+                lastEvaluatedMetrics.remove(resourceId);
                 return false;
             }
 
@@ -107,13 +115,13 @@ public class JobAutoScalerImpl implements JobAutoScaler {
             }
 
             LOG.debug("Evaluating scaling metrics for {}", collectedMetrics);
-            var evaluatedMetrics = evaluator.evaluate(conf, collectedMetrics);
+            evaluatedMetrics = evaluator.evaluate(conf, collectedMetrics);
+            initRecommendedParallelism(evaluatedMetrics);
             LOG.debug("Scaling metrics evaluated: {}", evaluatedMetrics);
-            lastEvaluatedMetrics.put(resourceId, evaluatedMetrics);
-            flinkMetrics.registerScalingMetrics(() -> lastEvaluatedMetrics.get(resourceId));
 
             if (!collectedMetrics.isFullyCollected()) {
                 // We have done an upfront evaluation, but we are not ready for scaling.
+                resetRecommendedParallelism(evaluatedMetrics);
                 autoScalerInfo.replaceInKubernetes(kubernetesClient);
                 return false;
             }
@@ -129,7 +137,7 @@ public class JobAutoScalerImpl implements JobAutoScaler {
             return specAdjusted;
         } catch (Throwable e) {
             LOG.error("Error while scaling resource", e);
-            getOrInitAutoscalerFlinkMetrics(ctx, resourceId).numErrors.inc();
+            flinkMetrics.numErrors.inc();
             eventRecorder.triggerEvent(
                     resource,
                     EventRecorder.Type.Warning,
@@ -137,6 +145,12 @@ public class JobAutoScalerImpl implements JobAutoScaler {
                     EventRecorder.Component.Operator,
                     e.getMessage());
             return false;
+        } finally {
+            if (evaluatedMetrics != null) {
+                LOG.debug("Storing evaluated metrics {}", evaluatedMetrics);
+                lastEvaluatedMetrics.put(resourceId, evaluatedMetrics);
+                flinkMetrics.registerScalingMetrics(() -> lastEvaluatedMetrics.get(resourceId));
+            }
         }
     }
 
@@ -147,5 +161,21 @@ public class JobAutoScalerImpl implements JobAutoScaler {
                 id ->
                         new AutoscalerFlinkMetrics(
                                 ctx.getResourceMetricGroup().addGroup("AutoScaler")));
+    }
+
+    private void initRecommendedParallelism(
+            Map<JobVertexID, Map<ScalingMetric, EvaluatedScalingMetric>> evaluatedMetrics) {
+        evaluatedMetrics.forEach(
+                (jobVertexID, evaluatedScalingMetricMap) ->
+                        evaluatedScalingMetricMap.put(
+                                RECOMMENDED_PARALLELISM,
+                                evaluatedScalingMetricMap.get(PARALLELISM)));
+    }
+
+    private void resetRecommendedParallelism(
+            Map<JobVertexID, Map<ScalingMetric, EvaluatedScalingMetric>> evaluatedMetrics) {
+        evaluatedMetrics.forEach(
+                (jobVertexID, evaluatedScalingMetricMap) ->
+                        evaluatedScalingMetricMap.put(RECOMMENDED_PARALLELISM, null));
     }
 }
--- a/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/ScalingMetricCollector.java
+++ b/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/ScalingMetricCollector.java
@@ -83,7 +83,6 @@ public abstract class ScalingMetricCollector {
             Configuration conf)
             throws Exception {
 
-        var topology = getJobTopology(flinkService, cr, conf, autoscalerInfo);
         var resourceID = ResourceID.fromResource(cr);
         var now = clock.instant();
 
@@ -101,6 +100,7 @@ public abstract class ScalingMetricCollector {
             metricHistory.clear();
             metricCollectionStartTs = now;
         }
+        var topology = getJobTopology(flinkService, cr, conf, autoscalerInfo);
 
         // Trim metrics outside the metric window from metrics history
         var metricWindowSize = getMetricWindowSize(conf);
--- a/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/ScalingMetricEvaluator.java
+++ b/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/ScalingMetricEvaluator.java
@@ -139,10 +139,10 @@ public class ScalingMetricEvaluator {
 
         evaluatedMetrics.put(
                 PARALLELISM, EvaluatedScalingMetric.of(topology.getParallelisms().get(vertex)));
+
         evaluatedMetrics.put(
                 MAX_PARALLELISM,
                 EvaluatedScalingMetric.of(topology.getMaxParallelisms().get(vertex)));
-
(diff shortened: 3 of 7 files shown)


morhidi:
@mxm @gyfora this is my second attempt on adding the `RECOMMENDED_PARALLELISM` let me know what you think

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/JobAutoScalerImpl.java
```diff
@@ -102,10 +108,8 @@ public boolean scale(FlinkResourceContext<? extends AbstractFlinkResource<?, ?>>
...
 
             LOG.debug("Evaluating scaling metrics for {}", collectedMetrics);
-            var evaluatedMetrics = evaluator.evaluate(conf, collectedMetrics);
+            evaluatedMetrics = evaluator.evaluate(conf, collectedMetrics);
             LOG.debug("Scaling metrics evaluated: {}", evaluatedMetrics);
-            lastEvaluatedMetrics.put(resourceId, evaluatedMetrics);
-            flinkMetrics.registerScalingMetrics(() -> lastEvaluatedMetrics.get(resourceId));
 
             if (!collectedMetrics.isFullyCollected()) {
                 // We have done an upfront evaluation, but we are not ready for scaling.
```
gyfora, line 115: I think we should somehow null out the RECOMMENDED_PARALLELISM (or set it to Double.nan) if the metrics are not fully collected because at that point we are not actually recommending anything. It would be then consistent of not reporting the values when the job is not running / still stabilizing.

flink-kubernetes-operator-autoscaler/src/test/java/org/apache/flink/kubernetes/operator/autoscaler/RecommendedParallelismTest.java
```diff
@@ -0,0 +1,296 @@
...
+import java.util.stream.Collectors;
+
+import static org.apache.flink.kubernetes.operator.autoscaler.metrics.ScalingMetric.PARALLELISM;
+import static org.apache.flink.kubernetes.operator.autoscaler.metrics.ScalingMetric.RECOMMENDED_PARALLELISM;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+
+/** Test for recommended parallelism. */
+@EnableKubernetesMockClient(crud = true)
+public class RecommendedParallelismTest extends OperatorTestBase {
```
gyfora, line 62: I generally like the idea of having a separate test but this basically duplicates one of our worst / hardest-to-maintain tests that we wanted to rethink over time.

Would it be possible to just add the metric assertions to the existing tests? I know it is not ideal but maybe that will cause less pain over time.

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/JobAutoScalerImpl.java
morhidi: We can do that, and I feel like null is a more safe choice the Double.nan. @mxm what do you think? Shall we keep or wipe the metrics during the metric window? Since the recommended parallelism is set to the current parallelism to me both are ok.

gyfora: My point was that you cannot simply `null` them out here because that may interfere with how the metrics are initialized.

flink-kubernetes-operator-autoscaler/src/test/java/org/apache/flink/kubernetes/operator/autoscaler/RecommendedParallelismTest.java
gyfora: The problem with these tests are that they tend to break when we touch the scaling logic / default configs. So now instead of one test breaking and hard 2 fix, we have 2 tests that can break and are hard to fix

mxm: I agree with both of you that (a) having a dedicated test is preferable (b) duplicating most of the logic of an already hard to maintain test should be avoided. I think a compromise would be to add assertions to the existing end to end test and add a dedicated unit/integration test which tests only the core assumptions without going through the entire lifecycle.

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/ScalingMetricEvaluator.java
```diff
@@ -137,12 +138,13 @@ private Map<ScalingMetric, EvaluatedScalingMetric> evaluateMetrics(
                         latestVertexMetrics.get(TRUE_PROCESSING_RATE),
                         getAverage(TRUE_PROCESSING_RATE, vertex, metricsHistory)));
 
-        evaluatedMetrics.put(
-                PARALLELISM, EvaluatedScalingMetric.of(topology.getParallelisms().get(vertex)));
+        var parallelism = EvaluatedScalingMetric.of(topology.getParallelisms().get(vertex));
+        evaluatedMetrics.put(PARALLELISM, parallelism);
+        evaluatedMetrics.put(RECOMMENDED_PARALLELISM, parallelism);
```
mxm, line 143: I feel like we can't add it here yet because we don't know yet what the recommended parallelism is.

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/ScalingMetricCollector.java
```diff
@@ -101,6 +100,7 @@ public CollectedMetricHistory updateMetrics(
             metricHistory.clear();
             metricCollectionStartTs = now;
         }
+        var topology = getJobTopology(flinkService, cr, conf, autoscalerInfo);
```
mxm, line 103: Any reason for moving this?

flink-kubernetes-operator-autoscaler/src/test/java/org/apache/flink/kubernetes/operator/autoscaler/RecommendedParallelismTest.java
mxm: I agree with both of you that (a) having a dedicated test is preferable (b) duplicating most of the logic of an already hard to maintain test should be avoided. I think a compromise would be to add assertions to the existing end to end test and add a dedicated unit/integration test which tests only the core assumptions without going through the entire lifecycle. 

Adding a test method to ScalingExecutor should be sufficient IMHO to test that the recommended parallelism is added. The cleanup can be tested in the end to end test. If we want to go all the way, we could test the cleanup in JobAutoscalerImpl without the entire end to end setup, e.g. look at JobAutoscalerImplTest#testErrorReporting which only tests the error reporting. Since you are cleaning up in the finally block, the test setup should work similarly.

mxm (commented):
Thanks for the v2 @morhidi :)

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/ScalingMetricEvaluator.java
gyfora: We could simply set this in the scaling executor only (but initialize to null for the metric registration)

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/ScalingMetricCollector.java
morhidi: We query than wipe out the topology along with the metrics according to the original logic.

flink-kubernetes-operator-autoscaler/src/test/java/org/apache/flink/kubernetes/operator/autoscaler/RecommendedParallelismTest.java
morhidi: I discussed this with Gyula and we agreed the e2e for the recommended parallelism can stay and will be refactored together with the backlog e2e.

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/ScalingMetricEvaluator.java
morhidi: Updated the requested changes

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/JobAutoScalerImpl.java
```diff
@@ -148,4 +162,22 @@ private AutoscalerFlinkMetrics getOrInitAutoscalerFlinkMetrics(
...
     }
+
+    private void initRecommendedParallelism(
+            Map<JobVertexID, Map<ScalingMetric, EvaluatedScalingMetric>> evaluatedMetrics) {
+        evaluatedMetrics.forEach(
+                (jobVertexID, evaluatedScalingMetricMap) ->
+                        evaluatedScalingMetricMap.put(
+                                RECOMMENDED_PARALLELISM,
+                                evaluatedScalingMetricMap.get(PARALLELISM)));
+        LOG.debug("Recommended parallelisms where initialized  {}", evaluatedMetrics);
```
gyfora, line 173: I think the initialized / reset debug logs are an overkill. They don't provide any valuable info and are called in every iteration

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/JobAutoScalerImpl.java
```diff
@@ -79,6 +85,8 @@ public boolean scale(FlinkResourceContext<? extends AbstractFlinkResource<?, ?>>
         var conf = ctx.getObserveConfig();
         var resource = ctx.getResource();
         var resourceId = ResourceID.fromResource(resource);
+        var flinkMetrics = getOrInitAutoscalerFlinkMetrics(ctx, resourceId);
```
gyfora, line 88: we could now use the `flinkMetrics` in the catch block instead of reinitializing it

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/JobAutoScalerImpl.java
```diff
@@ -137,6 +145,12 @@ public boolean scale(FlinkResourceContext<? extends AbstractFlinkResource<?, ?>>
                     EventRecorder.Component.Operator,
                     e.getMessage());
             return false;
+        } finally {
+            if (evaluatedMetrics != null) {
+                LOG.debug("Storing evaluated metrics {}", evaluatedMetrics);
```
gyfora, line 150: I think either this or the logging of the evaluated metrics at line 120 should be removed, otherwise this is going to be crazy verbose

gyfora (approved):
The changes look good, added a few nits about logging, minor cleanups.

flink-kubernetes-operator-autoscaler/src/test/java/org/apache/flink/kubernetes/operator/autoscaler/RecommendedParallelismTest.java
mxm: Why did we not consider any of the proposed alternatives? What is the point of a review if no discussion takes place?

gyfora: @mxm I discussed this with Matyas and he prefered to keep this functionality independently and comprehensively tested instead of adding bits and pieces to other tests. This functionality involves the full lifecycle of the scaling logic and the test makes sure that the recommendation is conistsent along the way.

As you (@mxm) and I discussed we should come up with a way to have better and more robust end to end tests but this is clearly out of scope here. I support the decision made here, as I feel the feedback was considered fairly :)

mxm: There is no point in taking the time to review and make suggestions if those are dismissed after internal discussions. I would appreciate if we had those discussions here. 

I think we need to consider what the right amount of test coverage would be. There isn't a single unit or integration test here. We cannot add an e2e test for every feature. IMHO the impact of this feature does not justify copying questionable end to end testing code. This could have been tested in a more light way with comparable test coverage.

gyfora: @mxm I think thats fair point. The review wasn't ignored in this case but you are completely right that no appropriate feedback was provided before going ahead with this change.

@morhidi I think the best course of action would be to work on simplifying the test and adopt a more unit test style as Max suggested. @morhidi could you please work on that as a separate PR?

morhidi: Thanks @mxm for your comprehensive review and feedbacks. I agree we can/should be more careful with tracking internal decisions upstream. I'll follow up on that @gyfora thanks.

PR #618, opened 2023-06-09
 [hotfix][docs] Fix ArgoCD CRDs application

kenankule:
`syncOptions` field should reside under `.spec.syncPolicy`

## Documentation

  no


kenankule:
Please refer to the following example from the ArgoCD docs: 
https://argo-cd.readthedocs.io/en/stable/user-guide/sync-options/#replace-resource-instead-of-applying-changes

haoxins:
I submitted another PR https://github.com/apache/flink-kubernetes-operator/pull/619 to simplify the Argo CD parts.
I added the solution because we will face the issue in the old Flink operator version.
We won't have this problem after separating the CRDs.

So that the docs will confuse the new guys to use Flink operator.

PR #621, opened 2023-06-14
[FLINK-32334] Also check if no taskmanager are running while waiting for cluster shutdown

ashangit:

This pull request add a check to ensure that no TaskManager pods are running while waiting for cluster shutdown.

With standalone mode the TaskManager pods are managed through a deployment managed by the operator.

If we do not check that no TaskManager pods are running we can face issue when redeploying the standalone app (during a config change for ex.) if the TaskManager deployment has not really being deleted.

## Brief change log

  - Ensure TM pods are not running while waiting for cluster shutdown

## Verifying this change

  - https://github.com/apache/flink-kubernetes-operator/blob/main/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/service/StandaloneFlinkServiceTest.java#L93

  - Core observer or reconciler logic that is regularly executed: no, only affect the check during cluster shutdown

## Documentation

### DIFF
--- a/flink-kubernetes-operator/pom.xml
+++ b/flink-kubernetes-operator/pom.xml
@@ -198,6 +198,13 @@ under the License.
             <artifactId>junit-jupiter-params</artifactId>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <version>${mockito.version}</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
--- a/flink-kubernetes-operator/pom.xml
+++ b/flink-kubernetes-operator/pom.xml
@@ -198,13 +198,6 @@ under the License.
             <artifactId>junit-jupiter-params</artifactId>
             <scope>test</scope>
         </dependency>
-
-        <dependency>
-            <groupId>org.mockito</groupId>
-            <artifactId>mockito-core</artifactId>
-            <version>${mockito.version}</version>
-            <scope>test</scope>
-        </dependency>
     </dependencies>
 
     <build>
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/NativeFlinkService.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/NativeFlinkService.java
@@ -93,6 +93,12 @@ public class NativeFlinkService extends AbstractFlinkService {
                 .list();
     }
 
+    @Override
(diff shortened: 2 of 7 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/NativeFlinkService.java
```diff
@@ -93,6 +93,12 @@ protected PodList getJmPodList(String namespace, String clusterId) {
                 .list();
     }
 
+    @Override
+    protected PodList getTmPodList(String namespace, String clusterId) {
+        // Native mode does not manage TaskManager
+        return null;
```
gyfora, line 99: Maybe return empty PodList to save the null check in the service?

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/service/StandaloneFlinkServiceTest.java
```diff
@@ -99,9 +99,12 @@ public void testDeleteClusterDeploymentWithHADelete() throws Exception {
         List<Deployment> deployments = kubernetesClient.apps().deployments().list().getItems();
         assertEquals(2, deployments.size());
 
+        var requestsBeforeDelete = mockServer.getRequestCount();
         flinkStandaloneService.deleteClusterDeployment(
                 flinkDeployment.getMetadata(), flinkDeployment.getStatus(), configuration, true);
 
+        assertEquals(6, mockServer.getRequestCount() - requestsBeforeDelete);
```
gyfora, line 106: I don't really feel that this adequately tests the functionality. We should make sure that the deletion is blocked until the service returns empty list right?

gyfora (changes requested):

flink-kubernetes-operator/pom.xml
```diff
@@ -198,6 +198,13 @@ under the License.
             <artifactId>junit-jupiter-params</artifactId>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <version>${mockito.version}</version>
+            <scope>test</scope>
+        </dependency>
```
gyfora, line 207: Please avoid using Mockito as per the Flink coding guidelines: https://flink.apache.org/how-to-contribute/code-style-and-quality-common/#avoid-mockito---use-reusable-test-implementations

I suggest overriding the getPodList method of the standalone flink service by hand or using the TestingFlinkService.

gyfora (changes requested):

gyfora:
The PR looks good but you have some check style validations before I can merge this

PR #622, opened 2023-06-19
[FLINK-32012] Provide rollback feature on savepoint upgrade mode if HA missing and JM never started

ashangit:

This pull request will provide rollback feature If the JobManagers have never started and we rely on savepoint upgrade mode. The operator will safely rollback relying on the provided savepoint

Also benefit from this change to unify rollback and upgrade code path

## Brief change log

- Provide rollback feature on savepoint upgrade mode if HA missing and JM never started
- Unify rollback and upgrade code path

## Verifying this change

 - flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/RollbackTest.java: Test different rollback cases

  - Manually verified the change by running a 4 node cluster with 3 JobManagers and 1 TaskManagers in standalone deployment mode:
    - First deployed a working app
    - Redeploy the app changing the parallelism to get 1 more TM
    - Then deploy the app by setting a not existing service account leading the JM to never start and requiring the job to be rolled back
    - Once rollbacked redeploy app decrease of parallelism and appropriate service account leading to app being redeployed

  - The public API, i.e., is any changes to the `CustomResourceDescriptors`: yes but in status only adding lastRollbackSpec

## Documentation

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
@@ -21,6 +21,7 @@ import org.apache.flink.api.common.JobStatus;
 import org.apache.flink.kubernetes.operator.api.FlinkDeployment;
 import org.apache.flink.kubernetes.operator.api.status.FlinkDeploymentStatus;
 import org.apache.flink.kubernetes.operator.api.status.JobManagerDeploymentStatus;
+import org.apache.flink.kubernetes.operator.api.status.ReconciliationState;
 import org.apache.flink.kubernetes.operator.config.FlinkConfigManager;
 import org.apache.flink.kubernetes.operator.exception.DeploymentFailedException;
 import org.apache.flink.kubernetes.operator.exception.ReconciliationException;
@@ -139,6 +140,12 @@ public class FlinkDeploymentController
                         flinkApp,
                         previousDeployment,
                         false);
+            }
+            // Rely on the last stable spec if under a roll back
+            if (flinkApp.getStatus().getReconciliationStatus().getState()
+                    == ReconciliationState.ROLLING_BACK) {
+                flinkApp.setSpec(
+                        flinkApp.getStatus().getReconciliationStatus().deserializeLastStableSpec());
             }
             statusRecorder.patchAndCacheStatus(flinkApp);
             reconcilerFactory.getOrCreate(flinkApp).reconcile(ctx);
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
@@ -141,7 +141,7 @@ public class FlinkDeploymentController
                         previousDeployment,
                         false);
             }
-            // Rely on the last stable spec if under a roll back
+            // Rely on the last stable spec if rolling back
             if (flinkApp.getStatus().getReconciliationStatus().getState()
                     == ReconciliationState.ROLLING_BACK) {
                 flinkApp.setSpec(
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
@@ -21,7 +21,6 @@ import org.apache.flink.api.common.JobStatus;
 import org.apache.flink.kubernetes.operator.api.FlinkDeployment;
 import org.apache.flink.kubernetes.operator.api.status.FlinkDeploymentStatus;
 import org.apache.flink.kubernetes.operator.api.status.JobManagerDeploymentStatus;
-import org.apache.flink.kubernetes.operator.api.status.ReconciliationState;
 import org.apache.flink.kubernetes.operator.config.FlinkConfigManager;
 import org.apache.flink.kubernetes.operator.exception.DeploymentFailedException;
 import org.apache.flink.kubernetes.operator.exception.ReconciliationException;
@@ -140,12 +139,6 @@ public class FlinkDeploymentController
                         flinkApp,
                         previousDeployment,
                         false);
-            }
-            // Rely on the last stable spec if rolling back
-            if (flinkApp.getStatus().getReconciliationStatus().getState()
-                    == ReconciliationState.ROLLING_BACK) {
-                flinkApp.setSpec(
-                        flinkApp.getStatus().getReconciliationStatus().deserializeLastStableSpec());
             }
             statusRecorder.patchAndCacheStatus(flinkApp);
             reconcilerFactory.getOrCreate(flinkApp).reconcile(ctx);
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
@@ -123,16 +123,20 @@ public class ReconciliationUtils {
         // Clear errors
         status.setError(null);
         reconciliationStatus.setReconciliationTimestamp(clock.instant().toEpochMilli());
-        var state = upgrading ? ReconciliationState.UPGRADING : ReconciliationState.DEPLOYED;
+        ReconciliationState state;
         if (status.getReconciliationStatus().getState() == ReconciliationState.ROLLING_BACK) {
             state = upgrading ? ReconciliationState.ROLLING_BACK : ReconciliationState.ROLLED_BACK;
+        } else {
+            state = upgrading ? ReconciliationState.UPGRADING : ReconciliationState.DEPLOYED;
         }
         reconciliationStatus.setState(state);
 
-        var clonedSpec = ReconciliationUtils.clone(spec);
+        SPEC clonedSpec;
         if (status.getReconciliationStatus().getState() == ReconciliationState.ROLLING_BACK
                 || status.getReconciliationStatus().getState() == ReconciliationState.ROLLED_BACK) {
             clonedSpec = reconciliationStatus.deserializeLastReconciledSpec();
+        } else {
+            clonedSpec = ReconciliationUtils.clone(spec);
         }
         if (spec.getJob() != null) {
             // For jobs we have to adjust the reconciled spec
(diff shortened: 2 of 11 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
```diff
@@ -281,6 +284,17 @@ public static boolean isUpgradeModeChangedToLastStateAndHADisabledPreviously(
...
     }
 
+    public static <SPEC extends AbstractFlinkSpec> SPEC getLastSpec(
+            AbstractFlinkResource<SPEC, ?> deployment) {
+        var reconciliationStatus = deployment.getStatus().getReconciliationStatus();
+        var reconciliationState = reconciliationStatus.getState();
+        if (reconciliationState != ReconciliationState.ROLLED_BACK) {
+            return reconciliationStatus.deserializeLastReconciledSpec();
+        } else {
+            return reconciliationStatus.deserializeLastRollbackSpec();
```
gyfora, line 294: We should use `lastStableSpec` and avoid introducing a new status field. The lastStableSpec is meant exactly for rollback purposes.

ashangit: I rely on the `lastStableSpec` to do the rollback.
It is indeed the spec used to roll back to a working state.
But once the job is rolled back I need to be able to detect changes to the current spec (the one which has lead to the rollback). As the job spec has not been updated we need to be able to compare it with a specific stored spec from which we will be able to do comparison and find changes
I can't rely on the `lastReconcileSpec` or `lastStableSpec` as they are not align with the job spec.
This is why I introduce this `lastRollbackSpec` which contains the job spec leading to failure.

gyfora: You should always compare to the `lastReconciledSpec` I think that was the last spec that was reconciled and submitted by the user.

gyfora: As long as you do not update the `lastReconciledSpec` during the rollback the logic should be good

ashangit: The lastReconcileSpec is updated now as we rely on same code path than the upgrade one.
I will take a look to not update it but it looks quite strange to me to finally not have the lastReconcile being aligned with the "current" running spec

gyfora: `lastReconciled` spec is not `current` spec. Current spec is already in the spec, lastReconciled is what was last reconciled by the operator based on what the user provided. Its sole purpose is to be able to do the above mentioned comparison. We should try not to change the semantics if possible.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
```diff
@@ -140,6 +141,12 @@ public UpdateControl<FlinkDeployment> reconcile(FlinkDeployment flinkApp, Contex
                         previousDeployment,
                         false);
             }
+            // Rely on the last stable spec if rolling back
+            if (flinkApp.getStatus().getReconciliationStatus().getState()
+                    == ReconciliationState.ROLLING_BACK) {
+                flinkApp.setSpec(
+                        flinkApp.getStatus().getReconciliationStatus().deserializeLastStableSpec());
```
gyfora, line 148: We might want to do a spec diff check there with the current spec vs the last reconciled spec. It could happen that we initiated the rollback (setting the state to ROLLING_BACK) and in the meantime the user actually submitted a new spec.

With the current logic the new spec would be first ignored and we perform a rollback and after that we reconcile the new spec. This is not entirely incorrect but it will lead to extra time and extra redeployments.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
```diff
@@ -125,12 +126,19 @@ private static <SPEC extends AbstractFlinkSpec> void updateStatusForSpecReconcil
...
+        if (status.getReconciliationStatus().getState() == ReconciliationState.ROLLING_BACK) {
+            state = upgrading ? ReconciliationState.ROLLING_BACK : ReconciliationState.ROLLED_BACK;
+        }
+        reconciliationStatus.setState(state);
 
+        var clonedSpec = ReconciliationUtils.clone(spec);
+        if (status.getReconciliationStatus().getState() == ReconciliationState.ROLLING_BACK
+                || status.getReconciliationStatus().getState() == ReconciliationState.ROLLED_BACK) {
+            clonedSpec = reconciliationStatus.deserializeLastReconciledSpec();
+        }
```
gyfora, line 139: Both of these if branches have some redundant code, setting the state/clonedSpec variables twice instead of once in case of redeployments.

gyfora (changes requested):
I think this turned out to be a pretty nice and simple implementation piggybacking on the existing logic. However we should add some new test cases that cover the scenarios we wished to enable here in the first place.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
ashangit: I've updated the PR to take in account this case.

I also set the state to Upgrading if the there is a spec change as we don't want to apply specific code path related to roll back mechanism: https://github.com/apache/flink-kubernetes-operator/pull/622/files#diff-29ea38a50cac5b4432dd0969bc3e2177e29a5507f8c7bb01b80f605a8740de41R146

I had here to somehow exclude a specific spec diff when it is due to job.state as we can override this job.state in https://github.com/apache/flink-kubernetes-operator/blob/main/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java#L358

gyfora:
Can you please rebase it to the latest changes on master so we can run the e2es?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
ashangit: My change is bad for now there are still some cases with job.state diff and some usually ignore changed (for ex on `kubernetes.operator.deployment.rollback.enabled`) that are seen as an upgrade change now and lead to an upgrade instead of roll abck

gyfora: I will take a look, it could be related to the latest changes we have made to the master regarding the rescaling logic.

ashangit: For this issue it is clearly on my side.
I should have found a good way to detect the case I'm currently rebasing from last master (few conflict to resolve) and will push it

ashangit: Rebased + updated with [working solution](https://github.com/apache/flink-kubernetes-operator/pull/622/files#diff-29ea38a50cac5b4432dd0969bc3e2177e29a5507f8c7bb01b80f605a8740de41R157)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractJobReconciler.java
```diff
@@ -113,9 +113,22 @@ protected boolean reconcileSpecChange(FlinkResourceContext<CR> ctx, Configuratio
...
+            currentDeploySpec.getJob().setUpgradeMode(upgradeMode);
+
+            // We must use LAST_STATE mode when rolling back from
+            // SAVEPOINT. But we don't want to set upgrade mode to LAST_STATE
+            // as we will rely on SAVEPOINT restoreJob mechanism
+            if (upgradeMode == UpgradeMode.SAVEPOINT
+                    && status.getReconciliationStatus().getState()
+                            == ReconciliationState.ROLLING_BACK) {
+                upgradeMode = UpgradeMode.LAST_STATE;
+            }
```
gyfora, line 129: Can you please explain what's going on here? I don't really understand why we change from savepoint -> last-state for the cancel operation specifically.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
```diff
@@ -138,14 +138,32 @@ public void reconcile(FlinkResourceContext<CR> ctx) throws Exception {
...
+            // current spec is not valid
+            lastReconciledSpec.getJob().setState(currentDeploySpec.getJob().getState());
+            var specDiffRollingBack =
+                    new ReflectiveDiffBuilder<>(
+                                    ctx.getDeploymentMode(), lastReconciledSpec, currentDeploySpec)
+                            .build();
+            if (DiffType.IGNORE != specDiffRollingBack.getType()) {
+                reconciliationStatus.setState(ReconciliationState.UPGRADING);
+            }
+        }
```
gyfora, line 159: It would be great to encapsulate this into a method that mutates the CR/status fields and returns the value of specChanged variable that we should set afterwards. 

It could be something like:

```
if (reconciliationStatus.getState() == ReconciliationState.ROLLING_BACK) {
  specChanged = prepareCrForRollback(...)
}
```

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
gyfora: The getAvailableUpgrade mode method would only return save point if the job is either already shut down / terminally failed or is running and can be upgraded with the save point. So if it returns SAVEPOINT it should be safe to call cancel with savepoint.

ashangit:
There was an issue with new set of commits due to `deployConfig` being created too early and missing the potential reset of spec in case of `rolling_back`
Unittest updated by checking that the submitted config has well roll back a specific conf

gyfora:
Thank you @ashangit for this great contribution :) good stuff

PR #623, opened 2023-06-22
[FLINK-32412] Reduce JobID collision chance

FabioWanner:
Instead of using Java's hashCode, which is an integer value (32bit), a long representation of the uid is used. This decreases the chance for an ID collision drastically:

For a 50% collision chance with random integers, 77000 numbers need to be generated. For a long value (64 bit) a 50% change of a collision needs 5.1×10^9 random longs. For details look up: "the birthday problem".

A test is added to increase awareness of the problem when changing this part of the code.

Lower the chance of JobID collisions in session job deployments.

## Brief change log

  - *Increase the lower part of the JobID to use the 64 most significant bits of the ressources uid instead of the 32 bit hashCode of the uid*

## Verifying this change

  - *Added unit test*
  - *Adapted existing tests*
  - *Manually tested the upgrade of the operator when having a job deployed with the old JobID generation and then switching to a operator with the fix. Upgrading the job after this worked without any problems.*

## Documentation

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/sessionjob/FlinkSessionJobObserver.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/sessionjob/FlinkSessionJobObserver.java
@@ -43,6 +43,8 @@ import java.util.List;
 import java.util.Optional;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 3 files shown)


gyfora:
> About the failing e2e test: I can not figure out the root cause for the failure... I ran the e2e tests locally and did not have any problems... I would be glad to get some pointers on what's actually going wrong. Thanks a lot!

I think that's just a flaky test unfortunately. I re-triggered it!

PR #641, opened 2023-08-03
[FLINK-32729] allow starting a flink job in a suspended state.

bjhaid:

In our setup we intend migrating flinkDeployments from one cluster to the other, as part of doing this we want to start the flinkDeploments in a suspended state with an initialSavePoint configured, and only configure it to start after we must have stopped flinkDeployment in the source cluster. This patch was agreed upon in this [slack thread](https://apache-flink.slack.com/archives/C03GV7L3G2C/p1690825041612599).

## Brief change log

allow starting a flink job in a suspended state. 

## Verifying this change

## Documentation


gyfora:
Hi @bjhaid ,
Your code doesn’t actually implement the logic necessary for this only removes the validation.

we have an open PR https://github.com/apache/flink-kubernetes-operator/pull/642 that already contains the full solution with the required test coverage . I suggest we go with that if you don’t mind

PR #644, opened 2023-08-07
[FLINK-32772] Add support of SET statements in the SQL Runner example

mehdid93:

This pull request adds a new feature to take in account `SET` statements in the SQL Runner example.

## Brief change log

- Adding support of SET statements to set parameters to the context for eg. parallelism

## Verifying this change

This change can be verified as follows:
- Add `SET 'parallelism.default' = '2';` on top of the SQL statements in https://github.com/apache/flink-kubernetes-operator/blob/main/examples/flink-sql-runner-example/sql-scripts/statement-set.sql
- Build and execute the example, and verify that, with the changes in the PR, the error `The main method caused an error: Unsupported SQL query! executeSql() only accepts a single SQL statement of type [...]` does not happen anymore and the Flink job is executed with parallelism at 2.

  - Dependencies (does it add or upgrade a dependency): (yes / no) **no**
  - The public API, i.e., is any changes to the `CustomResourceDescriptors`: (yes / no) **no**
  - Core observer or reconciler logic that is regularly executed: (yes / no) **no**

## Documentation

  - Does this pull request introduce a new feature? (yes / no) **yes**
  - If yes, how is the feature documented? (not applicable / docs / JavaDocs / not documented) **not applicable but it may be worth mentioning it in the release notes of the future operator version**


examples/flink-sql-runner-example/src/main/java/org/apache/flink/examples/SqlRunner.java
```diff
@@ -93,4 +107,4 @@ public static String formatSqlFile(String content) {
         formatted.append(LINE_DELIMITER);
         return formatted.toString();
     }
-}
+}
```
gyfora, line 110: missing new-line at the end of file

mehdid93: Right! Fixed

PR #650, opened 2023-08-11
[backport][FLINK-32772] Add support of SET statements in the SQL Runner example

mehdid93:

## Brief change log

*(for example:)*
  - *Periodic savepoint trigger is introduced to the custom resource*
  - *The operator checks on reconciliation whether the required time has passed*
  - *The JobManager's dispose savepoint API is used to clean up obsolete savepoints*

## Verifying this change

*(or)*

*(or)*

*(example:)*
  - *Added integration tests for end-to-end deployment with large payloads (100MB)*

## Documentation

### DIFF
--- a/examples/flink-sql-runner-example/src/main/java/org/apache/flink/examples/SqlRunner.java
+++ b/examples/flink-sql-runner-example/src/main/java/org/apache/flink/examples/SqlRunner.java
@@ -27,6 +27,8 @@ import org.slf4j.LoggerFactory;
 import java.io.File;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 /** Main class for executing SQL scripts. */
 public class SqlRunner {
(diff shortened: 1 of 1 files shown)


gyfora:
We don't need to backport this, examples would work across versions. Users can simply use from main

mehdid93:
Good to know thank you Gyula.

AdrianVasiliu:
@gyfora 
> We don't need to backport this, examples would work across versions. Users can simply use from main

I do think this would be fine in the vast majority of cases, for the time being.
However:
* The sample in master builds with flink 1.17-snapshot (https://github.com/apache/flink-kubernetes-operator/blob/main/examples/flink-sql-runner-example/pom.xml#L26). 
* Once master will evolve to go with flink 1.18-snapshot, it will be built against a new major version than the sample from operator's `release-1.5` branch.
* At some point in the future, the Flink APIs may evolve (not necessarily in an incompatible manner), which may lead to changes in operator's sample in the master branch, that users of operator's 1.5 version would consume even if they use the older Flink release.

Maybe none of these would be harmful in practice, but in my eyes it would be cleaner / safer if the sample changes would be backported to 1.6. Not necessarily merged now (as your at 1.6 rc2), it could be merged after the release of 1.6.0.

PR #654, opened 2023-08-18
[FLINK-32890] Correct HA patch check for zookeeper metadata store

ashangit:

* This pull request correct the path used to check if HA metadata are available in zookeeper. Previous path check was leading to operator taking action based on availability of those metadata while the jobs one were not available see https://issues.apache.org/jira/browse/FLINK-32890 for some case leading to bad action

## Brief change log

  - Correct HA patch check for zookeeper metadata store

## Verifying this change

This change is already covered by existing tests, such as `FlinkUtilsZookeeperHATest.zookeeperHaMetaDataCheckTest`.

## Documentation


gyfora:
we could also back port this to the `release-1.6` branch, in case we have a bug fix release it would be nice to have it included

This fix alone probably doesn't warrant the bug fix release but you never know what may come up next :)

ashangit:
I've created this [PR](https://github.com/apache/flink-kubernetes-operator/pull/658) to back port it on `release-1.6` branch

PR #655, opened 2023-08-21
[FLINK-32903] Add a load simulation example pipeline for canary testing

mxm:
The current provided example pipeline generates an infinite amount of load, only yielding only scaleup decisions. It would be desirable to have a pipeline which scales up and down in a periodic fashion. Users do not always have enough data available to do that. Hence, the load needs to be simulated.

The pipeline has been tested and proven to yield very accurate load numbers.

### DIFF
--- a/examples/autoscaling/autoscaling-canary.yaml
+++ b/examples/autoscaling/autoscaling-canary.yaml
@@ -1,0 +1,56 @@
+################################################################################
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+# limitations under the License.
+################################################################################
+
+apiVersion: flink.apache.org/v1beta1
+kind: FlinkDeployment
+metadata:
+  name: autoscaling-canary
+spec:
+  image: autoscaling-example
+  flinkVersion: v1_18
+  flinkConfiguration:
+    kubernetes.operator.job.autoscaler.enabled: "true"
+    kubernetes.operator.job.autoscaler.stabilization.interval: "1m"
+    kubernetes.operator.job.autoscaler.metrics.window: "15m"
+    kubernetes.operator.job.autoscaler.target.utilization: "0.5"
+    kubernetes.operator.job.autoscaler.target.utilization.boundary: "0.3"
+    pipeline.max-parallelism: "32"
+    taskmanager.numberOfTaskSlots: "4"
+  serviceAccount: flink
+  jobManager:
+    resource:
+      memory: "2048m"
+      cpu: 0.5
+  taskManager:
+    resource:
+      memory: "2048m"
+      cpu: 1
+  podTemplate:
+    spec:
+      containers:
+        - name: flink-main-container
+  job:
+    jarURI: local:///opt/flink/usrlib/autoscaling.jar
+    entryClass: autoscaling.LoadSimulationPipeline
+    parallelism: 1
+    upgradeMode: stateless
+    args:
+    - --maxLoadPerTask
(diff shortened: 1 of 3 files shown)


examples/autoscaling/src/main/java/autoscaling/LoadSimulationPipeline.java
```diff
@@ -0,0 +1,184 @@
...
+
+        env.execute(
+                "Load Simulation (repeats after "
+                        + Duration.of(repeatsAfterMs, ChronoUnit.MILLIS)
+                        + ")");
+    }
+
+    private static class ImpulseSource implements SourceFunction<Long> {
+        private final int maxSleepTimeMs;
+        boolean canceled;
```
gyfora, line 102: has to be `volatile`

examples/autoscaling/src/main/java/autoscaling/LoadSimulationPipeline.java
```diff
@@ -0,0 +1,184 @@
...
+
+        public ImpulseSource(int samplingInterval) {
+            this.maxSleepTimeMs = samplingInterval / 10;
+        }
+
+        @Override
+        public void run(SourceContext<Long> sourceContext) throws Exception {
+            while (!canceled) {
+                sourceContext.collect(42L);
```
gyfora, line 111: should collect under checkpoint lock

examples/autoscaling/src/main/java/autoscaling/LoadSimulationPipeline.java
```diff
@@ -0,0 +1,184 @@
...
+import org.apache.flink.util.Collector;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.time.Duration;
+import java.time.temporal.ChronoUnit;
+
+/**
+ * Example pipeline which simulates load fluctuating load from zero to a defined max, and
```
gyfora, line 36: `load fluctuating load` -> `fluctuating load`

examples/autoscaling/autoscaling-canary.yaml
```diff
@@ -0,0 +1,56 @@
...
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+# limitations under the License.
+################################################################################
+
+apiVersion: flink.apache.org/v1beta1
+kind: FlinkDeployment
+metadata:
+  name: autoscaling-canary
```
gyfora, line 22: I wonder if the name canary may be confused by the actual canary functionality in the operator. Maybe we could simply name it example2 or periodic load or something like that.

examples/autoscaling/autoscaling-canary.yaml
```diff
@@ -0,0 +1,56 @@
...
+spec:
+  image: autoscaling-example
+  flinkVersion: v1_18
+  flinkConfiguration:
+    kubernetes.operator.job.autoscaler.enabled: "true"
+    kubernetes.operator.job.autoscaler.stabilization.interval: "1m"
+    kubernetes.operator.job.autoscaler.metrics.window: "15m"
+    kubernetes.operator.job.autoscaler.target.utilization: "0.5"
+    kubernetes.operator.job.autoscaler.target.utilization.boundary: "0.3"
+    pipeline.max-parallelism: "32"
```
gyfora, line 32: should we enable the adaptive scheduler?

gyfora (changes requested):

examples/autoscaling/src/main/java/autoscaling/LoadSimulationPipeline.java
mxm: I'm not sure about that. This will still stop, there can be a small delay though.

examples/autoscaling/src/main/java/autoscaling/LoadSimulationPipeline.java
mxm: The lock is there to ensure correctness while checkpointing. What is being checkpointed? This is not stateful function, neither are the other operators.

examples/autoscaling/autoscaling-canary.yaml
mxm: We can but it is not essential for this pipeline.

examples/autoscaling/autoscaling-canary.yaml
mxm: Sure, we can rename to `autoscaling-dynamic`.

examples/autoscaling/src/main/java/autoscaling/LoadSimulationPipeline.java
gyfora: If checkpointing is off this doesn’t matter much

gyfora (approved):

PR #656, opened 2023-08-23
[FLINK-32904] Support cron expressions for periodic snapshots triggering

afedulov:

Current functionality to trigger savepoints and checkpoints is based on specifying a fixed interval. This approach gives little control to prevent potentially overlapping snapshots running simultaneously in adjacent deployments and can cause issues for high scale, large state jobs. Moreover, lastTrigger information is not retained across deployment restarts which leads to immediate firing based on creation time. In order to have better control over the exact schedule, triggering based on cron expressions should be added.

## Brief change log

  - Adds cron expression config options for checkpoint and savepoint triggering
  - Extends SnapshotUtils logic to process both interval- and cron-based triggers

## Verifying this change
  -  Added unit and integration tests

## Documentation

  - If yes, how is the feature documented? generated config docs

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
@@ -305,7 +305,22 @@ public class KubernetesOperatorConfigOptions {
                     .withDescription(
                             "Interval at which periodic savepoints will be triggered. "
                                     + "The triggering schedule is not guaranteed, savepoints will be "
-                                    + "triggered as part of the regular reconcile loop.");
+                                    + "triggered as part of the regular reconcile loop. "
+                                    + "WARNING: not intended to be used together with the cron-based "
+                                    + "periodic savepoint triggering");
+
+    @Documentation.Section(SECTION_DYNAMIC)
+    public static final ConfigOption<String> PERIODIC_SAVEPOINT_CRON =
+            operatorConfig("periodic.savepoint.cron")
+                    .stringType()
+                    .defaultValue(
+                            "0 0 0 29 2 ? 1999") // Default value that is guaranteed not to trigger
+                    .withDescription(
+                            "A cron expression that defines when to automatically trigger savepoints."
+                                    + "The precise triggering schedule is not guaranteed, savepoints will be "
+                                    + "triggered as part of the regular reconcile loop. "
+                                    + "WARNING: not intended to be used together with the interval-based "
+                                    + "periodic savepoint triggering");
 
     @Documentation.Section(SECTION_DYNAMIC)
     public static final ConfigOption<Duration> PERIODIC_CHECKPOINT_INTERVAL =
@@ -321,7 +336,32 @@ public class KubernetesOperatorConfigOptions {
                                     + "but to complement them in special cases. For instance, a "
                                     + "full checkpoint might need to be occasionally triggered to "
                                     + "break the chain of incremental checkpoints and consolidate "
-                                    + "the partial incremental files.");
+                                    + "the partial incremental files. "
+                                    + "WARNING: not intended to be used together with the cron-based "
+                                    + "periodic checkpoint triggering");
+
+    @Documentation.Section(SECTION_DYNAMIC)
+    public static final ConfigOption<String> PERIODIC_CHECKPOINT_CRON =
+            operatorConfig("periodic.checkpoint.cron")
+                    .stringType()
+                    .defaultValue(
+                            "0 0 0 29 2 ? 1999") // Default value that is guaranteed not to trigger (not a leap year)
+                    .withDescription(
+                            "A cron expression that defines when to automatically trigger checkpoints. "
+                                    + "The precise triggering schedule is not guaranteed, checkpoints will "
+                                    + "be triggered as part of the regular reconcile loop. Periodic triggering "
+                                    + "with frequency higher than the reconciliation loop "
+                                    + "("
+                                    + OPERATOR_RECONCILE_INTERVAL.key()
+                                    + ") will lead to trigger omissions. "
+                                    + "NOTE: checkpoints are generally managed by Flink. This "
+                                    + "setting isn't meant to replace Flink's checkpoint settings, "
+                                    + "but to complement them in special cases. For instance, a "
+                                    + "full checkpoint might need to be occasionally triggered to "
+                                    + "break the chain of incremental checkpoints and consolidate "
+                                    + "the partial incremental files. "
+                                    + "WARNING: not intended to be used together with the interval-based "
+                                    + "periodic checkpoint triggering");
 
     @Documentation.Section(SECTION_SYSTEM)
     public static final ConfigOption<String> OPERATOR_WATCHED_NAMESPACES =
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
@@ -304,7 +304,9 @@ public class KubernetesOperatorConfigOptions {
                     .defaultValue("")
                     .withDescription(
                             "Option to enable automatic savepoint triggering. Can be specified "
-                                    + "either as a Duration type (i.e. '10m') or as a cron expression."
+                                    + "either as a Duration type (i.e. '10m') or as a cron expression "
+                                    + "in Quartz format (6 or 7 positions, see "
+                                    + "http://www.quartz-scheduler.org/documentation/quartz-2.3.0/tutorials/crontrigger.html)."
                                     + "The triggering schedule is not guaranteed, savepoints will be "
                                     + "triggered as part of the regular reconcile loop. "
                                     + "WARNING: not intended to be used together with the cron-based "
@@ -317,7 +319,9 @@ public class KubernetesOperatorConfigOptions {
                     .defaultValue("")
                     .withDescription(
                             "Option to enable automatic checkpoint triggering. Can be specified "
-                                    + "either as a Duration type (i.e. '10m') or as a cron expression."
+                                    + "either as a Duration type (i.e. '10m') or as a cron expression "
+                                    + "in Quartz format (6 or 7 positions, see "
+                                    + "http://www.quartz-scheduler.org/documentation/quartz-2.3.0/tutorials/crontrigger.html)."
                                     + "The triggering schedule is not guaranteed, checkpoints will "
                                     + "be triggered as part of the regular reconcile loop. "
                                     + "NOTE: checkpoints are generally managed by Flink. This "
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
@@ -298,37 +298,26 @@ public class KubernetesOperatorConfigOptions {
                                     + "Expected format: headerKey1:headerValue1,headerKey2:headerValue2.");
 
     @Documentation.Section(SECTION_DYNAMIC)
-    public static final ConfigOption<Duration> PERIODIC_SAVEPOINT_INTERVAL =
+    public static final ConfigOption<String> PERIODIC_SAVEPOINT_INTERVAL =
             operatorConfig("periodic.savepoint.interval")
-                    .durationType()
-                    .defaultValue(Duration.ZERO)
+                    .stringType()
+                    .defaultValue("")
                     .withDescription(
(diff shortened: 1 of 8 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/SnapshotUtils.java
```diff
@@ -229,33 +228,85 @@ protected static Optional<SnapshotTriggerType> shouldTriggerSnapshot(
...
+                        "Triggering new periodic {} based on cron schedule '{}' due at {}",
+                        snapshotType.toString().toLowerCase(),
+                        cronExpressionString,
+                        nextValidTimeAfterLastTrigger);
+                return true;
+            } else {
+                return false;
+            }
+        } catch (ParseException e) {
+            LOG.warn("Invalid cron expression: " + cronExpressionString);
```
gyfora, line 287: should we send a Kube event so the user sees this?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
```diff
@@ -321,7 +336,32 @@ public static String operatorConfigKey(String key) {
...
+                                    + OPERATOR_RECONCILE_INTERVAL.key()
+                                    + ") will lead to trigger omissions. "
+                                    + "NOTE: checkpoints are generally managed by Flink. This "
+                                    + "setting isn't meant to replace Flink's checkpoint settings, "
+                                    + "but to complement them in special cases. For instance, a "
+                                    + "full checkpoint might need to be occasionally triggered to "
+                                    + "break the chain of incremental checkpoints and consolidate "
+                                    + "the partial incremental files. "
+                                    + "WARNING: not intended to be used together with the interval-based "
+                                    + "periodic checkpoint triggering");
```
mxm, line 364: Would the interval and the cron-based logic really not work together, or is this more about warning the user about unwanted checkpoint triggering?

mxm: For the sake of simplifying this, could we unify both options under `periodic.{checkpoint,savepoint}.interval`? We can parse the String option using `Duration`, if it fails (i.e. throws `DateTimeParseException`) we try the cron-based approach. I think this approach should be safe because a mixup between the two is not possible.

gyfora: I was thinking the same thing, +1 for using the single existing interval config for this.

afedulov: @mxm yes, the warning just relates to the fact that such mixing makes triggering not predictable, because there is just one last checkpoint/savepoint timestamp field that is shared between the two.

afedulov: > For the sake of simplifying this, could we unify both options under periodic.
I am not against it, but see the following downsides:
- We won’t be able to tell if the user misspelled something in the duration format or her cron expression is invalid.  Since Quartz supports both 6- and 7-positional expressions, it can get confusing sometimes when ? or * are used in the "wrong" place. Some combinations are invalid, but we won't able to point at the error with the cron format, since there is a possibility that the user wanted to specify Duration to begin with, but made a mistake there.
- The name would become a bit awkward:  `periodic.{checkpoint|savepoint}.interval` , whereas cron does not necessarily represent an interval, but an also be a point-in-time setting.  This probably leads to the necessity to deprecate the already-released `periodic.savepoint.interval`
- Although the absolutely valid intention is to make things simple for the user, the complexity of explaining the "Eierlegende Milchwollsau" functionality of the single setting in the docs plus the unclear naming seem to work against this intention.

afedulov: After thinking about this a bit more, I guess a generic error message that "something is wrong with the format, please make sure to either use a valid Duration data type of a cron expression" should also be OK.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/SnapshotUtils.java
afedulov: I guess we could. Would require quite some passing around of the `EventRecorder`, plus mocking it for unit tests. Do we do that for other misconfigured settings?

gyfora: No, you are right it’s actually better to validate this in the DefaultValidator

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
afedulov: @mxm, @gyfora 
PTAL https://github.com/apache/flink-kubernetes-operator/commit/412f9136c5a26b3160b50aefb7e19e4430d9c0f8

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
```diff
@@ -298,37 +298,26 @@ public static String operatorConfigKey(String key) {
...
+    public static final ConfigOption<String> PERIODIC_SAVEPOINT_INTERVAL =
             operatorConfig("periodic.savepoint.interval")
-                    .durationType()
-                    .defaultValue(Duration.ZERO)
+                    .stringType()
+                    .defaultValue("")
                     .withDescription(
-                            "Interval at which periodic savepoints will be triggered. "
+                            "Option to enable automatic savepoint triggering. Can be specified "
+                                    + "either as a Duration type (i.e. '10m') or as a cron expression."
```
mxm, line 307:
```suggestion
                                    + "either as a Duration type (i.e. '10 minutes') or as a cron expression (see https://en.wikipedia.org/wiki/Cron)."
```

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
```diff
@@ -298,37 +298,26 @@ public static String operatorConfigKey(String key) {
...
+    public static final ConfigOption<String> PERIODIC_CHECKPOINT_INTERVAL =
             operatorConfig("periodic.checkpoint.interval")
-                    .durationType()
-                    .defaultValue(Duration.ZERO)
+                    .stringType()
+                    .defaultValue("")
                     .withDescription(
-                            "Interval at which periodic checkpoints will be triggered. "
+                            "Option to enable automatic checkpoint triggering. Can be specified "
+                                    + "either as a Duration type (i.e. '10m') or as a cron expression."
```
mxm, line 320: Same comment as above.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/SnapshotUtils.java
```diff
@@ -259,6 +257,31 @@ protected static Optional<SnapshotTriggerType> shouldTriggerSnapshot(
...
+    private static boolean shouldTriggerAutomaticSnapshot(
+            SnapshotType snapshotType, String automaticTriggerExpression, Instant lastTrigger) {
+        if (StringUtils.isBlank(automaticTriggerExpression)) {
+            return false;
+        }
+
+        boolean shouldTrigger =
+                shouldTriggerIntervalBasedSnapshot(
+                                snapshotType, automaticTriggerExpression, lastTrigger)
+                        || shouldTriggerCronBasedSnapshot(
```
mxm, line 269: This condition means that theoretically still both can trigger. We should make sure that the condition guarantees mutual exclusion. So first check interval, then check cron based execution.

afedulov: I actually just realized that the logic is also false. We cannot rely on the true/false at all. It might be that both triggers are simply not due. We'll need to work with exceptions directly. I'll give it another go.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
afedulov: Added a link to Quartz cron syntax.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
afedulov: Added a link to Quartz cron syntax.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/SnapshotUtils.java
```diff
@@ -284,20 +307,27 @@ static boolean shouldTriggerCronBasedSnapshot(
...
 
     private static boolean shouldTriggerIntervalBasedSnapshot(
-            SnapshotType snapshotType, Duration interval, Instant lastTrigger) {
+            SnapshotType snapshotType, String triggerExpression, Instant lastTrigger) {
+
+        Duration interval;
+        try {
+            interval = ConfigurationUtils.convertValue(triggerExpression, Duration.class);
+        } catch (IllegalArgumentException exception) {
+            return false;
```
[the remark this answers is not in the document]
afedulov: Modified.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/SnapshotUtils.java
afedulov: Fixed the logic:
https://github.com/apache/flink-kubernetes-operator/pull/656/files#diff-deedaba89a69eefb0acf4c0fcbb250b5ec9415cd78a3323827a0ce0c162b8b49R261

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/SnapshotUtils.java
```diff
@@ -229,33 +228,140 @@ protected static Optional<SnapshotTriggerType> shouldTriggerSnapshot(
...
+        // expression at the same time.
+        if (expressionIsAnInterval && expressionIsACron) {
+            LOG.error(
+                    "Something went wrong with the automatic {} trigger expression {}. This setting cannot be simultaneously a valid Duration and a cron expression.",
+                    snapshotType,
+                    automaticTriggerExpression);
+            return false;
+        }
+
+        return shouldTriggerInterval || shouldTriggerCron;
```
mxm, line 307: I find this method a bit too complex. How about changing the structure here a bit:

```java
if (isInterval()) {
  // check Interval triggering
} else {
  // check cron triggering
}
```

Can we remove the check for both an interval and a cron? We can simply assume the interval always has precedence.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/SnapshotUtils.java
```diff
@@ -264,101 +264,103 @@ static boolean shouldTriggerAutomaticSnapshot(
...
 
         // This should never happen. The string cannot be both a valid Duration and a cron
         // expression at the same time.
-        if (expressionIsAnInterval && expressionIsACron) {
+        if (isInterval && isCron) {
             LOG.error(
                     "Something went wrong with the automatic {} trigger expression {}. This setting cannot be simultaneously a valid Duration and a cron expression.",
                     snapshotType,
                     automaticTriggerExpression);
             return false;
```
mxm, line 277: The question is whether we want to skip triggering entirely or we choose to go with the interval. I think semantically it is not possible for a proper cron to be an interval, so this is is an extreme edge case. Also we fail silently and continue operation which is also debatable. 

I would just go with whatever matches first, i.e. the interval, but maybe there is no right and wrong.

afedulov: I think if there is any such weird case possible, I'd prefer to uncover it. If a user configures an interval, it is reasonable tot expect that she would eventually check it is is firing. In that case t is arguably better not to fire at all and potentially get a bug submitted than trigger a random thing that matches, but is interpreted in the wrong way.

mxm (approved):

afedulov:
@mxm Thanks for the review!

PR #657, opened 2023-08-24
[FLINK-31406] Do not delete jobgraph on scale only last-state upgrades

SteNicholas:

Currently the Flink service always deletes the jobgraph from HA metadata so that it's regenerated for last-state upgrades. This is unnecessary for scale only operations. Keeping the jobgraph could greately speed up startup time for some jobs. `AbstractFlinkService` supports to delete jobgraph on non-scale operations.

## Brief change log

- `AbstractFlinkService` deletes jobgraph on non-scale operations.

## Verifying this change

  - Adds `ApplicationReconcilerTest#testScaleWithJobGraphKeptInHAMetadata` to verify whether to delete jobgraph on scale only last-state upgrades.

  - Core observer or reconciler logic that is regularly executed: (yes / **no**)

## Documentation

  - If yes, how is the feature documented? (**not applicable** / docs / JavaDocs / not documented)


SteNicholas:
@gyfora, PTAL.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
```diff
@@ -215,6 +221,18 @@ private void setJobIdIfNecessary(
...
+        var cr = ctx.getResource();
+        FlinkDeploymentSpec lastReconciledSpec =
+                cr.getStatus().getReconciliationStatus().deserializeLastReconciledSpec();
+        FlinkDeploymentSpec currentDeploySpec = cr.getSpec();
+        var specDiff =
+                new ReflectiveDiffBuilder<>(
+                                ctx.getDeploymentMode(), lastReconciledSpec, currentDeploySpec)
+                        .build();
+        return specDiff.getType() == DiffType.SCALE;
+    }
```
gyfora, line 234: we should pass this information to deploy instead of computing it again

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
```diff
@@ -176,23 +176,20 @@ public KubernetesClient getKubernetesClient() {
...
-            final String namespace = conf.get(KubernetesConfigOptions.NAMESPACE);
-            FlinkUtils.deleteJobGraphInKubernetesHA(clusterId, namespace, kubernetesClient);
-        } else if (FlinkUtils.isZookeeperHAActivated(conf)) {
-            FlinkUtils.deleteJobGraphInZookeeperHA(conf);
-        }
-
         if (requireHaMetadata) {
             validateHaMetadataExists(conf);
+            LOG.info(
+                    "Deleting job graph from HA metadata-{}",
```
gyfora, line 188: I think this logging is off...
Can we simply log `Deleting ....` or `Not deleting ...`?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkService.java
```diff
@@ -54,7 +54,8 @@ public interface FlinkService {
 
     KubernetesClient getKubernetesClient();
 
-    void submitApplicationCluster(JobSpec jobSpec, Configuration conf, boolean requireHaMetadata)
+    void submitApplicationCluster(
+            JobSpec jobSpec, Configuration conf, boolean requireHaMetadata, boolean requireScale)
```
gyfora, line 58: We could rename the new parameter to `deleteJobGraph`?

SteNicholas:
@gyfora, I have addressed above comments. PTAL.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
```diff
@@ -267,7 +270,8 @@ protected abstract boolean reconcileSpecChange(
      * @return True if any further reconciliation action was taken.
      * @throws Exception Error during reconciliation.
      */
-    protected abstract boolean reconcileOtherChanges(FlinkResourceContext<CR> ctx) throws Exception;
+    protected abstract boolean reconcileOtherChanges(
+            FlinkResourceContext<CR> ctx, boolean requireScale) throws Exception;
```
gyfora, line 274: Let's pass the difftype as an argument instead of the confusing boolean param I think.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
```diff
@@ -257,7 +259,8 @@ private void updateStatusBeforeFirstDeployment(
      * @return True if spec change reconciliation was executed
      */
     protected abstract boolean reconcileSpecChange(
-            FlinkResourceContext<CR> ctx, Configuration deployConfig) throws Exception;
+            FlinkResourceContext<CR> ctx, Configuration deployConfig, boolean requireScale)
```
gyfora, line 262: Let's pass the difftype as an argument instead of the confusing boolean param I think.

gyfora (changes requested):

SteNicholas:
@gyfora, sorry for naming a confusing parameter name. I have addressed rename comments and renamed to `DiffType diffType`. PTAL.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconcilerTest.java
```diff
@@ -847,6 +865,62 @@ public void cancelJob(
...
+                jobGraphVal,
+                kubernetesClient
+                        .configMaps()
+                        .inNamespace(TEST_NAMESPACE)
+                        .withLabels(configMapLabels)
+                        .list()
+                        .getItems()
+                        .get(0)
+                        .getData()
+                        .get(jobGraphKey));
```
gyfora, line 921: I think this test is incorrect in the sense that it doesn't validate the right things. 
The last-state-fallback config is unnecessary.

After the first reconcile you can use:
```
reconciler.reconcile(deployment, context);
        verifyAndSetRunningJobsToStatus(deployment, flinkService.listJobs());
```
Like in other tests. After that you should submit the scale change, verify that after first reconcile the job is suspended, and after the second reconcile (during deployment) the job graph is not deleted

SteNicholas: @gyfora, I have addressed the comments and corrected the test case. PTAL.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconcilerTest.java
```diff
@@ -873,6 +891,58 @@ public void cancelJob(
...
+        reconciler.reconcile(deployment, context);
+        assertEquals(JobState.SUSPENDED, getReconciledJobState(deployment));
+        assertFalse(deployment.getStatus().getReconciliationStatus().scalingInProgress());
+        getJobSpec(deployment).setState(JobState.SUSPENDED);
+        deployment
+                .getSpec()
+                .getFlinkConfiguration()
+                .put(PipelineOptions.PARALLELISM_OVERRIDES.key(), new JobVertexID() + ":1");
+        reconciler.reconcile(deployment, context);
+        assertEquals(
```
gyfora, line 933: Sorry but I still don't fully understand what is happening here.
1. You first change the parallelism (that is not a `SCALE` change in native mode only the parallelism overrides currently) and then you assert that the job is suspended? What is the point? Could we simply remove this part of the test?
2. You set the state to suspended and add parallelism overrides. That will not scale the job, that will suspend the job. We naturally don't delete things on suspend, only on recovery so there is nothing to test at this point but also once you restore the jobgrapgh should be deleted because this is not SCALE change because you changed the state as well

gyfora (changes requested):

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconcilerTest.java
gyfora: Could you please add some comments to the test case (step by step) just so that we can all understand what is being tested/verified in each step?

PR #660, opened 2023-08-29
[FLINK-32991] Ensure registration of all scaling metrics

mxm:
Not all metrics appear in the latest 1.6 release. This is because we report metrics as soon as they are available and the registration code assumes that they will all be available at once. In practice, some are only available after sampling data multiple times. For example, TARGET_DATA_RATE is only available after the source metrics have been aggregated and the lag has been computed.

### DIFF
--- a/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/AutoscalerFlinkMetrics.java
+++ b/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/AutoscalerFlinkMetrics.java
@@ -49,7 +49,7 @@ public class AutoscalerFlinkMetrics {
 
     final Counter numBalanced;
 
-    private final MetricGroup metricGroup;
+    final MetricGroup metricGroup;
 
     private boolean scalingMetricsInitialized;
 
--- a/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/AutoscalerFlinkMetrics.java
+++ b/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/AutoscalerFlinkMetrics.java
@@ -27,10 +27,9 @@ import org.apache.flink.runtime.jobgraph.JobVertexID;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.Set;
 import java.util.function.Supplier;
 
 import static org.apache.flink.kubernetes.operator.autoscaler.metrics.ScalingMetric.PARALLELISM;
@@ -52,7 +51,7 @@ public class AutoscalerFlinkMetrics {
 
     private final MetricGroup metricGroup;
 
-    private final Set<JobVertexID> vertexMetrics = new HashSet<>();
+    private boolean scalingMetricsInitialized;
 
     public AutoscalerFlinkMetrics(MetricGroup metricGroup) {
         this.numScalings = metricGroup.counter("scalings");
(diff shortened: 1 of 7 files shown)


flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/AutoscalerFlinkMetrics.java
```diff
@@ -62,54 +61,49 @@ public AutoscalerFlinkMetrics(MetricGroup metricGroup) {
...
+        if (scalingMetricsInitialized) {
+            // It is important that we only initialize these metrics once because the Flink API does
+            // not support registering counters / gauges multiple times.
+            // The metrics will be updated via the provided metricsSupplier.
+            return;
+        }
+        scalingMetricsInitialized = true;
+
+        LOG.info("Registering scaling metrics");
+        for (JobVertexID jobVertexID : jobVerticesSupplier.get()) {
```
1996fanrui, line 76: The `jobVerticesSupplier.get()` is just called once, and it's not the lazy call. so the `jobVerticesSupplier` parameter can be simplified to `jobVertices`, right?

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/AutoscalerFlinkMetrics.java
```diff
@@ -62,54 +61,49 @@ public AutoscalerFlinkMetrics(MetricGroup metricGroup) {
...
+                    metricsSupplier) {
+        if (scalingMetricsInitialized) {
+            // It is important that we only initialize these metrics once because the Flink API does
+            // not support registering counters / gauges multiple times.
+            // The metrics will be updated via the provided metricsSupplier.
+            return;
+        }
+        scalingMetricsInitialized = true;
+
+        LOG.info("Registering scaling metrics");
```
1996fanrui, line 75: Is this log necessary?

If yes, could we add the resourceId(similar to job information)? Otherwise we don't know which job is registering scaling metrics.

flink-kubernetes-operator-autoscaler/src/test/java/org/apache/flink/kubernetes/operator/autoscaler/AutoScalerFlinkMetricsTest.java
```diff
@@ -66,8 +67,10 @@ public void testMetricsRegistration() {
         initRecommendedParallelism(evaluatedMetrics);
         lastEvaluatedMetrics.put(resourceID, evaluatedMetrics);
 
-        metrics.registerScalingMetrics(() -> lastEvaluatedMetrics.get(resourceID));
-        metrics.registerScalingMetrics(() -> lastEvaluatedMetrics.get(resourceID));
+        metrics.registerScalingMetrics(
+                () -> List.of(jobVertexID), () -> lastEvaluatedMetrics.get(resourceID));
+        metrics.registerScalingMetrics(
+                () -> List.of(jobVertexID), () -> lastEvaluatedMetrics.get(resourceID));
```
1996fanrui, line 73: Could you help adding a test that some metrics are collected after `registerScalingMetrics`?

1996fanrui (commented):
Hi @mxm , thanks for your contribution!

Overall looks good to me, I left a couple of minor comments, please take a look in your free time, thanks~

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/AutoscalerFlinkMetrics.java
mxm: Getting the job vertices is an expensive operation. We want to avoid calling it on every call to this method which will happen for every reconciliation loop.

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/AutoscalerFlinkMetrics.java
mxm: The logger is already scoped by resource id.

mxm: The logging was there before and I think it is helpful to print it once after resource creation.

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/AutoscalerFlinkMetrics.java
1996fanrui: Thanks for the clarification, the optimization makes sense to me.

Could you help adding a comment here? This optimization is implicit, and adding comments can prevent other developers from removing this optimization during refactoring in the future.

flink-kubernetes-operator-autoscaler/src/test/java/org/apache/flink/kubernetes/operator/autoscaler/AutoScalerFlinkMetricsTest.java
mxm: Sure I can do that, although this is not the scope of this JIRA. The scope here is to register all metrics, not check that they are reported. There are already tests that do that.

mxm:
Thanks for the review @1996fanrui. I addressed your comments.

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/AutoscalerFlinkMetrics.java
mxm: Good point. I've refactored the code, to cache the topological ordering because we call this at multiple places already. That way we can remove the Supplier for the vertices.

flink-kubernetes-operator-autoscaler/src/test/java/org/apache/flink/kubernetes/operator/autoscaler/AutoScalerFlinkMetricsTest.java
1996fanrui: Sorry, Max. I didn't express clear.

This PR you fix a bug that some metrics aren't ready when  `registerScalingMetrics` is called, so these metrics cannot be registerd even if they are ready in the future.

Could we adding a test for this case? This test should fail without your PR, and it should success with your PR.

Please let me know if it's not clear, thanks~

mxm: I added this test in https://github.com/apache/flink-kubernetes-operator/pull/660/commits/7cc8bbc51cbf3aeba9d035f50bf0d1b76e8b5e8f

1996fanrui (approved):
Thanks @mxm for the fix and update, LGTM.

PR #662, opened 2023-09-01
[FLINK-33008] Refactor to use KubernetesClient from JOSDK Context

darenwkt:

Refactor code to use KubernetesClient from JOSDK Context

## Brief change log

KubernetesClient is passed from josdkContext to flinkResourceContext. The rest of the code get client from either context.

## Verifying this change

## Documentation

### DIFF
--- a/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/AutoscalerInfoManager.java
+++ b/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/AutoscalerInfoManager.java
@@ -42,13 +42,11 @@ public class AutoscalerInfoManager {
 
     private final ConcurrentHashMap<ResourceID, Optional<AutoScalerInfo>> cache =
             new ConcurrentHashMap<>();
-    private final KubernetesClient kubernetesClient;
 
-    public AutoscalerInfoManager(KubernetesClient kubernetesClient) {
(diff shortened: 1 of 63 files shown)


flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/listener/FlinkResourceListener.java
```diff
@@ -45,8 +44,6 @@ public interface FlinkResourceListener extends Plugin {
     interface ResourceContext<R extends AbstractFlinkResource<?, ?>> {
         R getFlinkResource();
 
-        KubernetesClient getKubernetesClient();
```
gyfora, line 48: I think we should not touch this here, I think the FlinkResourceListener interfaces should be reworked but probably not as part of this ticket.

gyfora (changes requested):
Added a minor comment about the resource listener interface, otherwise looks good.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/listener/FlinkResourceListener.java
darenwkt: Thanks for the comment, have addressed it in latest commit

darenwkt:
Thanks Gyula for the review, have addressed the comment in latest commit

gyfora (approved):

PR #663, opened 2023-09-05
[FLINK-32777] add config for opt out okhttp hostname verification

HuangZhenQiu:
Currently, Flink operator can't run in IPV6 environment due to a OKhttp host name verification bug. To mitigate the issue reported by users, this change enable a config for user to choose disable host name verification in http client

## Brief change log
- add a configuration to disble okhttp host name verification for IPV6 

  - The public API, i.e., is any changes to the `CustomResourceDescriptors`: (yes)

## Documentation


gyfora (changes requested):
Could you please link the issue that this is trying to solve?

There are no tests and no clear explanation why this route was taken. In general I would prefer to solve it in a different way, such as upgrading the okhttp client to a version that is fixed or determine the verification setting automatically.

I think it's not very user friendly to introduce new configs to work around bugs in dependencies this way.

HuangZhenQiu:
@gyfora 
The user reported a work around by using environment variable before the new Okhttp version upgrade. If so, the config is not needed. I will create another PR to add how to run operator IPV6 in the doc.

PR #664, opened 2023-09-05
[FLINK-33027] Allow users to override parallelism of excluded vertices

gyfora:

1. Move parallelism override application to JobAutoscaler instead of reconciler to not leak autoscaler logic
2. Allow users to override excluded vertex parallelisms to avoid the corner case where excluded vertex parallelisms are stuck

## Verifying this change

Improved and extended unit test coverage

## Documentation

### DIFF
--- a/flink-kubernetes-operator-autoscaler/src/test/java/org/apache/flink/kubernetes/operator/autoscaler/JobAutoScalerImplTest.java
+++ b/flink-kubernetes-operator-autoscaler/src/test/java/org/apache/flink/kubernetes/operator/autoscaler/JobAutoScalerImplTest.java
@@ -20,9 +20,11 @@ package org.apache.flink.kubernetes.operator.autoscaler;
 import org.apache.flink.api.common.JobID;
 import org.apache.flink.api.common.JobStatus;
 import org.apache.flink.configuration.Configuration;
+import org.apache.flink.configuration.PipelineOptions;
 import org.apache.flink.kubernetes.operator.OperatorTestBase;
 import org.apache.flink.kubernetes.operator.TestUtils;
 import org.apache.flink.kubernetes.operator.api.FlinkDeployment;
+import org.apache.flink.kubernetes.operator.autoscaler.config.AutoScalerOptions;
 import org.apache.flink.kubernetes.operator.autoscaler.metrics.FlinkMetric;
 import org.apache.flink.kubernetes.operator.autoscaler.metrics.ScalingMetric;
 import org.apache.flink.kubernetes.operator.autoscaler.topology.JobTopology;
@@ -48,6 +50,7 @@ import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 import java.time.Instant;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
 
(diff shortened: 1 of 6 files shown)


flink-kubernetes-operator-autoscaler/src/test/java/org/apache/flink/kubernetes/operator/autoscaler/JobAutoScalerImplTest.java
```diff
@@ -196,4 +199,75 @@ void testParallelismOverrides() throws Exception {
...
+                .put(PipelineOptions.PARALLELISM_OVERRIDES.key(), v1 + ":1");
+        autoscaler.applyParallelismOverrides(getResourceContext(deployment));
+        assertEquals(specClone, deployment.getSpec());
+
+        // Define partly overlapping overrides
+        var v2 = new JobVertexID();
+        deployment
+                .getSpec()
+                .getFlinkConfiguration()
+                .put(PipelineOptions.PARALLELISM_OVERRIDES.key(), v1 + ":1," + v2 + ":4");
```
gaborgsomogyi, line 249: OFFTOPIC: As a general user it would be good to see in the doc how to define maps in the config.

flink-kubernetes-operator-autoscaler/src/test/java/org/apache/flink/kubernetes/operator/autoscaler/JobAutoScalerImplTest.java
```diff
@@ -196,4 +199,75 @@ void testParallelismOverrides() throws Exception {
...
+        assertEquals(
+                Map.of(v1.toHexString(), "2"),
+                getResourceContext(deployment)
+                        .getDeployConfig(deployment.getSpec())
+                        .get(PipelineOptions.PARALLELISM_OVERRIDES));
+
+        // Test when there are also user overrides, autoscaler should take precedence
+        specClone = ReconciliationUtils.clone(deployment.getSpec());
+
+        // This should be ignored
```
gaborgsomogyi, line 236: It would be good to add a minimal explanation why: `, because ...`

gaborgsomogyi (approved):
LGTM. Some explanation would be good to speed up dev's/user's understanding but that's optional.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/JobAutoScaler.java
```diff
@@ -30,6 +28,6 @@ public interface JobAutoScaler {
     /** Called when the custom resource is deleted. */
     void cleanup(FlinkResourceContext<?> ctx);
 
-    /** Get the current parallelism overrides for the job. */
-    Map<String, String> getParallelismOverrides(FlinkResourceContext<?> ctx);
+    /** Apply current parallelism overrides. */
+    void applyParallelismOverrides(FlinkResourceContext<?> ctx);
```
mxm, line 32: Is this interface method really necessary? Overrides should be applied in the scale method, no?

gyfora: This is based on earlier discussion in https://issues.apache.org/jira/browse/FLINK-32589 
Overrides need to be applied separately (and it was implemented already this way I just relocated the logic) so that we do not lose them on spec changes.

mxm: I couldn't find any discussion in that thread which would explain why the overrides couldn't be applied without the method here. Perhaps the code would need refactoring (which is done here already) to apply the scale method before reconciling other parts of the spec.

mxm (approved):
Thanks for the cleanup! Ideally, autoscaling would manage the overrides end-to-end but this is already cleaner than before.

gyfora:
> Thanks for the cleanup! Ideally, autoscaling would manage the overrides end-to-end but this is already cleaner than before.

I agree, but for that as you said we need to first call the scale method before anything else. Currently we do the opposite and call scale after everything else. We have to think this through to make sure we don't introduce some problem in the general reconciler / autoscaler logic

PR #666, opened 2023-09-08
[FLINK-33067] Expose rate limiter config and enable by default

gyfora:

There are certain cases currently when the operator can stuck in an infinite no wait retry loop under certain circumstances (see https://github.com/operator-framework/java-operator-sdk/issues/2046 for details).

As a solution we introduce rate limiter configuration with a default that should already safeguard against prod problems that may arise without it.

The PR also fixes and improves the default retry configuration and fixes a bug with exponential backoff.

## Brief change log

  - *Expose rate limiter config and enable by default*
  - *Update default retry config*
  - *Unit Tests*

## Verifying this change

Unit tests, manual verification

## Documentation

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkOperatorConfiguration.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkOperatorConfiguration.java
@@ -25,6 +25,8 @@ import org.apache.flink.kubernetes.operator.utils.EnvUtils;
 
 import io.fabric8.kubernetes.api.model.DeletionPropagation;
 import io.javaoperatorsdk.operator.api.config.LeaderElectionConfiguration;
+import io.javaoperatorsdk.operator.processing.event.rate.LinearRateLimiter;
+import io.javaoperatorsdk.operator.processing.event.rate.RateLimiter;
 import io.javaoperatorsdk.operator.processing.retry.GenericRetry;
 import lombok.Value;
 import org.apache.commons.lang3.StringUtils;
@@ -63,6 +65,7 @@ public class FlinkOperatorConfiguration {
     Integer savepointHistoryCountThreshold;
     Duration savepointHistoryAgeThreshold;
     GenericRetry retryConfiguration;
+    RateLimiter<?> rateLimiter;
     boolean exceptionStackTraceEnabled;
     int exceptionStackTraceLengthThreshold;
     int exceptionFieldLengthThreshold;
@@ -174,6 +177,7 @@ public class FlinkOperatorConfiguration {
                         KubernetesOperatorMetricOptions.OPERATOR_METRICS_HISTOGRAM_SAMPLE_SIZE);
 
         GenericRetry retryConfiguration = getRetryConfig(operatorConfig);
+        RateLimiter rateLimiter = getRateLimiter(operatorConfig);
 
         String labelSelector =
                 operatorConfig.getString(KubernetesOperatorConfigOptions.OPERATOR_LABEL_SELECTOR);
@@ -200,6 +204,7 @@ public class FlinkOperatorConfiguration {
                 savepointHistoryCountThreshold,
                 savepointHistoryAgeThreshold,
                 retryConfiguration,
+                rateLimiter,
                 exceptionStackTraceEnabled,
                 exceptionStackTraceLengthThreshold,
                 exceptionFieldLengthThreshold,
@@ -211,17 +216,36 @@ public class FlinkOperatorConfiguration {
     }
 
     private static GenericRetry getRetryConfig(Configuration conf) {
-        return new GenericRetry()
-                .setMaxAttempts(
-                        conf.getInteger(
-                                KubernetesOperatorConfigOptions.OPERATOR_RETRY_MAX_ATTEMPTS))
-                .setInitialInterval(
-                        conf.get(KubernetesOperatorConfigOptions.OPERATOR_RETRY_INITIAL_INTERVAL)
-                                .toMillis())
-                .setIntervalMultiplier(
-                        conf.getDouble(
-                                KubernetesOperatorConfigOptions
-                                        .OPERATOR_RETRY_INTERVAL_MULTIPLIER));
+        var genericRetry =
+                new GenericRetry()
+                        .setMaxAttempts(
+                                conf.getInteger(
+                                        KubernetesOperatorConfigOptions
+                                                .OPERATOR_RETRY_MAX_ATTEMPTS))
+                        .setInitialInterval(
+                                conf.get(
+                                                KubernetesOperatorConfigOptions
+                                                        .OPERATOR_RETRY_INITIAL_INTERVAL)
+                                        .toMillis())
+                        .setIntervalMultiplier(
+                                conf.getDouble(
+                                        KubernetesOperatorConfigOptions
+                                                .OPERATOR_RETRY_INTERVAL_MULTIPLIER));
+
+        if (conf.contains(KubernetesOperatorConfigOptions.OPERATOR_RETRY_MAX_INTERVAL)) {
+            genericRetry.setMaxInterval(
+                    conf.get(KubernetesOperatorConfigOptions.OPERATOR_RETRY_MAX_INTERVAL)
+                            .toMillis());
+        } else {
+            genericRetry.withoutMaxInterval();
+        }
+        return genericRetry;
+    }
+
+    private static RateLimiter<?> getRateLimiter(Configuration conf) {
+        return new LinearRateLimiter(
+                conf.get(KubernetesOperatorConfigOptions.OPERATOR_RATE_LIMITER_PERIOD),
+                conf.get(KubernetesOperatorConfigOptions.OPERATOR_RATE_LIMITER_LIMIT));
     }
 
     private static LeaderElectionConfiguration getLeaderElectionConfig(Configuration conf) {
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
@@ -362,24 +362,44 @@ public class KubernetesOperatorConfigOptions {
             operatorConfig("retry.initial.interval")
                     .durationType()
                     .defaultValue(Duration.ofSeconds(5))
-                    .withDescription(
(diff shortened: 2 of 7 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
```diff
@@ -365,22 +365,45 @@ public static String operatorConfigKey(String key) {
...
+                    .defaultValue(Duration.ofSeconds(15))
+                    .withDescription("Operator rate limiter refresh period for each resource.");
+
+    @Documentation.Section(SECTION_SYSTEM)
+    public static final ConfigOption<Integer> OPERATOR_RATE_LIMITER_LIMIT =
+            operatorConfig("rate-limiter.limit")
+                    .intType()
+                    .defaultValue(5)
+                    .withDescription(
+                            "Max number of reconcile loops triggered in within the rate limiter refresh period for each resource. Setting the limit <= 0 disables the limiter.");
```
mbalassi, line 405: nit: `triggered within`

mbalassi (approved):
Thanks, great. Could you please explain the default behavior in plain English now given the default values you propose? (How many request would we send/resources/sec)

gyfora:
> Thanks, great. Could you please explain the default behavior in plain English now given the default values you propose? (How many request would we send/resources/sec)

In plain English the default is:
At most 5 reconciles for a given resource every 15 seconds. This should be more than enough to not have any visible effect on the normal behaviour (upgrades would go through instantaneously as they need only 2/3 reconciles)

But if it gets stuck in an error loop, it will only do 5 loops, then wait for 15 seconds, rinse and repeat

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
```diff
@@ -365,22 +365,45 @@ public static String operatorConfigKey(String key) {
                     .withDescription(
                             "Initial interval of automatic reconcile retries on recoverable errors.");
 
+    @Documentation.Section(SECTION_SYSTEM)
+    public static final ConfigOption<Duration> OPERATOR_RETRY_MAX_INTERVAL =
+            operatorConfig("retry.max.interval")
+                    .durationType()
+                    .defaultValue(Duration.ZERO)
+                    .withDescription(
+                            "Max interval of automatic reconcile retries on recoverable errors, if set to <=0, no limit is applied.");
```
mxm, line 374: How are recoverable errors defined?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
```diff
@@ -365,22 +365,45 @@ public static String operatorConfigKey(String key) {
...
+
     @Documentation.Section(SECTION_SYSTEM)
     public static final ConfigOption<Double> OPERATOR_RETRY_INTERVAL_MULTIPLIER =
             operatorConfig("retry.interval.multiplier")
                     .doubleType()
-                    .defaultValue(2.0)
+                    .defaultValue(1.5)
```
mxm, line 380: Why are we changing this value here? It's unrelated to the rate limiting

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
```diff
@@ -365,22 +365,45 @@ public static String operatorConfigKey(String key) {
...
                     .withDescription(
                             "Interval multiplier of automatic reconcile retries on recoverable errors.");
 
     @Documentation.Section(SECTION_SYSTEM)
     public static final ConfigOption<Integer> OPERATOR_RETRY_MAX_ATTEMPTS =
             operatorConfig("retry.max.attempts")
                     .intType()
-                    .defaultValue(10)
+                    .defaultValue(15)
```
mxm, line 388: Same here.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
```diff
@@ -365,22 +365,45 @@ public static String operatorConfigKey(String key) {
...
                     .withDescription(
                             "Max attempts of automatic reconcile retries on recoverable errors.");
 
+    @Documentation.Section(SECTION_SYSTEM)
+    public static final ConfigOption<Duration> OPERATOR_RATE_LIMITER_PERIOD =
+            operatorConfig("rate-limiter.refresh-period")
+                    .durationType()
+                    .defaultValue(Duration.ofSeconds(15))
+                    .withDescription("Operator rate limiter refresh period for each resource.");
```
mxm, line 397:
```suggestion
            operatorConfig("rate-limiter.period")
                    .durationType()
                    .defaultValue(Duration.ofSeconds(15))
                    .withDescription("Operator rate limiter period for each resource.");
```

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkOperatorConfiguration.java
```diff
@@ -220,8 +227,14 @@ private static GenericRetry getRetryConfig(Configuration conf) {
                                 .toMillis())
                 .setIntervalMultiplier(
                         conf.getDouble(
-                                KubernetesOperatorConfigOptions
-                                        .OPERATOR_RETRY_INTERVAL_MULTIPLIER));
+                                KubernetesOperatorConfigOptions.OPERATOR_RETRY_INTERVAL_MULTIPLIER))
+                .setMaxInterval(maxInterval > 0 ? maxInterval : Long.MAX_VALUE);
```
mxm, line 231: The default value `Long.MAX_VALUE` means, retry as often as you like? Any value than that means wait at least that amount of time in milliseconds?

mxm: It's a bit tricky because the JOSDK treats Long.MAX_VALUE as unset...

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
mxm: I guess it means, that we refresh 15 seconds after hitting the limit. Current name makes sense then.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
mxm: AFAIK the current default wasn't modified by adding this configuration option, correct?

mxm (approved):
Generally, looks good. Just some clarifications above.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkOperatorConfiguration.java
gyfora: I don't really see this MAX_VALUE behaviour that you mention, in any case I changed this to -1 which is the right way to set infinite

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
gyfora: Previously it was infinite, and a recent change actually introduced a 15sec limit accidentally. This is a correction of that. So with this, no change since the last release :)

gyfora: As for the "recoverable error", I just lifted the wording from the other config. It basically means any error that is not handled by the operator logic directly but is "thrown" and therefore should be retried :)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
gyfora: Based on some testing I felt that the current 2 multiplier for the exponential backoff was a bit too aggressive in some cases. So 1.5 with a little longer max retries felt more appropriate. If you feel we shouldn't change this I remove it from the PR

gyfora:
@mxm I have updated the PR based on the comments, also simplified wording for the retry related configs. PTAL

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkOperatorConfiguration.java
mxm: See here for default `UNSET_VALUE`: https://github.com/operator-framework/java-operator-sdk/blob/a3fb3ad9ef84ad60e456ac2b3714b070b37e055b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/retry/GenericRetry.java#L103 

which is set to `Long.MAX_VALUE`: https://github.com/operator-framework/java-operator-sdk/blob/a3fb3ad9ef84ad60e456ac2b3714b070b37e055b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/retry/GradualRetry.java#L19

mxm: If set to `Long.MAX_VALUE`, this default will be applied: https://github.com/operator-framework/java-operator-sdk/blob/a3fb3ad9ef84ad60e456ac2b3714b070b37e055b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/retry/GradualRetry.java#L16

gyfora: Oh I see what you meant, this wouldn't affect us as we are not using the `@GradualRetry` annotation, that is specific to that. In any case I removed this to be less confusing

PR #670, opened 2023-09-12
[FLINK-31871] Interpret Flink MemoryUnits according to the actual user input

srpraneeth:
JM and TM memory settings is only supported in bibytes format. This change will help interpreting the flink memory according to the actual user input.

## Brief change log

- Added the implementation 
- Added tests

## Verifying this change

## Documentation

Example values for the change

The below table property applies for the 

- jobManager.resource.memory
- taskManager.resource.memory

| Configuration | Previous InterpretedValue 			| New InterpretedValue 	|
|---------------|---------------------------------------|-----------------------|
| 2g            | 2147483648 b              			| 2000000000 b          |
| 2gb           | 2147483648 b              			| 2000000000 b          |
| 2G            | 2147483648 b              			| 2000000000 b          |
| 2 g           | 2147483648 b              			| 2000000000 b 			|
| 512m          | 536870912 b							| 512000000 b           |
| 2gi           | Fail (Could not parse value '2gi') 	| 2147483648 b 			|
| 2Gi           | Fail (Could not parse value '2Gi')	| 2147483648 b          |
| 2gib          | Fail (Could not parse value '2gi') 	| 2147483648 b 			|
| 2 Gi          | Fail (Could not parse value '2 Gi')	| 2147483648 b 			|
| 512mi         | Fail (Could not parse value '512mi')	| 536870912 b 			|
| 1000000b      | 1000000b								| 1000000b b 			|
| 1000000 b     | 1000000b								| 1000000b b 			|
| 1e6 b     	| Fail (Could not parse value '1e6 b')  | 1000000b b 			|
| 1e6 		    | Fail (Could not parse value '1e6')    | 1000000b b 			|

All these cases are covered with tests.

### DIFF
--- a/e2e-tests/data/flinkdep-cr.yaml
+++ b/e2e-tests/data/flinkdep-cr.yaml
@@ -72,7 +72,7 @@ spec:
       cpu: 0.5
   taskManager:
     resource:
-      memory: "1024m"
+      memory: "1Gi"
       cpu: 0.5
   job:
     jarURI: local:///opt/flink/usrlib/myjob.jar
--- a/e2e-tests/data/sessionjob-cr.yaml
+++ b/e2e-tests/data/sessionjob-cr.yaml
@@ -55,7 +55,7 @@ spec:
             claimName: session-cluster-1-pvc
   jobManager:
     resource:
-      memory: "1024m"
+      memory: "1Gi"
       cpu: 0.5
   taskManager:
     resource:
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
@@ -27,6 +27,7 @@ import org.apache.flink.configuration.DeploymentOptions;
 import org.apache.flink.configuration.DeploymentOptionsInternal;
 import org.apache.flink.configuration.HighAvailabilityOptions;
 import org.apache.flink.configuration.JobManagerOptions;
+import org.apache.flink.configuration.MemorySize;
 import org.apache.flink.configuration.PipelineOptions;
 import org.apache.flink.configuration.TaskManagerOptions;
 import org.apache.flink.kubernetes.configuration.KubernetesConfigOptions;
@@ -420,11 +421,33 @@ public class FlinkConfigBuilder {
                             ? JobManagerOptions.TOTAL_PROCESS_MEMORY
                             : TaskManagerOptions.TOTAL_PROCESS_MEMORY;
             if (resource.getMemory() != null) {
-                effectiveConfig.setString(memoryConfigOption.key(), resource.getMemory());
+                effectiveConfig.setString(
+                        memoryConfigOption.key(), parseResourceMemoryString(resource.getMemory()));
             }
 
             configureCpu(resource, effectiveConfig, isJM);
         }
+    }
+
+    // Using the K8s units specification for the JM and TM memory settings
+    public static String parseResourceMemoryString(String memory) {
+        try {
+            return MemorySize.parse(memory).toString();
+        } catch (IllegalArgumentException e) {
+            var memoryQuantity = formatMemoryStringForK8sSpec(memory);
+            return Quantity.parse(memoryQuantity).getNumericalAmount() + "";
+        }
+    }
+
+    private static String formatMemoryStringForK8sSpec(String memory) {
+        var memoryQuantity = memory.trim().replaceAll("\\s", "").toUpperCase();
+        if (memoryQuantity.endsWith("B")) {
+            memoryQuantity = memoryQuantity.substring(0, memoryQuantity.length() - 1);
+        }
+        if (memoryQuantity.endsWith("I")) {
+            memoryQuantity = memoryQuantity.substring(0, memoryQuantity.length() - 1) + "i";
+        }
+        return memoryQuantity;
     }
 
     private void configureCpu(Resource resource, Configuration conf, boolean isJM) {
--- a/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilderTest.java
+++ b/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilderTest.java
@@ -429,7 +429,7 @@ public class FlinkConfigBuilderTest {
                         .build();
 
         assertEquals(
-                MemorySize.parse("2048m"),
+                MemorySize.parse("2 gb"),
                 configuration.get(JobManagerOptions.TOTAL_PROCESS_MEMORY));
         assertEquals(Double.valueOf(1), configuration.get(KubernetesConfigOptions.JOB_MANAGER_CPU));
         assertEquals(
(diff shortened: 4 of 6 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
```diff
@@ -420,13 +421,24 @@ private void setResource(Resource resource, Configuration effectiveConfig, boole
...
             }
 
             configureCpu(resource, effectiveConfig, isJM);
         }
     }
 
+    // Using the K8s units specification for the JM and TM memory settings
+    private String parseResourceMemoryString(String memory) {
+        try {
+            return MemorySize.parse(memory).toString();
```
morhidi, line 435: Mmm in my head we should aim to support all memory configurations and notations kubernetes supports and convert them to Flinks proprietary settings unless those are specifically provided by Flink properties.

morhidi: @mxm wdyt?

morhidi:
Thanks for contributing @srpraneeth. Could you please update the PR and help understand the goal here with specific examples CR examples?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
morhidi: What are we handling with the current implementation exactly?

srpraneeth: Handled all the backward compatible cases. But the semantics of those cases itself will change due to the issue.
For ex: 
taskManager.resource.memory = 2g before the change means that the TM memory will be 2147483648 b but with this change it will be 2000000000 b.

Also added a table with the CR configuration, Previously Interpreted value, New Interpreted Value for all the cases handled. Please check and let me know if you think we need to handle more cases.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilderTest.java
```diff
@@ -478,6 +478,50 @@ public void testTaskManagerSpec() {
                 Double.valueOf(1), configuration.get(KubernetesConfigOptions.TASK_MANAGER_CPU));
     }
 
+    @Test
+    public void testApplyJobManagerSpecWithBiByteMemorySetting() {
```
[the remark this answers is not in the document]
srpraneeth: Added more tests. 
Please refer to the PR description for all list of handled cases and tests.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
srpraneeth: @morhidi The issue is that Flink interprets values only in Bibytes (gi, mi, ki) format. For instance, a simple value like '2g' is currently interpreted as '2gi = 2147483648 b' and there is no way to use both Bibyte and decimal byte formats as in K8s Spec.

This change will ensure that '2gi' is interpreted as '2g' accordingly, and the appropriate memory will be used. I assume that by 'backward change,' here means that the existing CR memory configuration should remain functional and not be disrupted, even though the values interpreted by the operator will change.

morhidi:
The last column looks good to me, we're not failing on any previous values which is important for backward compatibility reasons. I'm leaning towards accepting the corrected logic, while loosing the difference between `2147483648 b` vs `2000000000 b`, where users intention was probably was to have `2000000000 b` anyways.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilderTest.java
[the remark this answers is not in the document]
morhidi: Could you please break up that large test case to smaller ones? You can follow the table you put together in the description.

morhidi:
What happens when users overwrite the configuration with standard memory parameters? Could you please handle these cases and cover with tests?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
gyfora: I am completely confused here. You force pushed the branch with a completely different logic which I think is not fully backward compatible in the sense that the same spec will lead to now different memory setting.

In this case it would have been nice not to force push it so we can see the 2 alternatives side by side (in 2 commits).

In any case I thought the cleanest solution would be:
```
try {
 // Parse like before with Flink memory utils
} catch (parse error) {
// Parse remaining cases that were previously uncovered
}
```
(and I think this was the original code in this PR)

In your example with `2g` your current code will silently reduce the available memory by 7%. I think 7% reduction may have some critical prod impact. I understand that it would be better to have this consistent with k8s than with Flink but  this is not something we can afford to do in a minor release.

morhidi: We should probably compare it to the original value, but anything greater than `5%` is large enough to consider not to break it with this change. I don't have a strong argue here tbh, we can improve this further with the upcoming GA release. There will be still confusion caused by the custom config overrides anyways, we can't completely fix that.

srpraneeth: @gyfora Sorry my bad I should not have force pushed. 

Just for the sake of understanding I am using the term Approach 1 with the current approach and Approach2 using try/catch as suggested by Gyula(My previous approach). 
The interpreted values for different configurations is listed below. 

| Configuration | Previous InterpretedValue 			| New InterpretedValue (Approach1) 	| New InterpretedValue (Approach2)  															  | 
|---------------|---------------------------------------|-----------------------------------|-------------------------------------------------------------------------------------------------|
| 2g            | 2147483648 b              			| 2000000000 b          			| 2147483648 b 																					  |
| 2gb           | 2147483648 b              			| 2000000000 b          			| Fail (Could not parse the syntax i by the MemorySize, could not parse the syntax g by K8s spec) |
| 2G            | 2147483648 b              			| 2000000000 b          			| 2147483648 b 																					  |
| 2 g           | 2147483648 b              			| 2000000000 b 						| 2147483648 b 																					  |
| 512m          | 536870912 b							| 512000000 b           			| 536870912 b 																					  |
| 2gi           | Fail (Could not parse value '2gi') 	| 2147483648 b 						| Fail (Could not parse the syntax i by the MemorySize, could not parse the syntax g by K8s spec) |
| 2Gi           | Fail (Could not parse value '2Gi')	| 2147483648 b          			| 2147483648 b 																					  |
| 2gib          | Fail (Could not parse value '2gi') 	| 2147483648 b 						| Fail (Could not parse the syntax i by the MemorySize, could not parse the syntax g by K8s spec) |
| 2 Gi          | Fail (Could not parse value '2 Gi')	| 2147483648 b 						| 2147483648 b 																					  |
| 512mi         | Fail (Could not parse value '512mi')	| 536870912 b 						| Fail (Could not parse the syntax i by the MemorySize, could not parse the syntax g by K8s spec) |
| 1000000b      | 1000000b								| 1000000 b 						| 1000000 b 																					  |
| 1000000 b     | 1000000b								| 1000000 b 						| 1000000 b 																					  |
| 1e6 b     	| Fail (Could not parse value '1e6 b')  | 1000000 b 						| Fail (Could not parse the syntax i by the MemorySize, could not parse the syntax g by K8s spec) |
| 1e6 		    | Fail (Could not parse value '1e6')    | 1000000 b 						| 1000000 b 																					  |


The problem with the Approach 2 is the Flink MemorySize interprets the value incorrectly.
So if user try to configure 2g or 2G then the flink memorysize will parse it successfully but would interpret it as 2 gibibytes instead of 2 Gigabytes. The only thing that this cant parse is the configurations with i syntax like 2Gi and when K8s Quantity.parse would parse this it will interpret again it as 2 gibibytes and because of this problem the user would never be able to configure the 2 gigabytes (2000000000 b) with Approach2.
So both the try and catch part of the code would result in same interpretation of the values which is Bi bytes. 


So I dont see a way to make this change happen without breaking users or coming up with another syntax for configuring the memory in decimal bytes.
But I do agree with @gyfora 7% difference in the allocation may create issues for users.

gyfora: Even if Flink memoryutils interpret the decimal bytes incorrectly, we should not silently change the interpretation as it may cause failure in currently deployed prod jobs. 

We can fix this when we change "major version" ie bump the crd to v1. For now I think it is fine to interpret the current configs as before (even if technically incorrect at times) and open a follow up JIRA to fix this for v1.

Wdyt?

srpraneeth: I agree with that approach, rather than breaking user jobs silently making these changes, it better to do it as part of a breaking change in CR.

gyfora (commented):
@srpraneeth could you please change one of the e2e CRs to use the newly available memory format somewhere?

That will test the validation logic and that this doesn't get blocked by some incorrect validation in the future.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
srpraneeth: After addressing the comments and using the Quantity as fall back logic, below is the behaviour for different cases. 
Please review the same

| Configuration | Previous InterpretedValue 			| New InterpretedValue (Approach2) 	|
|---------------|---------------------------------------|-----------------------------------|
| 2g            | 2147483648 b              			| 2147483648 b          			|
| 2gb           | 2147483648 b              			| 2147483648 b          			|
| 2G            | 2147483648 b              			| 2147483648 b          			|
| 2 g           | 2147483648 b              			| 2147483648 b 						|
| 512m          | 536870912 b							| 536870912 b           			|
| 2gi           | Fail (Could not parse value '2gi') 	| 2147483648 b 						|
| 2Gi           | Fail (Could not parse value '2Gi')	| 2147483648 b          			|
| 2gib          | Fail (Could not parse value '2gi') 	| 2147483648 b 						|
| 2 Gi          | Fail (Could not parse value '2 Gi')	| 2147483648 b 						|
| 512mi         | Fail (Could not parse value '512mi')	| 536870912 b 						|
| 1000000b      | 1000000b								| 1000000 b 						|
| 1000000 b     | 1000000b								| 1000000 b 						|
| 1e6 b     	| Fail (Could not parse value '1e6 b')  | 1000000 b 						|
| 1e6 		    | Fail (Could not parse value '1e6')    | 1000000 b 						|


The only way for user to configure the giga(decimal) bytes is to convert into the numerical bytes and then use it.
For ex: 2G = 2 * 1000 * 1000 * 1000 b = 2000000000 b

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilderTest.java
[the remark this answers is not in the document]
srpraneeth: Fixed

e2e-tests/data/flinkdep-cr.yaml
```diff
@@ -72,7 +72,7 @@ spec:
       cpu: 0.5
   taskManager:
     resource:
-      memory: "1024m"
+      memory: "2Gi"
```
gyfora, line 75: Can we maybe use `1Gi` to not create too big pods?

e2e-tests/data/sessionjob-cr.yaml
```diff
@@ -55,7 +55,7 @@ spec:
             claimName: session-cluster-1-pvc
   jobManager:
     resource:
-      memory: "1024m"
+      memory: "512Mi"
```
gyfora, line 58: Can we maybe use `1Gi` to stay closer to what we had before?

gyfora (commented):
Looks good, some last minor comments :)

PR #675, opened 2023-09-18
[hotfix] Removing the obsolete comment of JobAutoScaler#scale

1996fanrui:

After FLINK-33081, the `JobAutoScaler#scale` doesn't return value anymore, so the comment can be updated.

## Brief change log

[hotfix] Removing the obsolete comment of `JobAutoScaler#scale`


gyfora:
Could we include this in some other commit instead of having it separately?

1996fanrui:
> Could we include this in some other commit instead of having it separately?

Sure, I will do it in the subsequent commit. I will close this PR.

PR #677, opened 2023-09-18
[FLINK-33097][autoscaler] Initialize the generic autoscaler module and interfaces

1996fanrui:

The first subtask of [FLIP-334](https://cwiki.apache.org/confluence/x/x4qzDw).

## Brief change log

- [FLINK-33097][autoscaler] Initialize the generic autoscaler module and interfaces
  - User will re-implement these `EventHandler`, `StateStore` and `ScalingRealizer`, so mark these classes to the `@Experimental`.
  - User will use the `Flink-Kubenetes-Operator` or `StandaloneAutoscaler`, and don't use the `JobAutoScaler` directly, so  mark `JobAutoScaler` as the `@Internal`.

- [FLINK-33098][autoscaler] Move non-kubernetes related autoscaler classes to flink-autoscaler module
- [FLINK-33098][autoscaler] Support kubernetes autoscaler using generic interface
  - Migrate all generic classes from flink-kubernetes-operator-autoscaler module to autoscaler moudle
  - Migrate all kubernetes implementation classes from flink-kubernetes-operator-autoscaler module to flink-kubernetes-operator moudle
  - Support all old tests
  - Removing the flink-kubernetes-operator-autoscaler module
- [FLINK-33098][autoscaler] Removing the operator prefix for all autoscaler options
  - Removing the option prefix(kubernetes.operator.) for all options and update the doc(All old option names are marked with withDeprecatedKeys to ensure the compatibility.)

## Verifying this change

All old tests have been supported.

  - The public API: yes

## Documentation


flink-autoscaler/src/main/java/org/apache/flink/autoscaler/state/AutoScalerStateStore.java
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * The state store is responsible for store all states during scaling.
+ *
+ * @param <KEY> The job key.
+ * @param <Context> Instance of JobAutoScalerContext.
+ */
+@Experimental
+public interface AutoScalerStateStore<KEY, Context extends JobAutoScalerContext<KEY>> {
+
+    void storeScalingHistory(Context jobContext, String scalingHistory);
```
1996fanrui, line 32: Hi @gyfora  @mxm , because of some offline comments and ease of review, I split the decoupling into the FLINK-33097 and FLINK-33098. Of course, if you think they should be merged to one PR. I will go ahead at this PR with multiple commits.

@Samrat002 provided 2 improvements about the `StateStore`.

## 1. Using the structured class instead of `String`

The structured class is clearer than String.

First of all, we define the `SerializableState` interface to abstract the `serialize` and `deserialize`.

```
    interface SerializableState<State extends SerializableState> {

        String serialize();

        State deserialize(String serializedResult);
    }
```

And then, define a `ScalingHistory` class, it implement the `SerializableState`.


## 2. Using the `byte[]` instead of `String` as the serialized result

Reason: In the future there may be some complex state objects that cannot be serialized to String.


Hi @Samrat002 , please correct me if my description is wrong, thanks~

1996fanrui: > improvement1

The improvement1 sounds make sense to me, and I need to improve the `SerializableState`. The current `deserialize` method needs to create a object first, and then call `object.deserialize(serializedResult)`. In general, the serialize is a separate object.

I'm afraid whether it's too complex If we introduce 2 class for each state.

> improvement2

For improvement2, my concern is the serialized type is changed, and all old jobs cannot be compatible directly. 

The compatibility of `byte[]` must be stronger than String, but the benefits it brings are uncertain (because there may not be classes that can only be serialized into `byte[]` in the future).

The negative impact is certain, and it will bring additional migration costs to historical users.

Samrat002: Much appreciated, @1996fanrui , for summarizing our discussion.

I would like to add one more point: the `AutoScalerEventHandler` currently offers two types, namely `warning` and `normal`. In my opinion, it would be beneficial to include an `Error` or `Fatal` option as part of the interface. This flexibility would allow different users implementing this autoscaler module to define and use these options according to their specific requirements.

I would definitely appreciate hearing the thoughts of @mxm  and @gyfora  regarding the adoption of structured classes over string and the proposal to introduce a new type called Fatal in `AutoScalerEventHandler`, as described in this thread.

1996fanrui: Hi @Samrat002 , thanks for your feedback!

I have updated these method parameters of `AutoScalerStateStore` to the specific class instead of String, such as: `Map<JobVertexID, SortedMap<Instant, ScalingSummary>> scalingHistory`.

```
public interface AutoScalerStateStore<KEY, Context extends JobAutoScalerContext<KEY>> {

    void storeScalingHistory(
            Context jobContext,
            Map<JobVertexID, SortedMap<Instant, ScalingSummary>> scalingHistory);

    Optional<Map<JobVertexID, SortedMap<Instant, ScalingSummary>>> getScalingHistory(
            Context jobContext);

    void removeScalingHistory(Context jobContext);
}
```

The PR has been updated as well.

Do you think is it ok? It means the state store is responsible for how to serialize and deserialize, for example:

- The default `KubernetesAutoScalerStateStore` will serialize all states to String inside of `KubernetesAutoScalerStateStore`
- As you mentioned before: if there is any complex type in the future. Each state store to determine how to serialize them.

Also, let me add a reason why update these parameters here:

Currently, all states are stored at ConfigMap, and it has size limitation. The size limitation should just work with `KubernetesAutoScalerStateStore`, and size limitation is a part of serialization. So we should move the serialization and deserialization in the `AutoScalerStateStore`.

XComp (commented):
Hi @1996fanrui, sorry for commenting on your draft PR. I was just curious about the `RestClusterClient` usage (since it's also documented in FLIP-334). I found this PR to be the better location for the question because the ML thread discussion is already closed. See my comment below.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerContext.java
```diff
@@ -0,0 +1,44 @@
...
+    /** The identifier of each flink job. */
+    KEY getJobKey();
+
+    JobID getJobID();
+
+    Configuration getConfiguration();
+
+    MetricGroup getMetricGroup();
+
+    RestClusterClient<String> getRestClusterClient() throws Exception;
```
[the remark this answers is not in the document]
1996fanrui: >  I am wondering why we're relying on `RestClusterClient` here? Shouldn't we use the `ClusterClient` interface, instead, to keep it generic and improve testability?

After I check, `ClusterClient` has only some commonly used interfaces, such as: `submitJob`, `getJobStatus`, `cancelJob` and a series of savepoint releated interfaces, etc.

Autoscaler is using a couple of interfaces that request rest api, such as: 

- [getJobTopology](https://github.com/apache/flink-kubernetes-operator/blob/662fa612a8ab352e43ab8a99fa61aadfbe41e4d7/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/ScalingMetricCollector.java#L106) 
- [getJobDetailsInfo](https://github.com/apache/flink-kubernetes-operator/blob/662fa612a8ab352e43ab8a99fa61aadfbe41e4d7/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/ScalingMetricCollector.java#L97)
- [Query generic metrics ](https://github.com/apache/flink-kubernetes-operator/blob/662fa612a8ab352e43ab8a99fa61aadfbe41e4d7/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/ScalingMetricCollector.java#L433)

Also, the new `rescale rest api (FLIP-291)`[1] will be introduced in this FLIP. That's why we using the `RestClusterClient` here. 

These information cannot be fetched by `ClusterClient`, right?

Please correct me if I misunderstand here, thanks a lot.

[1] https://cwiki.apache.org/confluence/display/FLINK/FLIP-291%3A+Externalized+Declarative+Resource+Management

1996fanrui (commented):
Hi @XComp , thanks for your comments here.

I have replied your comments, please take a look in your free time, thanks~

1996fanrui (commented):
Hi @gyfora @mxm , this PR is ready to review, please help review in your free time, thanks~

The third commit is that support kubernetes autoscaler using generic interface, it's really a huge commit. And it's hard to split to multiple commits if we want to ensure each commits can be compiled.

mateczagany (changes requested):
Hi, thank you for this PR, I think it's a great idea to share the ability to auto-scale jobs to YARN users. It's quite a big PR so I could not go through it all yet, but I've found 2 issues that I think we should look into.

I will try to take a more extensive look at this PR next week.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
```diff
@@ -174,6 +182,32 @@ public void reconcile(FlinkResourceContext<CR> ctx) throws Exception {
...
+
+        if (autoscalerDisabled(ctx)) {
+            autoScalerContext.getConfiguration().set(AUTOSCALER_ENABLED, false);
+            resourceScaler.scale(autoScalerContext);
+            return;
+        }
+        if (waitingForRunning(ctx.getResource().getStatus())) {
+            LOG.info("Autoscaler is waiting for  stable, running state");
+            resourceScaler.cleanup(autoScalerContext.getJobKey());
+            return;
```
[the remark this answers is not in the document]
1996fanrui: Thanks for pointing it out.

If the `applyParallelismOverrides` is expected here, we can call the related code here. The `applyParallelismOverrides` just calls the `stateStore.getParallelismOverrides(ctx)` and `scalingRealizer.realize(ctx, userOverrides);`, and `stateStore` and `scalingRealizer` can be reached here. 

I plan to extract the `org.apache.flink.autoscaler.JobAutoScalerImpl#applyParallelismOverrides` to a static method, WDYT?

flink-autoscaler/pom.xml
```diff
@@ -45,6 +45,32 @@ under the License.
...
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-params</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <!-- TODO FLINK-33098: These jackson dependencies can be replaced with flink shaded jackson. It can be done
```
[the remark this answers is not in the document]
gyfora: I think we should use the shaded dependency that comes with the Flink client version we are depending on. That is the simplest way to avoid any potential issues :)

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/AbstractJobAutoScalerContext.java
```diff
@@ -0,0 +1,80 @@
...
+import org.apache.flink.configuration.Configuration;
+import org.apache.flink.metrics.MetricGroup;
+import org.apache.flink.util.function.SupplierWithException;
+
+/**
+ * The abstract job autoscaler context.
+ *
+ * @param <KEY> The job key.
+ */
+public abstract class AbstractJobAutoScalerContext<KEY> implements JobAutoScalerContext<KEY> {
```
gyfora, line 31: I would simply rename this as `JobAutoScalerContext<KEY>`, delete constructor , make all fields simply final and put `@Value` on it, to generate all the getters, toString etc.

We don't need to have interfaces for everything, too much boilerplate

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/DefaultJobAutoScalerContext.java
```diff
@@ -0,0 +1,36 @@
...
+package org.apache.flink.autoscaler;
+
+import org.apache.flink.api.common.JobID;
+import org.apache.flink.client.program.rest.RestClusterClient;
+import org.apache.flink.configuration.Configuration;
+import org.apache.flink.metrics.MetricGroup;
+import org.apache.flink.util.function.SupplierWithException;
+
+/** The default job autoscaler context, the jobKey is JobID. */
+public class DefaultJobAutoScalerContext extends AbstractJobAutoScalerContext<JobID> {
```
gyfora, line 27: Let's simply delete the class and simply have a non-abstract `JobAutoscalerContext<JobId>`

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerContext.java
```diff
@@ -0,0 +1,44 @@
...
+import org.apache.flink.configuration.Configuration;
+import org.apache.flink.metrics.MetricGroup;
+
+/**
+ * The job autoscaler context, it includes all details related to the current job.
+ *
+ * @param <KEY> The job key.
+ */
+@Experimental
+public interface JobAutoScalerContext<KEY> {
```
gyfora, line 32: Let's delete the interface for now (see my previous comment) we can re-add this later if necessary but I doubt it.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/state/HeapedAutoScalerStateStore.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.SortedMap;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * The state store based on the Java Heap.
+ *
+ * @param <KEY>
+ * @param <Context>
+ */
+public class HeapedAutoScalerStateStore<KEY, Context extends JobAutoScalerContext<KEY>>
```
gyfora, line 37: Let's rename this `HeapedAutoScalerStateStore` -> `InMemoryAutoScalerStateStore`

gyfora (changes requested):
I added some initial comment, to simplify the context. Let's try to keep the code simple and not introduce new interfaces unless absolutely necessary. And also please use the Lombok annotations so we can get rid of constructors, getters etc.

1996fanrui (commented):
> I added some initial comment, to simplify the context. Let's try to keep the code simple and not introduce new interfaces unless absolutely necessary. And also please use the Lombok annotations so we can get rid of constructors, getters etc.

Hi @gyfora , thanks for your review and good suggestions! I have updated all contexts according to your suggestions.

gyfora (commented):
The refactor also completely changes how the autoscaler module is loaded in the operator. Previously we intentionally loaded this as a service so it would be possible to potentially upgrade the autoscaler without the rest of the operator.

Now @1996fanrui you moved the operator related autoscaler classes in the main operator module resulting in a tight coupling. 

@mxm how do you feel about this? Originally you were the one who introduced the service loader approach :)

1996fanrui:
> The refactor also completely changes how the autoscaler module is loaded in the operator. Previously we intentionally loaded this as a service so it would be possible to potentially upgrade the autoscaler without the rest of the operator.
> 
> Now @1996fanrui you moved the operator related autoscaler classes in the main operator module resulting in a tight coupling.
> 
> @mxm how do you feel about this? Originally you were the one who introduced the service loader approach :)

Sorry for breaking this due to I don't know the background before. And thanks to @gyfora for pointing it out.

If it's still needed, I can restore the `flink-kubernetes-operator-autoscaler` module. And moving all kubernetes-autoscaler related classes to this module. WDYT?

cc @mxm

gyfora:
@1996fanrui Sorry I merged a PR related to event triggering from the ScalingExecutor. I will try to avoid merging further autoscaler related PRs until this is finalised. Please take care when rebasing

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/KubernetesJobAutoScalerContext.java
```diff
@@ -0,0 +1,62 @@
...
+import org.apache.flink.client.program.rest.RestClusterClient;
+import org.apache.flink.configuration.Configuration;
+import org.apache.flink.kubernetes.operator.api.AbstractFlinkResource;
+import org.apache.flink.metrics.MetricGroup;
+import org.apache.flink.util.function.SupplierWithException;
+
+import io.fabric8.kubernetes.client.KubernetesClient;
+import io.javaoperatorsdk.operator.processing.event.ResourceID;
+
+/** The kubernetes job autoscaler context. */
```
mxm, line 31:
```suggestion
/** An implementation of JobAutoscalerContext for Kubernetes. */
```

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/KubernetesScalingRealizer.java
```diff
@@ -17,24 +17,26 @@
...
+import org.apache.flink.autoscaler.realizer.ScalingRealizer;
+import org.apache.flink.configuration.ConfigurationUtils;
+import org.apache.flink.configuration.PipelineOptions;
 
-import com.google.auto.service.AutoService;
+import io.javaoperatorsdk.operator.processing.event.ResourceID;
+
+import java.util.Map;
+
+/** The kubernetes scaling realizer. */
```
mxm, line 28:
```suggestion
/** The Kubernetes implementation for applying parallelism overrides. */
```

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkResourceContext.java
```diff
@@ -47,6 +50,36 @@ public abstract class FlinkResourceContext<CR extends AbstractFlinkResource<?, ?
     private FlinkOperatorConfiguration operatorConfig;
     private Configuration observeConfig;
     private FlinkService flinkService;
+    private KubernetesJobAutoScalerContext autoScalerContext;
+
+    /**
+     * Get or create the autoscaler context.
+     *
+     * @return autoScalerContext.
+     */
```
mxm, line 59:
```suggestion
```

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
```diff
@@ -174,6 +182,32 @@ public void reconcile(FlinkResourceContext<CR> ctx) throws Exception {
...
+            autoScalerContext.getConfiguration().set(AUTOSCALER_ENABLED, false);
+            resourceScaler.scale(autoScalerContext);
+            return;
+        }
+        if (waitingForRunning(ctx.getResource().getStatus())) {
+            LOG.info("Autoscaler is waiting for stable, running state");
+            resourceScaler.cleanup(autoScalerContext.getJobKey());
+            return;
+        }
+        resourceScaler.scale(autoScalerContext);
```
mxm, line 198:
```suggestion
        if (autoscalerDisabled(ctx)) {
            autoScalerContext.getConfiguration().set(AUTOSCALER_ENABLED, false);
            resourceScaler.scale(autoScalerContext);
        } else if (waitingForRunning(ctx.getResource().getStatus())) {
            LOG.info("Autoscaler is waiting for stable, running state");
            resourceScaler.cleanup(autoScalerContext.getJobKey());
        } else {
            resourceScaler.scale(autoScalerContext);
        }
```

Much easier to read for me.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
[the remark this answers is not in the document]
mxm: We can internally reset the state, but the effective overrides on the deployment should not be cleared during the waiting period.

mxm:
> If it's still needed, I can restore the `flink-kubernetes-operator-autoscaler` module. And moving all kubernetes-autoscaler related classes to this module. WDYT?

Given that the autoscaler module is now decoupled from the operator version, it would make more sense than ever to have it pluggable. But given that the backends are not pluggable, the pluggability wouldn't be as useful anymore. I'm ok with removing the support for now if it proves difficult to maintain it.

flink-autoscaler/pom.xml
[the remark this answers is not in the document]
mxm: Yep, let's use the version bundled with the client.

mxm (commented):
Excellent work @1996fanrui! I've left some comments. All in all, this looks very solid to me. We will have to perform some manual testing post merge to ensure we don't hit any regressions.

flink-autoscaler/src/test/java/org/apache/flink/autoscaler/RestApiMetricsCollectorTest.java
```diff
@@ -0,0 +1,113 @@
...
+                        jobID,
+                        jobID,
+                        conf,
+                        new UnregisteredMetricsGroup(),
+                        () -> restClusterClient);
+
+        Map<JobVertexID, Map<FlinkMetric, AggregatedMetric>> jobVertexIDMapMap =
+                collector.queryAllAggregatedMetrics(context, metrics);
+
+        System.out.println(jobVertexIDMapMap);
```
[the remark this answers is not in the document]
1996fanrui: Sorry, what's the `left-over` here? I just copy it from old code. 😂

flink-autoscaler/src/test/java/org/apache/flink/autoscaler/RecommendedParallelismTest.java
```diff
@@ -217,73 +198,38 @@ public void endToEnd() throws Exception {
...
-    private void restart(Instant now) {
-        metricsCollector.setJobUpdateTs(now);
-        app.getStatus().getJobStatus().setState(JobStatus.CREATED.name());
-    }
-
-    private void running(Instant now) {
-        metricsCollector.setJobUpdateTs(now);
-        app.getStatus().getJobStatus().setState(JobStatus.RUNNING.name());
-    }
-
```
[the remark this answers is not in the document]
1996fanrui: This comment is a same question with the previous comment. I reply it together.

The reason is the `waitingForRunning` logic is moved from `autoscaler` module to `org.apache.flink.kubernetes.operator.reconciler.deployment.AbstractFlinkResourceReconciler#waitingForRunning`[1].

This logic is not a part of generic `autoscaler` now. I will move this test to `flink-kubernetes-operator` module later.

[1] https://github.com/apache/flink-kubernetes-operator/pull/677#discussion_r1341955135

1996fanrui (commented):
Hi @mxm , thanks a lot for your detailed review! I have address your comments.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
```diff
@@ -72,7 +79,7 @@ public abstract class AbstractFlinkResourceReconciler<
 
     protected final EventRecorder eventRecorder;
     protected final StatusRecorder<CR, STATUS> statusRecorder;
-    protected final JobAutoScaler resourceScaler;
+    protected final JobAutoScaler<ResourceID, KubernetesJobAutoScalerContext> resourceScaler;
```
mxm, line 82: Can we rename this?

```suggestion
    protected final JobAutoScaler<ResourceID, KubernetesJobAutoScalerContext> autoscaler;
```

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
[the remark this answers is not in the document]
mxm: Preferably, I would like any logic related to applying parallelism inside the autoscaler implementation. This shouldn't change when the autoscaler is waiting for the running state. In fact, the job state checks should also be performed by the autoscaler, not by the reconciler. The current code mixes control over the parallelism overrides between the reconciler and the autoscaler.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
1996fanrui: done

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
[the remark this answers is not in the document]
1996fanrui: Thanks to @mxm for the suggestion!

I have moved the `waitingForRunning` logic back to autoscaler module, and related tests have been recovered.

Also, these 2 comments[1][2] have been recovered as well.

[1] https://github.com/apache/flink-kubernetes-operator/pull/677#discussion_r1350082356
[2] https://github.com/apache/flink-kubernetes-operator/pull/677#discussion_r1350079573

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkResourceContext.java
```diff
@@ -47,6 +55,48 @@ public abstract class FlinkResourceContext<CR extends AbstractFlinkResource<?, ?
...
+    private JobStatus generateJobStatusEnum(CommonStatus<?> status) {
+        if (status.getLifecycleState() != ResourceLifecycleState.STABLE) {
+            return null;
+        }
+
+        String state = status.getJobStatus().getState();
+        if (state == null) {
+            return null;
+        }
+        return JobStatus.valueOf(state);
```
1996fanrui, line 98: Hi @mxm  @gyfora , could the type of `org.apache.flink.kubernetes.operator.api.status.JobStatus#state` be changed from `String` to `org.apache.flink.api.common.JobStatus`?

I see all setters respect it. If it can be changed, the `JobStatus.valueOf(state)` isn't necessary.

gyfora: Let's not change this as part of this PR but feel free to open a JIRA for this

1996fanrui: Sure, I didn't change it at this PR, and I created [FLINK-33237](https://issues.apache.org/jira/browse/FLINK-33237) to optimize this after this PR~

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerContext.java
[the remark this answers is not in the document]
XComp: Sorry for responding with a delay due to vacation and other things. I guess, you're right. The `ClusterClient` interface doesn't cover all the methods that are provided by the `RestClusterClient`. It feels like we would have to revisit the `ClusterClient` to align things again. One workaround would be to provide a wrapper interface that only reveals the methods that are need by the Kubernetes Operator. WDYT?

My concern is that using a concrete implementation might make testing things harder.

Btw. have we thought of relying on the OpenAPI definition to generate a REST client?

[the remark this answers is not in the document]
gyfora: We intentionally avoided too many wrappers, generated clients etc whenever possible. Testing is not too bad and we would like to avoid the interface/factory hell that is currently present in core Flink :)

flink-autoscaler/pom.xml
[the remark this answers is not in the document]
XComp: hm, thinking about it: Doesn't we add an extra layer of complexity here? Let's say that there's a CVS in the dependency which requires us to update the version. Relying on Flink's shaded dependency would mean that we would have to wait for `flink-shaded` to be updated.

Taking one step back - isn't the question: Do we need to shade the jackson dependency?
If the answer is yes, shouldn't the Kubernetes Operator shade its dependency to be independent of `flink-shaded` releases and be able to react faster when having to update those dependencies?
If the answer is no, we could just use the dependency as is and don't need to worry about excluding anything from Flink (because its dependency is shaded already).

Its the same approach how we want to deal with it for external connectors as far as I understand. And if I understand the motivation correctly, shading should be necessary so that downstream projects (i.e. custom auto-scaler implementations) would be able to handle their own dependency versions. Does that make sense?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkResourceContext.java
mxm: Sure, if that does not cause any problems with the CRD state, that is fine with me.

mxm: (Didn't see your comments until I hit refresh). Fine to defer this.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
```diff
@@ -166,6 +197,13 @@ private static ConfigOptions.OptionBuilder autoScalerConfig(String key) {
...
                     .defaultValues()
+                    .withDeprecatedKeys(deprecatedOperatorConfigKey("vertex.exclude.ids"))
                     .withDescription(
                             "A (semicolon-separated) list of vertex ids in hexstring for which to disable scaling. Caution: For non-sink vertices this will still scale their downstream operators until https://issues.apache.org/jira/browse/FLINK-31215 is implemented.");
+
+    public static final ConfigOption<Duration> FLINK_CLIENT_TIMEOUT =
+            autoScalerConfig("flink.client.timeout")
+                    .durationType()
+                    .defaultValue(Duration.ofSeconds(10))
+                    .withDescription("The timeout for waiting the flink rest client to return.");
```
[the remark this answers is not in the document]
mxm: Let's keep the existing option. We can move it here like you did but it should continue to support `flink.client.timeout`. We can add it as a deprecated key.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/event/AutoScalerEventHandler.java
```diff
@@ -0,0 +1,55 @@
...
+            Type type,
+            String reason,
+            String message,
+            @Nullable String messageKey,
+            @Nullable Duration interval);
+
+    /** The type of the events. */
+    enum Type {
+        Normal,
+        Error
```
mxm, line 53: Why not Normal, _Warning_, Error?

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
```diff
@@ -0,0 +1,210 @@
...
+                return;
+            }
+
+            if (ctx.getJobStatus() != JobStatus.RUNNING) {
+                lastEvaluatedMetrics.remove(ctx.getJobKey());
+                return;
+            }
+
+            runScalingLogic(ctx, autoscalerMetrics);
+            stateStore.flush(ctx);
```
mxm, line 98: This is not flushing anymore after line 88 and line 93 but it was before.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/metrics/ScalingHistoryUtils.java
```diff
@@ -0,0 +1,117 @@
...
+                            while (entryIt.hasNext()) {
+                                var entry = entryIt.next();
+                                // Limit how long past scaling decisions are remembered
+                                entry.setValue(
+                                        entry.getValue()
+                                                .tailMap(
+                                                        now.minus(
+                                                                conf.get(
+                                                                        AutoScalerOptions
+                                                                                .VERTEX_SCALING_HISTORY_AGE))));
```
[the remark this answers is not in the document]
mxm: I see! It's hard to tell what has been moved and what not. I was concerned that returning a submap here can cause issues. It's better to clear the part from the map we do not need anymore and return the full map.

flink-autoscaler/src/test/java/org/apache/flink/autoscaler/RestApiMetricsCollectorTest.java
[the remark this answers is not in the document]
mxm: We should probably remove the println anyways, even if we copy this over.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/metrics/ScalingHistoryUtils.java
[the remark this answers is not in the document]
1996fanrui: Thanks for the clarification. Updated it by new a TreeMap.

flink-autoscaler/pom.xml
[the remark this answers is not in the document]
1996fanrui: Thanks @XComp for the detailed analysis.

> Do we need to shade the jackson dependency?

IIUC, it's not required now. If it's required, `flink-kubernetes-operator` must have completed the shade before.

I prefer to use the shaded version in the future, and the motivation as your mention: we have custom auto-scaler implementations.

The flink shaded version should be enough, so I agree with @gyfora and @mxm : let's version bundled with the flink client. And we can re-shade it if it doesn't work well in the future. WDYT?


-----------------------------

The second question:

This PR still using the jackson version of `flink-kubernetes-operator` instead of flink shaded version. I want to update it after `flink-1.18` released.

The reason is : current autoscaler is using the `loaderOptions` to limit the serialized size. 

- The shaded jackson version of `flink-1.17` is `2.13.4-16.1`, it doesn't support the `loaderOptions`. 
- The shaded jackson version of `flink-1.18` is `2.14.2-17.0`, it supports the `loaderOptions`. 

What do you think about updating it after `flink-1.18` is released, it should be released soon. If it makes sense, I can create a JIRA to follow it.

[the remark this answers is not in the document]
XComp: There's gonna be another release candidate (RC2) by the end of the week according to the release managers. I hope that 1.18.0 will be released within October. 

Using the shaded jackson dependency from Flink (in contrast to shading the dependency within the Kubernetes operator) comes with the benefit that you don't have to maintain a dedicated entry in the NOTICE file, as far as I understand. So, it has "some advantage".

Anyway, the approach can be changed later on if it becomes a problem. So, feel free to follow what you agreed on in this PR.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
mxm: Looks good to me. Just wanted to double-check! Thank you.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/event/AutoScalerEventHandler.java
mxm: Fine with me. Thanks for the explanation!

1996fanrui: Thanks for the feedback, I have reverted the `Error` to `Warning` by a force push.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java
```diff
@@ -441,15 +442,21 @@ protected Collection<AggregatedMetric> queryAggregatedMetricNames(
...
-                    Configuration conf,
+                    Context ctx,
                     Map<JobVertexID, Map<String, FlinkMetric>> filteredVertexMetricNames);
 
-    public void cleanup(AbstractFlinkResource<?, ?> cr) {
-        var resourceId = ResourceID.fromResource(cr);
-        histories.remove(resourceId);
-        availableVertexMetricNames.remove(resourceId);
+    public JobDetailsInfo getJobDetailsInfo(
+            JobAutoScalerContext<KEY> context, Duration clientTimeout) throws Exception {
```
mateczagany, line 449: I don't expect it to be fixed in this PR, just want to note that this is an abstract class with `RestApiMetricsCollector` extending it. This makes it seem like it would be possible to extend this class to be able to collect metrics via methods other than REST API.

This would already be a challenge as there are several REST API calls already in this class, but with this PR, we will also tie this class to REST API with `JobAutoScalerContext`.

Perhaps it makes sense to remove `RestApiMetricsCollector` at this point?

mateczagany (changes requested):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/KubernetesJobAutoScalerContext.java
```diff
@@ -0,0 +1,67 @@
...
+
+/** An implementation of JobAutoscalerContext for Kubernetes. */
+public class KubernetesJobAutoScalerContext extends JobAutoScalerContext<ResourceID> {
+
+    private final AbstractFlinkResource<?, ?> resource;
+
+    private final KubernetesClient kubernetesClient;
+
+    public KubernetesJobAutoScalerContext(
+            JobID jobID,
```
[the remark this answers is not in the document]
1996fanrui: Thanks for pointing it out, updated.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java
1996fanrui: Thanks @mateczagany for this comment.

IIUC, you mean `ScalingMetricCollector` is using the `RestClusterClient`, and `RestApiMetricsCollector` is totally based on `RestClusterClient`, so these 2 classes can be merged into one classes, right?

If so, I try to explain the difference between : `RestApiMetricsCollector` and `ScalingMetricCollector`.

- `RestApiMetricsCollector` calls `RestClusterClient`, and it's used to fetch specific metrics.
- `ScalingMetricCollector` calls `RestClusterClient` and is not used to fetch specific metrics.
- `RestClusterClient` is used in `ScalingMetricCollector` to get some job metadata, such as: `getJobDetailsInfo` to generate the `JobTopology`, `queryFilteredMetricNames`, `updateKafkaSourceMaxParallelisms`.
- The JobTopology is the metadata of Job, and it cannot be fetched from metrics. That means the `RestClusterClient` is needed even if we query specific metrics from other system.

Based on them, it may be better to keep `ScalingMetricCollector` as abstract class and not remove `RestApiMetricsCollector`. It's easy to fetch specific metrics from other system in the future.

Also, we can see the `ScalingMetricCollector` also used the `RestClusterClient` on the current master branch.

WDYT? And please correct me if my understanding is wrong, thanks~

[1] https://github.com/apache/flink-kubernetes-operator/blob/305498a9ab2e04ab71a4c2d87f2edb746373df1a/flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/ScalingMetricCollector.java#L368C45-L368C45

1996fanrui:
Hi @mxm @gyfora , thanks for your hard review! 

I addressed all comments of this PR so now. Some interfaces, code design and code style are improved, and no big changes during reviewing, I have updated the latest interfaces to the FLIP doc.

About manual testing, I rebuilt the image to check whether it's fine just now, and run the `AutoscalingExample` with `autoscaling.yaml`. In general, it works well. There are some warn log due to these demo job still use the deprecated configuration key, and I have updated them in the last fix commit. 

Please help review again in your free time, big thanks~ ❤️ 

I can squash the last commit after your check. Also, please let me know if you think more manual testing is needed, thanks.

```
2023-10-17 15:33:56,930 o.a.f.c.Configuration          [WARN ] [default.autoscaling-example] Config uses deprecated configuration key 'kubernetes.operator.job.autoscaler.enabled' instead of proper key 'job.autoscaler.enabled'
2023-10-17 15:33:56,931 o.a.f.c.Configuration          [WARN ] [default.autoscaling-example] Config uses deprecated configuration key 'kubernetes.operator.job.autoscaler.enabled' instead of proper key 'job.autoscaler.enabled'
2023-10-17 15:33:57,205 o.a.f.a.ScalingMetricCollector [INFO ] [default.autoscaling-example] Job updated at 2023-10-17T07:32:56.133Z. Clearing metrics.
2023-10-17 15:33:57,331 o.a.f.c.Configuration          [WARN ] [default.autoscaling-example] Config uses deprecated configuration key 'kubernetes.operator.job.autoscaler.metrics.window' instead of proper key 'job.autoscaler.metrics.window'
2023-10-17 15:33:57,332 o.a.f.c.Configuration          [WARN ] [default.autoscaling-example] Config uses deprecated configuration key 'kubernetes.operator.job.autoscaler.stabilization.interval' instead of proper key 'job.autoscaler.stabilization.interval'
2023-10-17 15:33:57,397 o.a.f.a.ScalingMetricCollector [WARN ] [default.autoscaling-example] pendingRecords metric for cbc357ccb763df2852fee8c4fc7d55f2 could not be found. Either a legacy source or an idle source. Assuming no pending records.
2023-10-17 15:33:57,620 o.a.f.a.ScalingMetricCollector [INFO ] [default.autoscaling-example] Metric window not full until 2023-10-17T07:36:56.931915Z
2023-10-17 15:33:57,623 o.a.f.c.Configuration          [WARN ] [default.autoscaling-example] Config uses deprecated configuration key 'kubernetes.operator.job.autoscaler.target.utilization.boundary' instead of proper key 'job.autoscaler.target.utilization.boundary'
2023-10-17 15:33:57,624 o.a.f.c.Configuration          [WARN ] [default.autoscaling-example] Config uses deprecated configuration key 'kubernetes.operator.job.autoscaler.target.utilization.boundary' instead of proper key 'job.autoscaler.target.utilization.boundary'
```

gyfora:
> Hi @mxm @gyfora , thanks for your hard review!
> 
> I addressed all comments of this PR so now. Some interfaces, code design and code style are improved, and no big changes during reviewing, I have updated the latest interfaces to the FLIP doc.
> 
> About manual testing, I rebuilt the image to check whether it's fine just now, and run the `AutoscalingExample` with `autoscaling.yaml`. In general, it works well. There are some warn log due to these demo job still use the deprecated configuration key, and I have updated them in the last fix commit.
> 
> Please help review again in your free time, big thanks~ ❤️
> 
> I can squash the last commit after your check. Also, please let me know if you think more manual testing is needed, thanks.
> 
> ```
> 2023-10-17 15:33:56,930 o.a.f.c.Configuration          [WARN ] [default.autoscaling-example] Config uses deprecated configuration key 'kubernetes.operator.job.autoscaler.enabled' instead of proper key 'job.autoscaler.enabled'
> 2023-10-17 15:33:56,931 o.a.f.c.Configuration          [WARN ] [default.autoscaling-example] Config uses deprecated configuration key 'kubernetes.operator.job.autoscaler.enabled' instead of proper key 'job.autoscaler.enabled'
> 2023-10-17 15:33:57,205 o.a.f.a.ScalingMetricCollector [INFO ] [default.autoscaling-example] Job updated at 2023-10-17T07:32:56.133Z. Clearing metrics.
> 2023-10-17 15:33:57,331 o.a.f.c.Configuration          [WARN ] [default.autoscaling-example] Config uses deprecated configuration key 'kubernetes.operator.job.autoscaler.metrics.window' instead of proper key 'job.autoscaler.metrics.window'
> 2023-10-17 15:33:57,332 o.a.f.c.Configuration          [WARN ] [default.autoscaling-example] Config uses deprecated configuration key 'kubernetes.operator.job.autoscaler.stabilization.interval' instead of proper key 'job.autoscaler.stabilization.interval'
> 2023-10-17 15:33:57,397 o.a.f.a.ScalingMetricCollector [WARN ] [default.autoscaling-example] pendingRecords metric for cbc357ccb763df2852fee8c4fc7d55f2 could not be found. Either a legacy source or an idle source. Assuming no pending records.
> 2023-10-17 15:33:57,620 o.a.f.a.ScalingMetricCollector [INFO ] [default.autoscaling-example] Metric window not full until 2023-10-17T07:36:56.931915Z
> 2023-10-17 15:33:57,623 o.a.f.c.Configuration          [WARN ] [default.autoscaling-example] Config uses deprecated configuration key 'kubernetes.operator.job.autoscaler.target.utilization.boundary' instead of proper key 'job.autoscaler.target.utilization.boundary'
> 2023-10-17 15:33:57,624 o.a.f.c.Configuration          [WARN ] [default.autoscaling-example] Config uses deprecated configuration key 'kubernetes.operator.job.autoscaler.target.utilization.boundary' instead of proper key 'job.autoscaler.target.utilization.boundary'
> ```

Thank you @1996fanrui for the persistence and all the nice work. I will do a final pass on this later today, let's try to merge this as soon as possible to unblock current work on the autoscaler :)

mateczagany:
I've also manually tested this change on Flink 1.18 (both reactive and non-reactive), and worked flawlessly. 
Thank you for the effort you've put into this, I really like the direction the autoscaler is heading.

1996fanrui:
> I've also manually tested this change on Flink 1.18 (both reactive and non-reactive), and worked flawlessly. Thank you for the effort you've put into this, I really like the direction the autoscaler is heading.

Thanks @mateczagany  for the test!

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkResourceContext.java
```diff
@@ -47,6 +55,48 @@ public abstract class FlinkResourceContext<CR extends AbstractFlinkResource<?, ?
...
+            return autoScalerContext;
+        }
+        return autoScalerContext = createJobAutoScalerContext();
+    }
+
+    private KubernetesJobAutoScalerContext createJobAutoScalerContext() {
+        Configuration conf = new Configuration(getObserveConfig());
+        conf.set(
+                AutoScalerOptions.FLINK_CLIENT_TIMEOUT,
+                getOperatorConfig().getFlinkClientTimeout());
```
gyfora, line 71: Is this really necessary? seems costly, and we already have client timeout config in the operator itself which is already applied

flink-autoscaler/src/test/java/org/apache/flink/autoscaler/state/FlushCountableAutoscalerStateStore.java
```diff
@@ -15,19 +15,23 @@
...
-package org.apache.flink.kubernetes.operator.autoscaler;
+package org.apache.flink.autoscaler.state;
 
-import org.apache.flink.kubernetes.operator.api.AbstractFlinkResource;
+import org.apache.flink.autoscaler.JobAutoScalerContext;
 
-import io.fabric8.kubernetes.client.KubernetesClient;
-import lombok.SneakyThrows;
+/** The state store counts the flush. */
+public class FlushCountableAutoscalerStateStore<KEY, Context extends JobAutoScalerContext<KEY>>
```
gyfora, line 23: Could we call this `TestingAutoscalerStateStore` that way the intention is clear.

flink-autoscaler/src/test/java/org/apache/flink/autoscaler/realizer/ScalingRealizerCollector.java
```diff
@@ -0,0 +1,56 @@
...
+
+package org.apache.flink.autoscaler.realizer;
+
+import org.apache.flink.autoscaler.JobAutoScalerContext;
+
+import java.util.LinkedList;
+import java.util.Map;
+
+/** The event handler for collecting scaling events. */
+public class ScalingRealizerCollector<KEY, Context extends JobAutoScalerContext<KEY>>
```
gyfora, line 26: Could we call this `TestingScalingRealizer` that way the intention is clear.

flink-autoscaler/src/test/java/org/apache/flink/autoscaler/event/EventCollector.java
```diff
@@ -0,0 +1,110 @@
...
+
+import java.time.Duration;
+import java.time.Instant;
+import java.util.LinkedList;
+import java.util.Map;
+import java.util.Objects;
+import java.util.concurrent.ConcurrentHashMap;
+
+/** The event handler for collecting events. */
+public class EventCollector<KEY, Context extends JobAutoScalerContext<KEY>>
```
gyfora, line 34: Could we call this `TestingEventCollector` that way the intention is clear.

gyfora (approved):
Overall looks good, please rename the testing implementations for the components to `Testing...` this aligns it with other components and makes it clearer.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkResourceContext.java
1996fanrui: @mxm  also mentioned this, I have answered at this comment[1].

In brief, operator has client timeout config, however:
- `autoscaler` module cannot depend on operator code.
- The client timeout config of operator is used in many places(not only autoscaler), so move it into `autoscaler` module is strange.

[1] https://github.com/apache/flink-kubernetes-operator/pull/677#discussion_r1350028811

gyfora: Ah sorry, now I see. Let's keep this for now

flink-autoscaler/src/test/java/org/apache/flink/autoscaler/event/EventCollector.java
1996fanrui: Thanks for these suggestions, these 3 classes are renamed at the last commit.

gyfora (approved):

mxm (approved):

PR #681, opened 2023-09-21
[FLINK-33107] Use correct upgrade mode when executing rollback, simpl…

gyfora:

Rollbacks have recently been improved to use the same `upgrade` mechanism that we use during user spec changes to cover more cases.

However the previous implementation has a few shortcomings / bugs:
 - The code is a bit scattered around in the reconciler, making it difficult to understand
 - The upgradeMode was picked up from the last-stable spec, which in many cases can cause unintentional state loss. It should be picked up from the lastReconciledSpec
 - During upgrade the shutdown upgrade mode is not recorded correctly anywhere which may cause state loss or unrecoverable situations

This PR aims at improving the rollback flow and solve the current issues highlighted previously.

## Brief change log

  - *Refactor upgrade flow and remove extra reconcile cycle previously added when rollback was initiated*
  - *Use upgradeMode from lastReconciledSpec instead of lastStableSpec when rolling back*
  - *Record upgradeMode in lastReconciledSpec correctly when suspend was performed in ROLLING_BACK state*

## Verifying this change

 - Existing unit tests have been extended + new unit tests added to cover the previously not working corner cases.
 - Manual verification on local k8s

## Documentation


gyfora:
cc @ashangit

ashangit (approved):
Nice improvement

srpraneeth (approved):
LGTM

PR #682, opened 2023-09-30
[FLINK-33007] Integrate autoscaler config validation into the general validator flow

srpraneeth:

*Currently the AutoScaler configurations are not validated, hence they fail at runtime. This change intends to introduce a validator plugin for validating AutoScaler configuration*

## Brief change log

  - *AutoScaler Validator implemented using FlinkResourceValidator plugin*
  - *Unit tests added for the same, please find below details of all the rules being validated*
  - *Added the META-INF/services file for the validator to pick the right autoscaler plugin while available in classpath*

## Verifying this change

  - *Added Unit tests to cover all the rules validated on the autoscaler configuration.*
  - *Extended e2e-tests to make sure the AutoScaler validator plugin is picked and validated on the tests*
  - *Manually verified the change by running a test on the operator with a flinkdeployment with the CR voilating the validation rules.*

## Documentation

## Properties and Rules validated by the AutoScaler Configuration

|| Rule      | Affecting Property | MinValue | MaxValue | 
|-| ----------- | ----------- | ----------- | ----------- | 
|1| validateMaxScaleDownFactor | kubernetes.operator.job.autoscaler.scale-down.max-factor | 0.0d | 1.0d |
|2| validateMaxScaleUpFactor   | kubernetes.operator.job.autoscaler.scale-up.max-factor | 1.0d | 100000.0d |
|3| validateTargetUtilization   | kubernetes.operator.job.autoscaler.target.utilization | 0.0d | 1.0d | 
|4| validateTargetUtilizationBoundary   | kubernetes.operator.job.autoscaler.target.utilization.boundary | 0.0d | 1.0d |

### DIFF
--- a/e2e-tests/data/autoscaler.yaml
+++ b/e2e-tests/data/autoscaler.yaml
@@ -41,6 +41,12 @@ spec:
     job.autoscaler.stabilization.interval: "5s"
     job.autoscaler.metrics.window: "1m"
 
+#    Invalid Validations for testing autoscaler configurations
+#    kubernetes.operator.job.autoscaler.scale-down.max-factor: "-0.6"
+#    kubernetes.operator.job.autoscaler.scale-up.max-factor: "-1.0"
+#    kubernetes.operator.job.autoscaler.target.utilization: "-10.7"
+#    kubernetes.operator.job.autoscaler.target.utilization.boundary: "-Hundred"
+
     jobmanager.scheduler: adaptive
   serviceAccount: flink
   podTemplate:
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultValidator.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultValidator.java
@@ -17,7 +17,9 @@
 
 package org.apache.flink.kubernetes.operator.validation;
 
+import org.apache.flink.autoscaler.config.AutoScalerOptions;
 import org.apache.flink.configuration.CheckpointingOptions;
+import org.apache.flink.configuration.ConfigOption;
 import org.apache.flink.configuration.Configuration;
 import org.apache.flink.configuration.HighAvailabilityOptions;
 import org.apache.flink.configuration.JobManagerOptions;
@@ -60,6 +62,7 @@ import java.util.regex.Pattern;
 
 /** Default validator implementation for {@link FlinkDeployment}. */
 public class DefaultValidator implements FlinkResourceValidator {
+
     private static final Pattern DEPLOYMENT_NAME_PATTERN =
             Pattern.compile("[a-z]([-a-z\\d]{0,43}[a-z\\d])?");
     private static final String[] FORBIDDEN_CONF_KEYS =
@@ -102,7 +105,8 @@ public class DefaultValidator implements FlinkResourceValidator {
                 validateJmSpec(spec.getJobManager(), effectiveConfig),
                 validateTmSpec(spec.getTaskManager(), effectiveConfig),
                 validateSpecChange(deployment, effectiveConfig),
-                validateServiceAccount(spec.getServiceAccount()));
+                validateServiceAccount(spec.getServiceAccount()),
+                validateAutoScalerFlinkConfiguration(effectiveConfig));
     }
 
     @SafeVarargs
@@ -487,7 +491,8 @@ public class DefaultValidator implements FlinkResourceValidator {
         return firstPresent(
                 validateNotApplicationCluster(sessionCluster),
                 validateSessionClusterId(sessionJob, sessionCluster),
-                validateJobSpec(sessionJob.getSpec().getJob(), null, effectiveConfig));
+                validateJobSpec(sessionJob.getSpec().getJob(), null, effectiveConfig),
+                validateAutoScalerFlinkConfiguration(effectiveConfig));
     }
 
     private Optional<String> validateJobNotEmpty(FlinkSessionJob sessionJob) {
@@ -553,5 +558,53 @@ public class DefaultValidator implements FlinkResourceValidator {
                     "spec.serviceAccount must be defined. If you use helm, its value should be the same with the name of jobServiceAccount.");
         }
         return Optional.empty();
+    }
+
+    public static Optional<String> validateAutoScalerFlinkConfiguration(
+            Map<String, String> effectiveConfig) {
+        if (effectiveConfig == null) {
+            return Optional.empty();
+        }
+        Configuration flinkConfiguration = Configuration.fromMap(effectiveConfig);
+        if (!flinkConfiguration.getBoolean(AutoScalerOptions.AUTOSCALER_ENABLED)) {
+            return Optional.empty();
+        }
+        return firstPresent(
+                validateNumber(flinkConfiguration, AutoScalerOptions.MAX_SCALE_DOWN_FACTOR, 0.0d),
+                validateNumber(flinkConfiguration, AutoScalerOptions.MAX_SCALE_UP_FACTOR, 0.0d),
+                validateNumber(flinkConfiguration, AutoScalerOptions.TARGET_UTILIZATION, 0.0d),
+                validateNumber(
+                        flinkConfiguration, AutoScalerOptions.TARGET_UTILIZATION_BOUNDARY, 0.0d));
+    }
+
+    private static <T extends Number> Optional<String> validateNumber(
+            Configuration flinkConfiguration,
+            ConfigOption<T> autoScalerConfig,
+            Double min,
+            Double max) {
+        try {
+            var configValue = flinkConfiguration.get(autoScalerConfig);
+            if (configValue != null) {
+                double value = configValue.doubleValue();
+                if ((min != null && value < min) || (max != null && value > max)) {
+                    return Optional.of(
+                            String.format(
+                                    "The AutoScalerOption %s is invalid, it should be a value within the range [%s, %s]",
+                                    autoScalerConfig.key(),
+                                    min != null ? min.toString() : "-Infinity",
+                                    max != null ? max.toString() : "+Infinity"));
+                }
+            }
+            return Optional.empty();
+        } catch (IllegalArgumentException e) {
+            return Optional.of(
+                    String.format(
+                            "Invalid value in the autoscaler config %s", autoScalerConfig.key()));
+        }
+    }
+
+    private static <T extends Number> Optional<String> validateNumber(
+            Configuration flinkConfiguration, ConfigOption<T> autoScalerConfig, Double min) {
+        return validateNumber(flinkConfiguration, autoScalerConfig, min, null);
     }
 }
--- a/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/validation/DefaultValidatorTest.java
+++ b/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/validation/DefaultValidatorTest.java
@@ -17,7 +17,9 @@
 
 package org.apache.flink.kubernetes.operator.validation;
 
+import org.apache.flink.autoscaler.config.AutoScalerOptions;
 import org.apache.flink.configuration.CheckpointingOptions;
+import org.apache.flink.configuration.ConfigOption;
 import org.apache.flink.configuration.Configuration;
 import org.apache.flink.configuration.HighAvailabilityOptions;
 import org.apache.flink.kubernetes.configuration.KubernetesConfigOptions;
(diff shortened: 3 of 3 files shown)


srpraneeth:
@gyfora Can you please take a look at this PR

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/validation/AutoScalerValidator.java
```diff
@@ -0,0 +1,103 @@
...
+        return Optional.empty();
+    }
+
+    private Optional<String> validateAutoScalerFlinkConfiguration(
+            Configuration flinkConfiguration) {
+        return firstPresent(
+                validateNumber(
+                        flinkConfiguration, AutoScalerOptions.MAX_SCALE_DOWN_FACTOR, 0.0d, 1.0d),
+                validateNumber(
+                        flinkConfiguration, AutoScalerOptions.MAX_SCALE_UP_FACTOR, 1.0d, 100000.0d),
```
mateczagany, line 68: It is possible that some users have already set this to a higher value than 100.000 in their setup when they don't want to limit the scale-up, wouldn't it make sense to set the max value of this config to `Double.MAX_VALUE`? Just to make sure we don't introduce breaking changes in these scenarios.

gyfora: I agree let's not validate numbers that do not actually cause problems

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/validation/AutoScalerValidator.java
```diff
@@ -0,0 +1,103 @@
...
+            return validateAutoScalerFlinkConfiguration(flinkConfiguration);
+        }
+        return Optional.empty();
+    }
+
+    private Optional<String> validateAutoScalerFlinkConfiguration(
+            Configuration flinkConfiguration) {
+        return firstPresent(
+                validateNumber(
+                        flinkConfiguration, AutoScalerOptions.MAX_SCALE_DOWN_FACTOR, 0.0d, 1.0d),
```
gyfora, line 66: Let's not validate the max value here, only that is not negative.

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/validation/AutoScalerValidator.java
```diff
@@ -0,0 +1,103 @@
...
+                validateNumber(
+                        flinkConfiguration, AutoScalerOptions.MAX_SCALE_UP_FACTOR, 1.0d, 100000.0d),
+                validateNumber(
```
gyfora, line 69: Same as before let's not validate the max

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/validation/AutoScalerValidator.java
```diff
@@ -0,0 +1,103 @@
...
+                        flinkConfiguration, AutoScalerOptions.TARGET_UTILIZATION, 0.0d, 1.0d),
+                validateNumber(
+                        flinkConfiguration,
+                        AutoScalerOptions.TARGET_UTILIZATION_BOUNDARY,
```
gyfora, line 73: Same as before let's not validate the max

gyfora (changes requested):

srpraneeth:
@mateczagany @gyfora 

Can you please review again. I have fixed the comments and the latest rules validated as below

|| Rule      | Affecting Property | MinValue | MaxValue | 
|-| ----------- | ----------- | ----------- | ----------- | 
|1| validateMaxScaleDownFactor | kubernetes.operator.job.autoscaler.scale-down.max-factor | 0.0d | None |
|2| validateMaxScaleUpFactor   | kubernetes.operator.job.autoscaler.scale-up.max-factor | 1.0d | None |
|3| validateTargetUtilization   | kubernetes.operator.job.autoscaler.target.utilization | 0.0d | None | 
|4| validateTargetUtilizationBoundary   | kubernetes.operator.job.autoscaler.target.utilization.boundary | 0.0d | 1.0d |

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/validation/AutoScalerValidator.java
srpraneeth: Fixed

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/validation/AutoScalerValidator.java
srpraneeth: Fixed

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/validation/AutoScalerValidator.java
srpraneeth: Fixed

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/validation/AutoScalerValidator.java
srpraneeth: Fixed

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/validation/AutoScalerValidator.java
```diff
@@ -0,0 +1,109 @@
...
+                        0.0d,
+                        1.0d));
+    }
+
+    private <T extends Number> Optional<String> validateNumber(
+            Configuration flinkConfiguration,
+            ConfigOption<T> autoScalerConfig,
+            Double min,
+            Double max) {
+        var configValue = flinkConfiguration.get(autoScalerConfig);
```
gyfora, line 80: This line may throw an exception if the config value is not actually a number, we should catch and turn it into a validation error probably.

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/validation/AutoScalerValidator.java
```diff
@@ -0,0 +1,109 @@
...
+        if (spec.getFlinkConfiguration() != null) {
+            var flinkConfiguration = Configuration.fromMap(spec.getFlinkConfiguration());
+            return validateAutoScalerFlinkConfiguration(flinkConfiguration);
+        }
+        return Optional.empty();
+    }
+
+    @Override
+    public Optional<String> validateDeployment(FlinkDeployment deployment) {
+        LOG.info("AutoScaler Configurations validator {}", deployment);
```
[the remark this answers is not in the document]
gyfora: For this you may also need the default config, you can take a look at how to get that based on the `DefaultValidator`

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/validation/AutoScalerValidator.java
```diff
@@ -0,0 +1,109 @@
...
+            Configuration flinkConfiguration) {
+        return firstPresent(
+                validateNumber(flinkConfiguration, AutoScalerOptions.MAX_SCALE_DOWN_FACTOR, 0.0d),
+                validateNumber(flinkConfiguration, AutoScalerOptions.MAX_SCALE_UP_FACTOR, 1.0d),
+                validateNumber(flinkConfiguration, AutoScalerOptions.TARGET_UTILIZATION, 0.0d),
+                validateNumber(
+                        flinkConfiguration,
+                        AutoScalerOptions.TARGET_UTILIZATION_BOUNDARY,
+                        0.0d,
+                        1.0d));
```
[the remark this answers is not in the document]
srpraneeth: Fixed

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/validation/AutoScalerValidator.java
srpraneeth: The IllegalArgEx is caught and handled as a validation error.

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/validation/AutoScalerValidator.java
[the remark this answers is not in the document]
srpraneeth: I have added a check to validate only if the autoscaler is enabled. I added unit test and tested through e2e test to make sure that it works. 

Can you please explain what is default config for. 
I see that in the validator utils the default config is plugged in through a FlinkConfigManager (through constructor in DefaultValidator) but that may not be straight forward in case of AutoScalerValidator as AutoScalerValidator is not present in the operator module. It is discovered through ServiceLoader instead from the autoscaler module at runtime. 

I added a UT to make sure that with out the default config still the flinkConfiguration.get would get me the default value for the autoscaler configurations. 
Let me know if you still think I should add the default config I can look into it.

e2e-tests/data/autoscaler.yaml
```diff
@@ -41,6 +41,13 @@ spec:
     kubernetes.operator.job.autoscaler.stabilization.interval: "5s"
     kubernetes.operator.job.autoscaler.metrics.window: "1m"
 
+#    Invalid Validations for testing autoscaler configurations
+  #    kubernetes.operator.job.autoscaler.scale-down.max-factor: "-0.6"
+  #    kubernetes.operator.job.autoscaler.scale-up.max-factor: "100001.0"
+  #    pipeline.max-parallelism: "2"
+  #    kubernetes.operator.job.autoscaler.target.utilization: "10.7"
```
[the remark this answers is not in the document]
srpraneeth: Updated. Thanks for catching this..

flink-kubernetes-operator-autoscaler/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/validation/AutoScalerValidator.java
```diff
@@ -0,0 +1,109 @@
...
+            return validateAutoScalerFlinkConfiguration(flinkConfiguration);
+        }
+        return Optional.empty();
+    }
+
+    private Optional<String> validateAutoScalerFlinkConfiguration(
+            Configuration flinkConfiguration) {
+        return firstPresent(
+                validateNumber(flinkConfiguration, AutoScalerOptions.MAX_SCALE_DOWN_FACTOR, 0.0d),
+                validateNumber(flinkConfiguration, AutoScalerOptions.MAX_SCALE_UP_FACTOR, 1.0d),
```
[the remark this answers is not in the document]
srpraneeth: Done.. Fixed.. Now all the above options are just validated for the non negative number. 

|| Rule      | Affecting Property | MinValue | MaxValue | 
|-| ----------- | ----------- | ----------- | ----------- | 
|1| validateMaxScaleDownFactor | kubernetes.operator.job.autoscaler.scale-down.max-factor | 0.0d | None |
|2| validateMaxScaleUpFactor   | kubernetes.operator.job.autoscaler.scale-up.max-factor | 0.0d | None |
|3| validateTargetUtilization   | kubernetes.operator.job.autoscaler.target.utilization | 0.0d | None | 
|4| validateTargetUtilizationBoundary   | kubernetes.operator.job.autoscaler.target.utilization.boundary | 0.0d | None |

srpraneeth:
Refactored after the merge of PR #677 
@gyfora

gyfora (changes requested):
Hey @srpraneeth !
Sorry for iterating too much on this :) 

I think it would actually make sense to (instead of implementing the FlinkResourceValidator interface) simply embed this logic into the DefaultValidator.

We could then just keep the `validateAutoScalerFlinkConfiguration` as a static utility and call it from the `DefaultValidator` .

That way we benefit from the config handling already present in the DefaultValidator that properly handles / default and config overrides. We would also avoid these costly conversions between configs/maps etc.

Furthermore the config validation utility then can be moved to the autoscaler module where it logically belongs and other autoscaler imeplementaitons can also use it in the future (not only Kubernetes). I think this would simplify the code a lot actually and make the whole thing better.

srpraneeth:
> Hey @srpraneeth ! Sorry for iterating too much on this :)
> 
> I think it would actually make sense to (instead of implementing the FlinkResourceValidator interface) simply embed this logic into the DefaultValidator.
> 
> We could then just keep the `validateAutoScalerFlinkConfiguration` as a static utility and call it from the `DefaultValidator` .
> 
> That way we benefit from the config handling already present in the DefaultValidator that properly handles / default and config overrides. We would also avoid these costly conversions between configs/maps etc.
> 
> Furthermore the config validation utility then can be moved to the autoscaler module where it logically belongs and other autoscaler imeplementaitons can also use it in the future (not only Kubernetes). I think this would simplify the code a lot actually and make the whole thing better.


@gyfora Thanks for the thoroughly checking on PR. Agree this makes complete sense. 
I will move the autoscaler validation to the DefaultValidator and update the PR.

gyfora (approved):

PR #685, opened 2023-10-18
[FLINK-33187] using hashcode for parallelism map comparison

clarax:

This is to implement the scaling report comparison algorithm for event suppression by using hashcodes of the parallelism maps. Originally I used the full string of report message which contains the metrics fluctuating with current load. As long as parallelism map doesn't change, we don't need to generate new events within the defined interval.

## Brief change log
  - Save hashcode of the parallelism map to metadata labels
  - Use the hashcode to compare two scaling report for advice

## Verifying this change
  - Updated and added unit tests.
  - Verified in integration test env.

  - Dependencies (does it add or upgrade a dependency): No
  - The public API, i.e., is any changes to the `CustomResourceDescriptors`: No
  - Core observer or reconciler logic that is regularly executed: No

## Documentation

  - Does this pull request introduce a new feature? No
  - If yes, how is the feature documented? N/A

### DIFF
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
@@ -235,6 +235,7 @@ public class AutoScalerOptions {
             autoScalerConfig("scaling.event.interval")
                     .durationType()
                     .defaultValue(Duration.ofSeconds(1800))
+                    .withDeprecatedKeys(deprecatedOperatorConfigKey("scaling.event.interval"))
                     .withDescription("Time interval to resend the identical event");
 
     public static final ConfigOption<Duration> FLINK_CLIENT_TIMEOUT =
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
@@ -231,8 +231,8 @@ public class AutoScalerOptions {
                     .withDescription(
                             "A (semicolon-separated) list of vertex ids in hexstring for which to disable scaling. Caution: For non-sink vertices this will still scale their downstream operators until https://issues.apache.org/jira/browse/FLINK-31215 is implemented.");
 
-    public static final ConfigOption<Duration> SCALING_REPORT_INTERVAL =
-            autoScalerConfig("scaling.report.interval")
+    public static final ConfigOption<Duration> SCALING_EVENT_INTERVAL =
+            autoScalerConfig("scaling.event.interval")
                     .durationType()
                     .defaultValue(Duration.ofSeconds(1800))
                     .withDescription("Time interval to resend the identical event");
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/event/AutoScalerEventHandler.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/event/AutoScalerEventHandler.java
@@ -24,9 +24,9 @@ import org.apache.flink.runtime.jobgraph.JobVertexID;
 
 import javax.annotation.Nullable;
 
+import java.time.Duration;
 import java.util.Map;
 
-import static org.apache.flink.autoscaler.config.AutoScalerOptions.SCALING_ENABLED;
 import static org.apache.flink.autoscaler.metrics.ScalingMetric.EXPECTED_PROCESSING_RATE;
 import static org.apache.flink.autoscaler.metrics.ScalingMetric.TARGET_DATA_RATE;
 import static org.apache.flink.autoscaler.metrics.ScalingMetric.TRUE_PROCESSING_RATE;
@@ -44,18 +44,42 @@ public interface AutoScalerEventHandler<KEY, Context extends JobAutoScalerContex
     String SCALING_SUMMARY_HEADER_SCALING_DISABLED = "Recommended parallelism change:";
     String SCALING_SUMMARY_HEADER_SCALING_ENABLED = "Scaling vertices:";
     String SCALING_REPORT_REASON = "ScalingReport";
-    String EVENT_MESSAGE_KEY = "ScalingExecutor";
+    String SCALING_REPORT_KEY = "ScalingExecutor";
 
-    /** Handle the event. */
+    /**
+     * Handle the event.
+     *
+     * @param interval Define the interval to suppress duplicate events. No dedupe if null.
+     */
     void handleEvent(
-            Context context, Type type, String reason, String message, @Nullable String messageKey);
+            Context context,
+            Type type,
+            String reason,
+            String message,
+            @Nullable String messageKey,
+            @Nullable Duration interval);
 
+    /**
+     * Handle scaling reports.
+     *
+     * @param interval Define the interval to suppress duplicate events.
+     * @param scaled Whether AutoScaler actually scaled the Flink job or just generate advice for
+     *     scaling.
+     */
     default void handleScalingEvent(
-            Context context, Map<JobVertexID, ScalingSummary> scalingSummaries) {
+            Context context,
+            Map<JobVertexID, ScalingSummary> scalingSummaries,
+            boolean scaled,
+            Duration interval) {
         // Provide default implementation without proper deduplication
-        var scalingReport =
-                scalingReport(scalingSummaries, context.getConfiguration().get(SCALING_ENABLED));
-        handleEvent(context, Type.Normal, SCALING_REPORT_REASON, scalingReport, EVENT_MESSAGE_KEY);
+        var scalingReport = scalingReport(scalingSummaries, scaled);
+        handleEvent(
+                context,
+                Type.Normal,
+                SCALING_REPORT_REASON,
+                scalingReport,
+                SCALING_REPORT_KEY,
+                interval);
     }
 
     static String scalingReport(
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/event/AutoScalerEventHandler.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/event/AutoScalerEventHandler.java
@@ -19,10 +19,17 @@ package org.apache.flink.autoscaler.event;
 
 import org.apache.flink.annotation.Experimental;
 import org.apache.flink.autoscaler.JobAutoScalerContext;
+import org.apache.flink.autoscaler.ScalingSummary;
+import org.apache.flink.runtime.jobgraph.JobVertexID;
 
 import javax.annotation.Nullable;
 
-import java.time.Duration;
+import java.util.Map;
+
+import static org.apache.flink.autoscaler.config.AutoScalerOptions.SCALING_ENABLED;
+import static org.apache.flink.autoscaler.metrics.ScalingMetric.EXPECTED_PROCESSING_RATE;
+import static org.apache.flink.autoscaler.metrics.ScalingMetric.TARGET_DATA_RATE;
+import static org.apache.flink.autoscaler.metrics.ScalingMetric.TRUE_PROCESSING_RATE;
 
 /**
  * Handler for autoscaler events.
@@ -32,20 +39,45 @@ import java.time.Duration;
  */
 @Experimental
 public interface AutoScalerEventHandler<KEY, Context extends JobAutoScalerContext<KEY>> {
+    String SCALING_SUMMARY_ENTRY =
+            " Vertex ID %s | Parallelism %d -> %d | Processing capacity %.2f -> %.2f | Target data rate %.2f";
+    String SCALING_SUMMARY_HEADER_SCALING_DISABLED = "Recommended parallelism change:";
+    String SCALING_SUMMARY_HEADER_SCALING_ENABLED = "Scaling vertices:";
+    String SCALING_REPORT_REASON = "ScalingReport";
+    String EVENT_MESSAGE_KEY = "ScalingExecutor";
 
-    /**
-     * Handle the event.
-     *
-     * @param interval When interval is great than 0, events that repeat within the interval will be
-     *     ignored.
-     */
+    /** Handle the event. */
     void handleEvent(
-            Context context,
-            Type type,
-            String reason,
-            String message,
-            @Nullable String messageKey,
-            @Nullable Duration interval);
+            Context context, Type type, String reason, String message, @Nullable String messageKey);
+
+    default void handleScalingEvent(
+            Context context, Map<JobVertexID, ScalingSummary> scalingSummaries) {
+        // Provide default implementation without proper deduplication
+        var scalingReport =
+                scalingReport(scalingSummaries, context.getConfiguration().get(SCALING_ENABLED));
+        handleEvent(context, Type.Normal, SCALING_REPORT_REASON, scalingReport, EVENT_MESSAGE_KEY);
+    }
+
+    static String scalingReport(
+            Map<JobVertexID, ScalingSummary> scalingSummaries, boolean scalingEnabled) {
+        StringBuilder sb =
+                new StringBuilder(
+                        scalingEnabled
+                                ? SCALING_SUMMARY_HEADER_SCALING_ENABLED
+                                : SCALING_SUMMARY_HEADER_SCALING_DISABLED);
+        scalingSummaries.forEach(
+                (v, s) ->
+                        sb.append(
+                                String.format(
+                                        SCALING_SUMMARY_ENTRY,
+                                        v,
+                                        s.getCurrentParallelism(),
+                                        s.getNewParallelism(),
+                                        s.getMetrics().get(TRUE_PROCESSING_RATE).getAverage(),
+                                        s.getMetrics().get(EXPECTED_PROCESSING_RATE).getCurrent(),
+                                        s.getMetrics().get(TARGET_DATA_RATE).getAverage())));
+        return sb.toString();
+    }
 
     /** The type of the events. */
     enum Type {
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/KubernetesAutoScalerEventHandler.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/KubernetesAutoScalerEventHandler.java
@@ -17,19 +17,29 @@
 
 package org.apache.flink.kubernetes.operator.autoscaler;
 
+import org.apache.flink.autoscaler.ScalingSummary;
 import org.apache.flink.autoscaler.event.AutoScalerEventHandler;
 import org.apache.flink.kubernetes.operator.utils.EventRecorder;
+import org.apache.flink.runtime.jobgraph.JobVertexID;
 
 import io.javaoperatorsdk.operator.processing.event.ResourceID;
 
 import javax.annotation.Nullable;
 
-import java.time.Duration;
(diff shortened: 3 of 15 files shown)


flink-autoscaler/src/main/java/org/apache/flink/autoscaler/event/AutoScalerEventHandler.java
```diff
@@ -33,19 +35,15 @@
 @Experimental
 public interface AutoScalerEventHandler<KEY, Context extends JobAutoScalerContext<KEY>> {
 
-    /**
-     * Handle the event.
-     *
-     * @param interval When interval is great than 0, events that repeat within the interval will be
-     *     ignored.
-     */
+    /** Handle the event. */
```
gyfora, line 38: I think the entire eventHandler interface is getting pretty weird. It was already not very nice before this change but now it's even worse because we are leaking too many internals of the operator event triggering mechanism. 

I think we need to break up the handleEvent into two new methods.

```
// Would be the current logic but without the interval logic
handleGenericAutoscalerEvent(...)

// Handle scaling events separately. Then the operator can hide the complex interval/label triggering logic
handleScalingEvent(Context,  Map<JobVertexID, ScalingSummary> scalingSummaries, scalingEnabled)
```

gyfora: And we could even have a default implementation within the interface itself to turn it into a generic event in case we are not running on the operator

clarax: Thank you for calling out. I was not sure if I should have a separate method, not sure what this handler interfaceshould be handing. We can move the logic of handleGenericAutoscalerEven to the interface and make it an abstract class. What do you think? @gyfora

gyfora: I don't understand the suggestion @clarax what do you think my proposal? Have those 2 methods in the interface and add a default implementation for `handleScalingEvent`.

clarax: Updated for two methods.  I don't understand "a default implementation within the interface itself to turn it into a generic event in case we are not running on the operator". I made the second method a generic method for any suppression callers want to use.

gyfora:
Sorry @clarax I think I gave confusing feedback and didn't express myself clearly. I was thinking something like this:

```
interface AutoScalerEventHandler {

void handleGenericEvent(Ctx, Type, Reson, Message, Key, Interval);

default void handleScalingEvent(Ctx, Map<JobVertexId, ScalingSummary> summaries, boolean scalingEnabled, Interval) {
    // Provide default implementation without proper deduplication
   handleGenericEvent(...., toMessage(summaries), interval);
} 

}
```

Then for Kubernetes we have the custom label logic encapsulated:

```
class KubernetesAutoScalerEventHandler {

...

@Override 
void handleScalingEvent(Ctx, Map<JobVertexId, ScalingSummary> summaries, boolean scalingEnabled, Interval) {
   var labels = createLabels(summaries)
   getOldEvent, compare labels, whatever we need to do we do not leak to ScalingExecutor
}

}
```

This way the ScalingExecutor simply calls:

```
eventHandler.handleScalingEvent(...)
```

And we could remove the predicate/label logic that really do not belong there as that is very Kubernetes specific and move it to the `KubernetesAutoScalerEventHandler` implementation. 

What do you think?

clarax:
Refactored as @gyfora suggusted.

gyfora (approved):
I think it looks really good now, we have a much nicer separation of responsibilities in the different components. Please verify the latest version manually (locally or something) if you can @clarax 

@1996fanrui What do you think about the interface change?

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/event/AutoScalerEventHandler.java
1996fanrui: The reason for removing the interval from this interface is because interval just works for kubernetesEventHandler, right? 

I wanna whether it's needed for Standalone Autoscaler, if it's needed, we should keep the `interval` here.

Actually, I prefer keep it first. And removing it if we are sure it's really not needed in the future. WDYT? cc @gyfora 

Also, if the interval is not needed here. `AutoScalerOptions#SCALING_REPORT_INTERVAL` should be moved into `flink-kubernetes-operator` module.

gyfora: I agree, we can keep the interval as a fixed parameter to the reporter methods. That way we don't need custom config handling in the operator.

clarax: Updated the PR with the change. Added comments to explain too.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/KubernetesAutoScalerEventHandler.java
```diff
@@ -43,26 +52,71 @@ public void handleEvent(
...
+            boolean scaled,
+            @Nonnull Duration interval) {
+        if (scaled) {
+            AutoScalerEventHandler.super.handleScalingEvent(
+                    context, scalingSummaries, scaled, interval);
         } else {
-            eventRecorder.triggerEventByInterval(
+            var conf = context.getConfiguration();
+            var scalingReport = AutoScalerEventHandler.scalingReport(scalingSummaries, scaled);
+            var labels = Map.of(PARALLELISM_MAP_KEY, getParallelismHashCode(scalingSummaries));
```
1996fanrui, line 78: Hi @clarax ,IIUC, the current strategy is: when the scaled is disabled, we want to remove duplicates based on `getParallelismHashCode(scalingSummaries)` and interval, right?

If yes, how about we don't change the `AutoScalerEventHandler` and consider `getParallelismHashCode(scalingSummaries)` as the messageKey? 

And refactor ingthe `EventUtils`:

-  Ignoring repeating message based on the `generatedMessageKey` and `interval`
- `generatedMessageKey = messageKey != null ? messageKey : message;`
- For existed event,  when the {@link #MESSAGE_KEY} exists in label, it's messageKey, otherwise the message is messageKey.

Here is my POC commit[1], it maybe clearer. If the POC has the same effect as the current PR, I like this POC.

Because the interface is clearer and  `AutoScalerEventHandler` doesn't care the logic about `handleScalingEvent`.

Please correct me if my understanding is wrong, thanks~

[1] https://github.com/1996fanrui/flink-kubernetes-operator/commit/14e8f189f4d129eba23050ea37fe8c012f0a22d2

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/EventUtils.java
```diff
@@ -117,53 +119,53 @@ public static boolean createByInterval(
...
-        eventListener.accept(client.resource(event).createOrReplace());
+    private static void setLabels(Event existing, @Nullable Map<String, String> labels) {
+        if (existing.getMetadata() == null) {
+            var metaData = new ObjectMeta();
+            metaData.setLabels(labels);
+        } else if (existing.getMetadata().getLabels() == null) {
+            existing.getMetadata().setLabels(labels);
+        } else {
+            existing.getMetadata().setLabels(labels);
+        }
```
1996fanrui, line 168:
```suggestion
        } else {
            existing.getMetadata().setLabels(labels);
        }
```

Same behavior for different code conditions, they can be simplified.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/EventUtils.java
```diff
@@ -117,53 +119,53 @@ public static boolean createByInterval(
...
-            EventRecorder.Type type,
-            String reason,
-            String message,
-            EventRecorder.Component component,
-            Consumer<Event> eventListener,
-            String eventName) {
-        Event event = buildEvent(target, type, reason, message, component, eventName);
-        eventListener.accept(client.resource(event).createOrReplace());
+    private static void setLabels(Event existing, @Nullable Map<String, String> labels) {
+        if (existing.getMetadata() == null) {
```
1996fanrui, line 161: If labels is null, is this `setLabels` necessary?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/KubernetesAutoScalerEventHandler.java
1996fanrui: cc @gyfora

gyfora: @1996fanrui The problem with your approach is that it breaks the messageKey logic for scaling events (you set a new key every time the parallelism changes but only when scaling is off)

The general problem with the current interface is that it already contains many Kubernetes specific parts (messageKey) but not enough to properly implement new functionality (such as the one we are trying here to deduplicate events based on certain logic, like the actual parallelism overrides)

In other interfaces like the StateStore we decided to go with specific methods, and I think that's what we should do here as well. I prefer @clarax s solution as it allows us complete flexibility in the Kubernetes implementation without pushing anything on to the interface.

To give a bit more background on the `messageKey` , we use that to define the specific event when triggering in Kubernetes such that we keep the history (last triggered, etc) without always creating new event objects. This is completely independent of the interval / parallelism based triggering that are trying to do here.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/EventUtils.java
clarax: Good catch. Will update.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/KubernetesAutoScalerEventHandler.java
1996fanrui: Hi @gyfora , thanks for your explanation!

> you set a new key every time the parallelism changes but only when scaling is off
 
I set it based on the logic of current PR. Do you mean my POC logic isn't exactly same with the current PR? I may have some misunderstand here.

FYI: I do this POC due to I want to design an interface for support the kubernetes requirement, and let it as generic as possible. If design a generic interface is very hard here, the current design is good to me.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/EventUtils.java
clarax: Currently we don't call with non null value. But it doesn't hurt to allow caller to set it to null as needed. It will be considered as different labels from empty map so we don't accidentally dedupe.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/event/AutoScalerEventHandler.java
```diff
@@ -32,20 +40,69 @@
  */
 @Experimental
 public interface AutoScalerEventHandler<KEY, Context extends JobAutoScalerContext<KEY>> {
+    String SCALING_SUMMARY_ENTRY =
+            " Vertex ID %s | Parallelism %d -> %d | Processing capacity %.2f -> %.2f | Target data rate %.2f";
+    String SCALING_SUMMARY_HEADER_SCALING_DISABLED = "Recommended parallelism change:";
+    String SCALING_SUMMARY_HEADER_SCALING_ENABLED = "Scaling vertices:";
+    String SCALING_REPORT_REASON = "ScalingReport";
+    String EVENT_MESSAGE_KEY = "ScalingExecutor";
```
gyfora, line 48: This should be renamed to `SCALING_MESSAGE_KEY`

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/KubernetesAutoScalerEventHandler.java
gyfora: @1996fanrui in Claras PR the message key is fixed: `ScalingExecutor` in your POC PR the message key is derived from the parallelism, so it changes all the time. They are not the same.

clarax: Hi @1996fanrui, Thanks for the proposed alternative, just adding to @gyfora 's comment, 
"To give a bit more background on the messageKey , we use that to define the specific event when triggering in Kubernetes such that we keep the history (last triggered, etc) without always creating new event objects. This is completely independent of the interval / parallelism based triggering that are trying to do here.
"

Overriding messageKey will totally break the design of using messageKey to keep history of certain type of events at minimal cost. We keep using the same Kubernetes event object and just need to increment the count of the object and updating the messsages. The dedupe logic by interval is independent where we don't even update the kubenetes event object or trigger any operator events. The behaviors of your POC is different from this PR.

1996fanrui: Thanks for the clarification!

Sounds good to me.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/event/AutoScalerEventHandler.java
clarax: will update.

gyfora (changes requested):
Another thing we have been discussing with @1996fanrui 

The interval config should probably be renamed `scaling.report.interval->scaling.event.interval` this way we can use it generally in the future for autoscaler triggered events.

We should also make sure that the simple `handleEvent` method also respects the interval if specified. And we should probably use the interval also for ineffective scaling events. I know that some of these changes are not directly related to this PR but it may be better to clean it up so we leave it in a good state afterwards.

clarax:
> Another thing we have been discussing with @1996fanrui
> 
> The interval config should probably be renamed `scaling.report.interval->scaling.event.interval` this way we can use it generally in the future for autoscaler triggered events.
> 
> We should also make sure that the simple `handleEvent` method also respects the interval if specified. And we should probably use the interval also for ineffective scaling events. I know that some of these changes are not directly related to this PR but it may be better to clean it up so we leave it in a good state afterwards.

Resolved all requested changes.

1996fanrui:
Hi, the ci fails, and please run the `mvn clean install -DskipTests -Pgenerate-docs` again, thanks

https://github.com/apache/flink-kubernetes-operator/actions/runs/6621898931/job/17987047236?pr=685#step:5:19405

1996fanrui (approved):

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
```diff
@@ -201,8 +201,8 @@ private static ConfigOptions.OptionBuilder autoScalerConfig(String key) {
                     .withDescription(
                             "A (semicolon-separated) list of vertex ids in hexstring for which to disable scaling. Caution: For non-sink vertices this will still scale their downstream operators until https://issues.apache.org/jira/browse/FLINK-31215 is implemented.");
 
-    public static final ConfigOption<Duration> SCALING_REPORT_INTERVAL =
-            autoScalerConfig("scaling.report.interval")
+    public static final ConfigOption<Duration> SCALING_EVENT_INTERVAL =
+            autoScalerConfig("scaling.event.interval")
```
gyfora, line 205: Docs seem to be inconsistent with this and should regenerated after renaming to `event.interval`

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/EventUtils.java
```diff
@@ -197,4 +198,19 @@ private static Event buildEvent(
...
+                                Instant.parse(existing.getLastTimestamp())
+                                        .plusMillis(interval.toMillis()));
+    }
+
+    private static boolean labelCheck(
+            Event existing, Predicate<Map<String, String>> dedupePredicate) {
+        return dedupePredicate == null
+                || (existing.getMetadata() != null
+                        && dedupePredicate.test(existing.getMetadata().getLabels()));
+    }
```
gyfora, line 215: I may misunderstand something but seems like labels are basically ignored when the `interval == null` . In that case intervalCheck is always false. Is this intentional?

gyfora (changes requested):
The PR looks great @clarax , thank you!

I found 2 very minor things (see comments), otherwise it is good to go

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
```diff
@@ -201,8 +201,8 @@ private static ConfigOptions.OptionBuilder autoScalerConfig(String key) {
...
                     .durationType()
                     .defaultValue(Duration.ofSeconds(1800))
                     .withDescription("Time interval to resend the identical event");
```
1996fanrui, line 208:
```suggestion
                    .withDeprecatedKeys(
                            deprecatedOperatorConfigKey(
                                    "scaling.event.interval"))
                    .withDescription("Time interval to resend the identical event");
```

As this comment[1] mentioned: ` all config keys to work with the "old" syntax at least in the 1.7.0 release. `, so please update it, thanks!

Sorry for that, I didn't support the old key during decoupling autoscaler and kubernetes-operator.

https://github.com/apache/flink-kubernetes-operator/pull/686#discussion_r1369777698

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/EventUtils.java
clarax: when Interval == null, we don't dedupe. This is intentional. It is commented in the interface method param and unit  tested.
but when the config is null, we use the default value of 30 min. It is also documented and unit tested.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
clarax: updated.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/EventUtils.java
gyfora: makes sense

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
clarax: > As this comment[1] mentioned: `all config keys to work with the "old" syntax at least in the 1.7.0 release.`, so please update it, thanks!
> 
> Sorry for that, I didn't support the old key during decoupling autoscaler and kubernetes-operator.
> 
> [#686 (comment)](https://github.com/apache/flink-kubernetes-operator/pull/686#discussion_r1369777698)
updated.

PR #686, opened 2023-10-18
[FLINK-33306] Use observed source throughput as true processing rate

gyfora:

The aim of this PR is to address the cases when operator incorrectly reports low busy time (high idleness) for sources that are in fact cannot keep up due to the slowness of the reader/fetchers. As the metrics cannot be generally fixed on the Flink - connector side we have to detect this and handle it when collecting the metrics.

The main symptom of this problem is overestimation of the true processing rate and not triggering scaling even if lag is building up as the autoscaler thinks it will be able to keep up.

To tackle this we differentiate two different methods of TPR measurement:
 1. **Busy-time based TPR** (this is the current approach in the autoscaler) : computed from incoming records and busy time 
 2. **Observed TPR** : computed from incoming records and back pressure, measurable only when we assume full processing throughput (i.e during catch-up)

### Current behaviour

The operator currently always uses a busy-time based TPR calculation which is very flexible and allows for scaling up / down but is susceptible to overestimation due to the broken metrics. 

### Suggested new behaviour

Instead of using the busy-time based TPR we detect when TPR is overestimated (busy-time too low) and switch to observed TPR. 

To do this, whenever we there is lag for a source (during catchup, or lag-buildup) we measure both busy-time and observed TPR.

If the avg busy-time based TPR is off by a configured amount we switch to observed TPR for this source during metric evaluation. 

**Why not use observed TPR all the time?**
Observed TPR can only be measured when we are catching up (during stabilization) or when cannot keep up. This makes it harder to scale down or to detect changes in source throughput over time (before lag starts to build up). Instead of using observed TPR we switch to it only when we detect a problem with the busy-time (this is a rare case overall), to hopefully get the best of both worlds.

#### Fix metric name querying for sources

Metric names were only previously queried in topology change, however this is not good for sources where new metrics are dynamically added/removed. Now source metric names are always queried.

#### Fix infinity handling in average

Previously avg logic returned infinity on any infinite value in the array. This causes invalid scale downs when new records are indeed coming in. The logic has been fixed to only return infinity when all of them are infinity.

## Brief change log

  - *Fix source metric name querying (always refresh for sources to detect new partition metrics)*
  - *Collect metrics also during stabilization (but do not evaluate) to allow measuring observed TPR*
  - *Add new TPR measurement logic for sources*
  - *Change TPR evaluation logic to switch between observed vs busy time based true processing rate*
  - *Fix infinity handling for average (only return infinity if all infinite otherwise ignore)*

## Verifying this change

## Documentation

### DIFF
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java
@@ -92,7 +92,7 @@ public abstract class ScalingMetricCollector<KEY, Context extends JobAutoScalerC
                         jobKey,
                         (k) -> {
                             try {
-                                return stateStore.getEvaluatedMetrics(ctx).orElse(new TreeMap<>());
+                                return stateStore.getCollectedMetrics(ctx).orElse(new TreeMap<>());
                             } catch (Exception exception) {
                                 throw new RuntimeException(
                                         "Get evaluated metrics failed.", exception);
@@ -106,7 +106,7 @@ public abstract class ScalingMetricCollector<KEY, Context extends JobAutoScalerC
         // metric timestamp
         if (!metricHistory.isEmpty() && jobUpdateTs.isAfter(metricHistory.firstKey())) {
             LOG.info("Job updated at {}. Clearing metrics.", jobUpdateTs);
-            stateStore.removeEvaluatedMetrics(ctx);
+            stateStore.removeCollectedMetrics(ctx);
             cleanup(ctx.getJobKey());
             metricHistory.clear();
         }
@@ -130,10 +130,10 @@ public abstract class ScalingMetricCollector<KEY, Context extends JobAutoScalerC
 
         // Add scaling metrics to history if they were computed successfully
         metricHistory.put(now, scalingMetrics);
-        stateStore.storeEvaluatedMetrics(ctx, metricHistory);
 
         if (now.isBefore(stableTime)) {
             LOG.info("Stabilizing until {}", stableTime);
+            stateStore.storeCollectedMetrics(ctx, metricHistory);
             return new CollectedMetricHistory(topology, Collections.emptySortedMap());
         }
 
@@ -145,6 +145,7 @@ public abstract class ScalingMetricCollector<KEY, Context extends JobAutoScalerC
             // Trim metrics outside the metric window from metrics history
             metricHistory.headMap(now.minus(metricWindowSize)).clear();
         }
+        stateStore.storeCollectedMetrics(ctx, metricHistory);
         return collectedMetrics;
     }
 
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java
@@ -363,25 +363,13 @@ public abstract class ScalingMetricCollector<KEY, Context extends JobAutoScalerC
                                         queryFilteredMetricNames(
                                                 ctx,
                                                 topology,
-                                                vertices.stream()
-                                                        .filter(topology::isSource)
-                                                        .filter(
-                                                                v ->
-                                                                        !topology.getFinishedVertices()
-                                                                                .contains(v)));
+                                                vertices.stream().filter(topology::isSource));
                                 newMetricNames.putAll(sourceMetricNames);
                                 return newMetricNames;
                             }
 
                             // Query all metric names
-                            return queryFilteredMetricNames(
-                                    ctx,
-                                    topology,
-                                    vertices.stream()
-                                            .filter(
-                                                    v ->
-                                                            !topology.getFinishedVertices()
-                                                                    .contains(v)));
+                            return queryFilteredMetricNames(ctx, topology, vertices.stream());
                         });
         names.keySet().removeAll(topology.getFinishedVertices());
         return names;
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java
@@ -115,14 +115,8 @@ public abstract class ScalingMetricCollector<KEY, Context extends JobAutoScalerC
 
         // Calculate timestamp when the metric windows is full
         var metricWindowSize = getMetricWindowSize(conf);
-        var metricsAfterStable = metricHistory.tailMap(stableTime);
         var windowFullTime =
-                metricsAfterStable.isEmpty()
-                        ? now.plus(metricWindowSize)
-                        : metricsAfterStable.firstKey().plus(metricWindowSize);
-
-        // Trim metrics outside the metric window from metrics history
-        metricHistory.headMap(now.minus(metricWindowSize)).clear();
+                getWindowFullTime(metricHistory.tailMap(stableTime), now, metricWindowSize);
 
         // The filtered list of metrics we want to query for each vertex
         var filteredVertexMetricNames = queryFilteredMetricNames(ctx, topology);
@@ -147,15 +141,24 @@ public abstract class ScalingMetricCollector<KEY, Context extends JobAutoScalerC
         if (now.isBefore(windowFullTime)) {
             LOG.info("Metric window not full until {}", windowFullTime);
         } else {
-            // We clear any metrics from the stabilization interval once the metric window is full
-            metricHistory.headMap(stableTime).clear();
             collectedMetrics.setFullyCollected(true);
+            // Trim metrics outside the metric window from metrics history
+            metricHistory.headMap(now.minus(metricWindowSize)).clear();
         }
         return collectedMetrics;
     }
 
     protected Duration getMetricWindowSize(Configuration conf) {
         return conf.get(AutoScalerOptions.METRICS_WINDOW);
+    }
+
+    private static Instant getWindowFullTime(
+            SortedMap<Instant, CollectedMetrics> metricsAfterStable,
+            Instant now,
+            Duration metricWindowSize) {
+        return metricsAfterStable.isEmpty()
+                ? now.plus(metricWindowSize)
+                : metricsAfterStable.firstKey().plus(metricWindowSize);
     }
 
     @VisibleForTesting
@@ -287,7 +290,9 @@ public abstract class ScalingMetricCollector<KEY, Context extends JobAutoScalerC
                             observedTprAvg(
                                     jobVertexID,
                                     metricHistory,
-                                    conf.get(AutoScalerOptions.OBSERVED_TPR_MIN_OBSERVATIONS)));
+                                    conf.get(
+                                            AutoScalerOptions
+                                                    .OBSERVED_TRUE_PROCESSING_RATE_MIN_OBSERVATIONS)));
                     vertexScalingMetrics
                             .entrySet()
                             .forEach(e -> e.setValue(ScalingMetrics.roundMetric(e.getValue())));
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java
@@ -59,8 +59,10 @@ import java.util.SortedMap;
 import java.util.TreeMap;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import static org.apache.flink.autoscaler.metrics.ScalingHistoryUtils.updateVertexList;
 import static org.apache.flink.autoscaler.utils.AutoScalerUtils.excludeVerticesFromScaling;
@@ -97,32 +99,30 @@ public abstract class ScalingMetricCollector<KEY, Context extends JobAutoScalerC
                             }
                         });
 
-        // The timestamp of the first metric observation marks the start
-        // If we haven't collected any metrics, we are starting now
-        var metricCollectionStartTs = metricHistory.isEmpty() ? now : metricHistory.firstKey();
-
         var jobDetailsInfo =
                 getJobDetailsInfo(ctx, conf.get(AutoScalerOptions.FLINK_CLIENT_TIMEOUT));
         var jobUpdateTs = getJobUpdateTs(jobDetailsInfo);
-        if (jobUpdateTs.isAfter(metricCollectionStartTs)) {
+        // We detect job change compared to our collected metrics by checking against the earliest
+        // metric timestamp
+        if (!metricHistory.isEmpty() && jobUpdateTs.isAfter(metricHistory.firstKey())) {
             LOG.info("Job updated at {}. Clearing metrics.", jobUpdateTs);
             stateStore.removeEvaluatedMetrics(ctx);
             cleanup(ctx.getJobKey());
             metricHistory.clear();
-            metricCollectionStartTs = now;
         }
         var topology = getJobTopology(ctx, stateStore, jobDetailsInfo);
+        var stableTime = jobUpdateTs.plus(conf.get(AutoScalerOptions.STABILIZATION_INTERVAL));
+
+        // Calculate timestamp when the metric windows is full
+        var metricWindowSize = getMetricWindowSize(conf);
+        var metricsAfterStable = metricHistory.tailMap(stableTime);
+        var windowFullTime =
+                metricsAfterStable.isEmpty()
+                        ? now.plus(metricWindowSize)
+                        : metricsAfterStable.firstKey().plus(metricWindowSize);
 
         // Trim metrics outside the metric window from metrics history
-        var metricWindowSize = getMetricWindowSize(conf);
         metricHistory.headMap(now.minus(metricWindowSize)).clear();
-
-        var stableTime = jobUpdateTs.plus(conf.get(AutoScalerOptions.STABILIZATION_INTERVAL));
-        if (now.isBefore(stableTime)) {
-            // As long as we are stabilizing, collect no metrics at all
-            LOG.info("Skipping metric collection during stabilization period until {}", stableTime);
-            return new CollectedMetricHistory(topology, Collections.emptySortedMap());
-        }
 
         // The filtered list of metrics we want to query for each vertex
         var filteredVertexMetricNames = queryFilteredMetricNames(ctx, topology);
@@ -138,15 +138,19 @@ public abstract class ScalingMetricCollector<KEY, Context extends JobAutoScalerC
         metricHistory.put(now, scalingMetrics);
         stateStore.storeEvaluatedMetrics(ctx, metricHistory);
 
-        var collectedMetrics = new CollectedMetricHistory(topology, metricHistory);
-
-        var windowFullTime = metricCollectionStartTs.plus(metricWindowSize);
-        collectedMetrics.setFullyCollected(!now.isBefore(windowFullTime));
-
-        if (!collectedMetrics.isFullyCollected()) {
-            LOG.info("Metric window not full until {}", windowFullTime);
+        if (now.isBefore(stableTime)) {
+            LOG.info("Stabilizing until {}", stableTime);
+            return new CollectedMetricHistory(topology, Collections.emptySortedMap());
         }
 
+        var collectedMetrics = new CollectedMetricHistory(topology, metricHistory);
+        if (now.isBefore(windowFullTime)) {
+            LOG.info("Metric window not full until {}", windowFullTime);
+        } else {
+            // We clear any metrics from the stabilization interval once the metric window is full
+            metricHistory.headMap(stableTime).clear();
+            collectedMetrics.setFullyCollected(true);
+        }
         return collectedMetrics;
     }
 
@@ -265,9 +269,11 @@ public abstract class ScalingMetricCollector<KEY, Context extends JobAutoScalerC
                     ScalingMetrics.computeLoadMetrics(
                             jobVertexID, vertexFlinkMetrics, vertexScalingMetrics, conf);
 
+                    var metricHistory =
+                            histories.getOrDefault(jobKey, Collections.emptySortedMap());
                     double lagGrowthRate =
                             computeLagGrowthRate(
-                                    jobKey,
+                                    metricHistory,
                                     jobVertexID,
                                     vertexScalingMetrics.get(ScalingMetric.LAG));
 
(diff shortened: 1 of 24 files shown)


1996fanrui:
Hi @gyfora , thanks for fix this issue.

IIUC, the root cause of this problem is that Source Task has two threads:
1. Fetcher thread: fetch data from external system into flink, and cache these data into a buffer or queue
2. Source Task thread: consume data from cached buffer or queue, and process these data.

If one of these two operations is slow, it will cause job lag. Currently we are only compatible with the Source Task thread, not the fetcher thread, right?

I didn't read the code in detail. I have a question after reading the purpose and solution:

> Observed TPR can only be measured when we are catching up (during stabilization) or when cannot keep up.

I doubt this problem can be completely fixed if `Observed TPR` cannot be measured all the time. For example, `Fetcher thread` can fetch 10 records/s, and `Source Task thread` can process 100 records/s, and the input rate of kafka topic is 200 records/s.

The job doesn't have log when parallelism >= 20. Meanwhile, :
- The `BusyRatio` is very low (Each subtask process 10records/s, the BusyRatio = 10/100= 10%)
- And `Observed TPR` cannot be measured. 

The autoscaler will scale down parallelism due to BusyRatio=10%. After scale down, the job has lag, so `Observed TPR` can be measured, then trigger scale up, right?

If so, the autoscaler will scale down and scale up over and over again.

gyfora:
> Hi @gyfora , thanks for fix this issue.
> 
> IIUC, the root cause of this problem is that Source Task has two threads:
> 
> 1. Fetcher thread: fetch data from external system into flink, and cache these data into a buffer or queue
> 2. Source Task thread: consume data from cached buffer or queue, and process these data.
> 
> If one of these two operations is slow, it will cause job lag. Currently we are only compatible with the Source Task thread, not the fetcher thread, right?
> 
> I didn't read the code in detail. I have a question after reading the purpose and solution:
> 
> > Observed TPR can only be measured when we are catching up (during stabilization) or when cannot keep up.
> 
> I doubt this problem can be completely fixed if `Observed TPR` cannot be measured all the time. For example, `Fetcher thread` can fetch 10 records/s, and `Source Task thread` can process 100 records/s, and the input rate of kafka topic is 200 records/s.
> 
> The job doesn't have log when parallelism >= 20. Meanwhile, :
> 
> * The `BusyRatio` is very low (Each subtask process 10records/s, the BusyRatio = 10/100= 10%)
> * And `Observed TPR` cannot be measured.
> 
> The autoscaler will scale down parallelism due to BusyRatio=10%. After scale down, the job has lag, so `Observed TPR` can be measured, then trigger scale up, right?
> 
> If so, the autoscaler will scale down and scale up over and over again.
> 
> Please correct me if my understanding is wrong, thanks~

@1996fanrui 
I think the scenario is mostly covered. In case we scale down due to the too low busy time, we will then start accumulating lag. If enough lag accumulated, the Observed TPR measurement will kick in and will soon rescale the job.

After the job was rescaled it will have a backlog to catch up (we know this because we only triggered the measurements when there was already a lag before the restart) so we will have observed TPR measurements AFTER the restart as well. These measurements will last until the catch up is completed and the value will remain in the collected metrics subsequently so we won't scale down again.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricEvaluator.java
```diff
@@ -154,6 +153,56 @@ private Map<ScalingMetric, EvaluatedScalingMetric> evaluateMetrics(
...
+        }
+
+        if (Double.isNaN(observedTprAvg)) {
+            return TRUE_PROCESSING_RATE;
+        }
+
+        double switchThreshold = conf.get(AutoScalerOptions.OBSERVED_TPR_SWITCH_THRESHOLD);
+        // If we could measure the observed tpr we decide whether to switch to using it
+        // instead of busy time based on the error / difference between the two
+        if (busyTimeTprAvg > observedTprAvg * (1 + switchThreshold)) {
```
mxm, line 193: Should we add a config-based switch here to turn this on / off?

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java
```diff
@@ -97,32 +99,30 @@ public CollectedMetricHistory updateMetrics(
...
+                        : metricsAfterStable.firstKey().plus(metricWindowSize);
 
-        var stableTime = jobUpdateTs.plus(conf.get(AutoScalerOptions.STABILIZATION_INTERVAL));
-        if (now.isBefore(stableTime)) {
-            // As long as we are stabilizing, collect no metrics at all
-            LOG.info("Skipping metric collection during stabilization period until {}", stableTime);
-            return new CollectedMetricHistory(topology, Collections.emptySortedMap());
-        }
+        // Trim metrics outside the metric window from metrics history
+        metricHistory.headMap(now.minus(metricWindowSize)).clear();
```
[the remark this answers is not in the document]
mxm: We could also drop this line entirely and handle metric cleanup in the cleanup block line 102.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java
```diff
@@ -138,15 +138,19 @@ public CollectedMetricHistory updateMetrics(
...
+            return new CollectedMetricHistory(topology, Collections.emptySortedMap());
+        }
 
-        if (!collectedMetrics.isFullyCollected()) {
+        var collectedMetrics = new CollectedMetricHistory(topology, metricHistory);
+        if (now.isBefore(windowFullTime)) {
             LOG.info("Metric window not full until {}", windowFullTime);
+        } else {
+            // We clear any metrics from the stabilization interval once the metric window is full
+            metricHistory.headMap(stableTime).clear();
```
mxm, line 151:
```suggestion
            // Trim metrics outside the metric window from metrics history
            metricHistory.headMap(now.minus(metricWindowSize)).clear();
```

mxm (commented):
I don't know why we would start collecting already in the stabilization period. That breaks the contract we have established. We can already collect metrics during the window built-up period and we can use those for observed rate measurements. Unless there is a fundamental reason, I would like to not alter ScalingMetricCollector.

gyfora:
> I don't know why we would start collecting already in the stabilization period. That breaks the contract we have established. We can already collect metrics during the window built-up period and we can use those for observed rate measurements. Unless there is a fundamental reason, I would like to not alter ScalingMetricCollector.

The reason I changed this is that we rely on lag/catching up job to be able to collect the observed metrics. Many jobs may already catch up during stabilisation period practically rendering the new functionality useless because the observed metric will never be collected. 

This change doesn't actually affect any other metrics / contract that we have established so far. Everything would be evaluated exactly the same way except the newly added observed metric. But by nature I think that makes sense to collect during stabilisation

mxm:
If we catch up during the stabilization period, then there is no overestimation of the true processing rate. I thought this change is to prevent scenarios where we are are overestimating our true processing rate, hence never catching up?

gyfora:
> If we catch up during the stabilization period, then there is no overestimation of the true processing rate. I thought this change is to prevent scenarios where we are are overestimating our true processing rate, hence never catching up?
@mxm 
Imagine the following scenario:
 1. The TPR is overestimated, lag starts to build up
 2. Once there is enough lag, observed TPR kicks in, we rescale
 3. We already catch up during stable
 4. Since there is no observed rate, we use busy time -> we overestimate (which was the original issue)
 5. We scale down due to overestimation -> Go to 1.

mxm:
I missed that we could scale down again after making the right adjustment using the observed rate. LGTM apart from the comments on the diffs.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricEvaluator.java
```diff
@@ -154,6 +153,56 @@ private Map<ScalingMetric, EvaluatedScalingMetric> evaluateMetrics(
...
+            LOG.debug(
```
mxm, line 194: Any reason for logging these as `debug`?

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricEvaluator.java
```diff
@@ -154,6 +153,56 @@ private Map<ScalingMetric, EvaluatedScalingMetric> evaluateMetrics(
...
+                    "Using observed tpr {} for {} as busy time based seems too large ({})",
+                    observedTprAvg,
+                    jobVertexID,
+                    busyTimeTprAvg);
+            return OBSERVED_TPR;
```
mxm, line 199: Any reason for logging as `debug`?

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java
```diff
@@ -97,32 +99,30 @@ public CollectedMetricHistory updateMetrics(
...
 
-        // Trim metrics outside the metric window from metrics history
+        // Calculate timestamp when the metric windows is full
         var metricWindowSize = getMetricWindowSize(conf);
-        metricHistory.headMap(now.minus(metricWindowSize)).clear();
+        var metricsAfterStable = metricHistory.tailMap(stableTime);
+        var windowFullTime =
+                metricsAfterStable.isEmpty()
+                        ? now.plus(metricWindowSize)
+                        : metricsAfterStable.firstKey().plus(metricWindowSize);
```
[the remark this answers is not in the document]
gyfora: good point

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java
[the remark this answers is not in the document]
gyfora: makes sense to simplify

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricEvaluator.java
gyfora: I am not completely sure, the user could set a large lag threshold / switch threshold to effectively disable the feature. But in general if this works correctly I am not sure why anyone would want to disable it, instead of simply tuning the threshold maybe.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
```diff
@@ -151,6 +151,31 @@ private static ConfigOptions.OptionBuilder autoScalerConfig(String key) {
                     .withDescription(
                             "Lag threshold which will prevent unnecessary scalings while removing the pending messages responsible for the lag.");
 
+    public static final ConfigOption<Duration> OBSERVE_TPR_LAG_THRESHOLD =
+            autoScalerConfig("observed-tpr.lag-threshold")
+                    .durationType()
+                    .defaultValue(Duration.ofSeconds(30))
+                    .withDeprecatedKeys(deprecatedOperatorConfigKey("observed-tpr.lag-threshold"))
+                    .withDescription("Lag threshold for enabling observed tpr measurements.");
```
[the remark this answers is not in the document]
gyfora: makes sense

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricEvaluator.java
gyfora: In general we don't log any metric calculations etc, to avoid polluting the logs on a constant loop. So this is inline with other logging in the collector / evaluator.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
```diff
@@ -151,6 +151,36 @@ private static ConfigOptions.OptionBuilder autoScalerConfig(String key) {
                     .withDescription(
                             "Lag threshold which will prevent unnecessary scalings while removing the pending messages responsible for the lag.");
 
+    public static final ConfigOption<Duration> OBSERVE_TRUE_PROCESSING_RATE_LAG_THRESHOLD =
+            autoScalerConfig("observed-true-processing-rate.lag-threshold")
+                    .durationType()
+                    .defaultValue(Duration.ofSeconds(30))
+                    .withDeprecatedKeys(
+                            deprecatedOperatorConfigKey(
+                                    "observed-true-processing-rate.lag-threshold"))
```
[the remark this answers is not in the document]
gyfora: I decided to keep this so that users have a smoother migration path to the new configs.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
```diff
@@ -151,6 +151,36 @@ private static ConfigOptions.OptionBuilder autoScalerConfig(String key) {
                     .withDescription(
                             "Lag threshold which will prevent unnecessary scalings while removing the pending messages responsible for the lag.");
 
+    public static final ConfigOption<Duration> OBSERVE_TRUE_PROCESSING_RATE_LAG_THRESHOLD =
+            autoScalerConfig("observed-true-processing-rate.lag-threshold")
```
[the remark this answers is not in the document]
gyfora: I think observed true processing rate is closer to the practical meaning of this. We always measure the "source-fetch-rate" but the question is what do we do with it and how we compute the true processing rate. 

I don't necessarily think that we have to be extremely specific here that it only applies to sources etc. We should try to convey the practical meaning to users, it's already pretty difficult to understand.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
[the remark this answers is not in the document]
gyfora: I would ideally prefer all config keys to work with the "old" syntax at least in the 1.7.0 release. After that we don't have to define the deprecated keys for new configs but it should be somewhat gradual :)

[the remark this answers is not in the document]
1996fanrui: Thank you, let's respect this rule: `all config keys to work with the "old" syntax at least in the 1.7.0 release`.

mxm (approved):

gyfora:
@1996fanrui @mxm I fixed a state store interface method naming error in https://github.com/apache/flink-kubernetes-operator/pull/686/commits/f1e77a762b34e289bcce6770b3f83574a40f1257

+ I also added a test to guard the changed collection logic during stabilisation

1996fanrui (approved):

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java
```diff
@@ -332,30 +357,51 @@ protected Map<JobVertexID, Map<String, FlinkMetric>> queryFilteredMetricNames(
...
                         });
         names.keySet().removeAll(topology.getFinishedVertices());
         return names;
     }
 
+    private Map<JobVertexID, Map<String, FlinkMetric>> queryFilteredMetricNames(
+            Context ctx, JobTopology topology, Stream<JobVertexID> vertexStream) {
+        try (var restClient = ctx.getRestClusterClient()) {
+            return vertexStream
+                    .filter(v -> !topology.getFinishedVertices().contains(v))
```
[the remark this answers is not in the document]
gyfora: good catch, will remove it

1996fanrui (approved):

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricEvaluator.java
trystanj: old PR, but fwiw I am now seeing a large _under_estimation of capacity:

`Using observed tpr 916327.7340000003 for 1385b0b55fdb909b073b014d9c682612 as busy time based seems too large (2018541.9166666663)`

In fact the actual capacity is indeed around 2M, if not higher. We are on 1.7, but perhaps 1.8 will have improved its observed TPR due to [this enhancement](https://flink.apache.org/2024/03/21/apache-flink-kubernetes-operator-1.8.0-release-announcement/#improved-accuracy-of-autoscaling-metrics)?

gyfora: The observed true processing rate is simply measured based on the num records processed when the job is catching up. Why do you think that your job as 2M/sec processing capacity?

Maybe when the job is catching up it cannot go 2M / sec ?

trystanj: It's possible if I am misunderstanding something!

During the same time that log was printed during a downscale, I have metrics showing that actual `flink_taskmanager_job_task_operator_numRecordsIn` (per sec) was above 1.2M / sec for 15 minutes before and after, and peaked at about 2.8M for around 3 minutes. The peak was due to the rescale.

It caused it to severely underestimate the capacity, scale way back up (higher than it was before the downscale), and then eventually settle back down to where it tried to go initially. At that point it was no longer seeing the discrepancy between observed and busy-time-based.

I am guessing (but probably wrong 😄) that it has more to do with _when_ it's measuring the "observed" TPR than it being wrong? Like it perhaps misses the window where it's happening. Our restart time is 30s, stabilization time 5m, window 15m. The whole recovery lasted maybe 3m. If those times are not aligned with reality (it likely restarted much quicker than 30s) then could it cause it to miss the peak observation TPR?

gyfora: Observed TPR is a bit of outlier metric for a "good" reason. It is actually computed regardless of the stabilisation time as you can see [here](https://github.com/apache/flink-kubernetes-operator/blob/main/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/metrics/ScalingMetrics.java#L104). There is a config threshold for the observed tpr computation which sets a minimum source lag when this can be computed (set at 30 seconds by default)

The rationale is that you can only really observe the true processing rate when the job is catching up (going full speed). So that's the reason here, maybe you can set that threshold a bit higher (1m) which would mean that we measure only when there is at least 1 minute lag. Higher threshold will mean more accurate measurement when it's triggered but less overall datapoints.

gyfora: If your job catches up super quick then you may find anomalies for these measurements because then you will likely only have 1-2 datapoints at most and that can be off arbitrarily

trystanj: Great idea, thanks for the tip! I saw that config setting but did not understand what it was actually used for. I will experiment with increasing that setting - our jobs are all over the place in how quickly they recover (some are a few seconds, some are several minutes) but I would bias towards accuracy in order to avoid the flapping in our case.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/metrics/ScalingMetrics.java
```diff
@@ -78,6 +88,45 @@ public static void computeDataRateMetrics(
...
+
+        double observedTpr =
+                computeObservedTprWithBackpressure(
+                        numRecordsInPerSecond,
+                        flinkMetrics.get(FlinkMetric.BACKPRESSURE_TIME_PER_SEC).getAvg());
+
+        return Double.isNaN(observedTpr) ? Optional.empty() : Optional.of(observedTpr);
+    }
+
+    public static double computeObservedTprWithBackpressure(
```
sharath1709, line 121: This is an old PR but why is backpressure taken into account to compute ObservedTpr? NumRecordsInPerSecond it self should be the observedTpr since the job is already under lag. I understand the theoretical reason for computing it as numRecordsInPerSecond/(1 - backpressuredFraction). However, clearly the job is not able to process the data at this rate as evidenced by numRecordsInPerSecond due to whatever reason.

mxm: True processing rate is the max processing rate in the absence of backpressure. When we use the task busyness metric, we compute it as `throughput / (busy_time_ms_per_second / 1000)`. Here, we have only the throughput which we assume to be the maximum throughput. When there is backpressure, we need to factor it out to correctly compute the true processing rate. We do this in both approaches, once by using the busy time metric, once by using the backpressure metric. 

The assumption is that the backpressure comes from downstream and that we can eliminate it in the next scaling, by resizing downstream operators, and scale up to the true processing rate.

sharath1709: Semantically, I agree that the true processing rate should be defined as the maximum rate in the absence of backpressure. However, the ObservedTpr is specifically used to estimate catch-up time. This is guaranteed to be inaccurate, since the job is already experiencing backpressure and the actual catch-up time will be higher than what the autoscaler computes. As a result, the autoscaler fails to meet the SLA defined by job.autoscaler.catch-up.duration

sharath1709: I understand that job.autoscaler.catch-up.duration refers to the time to catch up only after the scaling operation. However, I find the assumption that backpressure can always be eliminated in the next scaling to be inaccurate. In practice, I’ve observed cases where a task remains backpressured even when the downstream task shows low utilization, depending on the task’s business logic.

gyfora: observed TPR has not much to do with catch up time. It was introduced specifically to compute processing capacity (TPR) correctly for sources in cases when the busy time reported for Flink is completely off. You can read more about this in the description

sharath1709: Let me clarify my statement. Catch-up Time is used to [calculate](https://github.com/apache/flink-kubernetes-operator/blob/b72ad7f23658c29939682cffff77bf737fe7a1ba/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/utils/AutoScalerUtils.java#L45) the target processing capacity. The true processing capacity is then [calculated](https://github.com/apache/flink-kubernetes-operator/blob/b72ad7f23658c29939682cffff77bf737fe7a1ba/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricEvaluator.java#L223) based on either busy time or backpressure based on certain configurations.

Next, the scale factor is [computed](https://github.com/apache/flink-kubernetes-operator/blame/b72ad7f23658c29939682cffff77bf737fe7a1ba/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java#L186) as:

double scaleFactor = targetCapacity / averageTrueProcessingRate;


If these values match, the parallelism for the vertex remains unchanged. In effect, both the autoscaler and the user assume that the backlog will be cleared within job.autoscaler.catch-up.duration (per the description: “The target duration for fully processing any backlog after a scaling operation. Set to 0 to disable backlog-based scaling.”).

However, in practice, the catch-up time almost always exceeds job.autoscaler.catch-up.duration, since the actual consumption rate is lower than the computed true processing capacity. This makes it difficult to reason about the guarantees provided by the autoscaler.

gyfora: I agree that the name catch up duration is a bit incorrect, we can fix the name or tweak the behaviour but I would like to continue this discussion on a mailing list / JIRA instead. These old PRs are not the place for it :)

sharath1709: Agreed. I will move the discussion to JIRA/mailing list

PR #689, opened 2023-10-25
[Flink 31966] Flink Kubernetes operator lacks TLS support

tagarr:
If the FlinkDeployment CR is configured to use ssl, the operator is unable to properly reconcile the deployment as the underlying RestClient the operator uses tries to load the certificates defined in the config.

## Brief change log
  - When rest cluster client is created (in AbstractFlinkService) the config is checked to see if rest ssl is configured. If it is, find the relevant secret defined in the kubernetes.secret config and copy the secret to a local directory in the location of /tmp/operator/certs/{namespace}/{instanceName}. Copied config is then modified to point to the new ssl cert location before the rest client is created.
  - Creation of a new OperatorKubernetesClusterDescriptor so that a restclient created in the operator can be passed to it and prevent the creation of a restclient within the private createClusterClientProvider method
  - Modify the KubernetesStandaloneClusterDescriptor to be able to pass in a RestClient in same way as the above class

## Verifying this change

  - Added new tests for the new OperatorKubernetesClusterDescriptor class
  - Added a SecureFlinkServiceTest that checks that the copied certs are deleted when the cluster is deleted, that data in the secrets are copied into the relevant directory of the operator and additional test when relevant file does not exist or secret does not exist
  - Added new basic-secure.yaml example that provides a secured flink application and updated relevant README 

  - Core observer or reconciler logic that is regularly executed: (no) N.b. this only modifies the Flink service used by the observers and reconcilers

## Documentation

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -100,6 +100,17 @@ To try out this run the following command:
 kubectl apply -f basic-session-deployment-and-job.yaml
 ```
 
+#### Basic secured application
+
+This sample uses cert-manager to create a self-signed Issuer and jks certificate to secure the internal and rest 
+communication between pods. N.b. you may see RuntimeExceptions in the operator logs whilst the cert-manger creates the
+ssl certificate secret
+
+To try out this run the following command:
+```bash
+kubectl apply -f basic-secure.yaml
+```
+
 ### SQL runner
 
 For running Flink SQL scripts check this [example](flink-sql-runner-example/README.md).
--- a/examples/basic-secure.yaml
+++ b/examples/basic-secure.yaml
@@ -1,0 +1,84 @@
+################################################################################
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+# limitations under the License.
+################################################################################
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: basic-secure-cert-secret-creds
+type: Opaque
+data:
+  password: cGFzc3dvcmQxMjM0
+---
+# Source: flink-kubernetes-operator/templates/webhook.yaml
+apiVersion: cert-manager.io/v1
+kind: Issuer
+metadata:
+  name: basic-secure-selfsigned-issuer
+spec:
+  selfSigned: {}
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: basic-secure-cert
+spec:
+  dnsNames:
+    - '*.flink.svc'
+    - '*.svc.cluster.local'
+  keystores:
+    jks:
+      create: true
+      passwordSecretRef:
+        name: basic-secure-cert-secret-creds
+        key: password
+  issuerRef:
+    kind: Issuer
+    name: basic-secure-selfsigned-issuer
+  commonName: FlinkDeployment
+  secretName: basic-secure-cert
+---
+apiVersion: flink.apache.org/v1beta1
+kind: FlinkDeployment
+metadata:
+  name: basic-secure
+spec:
+  image: flink:1.17
+  flinkVersion: v1_17
+  flinkConfiguration:
+    taskmanager.numberOfTaskSlots: "2"
+    security.ssl.enabled: 'true'
+    security.ssl.truststore: /etc/tls/truststore.jks
+    security.ssl.truststore-password: password1234
+    security.ssl.keystore: /etc/tls/keystore.jks
+    security.ssl.keystore-password: password1234
+    security.ssl.key-password: password1234
+    kubernetes.secrets: 'basic-secure-cert:/etc/tls'
+  serviceAccount: flink
+  jobManager:
+    resource:
+      memory: "2048m"
+      cpu: 1
+  taskManager:
+    resource:
+      memory: "2048m"
+      cpu: 1
+  job:
+    jarURI: local:///opt/flink/examples/streaming/StateMachineExample.jar
+    parallelism: 2
+    upgradeMode: stateless
+---
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
@@ -193,9 +193,9 @@ public abstract class AbstractFlinkService implements FlinkService {
 
         // If Kubernetes or Zookeeper HA are activated, delete the job graph in HA storage so that
         // the newly changed job config (e.g. parallelism) could take effect
-        final String clusterId = conf.get(KubernetesConfigOptions.CLUSTER_ID);
-        final String namespace = conf.get(KubernetesConfigOptions.NAMESPACE);
         if (FlinkUtils.isKubernetesHAActivated(conf)) {
+            final String clusterId = conf.get(KubernetesConfigOptions.CLUSTER_ID);
+            final String namespace = conf.get(KubernetesConfigOptions.NAMESPACE);
             FlinkUtils.deleteJobGraphInKubernetesHA(clusterId, namespace, kubernetesClient);
         } else if (FlinkUtils.isZookeeperHAActivated(conf)) {
             FlinkUtils.deleteJobGraphInZookeeperHA(conf);
@@ -1041,7 +1041,9 @@ public abstract class AbstractFlinkService implements FlinkService {
             final String namespace = conf.get(KubernetesConfigOptions.NAMESPACE);
             try {
                 Path certPath = Paths.get(CERT_DIR, namespace, clusterId);
-                FileUtils.deleteDirectory(certPath.toFile());
+                if (Files.exists(certPath)) {
+                    FileUtils.deleteDirectory(certPath.toFile());
+                }
             } catch (IOException e) {
                 throw new RuntimeException(e);
             }
@@ -1097,6 +1099,10 @@ public abstract class AbstractFlinkService implements FlinkService {
     }
 
     private void copyTLSFiles(Configuration conf) throws IOException {
+        if (!conf.contains(KubernetesConfigOptions.KUBERNETES_SECRETS)) {
+            LOG.info("No secret mount found for ssl config");
+            return;
+        }
         final String clusterId = conf.get(KubernetesConfigOptions.CLUSTER_ID);
         final String namespace = conf.get(KubernetesConfigOptions.NAMESPACE);
         ConfigOption<String> configKeystore =
@@ -1133,10 +1139,7 @@ public abstract class AbstractFlinkService implements FlinkService {
         // make sure the parent directories exist
         Files.createDirectories(targetPath.getParent());
         Map<String, String> secretMounts = conf.get(KubernetesConfigOptions.KUBERNETES_SECRETS);
-        if (secretMounts == null) {
-            LOG.warn("No secret mount found for ssl config");
-            return;
-        }
+        // find the secret that provides the certificate mount point
         Optional<String> certSecretName =
                 secretMounts.entrySet().stream()
                         .filter(e -> Paths.get(e.getValue()).equals(certStorePath.getParent()))
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
@@ -101,6 +101,7 @@ import org.apache.flink.util.Preconditions;
 import io.fabric8.kubernetes.api.model.DeletionPropagation;
 import io.fabric8.kubernetes.api.model.ObjectMeta;
 import io.fabric8.kubernetes.api.model.PodList;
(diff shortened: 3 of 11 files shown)


tagarr:
Hi @gyfora what are your thoughts on the PR ? I guess the most contentious change is the introduction of the OperatorKubernetesClusterDescriptor class that I need so I can override the private method  createClusterClientProvider to prevent it creating a RestClusterClient

mbalassi:
Hi @tagarr! Thanks for looking into this. Many Flink committers are traveling this week and next week due to Flink Forward Seattle happening. I will ask @gaborgsomogyi to review this for you when he is back from vacation next week.

gaborgsomogyi:
Hi @tagarr ! Thanks for your efforts! Before we go deeply in the feature I would like to understand the problem statement clearly. If I understand correctly then the issue what we would like to solve here is the following:
* The operator submits a Flink workload which has the following config (amongst others)
```
    security.ssl.enabled: 'true'
    security.ssl.truststore: /etc/tls/truststore.jks
    security.ssl.truststore-password: password1234
    security.ssl.keystore: /etc/tls/keystore.jks
    security.ssl.keystore-password: password1234
    security.ssl.key-password: password1234
```
* The Flink app uses the mentioned configs and switches to secure mode (keystore for server side and truststore for client side)
* The operator is creating a REST client which uses the following configs to make the connection with the workload (keystore part is not used since it's only needed for server side):
```
    security.ssl.enabled: 'true'
    security.ssl.truststore: /etc/tls/truststore.jks
    security.ssl.truststore-password: password1234
```
* Since the operator doesn't have the mentioned file the communication fails

Have I understood the issue correctly?

gaborgsomogyi:
Then why not simply mounting `/etc/tls/` in the operator and put certs in the proper dir?
I mean one can add configs like:
```
    security.ssl.enabled: 'true'
    security.ssl.truststore: /etc/tls/{namespace}/{instanceName}/truststore.jks
    security.ssl.truststore-password: password1234
    security.ssl.keystore: /etc/tls/{namespace}/{instanceName}/keystore.jks
    security.ssl.keystore-password: password1234
    security.ssl.key-password: password1234
```
At the moment I not yet see the 1k line addition🤔

There is some generic comments apart from the use-case (correct me if anything understood in a bad way):
* If I understand correctly it's always checking the cert existence when a new REST client is created. Let's assume the operator died, it's resurrecting so the local disk doesn't have any certs and there are 10k+ Flink jobs. I think it will be extremely slow to recover.
* Can multiple threads write the same file locally when they run on the same cluster?
* Several code parts are just not having single responsibility. An example is that `createLocalFile` is changing the Flink config as a side effect.
* Maybe I've just done a super shallow first round but I don't see a assertion that the REST client is using SSL.
* There are unnecessary copy-pastes in the tests.

tagarr:
@gaborgsomogyi Thanks for looking at this.
So the problem is that user declared flinkdeployments could specify the cert directory wherever they want to, I didn't want to lock down the directory to a known location. As this location could be anywhere, just using the same location is therefore not possible. Also as our flink operator containers are root read-only mounted I decided to put them in /tmp and provide a naming convention to ensure multiple flinkdeployments would not overwrite the same file. 

Regarding your other points.

- I doubt the operator would be managing 100's of deployments let another orders of magnitude greater than that. I also wasn't sure about permanently persisting them as if the cert is rotated I would need a mechanism to renew the cert, with this method you just need to roll the pod to refresh the certs
- I only wanted to change the config if I have correctly found and created the file, hence setting it in the createLocalFile method. This way if users are doing a different way to mount their certs I wouldn't be overriding them
- The operator will need to use the restclient for many of its functions i.e. listJobs, cancelJobs, triggerSavepoints e.t.c. without the correct ssl the operator won't be able to do this. I have provided a sample, why don't you try it on the current codebase unless I'm misunderstanding your question
- I will tidy up the tests but would like confirmation that the methodology I'm using is acceptable to the community

gaborgsomogyi:
Please see my comments inline:

> So the problem is that user declared flinkdeployments could specify the cert directory wherever they want to, I didn't want to lock down the directory to a known location. As this location could be anywhere, just using the same location is therefore not possible. Also as our flink operator containers are root read-only mounted I decided to put them in /tmp and provide a naming convention to ensure multiple flinkdeployments would not overwrite the same file.

I think the operator can use a single root CA as truststore generated per k8s cluster (or some tiny changes in the granularity) and the workloads can use this root CA as signing CA for keystores.

> I doubt the operator would be managing 100's of deployments let another orders of magnitude greater than that. I also wasn't sure about permanently persisting them as if the cert is rotated I would need a mechanism to renew the cert, with this method you just need to roll the pod to refresh the certs

Doubt or not at big companies the operator is handling extreme amount of jobs so I'm not convinced with the actual proposal in this area.

> I only wanted to change the config if I have correctly found and created the file, hence setting it in the createLocalFile method. This way if users are doing a different way to mount their certs I wouldn't be overriding them

I'm speaking about structural code patterns and not concrete things. To be crystal clear when I see a function called `createLocalFile` then I expect such function to do exactly one thing, which is creating a local file based on parameters and nothing else. I know why the config is changed but that must be done in another place of the code or the function can be renamed. This applies to all code parts.

> The operator will need to use the restclient for many of its functions i.e. listJobs, cancelJobs, triggerSavepoints e.t.c. without the correct ssl the operator won't be able to do this. I have provided a sample, why don't you try it on the current codebase unless I'm misunderstanding your question

Please provide me the line where your new test code asserts on that the REST client is using SSL and then we're fine.

> I will tidy up the tests but would like confirmation that the methodology I'm using is acceptable to the community

Not sure what you mean `methodology I'm using`.
What I suggest is to add automated tests for each added functionality including positive and negative test cases too. I've not checked the coverage provided by the tests but I'm pretty sure copy-paste is a no go.

tagarr:
Hi @gaborgsomogyi I didn't appreciate that there could be 100's or more of FlinkDeployments running on a cluster. If that's the case then my solution wouldn't be the best. What if I provide an optional secret mount for the truststore and optional secretKeyRef env var for the store password. Then modify the config for creating the rest client for the operator to point to this store ?

Additionally, I only created the OperatorKubernetesClusterDescriptor class so that the call to deployClusterInternal didn't throw an exception as the config for the actual cluster was being used. If instead of doing this I caught the exception and checked that it was a ClusterRetrieveException I would be able to reduce the changes considerably. Do you think this would be acceptable ?

For reference the exception thrown by the operator is:
```
[m[33m2023-11-08 15:50:27,797[m [36mo.a.f.k.o.l.AuditUtils        [m [32m[INFO ][flink/basic-secure] >>> Event  | Warning | CLUSTERDEPLOYMENTEXCEPTION | org.apache.flink.client.deployment.ClusterRetrieveException: Could not create the RestClusterClient.
[m[33m2023-11-08 15:50:27,800[m [36mo.a.f.k.o.r.ReconciliationUtils[m [33m[WARN ][flink/basic-secure] Attempt count: 0, last attempt: false
[m[33m2023-11-08 15:50:27,886[m [36mo.a.f.k.o.l.AuditUtils        [m [32m[INFO ][flink/basic-secure] >>> Status | Error   | UPGRADING       | {"type":"org.apache.flink.kubernetes.operator.exception.ReconciliationException","message":"java.lang.RuntimeException: org.apache.flink.client.deployment.ClusterRetrieveException: Could not create the RestClusterClient.","additionalMetadata":{},"throwableList":[{"type":"java.lang.RuntimeException","message":"org.apache.flink.client.deployment.ClusterRetrieveException: Could not create the RestClusterClient.","additionalMetadata":{}},{"type":"org.apache.flink.client.deployment.ClusterRetrieveException","message":"Could not create the RestClusterClient.","additionalMetadata":{}}]} 
[m[33m2023-11-08 15:50:27,890[m [36mi.j.o.p.e.ReconciliationDispatcher[m [1;31m[ERROR][flink/basic-secure] Error during event processing ExecutionScope{ resource id: ResourceID{name='basic-secure', namespace='flink'}, version: 7366433} failed.
org.apache.flink.kubernetes.operator.exception.ReconciliationException: java.lang.RuntimeException: org.apache.flink.client.deployment.ClusterRetrieveException: Could not create the RestClusterClient.
	at org.apache.flink.kubernetes.operator.controller.FlinkDeploymentController.reconcile(FlinkDeploymentController.java:148)
	at org.apache.flink.kubernetes.operator.controller.FlinkDeploymentController.reconcile(FlinkDeploymentController.java:56)
	at io.javaoperatorsdk.operator.processing.Controller$1.execute(Controller.java:138)
	at io.javaoperatorsdk.operator.processing.Controller$1.execute(Controller.java:96)
	at org.apache.flink.kubernetes.operator.metrics.OperatorJosdkMetrics.timeControllerExecution(OperatorJosdkMetrics.java:80)
	at io.javaoperatorsdk.operator.processing.Controller.reconcile(Controller.java:95)
	at io.javaoperatorsdk.operator.processing.event.ReconciliationDispatcher.reconcileExecution(ReconciliationDispatcher.java:139)
	at io.javaoperatorsdk.operator.processing.event.ReconciliationDispatcher.handleReconcile(ReconciliationDispatcher.java:119)
	at io.javaoperatorsdk.operator.processing.event.ReconciliationDispatcher.handleDispatch(ReconciliationDispatcher.java:89)
	at io.javaoperatorsdk.operator.processing.event.ReconciliationDispatcher.handleExecution(ReconciliationDispatcher.java:62)
	at io.javaoperatorsdk.operator.processing.event.EventProcessor$ReconcilerExecutor.run(EventProcessor.java:414)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
	at java.base/java.lang.Thread.run(Unknown Source)
Caused by: java.lang.RuntimeException: org.apache.flink.client.deployment.ClusterRetrieveException: Could not create the RestClusterClient.
	at org.apache.flink.kubernetes.KubernetesClusterDescriptor.lambda$createClusterClientProvider$1(KubernetesClusterDescriptor.java:121)
	at org.apache.flink.kubernetes.KubernetesClusterDescriptor.deployApplicationCluster(KubernetesClusterDescriptor.java:217)
	at org.apache.flink.client.deployment.application.cli.ApplicationClusterDeployer.run(ApplicationClusterDeployer.java:67)
	at org.apache.flink.kubernetes.operator.service.NativeFlinkService.deployApplicationCluster(NativeFlinkService.java:104)
	at org.apache.flink.kubernetes.operator.service.AbstractFlinkService.submitApplicationCluster(AbstractFlinkService.java:189)
	at org.apache.flink.kubernetes.operator.reconciler.deployment.ApplicationReconciler.deploy(ApplicationReconciler.java:182)
	at org.apache.flink.kubernetes.operator.reconciler.deployment.ApplicationReconciler.deploy(ApplicationReconciler.java:60)
	at org.apache.flink.kubernetes.operator.reconciler.deployment.AbstractFlinkResourceReconciler.reconcile(AbstractFlinkResourceReconciler.java:120)
	at org.apache.flink.kubernetes.operator.controller.FlinkDeploymentController.reconcile(FlinkDeploymentController.java:136)
	... 13 more
Caused by: org.apache.flink.client.deployment.ClusterRetrieveException: Could not create the RestClusterClient.
	... 22 more
Caused by: org.apache.flink.util.ConfigurationException: Failed to initialize SSLContext for the REST client
	at org.apache.flink.runtime.rest.RestClientConfiguration.fromConfiguration(RestClientConfiguration.java:107)
	at org.apache.flink.client.program.rest.RestClusterClientConfiguration.fromConfiguration(RestClusterClientConfiguration.java:78)
	at org.apache.flink.client.program.rest.RestClusterClient.<init>(RestClusterClient.java:225)
	at org.apache.flink.client.program.rest.RestClusterClient.<init>(RestClusterClient.java:197)
	at org.apache.flink.kubernetes.KubernetesClusterDescriptor.lambda$createClusterClientProvider$1(KubernetesClusterDescriptor.java:114)
	... 21 more
Caused by: java.nio.file.NoSuchFileException: /etc/tls/truststore.jks
	at java.base/sun.nio.fs.UnixException.translateToIOException(Unknown Source)
	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(Unknown Source)
	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(Unknown Source)
	at java.base/sun.nio.fs.UnixFileSystemProvider.newByteChannel(Unknown Source)
	at java.base/java.nio.file.Files.newByteChannel(Unknown Source)
	at java.base/java.nio.file.Files.newByteChannel(Unknown Source)
	at java.base/java.nio.file.spi.FileSystemProvider.newInputStream(Unknown Source)
	at java.base/java.nio.file.Files.newInputStream(Unknown Source)
	at org.apache.flink.runtime.net.SSLUtils.getTrustManagerFactory(SSLUtils.java:218)
	at org.apache.flink.runtime.net.SSLUtils.createRestNettySSLContext(SSLUtils.java:400)
	at org.apache.flink.runtime.net.SSLUtils.createRestNettySSLContext(SSLUtils.java:367)
	at org.apache.flink.runtime.net.SSLUtils.createRestClientSSLEngineFactory(SSLUtils.java:161)
	at org.apache.flink.runtime.rest.RestClientConfiguration.fromConfiguration(RestClientConfiguration.java:105)
	... 25 more

```

gaborgsomogyi:
What you're suggesting sounds better. I think we need 2 additional parameters:
* `...trustore`
* `...trustore-password`

Not yet understand why keystore is needed, if you can elaborate we can be more on the same page.
In my understanding trustrore is used in client scenarios and keystore is used in server scenarios.
Since I'm not aware where the operator is behaving as server I don't see the need to add any keystore configs.
If you know stuff like that please point on to discuss.

The mount part is not yet clear why we would like to add any config for that. Until now it's designed to add
any custom mount in the helm chart which is still available. I would use that.

tagarr:
You're right truststore would normally only be needed, it depends on if we want to cater for mTLS. If we did we would need to include keystore info. 
Quick question if we wanted to cater for mTLS would you cater for truststore/keystore potentially being in different secrets or document they should be in the same secret ?

gaborgsomogyi:
Makes sense, mTLS is one reason what justifies to add the remaining 3 keystore related configs.

gaborgsomogyi:
Just for the reference for others: https://nightlies.apache.org/flink/flink-docs-master/docs/deployment/security/security-ssl/#rest-endpoints-external-connectivity

```
If mutual authentication is enabled, the keystore and the truststore are used by both, the server endpoint and the REST clients as with internal connectivity.
```

tagarr:
@gaborgsomogyi I'm going to close this PR and submit a new one with all the new changes

PR #691, opened 2023-10-25
[FLINK-33359][FLINK-25002] add-opens jvm args to support jdk17 for kubernetes operator

snuyanzin:
The PR applies similar logic as done for FLINK-25002 by introducing `surefire.module.config` to set module related `add-opens`/`add-exports` properties required by jdk17+

to support jdk17

## Brief change log

pom files

## Verifying this change

## Documentation

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,6 +73,7 @@ jobs:
         version: ["v1_18","v1_17","v1_16","v1_15","v1_14","v1_13"]
         namespace: ["default","flink"]
         mode: ["native", "standalone"]
+        java-version: [ 11, 17 ]
         test:
           - test_application_kubernetes_ha.sh
           - test_application_operations.sh
@@ -128,10 +129,10 @@ jobs:
     name: e2e_ci
     steps:
       - uses: actions/checkout@v2
-      - name: Set up JDK 11
+      - name: Set up JDK ${{ matrix.java-version }}
         uses: actions/setup-java@v2
         with:
-          java-version: '11'
+          java-version: ${{ matrix.java-version }}
           distribution: 'adopt'
       - name: Cache local Maven repository
         uses: actions/cache@v3
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,12 +24,15 @@ jobs:
   test_ci:
     runs-on: ubuntu-latest
     name: test_ci
+    strategy:
+      matrix:
+        java-version: [ 11, 17 ]
     steps:
       - uses: actions/checkout@v2
-      - name: Set up JDK 11
+      - name: Set up JDK ${{ matrix.java-version }}
         uses: actions/setup-java@v2
         with:
-          java-version: '11'
+          java-version: ${{ matrix.java-version }}
           distribution: 'adopt'
       - name: Cache local Maven repository
         uses: actions/cache@v3
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -126,6 +126,16 @@ jobs:
             test: test_dynamic_config.sh
           - version: v1_17
             test: test_dynamic_config.sh
+          - version: v1_13
+            java-version: 17
+          - version: v1_14
+            java-version: 17
+          - version: v1_15
+            java-version: 17
+          - version: v1_16
+            java-version: 17
+          - version: v1_17
+            java-version: 17
     name: e2e_ci
     steps:
       - uses: actions/checkout@v2
--- a/flink-autoscaler/pom.xml
+++ b/flink-autoscaler/pom.xml
@@ -31,6 +31,13 @@ under the License.
     <name>Flink Autoscaler</name>
     <packaging>jar</packaging>
 
+    <properties>
+        <surefire.module.config> <!-- required by
+            JobTopologyTest
+			-->--add-opens=java.base/java.lang=ALL-UNNAMED
+        </surefire.module.config>
+    </properties>
+
     <dependencies>
         <dependency>
             <groupId>org.apache.flink</groupId>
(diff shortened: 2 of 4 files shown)


snuyanzin:
@gyfora , @1996fanrui this makes kubernetes operator compiling with tests on java 17
also adds java 17 to ci for `test_ci` not sure whether need to add it for e2e

flink-autoscaler/pom.xml
```diff
@@ -31,6 +31,13 @@ under the License.
     <name>Flink Autoscaler</name>
     <packaging>jar</packaging>
 
+    <properties>
+        <surefire.module.config> <!-- required by
+        JobTopologyTest
+        -->--add-opens=java.base/java.lang=ALL-UNNAMED
+        </surefire.module.config>
```
1996fanrui, line 38:
```suggestion
        <!-- required by JobTopologyTest -->
        <surefire.module.config>--add-opens=java.base/java.lang=ALL-UNNAMED</surefire.module.config>
```

How about this? It's easy to read.

flink-kubernetes-operator/pom.xml
```diff
@@ -34,6 +34,10 @@ under the License.
     <properties>
         <awaitility.version>4.1.0</awaitility.version>
         <plugins.tmp.dir>${project.build.directory}/plugins</plugins.tmp.dir>
+        <surefire.module.config> <!-- required by
+        FlinkConfigManagerTest
+        -->--add-opens=java.base/java.util=ALL-UNNAMED
+        </surefire.module.config>
```
1996fanrui, line 40:
```suggestion
        <!-- required by FlinkConfigManagerTest -->
        <surefire.module.config> --add-opens=java.base/java.util=ALL-UNNAMED </surefire.module.config>
```

1996fanrui (approved):
Thanks @snuyanzin  for the contribution!

I have built it with java 17 on my Mac, it works well, great job!

I left 2 minor comments, please take a look in your free time, thanks~

gyfora:
@snuyanzin can you please merge your PRs related to Java 17 compatibility into one? It's not good to have many small commits that are related to the same jira

gyfora:
It would be great to have a single PR after which we have java 17 compatibility

flink-autoscaler/pom.xml
snuyanzin: in case of just one flag it probably makes sense

flink-kubernetes-operator/pom.xml
snuyanzin: in case of just one flag it probably makes sense

snuyanzin:
@gyfora i added here #693 
please have  a look once you have time

.github/workflows/ci.yml
```diff
@@ -24,12 +24,15 @@ jobs:
   test_ci:
     runs-on: ubuntu-latest
     name: test_ci
+    strategy:
+      matrix:
+        java-version: [ 11, 17 ]
```
gyfora, line 29: In addition to running the CI with both java version it would be great to somehow run e2e-s as well on different versions.

That may be a bit trickier because we would have to change the docker base image etc. What do you think? 
If it's a lot of work we can open a follow-up ticket for that as well , but we need that before we can say java 17 is supported

snuyanzin: I made it for my own fork, so it should not be complecated.

The only thing that stopped me from adding this to the PR is that after that there are 200+ jobs running in ci... and the total time is more than an hour

If it's ok, then i can add it here as well

gyfora: We should not run it for all jobs. We can target only the latest Flink version (1.18) for now

snuyanzin: ok, now done for 1.18 only

gyfora:
looks good! thank you :)

melin:
The spark operator can automatically add the "add-opens" JVM arguments based on the JDK version. Can Flink also automatically add these?

image

Dennis-Mircea:
> The spark operator can automatically add the "add-opens" JVM arguments based on the JDK version. Can Flink also automatically add these?
> 
> image

Hi @melin, two things, since this PR and your question are actually about different layers.

**This PR (test-time)**: PR #691 added `--add-opens` for the Maven Surefire JVM via the `surefire.module.config` property, so the operator's own unit tests pass on JDK 17. That setup is still in place today in `pom.xml`, `flink-autoscaler/pom.xml`, and `flink-kubernetes-operator/pom.xml`. The only thing that has shifted since the PR was opened is the CI JDK baseline, which moved from `[11, 17]` to `[17, 21]` (`.github/workflows/ci.yml`). No automatic JDK detection is needed here because the build's supported JDK range is fixed by CI.

**Your question (runtime, managed Flink jobs)**: For the JM/TM JVMs of managed `FlinkDeployment`s, the operator already auto-applies the JDK 17 `--add-exports` / `--add-opens` set, keyed on the Flink version, via the chart's default `flink-conf.yaml` ([conf/flink-conf.yaml#L26-L27](https://github.com/apache/flink-kubernetes-operator/blob/main/helm/flink-kubernetes-operator/conf/flink-conf.yaml#L26-L27)):

```yaml
kubernetes.operator.default-configuration.flink-version.v1_18.env.java.opts.all: ...
kubernetes.operator.default-configuration.flink-version.v1_19+.env.java.default-opts.all: ...
```

Keying on Flink version is sufficient because each version has a fixed supported JDK baseline (1.18+ requires JDK 11+), so the flags are always safe to apply.

The 1.18 vs 1.19+ split also handles user overrides cleanly:
- 1.18 (`env.java.opts.all`) is fully overridden if the user sets their own, so they must re-include the flags.
- 1.19+ (`env.java.default-opts.all`) is appended by Flink, so user-supplied `opts.all` keeps the defaults.

The operator JVM itself uses a separate knob (`jvmArgs.operator`), empty by default, since the operator code runs cleanly on JDK 17 without these flags.

PR #698, opened 2023-10-31
[FLINK-33099][autoscaler] Introduce the Standalone Autoscaler and support flink cluster

1996fanrui:

[FLINK-33099][autoscaler] Introduce the Standalone Autoscaler and support flink cluster

## Brief change log

- Introduce the Standalone Autoscaler
- Implement `LoggableEventHandler` and `RescaleApiScalingRealizer`
- Support flink cluster: `FlinkClusterJobListFetcher`

## Verifying this change

Still developing the test.

## Documentation

  - If yes, how is the feature documented? docs in the next PR


gyfora:
I think the standalone implementation and the new interfaces / tests etc specific to that should be in a different maven module (`flink-autoscaler-standalone`). That way we wouldn't pollute the core implementation with unnecessary interfaces such as the job list etc.

If we are not sure about the stability we could even put this under the examples for now but that may be an overkill.

What do you think @mxm ?

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/standalone/JobListFetcher.java
```diff
@@ -0,0 +1,31 @@
...
+import org.apache.flink.autoscaler.JobAutoScalerContext;
+import org.apache.flink.metrics.MetricGroup;
+
+import java.util.List;
+
+/** The JobListFetcher will fetch the jobContext of all jobs. */
+@Experimental
+public interface JobListFetcher<KEY> {
+
+    List<JobAutoScalerContext<KEY>> fetch(MetricGroup metricGroup) throws Exception;
```
gyfora, line 30: Do we really need this interface in the initial version? Can we simply go with the implementation?

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/standalone/JobListFetcherFactory.java
```diff
@@ -0,0 +1,30 @@
...
+import org.apache.flink.annotation.Experimental;
+import org.apache.flink.api.java.utils.ParameterTool;
+
+import java.time.Duration;
+
+/** The factory for {@link JobListFetcher}. */
+@Experimental
+public interface JobListFetcherFactory<KEY> {
+
+    JobListFetcher<KEY> create(ParameterTool parameters, Duration restClientTimeout);
```
gyfora, line 29: Do we really need this factory in the initial version?

1996fanrui (commented):
> I think the standalone implementation and the new interfaces / tests etc specific to that should be in a different maven module (flink-autoscaler-standalone). That way we wouldn't pollute the core implementation with unnecessary interfaces such as the job list etc.
> 
> If we are not sure about the stability we could even put this under the examples for now but that may be an overkill.

Sounds make sense to me. 

Could we just move the standalone package into the `flink-autoscaler-standalone` module and keep `LoggableEventHandler`, `RescaleApiScalingRealizer` and rest body related classes in the `flink-autoscaler` module?

gyfora:
The `RescaleApiScalingRealizer` should probably move to the standalone implementation as well. Instead of introducing a LoggableEventHandler, we could simply add default implementation to the interface methods directly.

gyfora (changes requested):
I think we should remove the new interfaces / factory interfaces. We don't need interfaces and factories for everything. This is a concrete standalone implementation let's keep the code change minimal. We can always add interfaces later if we want to extend this.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/standalone/flinkcluster/FlinkClusterJobListFetcher.java
```diff
@@ -0,0 +1,109 @@
...
+    private final String restServerAddress;
+    private final Duration restClientTimeout;
+
+    public FlinkClusterJobListFetcher(String host, int port, Duration restClientTimeout) {
+        this.restServerAddress = String.format("http://%s:%s", host, port);
+        this.restClientTimeout = restClientTimeout;
+    }
+
+    @Override
+    public List<JobAutoScalerContext<JobID>> fetch(MetricGroup metricGroup) throws Exception {
```
gyfora, line 50: I don't think `metricGroup` should be part of the method argument. The `FlinkClusterJobListFetcher` could under the hood simply pass the unregistered metric group.

gyfora (changes requested):

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/standalone/JobListFetcher.java
mxm: +1, let's just keep the implementation. There are no other consumers for this interface.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/standalone/JobListFetcherFactory.java
mxm: +1 We can construct the implementation directly.

mxm (commented):
Looks good! I would also like to see a dedicated module for these classes. Great work!

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/realizer/RescaleApiScalingRealizer.java
```diff
@@ -0,0 +1,153 @@
...
+
+        JobStatus jobStatus = context.getJobStatus();
+        JobID jobID = context.getJobID();
+        if (jobID == null
+                || jobStatus == null
+                || jobStatus.isGloballyTerminalState()
+                || JobStatus.RECONCILING == jobStatus) {
+            LOG.info("Job in terminal or reconciling state cannot be scaled in-place");
+            return;
+        }
```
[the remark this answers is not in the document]
1996fanrui: I try to keep the same logic with `NativeFlinkService#supportsInPlaceScaling`[1]. After thinking again, I think it can be simply: just check for `RUNNING` state.

Because `JobAutoScalerImpl#scale` won't runScalingLogic  if it isn't `RUNNING`.

[1] https://github.com/apache/flink-kubernetes-operator/blob/582c07aeec6b4bbcc68e0a85d31000d89dbd1958/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/NativeFlinkService.java#L260

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/standalone/flinkcluster/FlinkClusterJobListFetcher.java
1996fanrui: Good catch! It can be simply, it can be passed to `FlinkClusterJobListFetcher` by the constructor directly.

Also, I don't pass it in this PR, and `new UnregisteredMetricsGroup()` directly when creating the `JobAutoScalerContext`, because we don't support metric reporter for Standalone Autoscaler so far.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/standalone/JobListFetcher.java
1996fanrui: Of course, I can remove this interface in this PR first, and add it in FLINK-33100.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/standalone/JobListFetcherFactory.java
1996fanrui: I can remove this interface in this PR first, and add it in [FLINK-33100](https://issues.apache.org/jira/browse/FLINK-33100).

1996fanrui (commented):
Hi @gyfora  @mxm , thanks for the quick review.

I have updated all comments except https://github.com/apache/flink-kubernetes-operator/pull/698#discussion_r1377377454. 

And I will try to finish the tests asap.

flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/StandaloneAutoscalerEntrypoint.java
```diff
@@ -0,0 +1,135 @@
...
+
+    /** Arguments related to {@link FlinkClusterJobListFetcher}. */
+    public static final String FLINK_CLUSTER_HOST = "flinkClusterHost";
+
+    private static final String DEFAULT_FLINK_CLUSTER_HOST = "localhost";
+
+    public static final String FLINK_CLUSTER_PORT = "flinkClusterPort";
+    private static final int DEFAULT_FLINK_CLUSTER_PORT = 8081;
+
+    public static <KEY> void main(String[] args) {
```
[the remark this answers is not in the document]
gyfora: If we are adding the fatjar + also include the respective notice file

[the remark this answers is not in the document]
1996fanrui: Thanks for your reminder!

Samrat002 (commented):
thank you @1996fanrui , 
overall  +1 for the patch 

I have few queries:
1. can you please help understand how `YarnJobFetcher` can be used in hadoop cluster and deploy flink-yarn ? How exactly it can be used to serve autoscaling ? 

2. will `YarnJobFetcher` be a seperate process or demon in flink-yarn cluster ?

1996fanrui:
Hi @gyfora  @mxm , I have addressed all comments, added README, finished all unit tests and ran it locally with fat jar. Please take a look in your free time, thanks~

> If we are adding the fatjar + also include the respective notice file.

Hi @gyfora , I didn't understand the notice file. Could you elaborate on it?


-------------------------------


>  2.  will `YarnJobFetcher` be a seperate process or demon in flink-yarn cluster ?

Hi @Samrat002 , thank you for your review!

I have added a `README.md` to explain some details of `Standalone Autoscaler`, it includes how it works with `JobListFetcher`. `YarnJobListFetcher` is an implemetation of `JobListFetcher`, I think they are similar. you can take a look, thanks~


>     1. can you please help understand how `YarnJobFetcher` can be used in hadoop cluster and deploy flink-yarn ? How exactly it can be used to serve autoscaling ?

As I understand, users need to pass the `yarn resource manger address` to  `YarnJobListFetcher`, `YarnJobListFetcher` can fetch job list by 2 steps:
1 . `YarnJobListFetcher` can fetch all flink clusters from yarn cluster based on rest api. 
2. Step2 is totally same with current `FlinkClusterJobListFetcher`, because `FlinkClusterJobListFetcher` works with single flink cluster.

Please correct me if my understanding is wrong, thanks~

flink-autoscaler-standalone/README.md
```diff
@@ -0,0 +1,79 @@
+# Flink Autoscaler Standalone
+
+## What's the autoscaler standalone?
+
+`Flink Autoscaler Standalone` is an implementation of `Flink Autoscaler`, it runs as 
+a separate java process. It computes the reasonable parallelism of all job vertices 
+by monitoring the metrics, such as: processing rate, busy time, etc. Please see 
+[FLIP-271](https://cwiki.apache.org/confluence/display/FLINK/FLIP-271%3A+Autoscaling)
```
gyfora, line 8: Here can we link the operator docs autoscaler page instead? Lot of things have changed in the Flip and that page will never be updated.

flink-autoscaler-standalone/README.md
```diff
@@ -0,0 +1,79 @@
...
+## What's the autoscaler standalone?
+
+`Flink Autoscaler Standalone` is an implementation of `Flink Autoscaler`, it runs as 
+a separate java process. It computes the reasonable parallelism of all job vertices 
+by monitoring the metrics, such as: processing rate, busy time, etc. Please see 
+[FLIP-271](https://cwiki.apache.org/confluence/display/FLINK/FLIP-271%3A+Autoscaling) 
+for an overview of how autoscaling works.
+
+`Flink Autoscaler Standalone` rescales flink job in-place by rescale api of 
+[FLIP-291](https://cwiki.apache.org/confluence/x/9pRbDg).
```
gyfora, line 12: Here again we should link the docs instead: https://nightlies.apache.org/flink/flink-docs-release-1.18/docs/deployment/elastic_scaling/#externalized-declarative-resource-management

flink-autoscaler-standalone/README.md
```diff
@@ -0,0 +1,79 @@
+# Flink Autoscaler Standalone
+
+## What's the autoscaler standalone?
```
gyfora, line 3: We should add a paragraph clarifying that this should only be used in non-kubernetes environments. For Kubernetes the Operator provides a full integration that is preferable.

flink-autoscaler-standalone/pom.xml
```diff
@@ -0,0 +1,207 @@
...
+                </exclusion>
+                <exclusion>
+                    <artifactId>commons-compress</artifactId>
+                    <groupId>org.apache.commons</groupId>
+                </exclusion>
+                <exclusion>
+                    <artifactId>commons-text</artifactId>
+                    <groupId>org.apache.commons</groupId>
+                </exclusion>
+                <exclusion>
```
gyfora, line 75: In this entire pom we have groupId / artifactId , in the reverse order. While syntactically correct this is not how it is generally written. Can you please change them?

flink-autoscaler-standalone/pom.xml
```diff
@@ -0,0 +1,207 @@
...
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-shade-plugin</artifactId>
+                <executions>
+                    <!-- Run shade goal on package phase -->
+                    <execution>
```
gyfora, line 169: As we are bundling dependencies into a father and we are likely going to publish that to maven (right?) .

We need to add a NOTICE file to list the bundled dependencies by license

gyfora (changes requested):

flink-autoscaler-standalone/pom.xml
1996fanrui: Thanks @gyfora  for the quick review! The rest comments are addressed.

> As we are bundling dependencies into a father and we are likely going to publish that to maven (right?) .

During add the fat jar, I have a question: where do users to download this fat jar? In maven repo or dist?

> We need to add a NOTICE file to list the bundled dependencies by license

Sorry, I didn't do it before. Is the NOTICE file created by tools or manually?

gyfora: We could simply publish the fat jar to maven, that would be the simplest (instead of the slim jar)

As for the notice file creation, I don't know any tools, but I think you will not have too many things in it as it's mostly Flink (that do not need to be listed)

1996fanrui: > jar

I have deployed this PR to snapshot repo: https://repository.apache.org/content/repositories/snapshots/org/apache/flink/flink-autoscaler-standalone/1.7-SNAPSHOT/

Run it locally, it works well.

> NOTICE

done

mateczagany (changes requested):
I only have 2 very minor comments, other than that it looks great.

flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/StandaloneAutoscalerEntrypoint.java
```diff
@@ -0,0 +1,121 @@
...
+    public static final String FLINK_CLUSTER_HOST = "flinkClusterHost";
+
+    private static final String DEFAULT_FLINK_CLUSTER_HOST = "localhost";
+
+    public static final String FLINK_CLUSTER_PORT = "flinkClusterPort";
+    private static final int DEFAULT_FLINK_CLUSTER_PORT = 8081;
+
+    public static <KEY, Context extends JobAutoScalerContext<KEY>> void main(String[] args) {
+        var parameters = ParameterTool.fromArgs(args);
+        LOG.info("The standalone autoscaler is started, parameters: {}", parameters);
```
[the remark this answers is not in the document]
1996fanrui: Thanks @mateczagany for your review!

Good catch! I have updated, and tried it locally, it works well.

```
2023-11-03 18:18:54,002 o.a.f.a.s.StandaloneAutoscalerEntrypoint [INFO ] [.] The standalone autoscaler is started, parameters: {flinkClusterPort=8081, flinkClusterHost=localhost}

```

flink-autoscaler-standalone/src/test/java/org/apache/flink/autoscaler/standalone/StandaloneAutoscalerExecutorTest.java
```diff
@@ -0,0 +1,122 @@
...
+                            }
+
+                            @Override
+                            public void cleanup(JobID jobID) {
+                                fail("Should be called.");
+                            }
+                        });
+
+        // scaling shouldn't throw exception even if fetch fails
+        autoscalerExecutor.scaling();
```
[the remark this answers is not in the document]
1996fanrui: Technically, I don't know why need this. Because the test will fail if `autoscalerExecutor.scaling();` throw any exception.

From a code design perspective, explicitly declaring `doesNotThrow` may be more clear for code maintenance. So it sounds good to me. I updated.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/event/AutoScalerEventHandler.java
```diff
@@ -32,32 +35,45 @@
...
- * Handler for autoscaler events.
+ * Handler for autoscaler events, it logs events by default.
  *
  * @param <KEY> The job key.
  * @param <Context> Instance of JobAutoScalerContext.
  */
 @Experimental
-public interface AutoScalerEventHandler<KEY, Context extends JobAutoScalerContext<KEY>> {
-    String SCALING_SUMMARY_ENTRY =
+public class AutoScalerEventHandler<KEY, Context extends JobAutoScalerContext<KEY>> {
```
mxm, line 44: I think this should remain an interface because this is meant to be implemented by other backends. We already have two implementation of this: Kubernetes, Standalone, and Logging only.

flink-autoscaler-standalone/README.md
```diff
@@ -0,0 +1,83 @@
...
+```
+
+Updating the `flinkClusterHost` and `flinkClusterPort` based on your flink cluster. 
+In general, the host and port are the same as Flink WebUI.
+
+## Extensibility of autoscaler standalone
+
+`Autoscaler` as a generic autoscaler component defines a series of generic interfaces:
+
+- **AutoScalerEventHandler** : Handling autoscaler events, such as: ScalingReport,
```
mxm, line 65: This should be an interface, currently a class. Additionally, we can provide a logging implementation.

flink-autoscaler-standalone/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+- **AutoScalerEventHandler** : Handling autoscaler events, such as: ScalingReport, 
+  AutoscalerError, etc. It logs events by default.
+- **AutoScalerStateStore** : Storing all state during scaling. `InMemoryAutoScalerStateStore` 
+  is the default implementation, it's based on the Java Heap, so the state will be discarded 
+  after process restarts. We will implement persistent State Store in the future, such as
+  : `JdbcAutoScalerStateStore`.
+- **ScalingRealizer** : Applying scaling actions. `RescaleApiScalingRealizer` is the default
+  implementation, it uses the Rescale API to apply parallelism changes.
+- **JobAutoScalerContext** : Including all details related to the current job.
```
mxm, line 73: Line 61-73 should be part of the flink-autoscaler README.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/standalone/JobListFetcher.java
mxm: I'm also ok with keeping the interface.

mxm (approved):
Great work! LGTM, just some minor comments.

flink-autoscaler-standalone/README.md
1996fanrui: done

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/event/AutoScalerEventHandler.java
1996fanrui: Both are fine to me, updated the `AutoScalerEventHandler` as an interface and added a `LoggableEventHandler` as the logging  implementation.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/event/LoggableEventHandler.java
```diff
@@ -0,0 +1,52 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nullable;
+
+import java.time.Duration;
+
+/** Autoscaler event handler which logs events. */
+public class LoggableEventHandler<KEY, Context extends JobAutoScalerContext<KEY>>
```
gyfora, line 30: This should be called `LoggingEventHandler`

gyfora (approved):
Added one last minor comment, otherwise looks good!

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/event/LoggableEventHandler.java
1996fanrui: Thanks for the review, updated!

gyfora (approved):
🚢

1996fanrui:
Thanks a lot everyone for the review, merging~

PR #699, opened 2023-11-01
[FLINK-33429] Tolerate missing metrics in the stabilization phase

mxm:
The new code for the 1.7.0 release introduces metric collection during the stabilization phase to allow sampling the observed true processing rate. Metrics might not be fully initialized during that phase, as evident through the error metrics. The following error is thrown:

```
java.lang.RuntimeException: Could not find required metric
NUM_RECORDS_OUT_PER_SEC for 667f5d5aa757fb217b92c06f0f5d2bf2
```

To prevent these errors shadowing actual errors, we should detect and ignore this recoverable exception.

### DIFF
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java
@@ -20,9 +20,11 @@ package org.apache.flink.autoscaler;
 import org.apache.flink.annotation.VisibleForTesting;
 import org.apache.flink.api.common.JobID;
 import org.apache.flink.autoscaler.config.AutoScalerOptions;
+import org.apache.flink.autoscaler.exceptions.RecoverableException;
 import org.apache.flink.autoscaler.metrics.CollectedMetricHistory;
 import org.apache.flink.autoscaler.metrics.CollectedMetrics;
 import org.apache.flink.autoscaler.metrics.FlinkMetric;
+import org.apache.flink.autoscaler.metrics.MetricNotFoundException;
 import org.apache.flink.autoscaler.metrics.ScalingMetric;
 import org.apache.flink.autoscaler.metrics.ScalingMetrics;
 import org.apache.flink.autoscaler.state.AutoScalerStateStore;
@@ -112,6 +114,7 @@ public abstract class ScalingMetricCollector<KEY, Context extends JobAutoScalerC
         }
         var topology = getJobTopology(ctx, stateStore, jobDetailsInfo);
         var stableTime = jobUpdateTs.plus(conf.get(AutoScalerOptions.STABILIZATION_INTERVAL));
+        final boolean isStabilizing = now.isBefore(stableTime);
 
         // Calculate timestamp when the metric windows is full
         var metricWindowSize = getMetricWindowSize(conf);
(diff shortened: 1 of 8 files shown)


flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java
```diff
@@ -343,8 +346,19 @@ private double computeLagGrowthRate(
...
+                throw new RecoverableException(e);
+            }
+            throw e;
+        }
+    }
+
     /** Query the available metric names for each job vertex. */
-    @SneakyThrows
     protected Map<JobVertexID, Map<String, FlinkMetric>> queryFilteredMetricNames(
             Context ctx, JobTopology topology) {
```
1996fanrui, line 363: nit: could these 2 `queryFilteredMetricNames` methods be inlined? The old `queryFilteredMetricNames` method is only called by new `queryFilteredMetricNames` method.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/exceptions/RecoverableException.java
```diff
@@ -0,0 +1,26 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.autoscaler.exceptions;
+
+/** An exception for a recoverable scenario. */
+public class RecoverableException extends RuntimeException {
```
1996fanrui, line 21: How about rename it to `IgnorableException` or `NotReadyException`?

Recoverable generally means that it can be recovered and will be tried again later. `JobAutoScalerImpl#scale` doesn't retry it, and just doesn't report `onError`. And regardless of whether it is a RecoverableException this time, the next control loop will perform the same logic.

Based on it, I think the `RecoverableException` isn't very accurate here. WDYT?

1996fanrui (approved):
Hi @mxm , thanks for your fix!

This PR can reduce some unnecessary error events, it makes sense to me. I left 2 minor comments, please take a look in your free time, thanks~

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java
mxm: I did not inline because this method is overridden in tests, e.g. `TestingScalingCollector`. Inlining the newly-added method into the protected method would change the behavior. I wanted to make sure that the newly added method is never overridden.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/exceptions/RecoverableException.java
mxm: Fair. At one point I actually had `NotReadyException` :) Renamed to `NotReadyException`.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java
1996fanrui: thanks for the clarification, sounds make sense.

1996fanrui (approved):

gyfora (approved):
This should clean up the logs nicely :)

mxm:
> This should clean up the logs nicely :)

Yes. I was also noticing it in the error counter which kept going up.

PR #700, opened 2023-11-02
[FLINK-33427] Handle new autoscaler config keys like operator configs

gyfora:

An important logic in the operator is to "ignore" config changes that only affect operator behaviour and do not perform application upgrades when only those change. This is currently done based on the `kubernetes.operator` config prefix.

As we have renamed the autoscaler configs and removed the prefix, we need to adapt the operator side as well.

## Brief change log

 - Add `job.autoscaler` prefix to ignored list

## Verifying this change

Unit tests extended

## Documentation


mxm (approved):
Good catch!

PR #701, opened 2023-11-07
[FLINK-33471] Make flink kubernetes operator compilable with jdk21

snuyanzin:

The PR is aiming to support compilation with java 21

Since spotless and google java format needs to be updated to work with java 21
new spotless/google-java-format rules are applied as well

Also lombok dependency is bump to the version which supports  java 21

## Verifying this change
This change is a trivial rework

  - Dependencies (does it add or upgrade a dependency): (yes )

## Documentation

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ jobs:
     name: test_ci
     strategy:
       matrix:
-        java-version: [ 11, 17 ]
+        java-version: [ 11, 17, 21 ]
     steps:
       - uses: actions/checkout@v2
       - name: Set up JDK ${{ matrix.java-version }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,7 +73,7 @@ jobs:
         version: ["v1_18","v1_17","v1_16","v1_15"]
         namespace: ["default","flink"]
         mode: ["native", "standalone"]
-        java-version: [ 11, 17 ]
+        java-version: [ 11, 17, 21 ]
         test:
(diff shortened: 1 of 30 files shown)


snuyanzin:
@1996fanrui @gyfora since you've looked at similar PR for java 17 support may I ask you to have a look at java 21 support?

pom.xml
```diff
@@ -259,7 +259,7 @@ under the License.
                 <configuration>
                     <java>
                         <googleJavaFormat>
-                            <version>1.7</version>
+                            <version>1.17.0</version>
```
snuyanzin, line 262: minimum version supporting java 21

1996fanrui (approved):
Thanks @snuyanzin for the contribution!

LGTM.

.github/workflows/ci.yml
```diff
@@ -26,7 +26,7 @@ jobs:
     name: test_ci
     strategy:
       matrix:
-        java-version: [ 11, 17 ]
+        java-version: [ 11, 17, 21 ]
```
mxm, line 29: Is it worth to explode the test matrix by adding another Java version? Can't we test Java 21 compatibility without running all integration tests with it?

snuyanzin: it does not run all
it runs only against flink1.18 (same as for 17)
below there are exclusions for 1.16 and 1.17

I'm ok to reduce number of tests, however then would be great if you can provide a hint what else could be cut

mxm: Oh! I missed that you were excluding. This is fine then.

mxm (approved):

PR #707, opened 2023-11-12
[FLINK-33527][autoscaler] Clear all physical states after autoscaler is disabled

1996fanrui:

Currently, we just clear ParallelismOverrides  after autoscaler is disabled.

We should clear CollectedMetrics and ScalingHistory  as well to prevent state leak.

- CollectedMetrics can be cleared directly.
- ScalingHistory can be cleared based on trim logic( VERTEX_SCALING_HISTORY_AGE)

## Brief change log

[FLINK-33527][autoscaler] Clear all physical states after autoscaler is disabled

## Verifying this change

  - *JobAutoScalerImplTest#testAutoscalerDisabled*

## Documentation

### DIFF
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
@@ -29,10 +29,12 @@ import org.apache.flink.autoscaler.realizer.ScalingRealizer;
 import org.apache.flink.autoscaler.state.AutoScalerStateStore;
 import org.apache.flink.configuration.PipelineOptions;
 import org.apache.flink.runtime.jobgraph.JobVertexID;
+import org.apache.flink.util.Preconditions;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.time.Clock;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
@@ -41,6 +43,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import static org.apache.flink.autoscaler.config.AutoScalerOptions.AUTOSCALER_ENABLED;
 import static org.apache.flink.autoscaler.metrics.AutoscalerFlinkMetrics.initRecommendedParallelism;
 import static org.apache.flink.autoscaler.metrics.AutoscalerFlinkMetrics.resetRecommendedParallelism;
+import static org.apache.flink.autoscaler.metrics.ScalingHistoryUtils.trimScalingHistory;
 
 /** The default implementation of {@link JobAutoScaler}. */
 public class JobAutoScalerImpl<KEY, Context extends JobAutoScalerContext<KEY>>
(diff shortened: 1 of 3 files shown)


1996fanrui (commented):
Hey @gyfora  @mxm , this PR fixed some state leaks when autoscaler switches from enabled to disabled. Please help take a look in your free time thanks~

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
```diff
@@ -115,10 +123,39 @@ public void cleanup(KEY jobKey) {
...
+        }
+
+        Optional<SortedMap<Instant, CollectedMetrics>> collectedMetricsOpt =
+                stateStore.getCollectedMetrics(ctx);
+        if (collectedMetricsOpt.isPresent()) {
+            needFlush = true;
+            stateStore.removeCollectedMetrics(ctx);
+        }
+
+        Optional<Map<JobVertexID, SortedMap<Instant, ScalingSummary>>> scalingHistoryOpt =
```
gyfora, line 141: lets be consistent and use:
```
var scalingHistoryOpt = ...
...
var scalingHistory = ...
```

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
```diff
@@ -115,10 +123,39 @@ public void cleanup(KEY jobKey) {
...
+            if (scalingHistory.isEmpty()) {
+                // All scaling histories are trimmed.
+                needFlush = true;
+                stateStore.removeScalingHistory(ctx);
+            } else if (!scalingHistoryOpt.get().equals(scalingHistory)) {
+                // Some scaling histories are trimmed.
+                needFlush = true;
+                stateStore.storeScalingHistory(ctx, scalingHistory);
+            }
+        }
```
gyfora, line 156: This looks strange. Why not simply call ` stateStore.removeScalingHistory(ctx);` all the time? Then we wouldn't need to touch the trimming logic.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
1996fanrui: Updated

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
1996fanrui: Trimming logic will consider `VERTEX_SCALING_HISTORY_AGE` option, and the default value is `24 hours`.

My thought is: if users enable autoscaler within 24 hours, the `ScalingHistory` can be used for subsequent scaling. WDYT?

gyfora: I see, makes sense

PR #709, opened 2023-11-13
[FLINK-33097] Cleanup use of Optional for new interfaces

mxm:
The new interfaces make use of Optional in many places where returning the object directly would result in less and more readable code.

### DIFF
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/state/AutoScalerStateStore.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/state/AutoScalerStateStore.java
@@ -23,6 +23,8 @@ import org.apache.flink.autoscaler.ScalingSummary;
 import org.apache.flink.autoscaler.metrics.CollectedMetrics;
 import org.apache.flink.runtime.jobgraph.JobVertexID;
 
+import javax.annotation.Nonnull;
+
 import java.time.Instant;
 import java.util.Map;
 import java.util.SortedMap;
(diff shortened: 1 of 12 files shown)


flink-autoscaler/src/main/java/org/apache/flink/autoscaler/state/AutoScalerStateStore.java
```diff
@@ -41,23 +40,22 @@ void storeScalingHistory(
             Context jobContext, Map<JobVertexID, SortedMap<Instant, ScalingSummary>> scalingHistory)
             throws Exception;
 
-    Optional<Map<JobVertexID, SortedMap<Instant, ScalingSummary>>> getScalingHistory(
-            Context jobContext) throws Exception;
+    Map<JobVertexID, SortedMap<Instant, ScalingSummary>> getScalingHistory(Context jobContext)
```
1996fanrui, line 43: Could we mark these get interfaces as `@Nonnull`?

It's useful to let other developers know this limitation if they want to implement new state store. Otherwise, autoscaler will have a series of NullPointerException.

1996fanrui (approved):
Thanks @mxm  for the clean up.

I left one minor comment, please take a look in your free time, thanks!

gyfora (approved):

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/state/AutoScalerStateStore.java
mxm: Sure, that makes sense.

PR #710, opened 2023-11-13
[FLINK-33527] Simplify state store cleanup logic

mxm:
The state store implementation should take care of optimizing flush
operations. That is already the case for the Kubernetes-based
implementation. The in-memory state store does not require flushing.

The current logic tries to save on API calls but there is no actual savings
because the ConfigMap needs to be retrieved at least once to check whether its
empty. So instead of checking for emptiness, we can simply clear it.

Further, I've removed the optimization to keep some of the data when disabling
the autoscaler. The metric collection contract requires that the entire metric
window contains data. By removing some of the data, once turned on again, we
could add metric gaps to the metric history if there are still metrics left from
the last metric window which overlap with the current. The reason is that the
metric window is counted from the oldest timestamp found in the metric history.


gyfora (approved):
This goes a bit against what @1996fanrui did but overall I prefer this approach due to the simplicity.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/ConfigMapStore.java
```diff
@@ -67,19 +67,14 @@ protected Optional<String> getSerializedState(
...
-                                    "The configMap isn't created, so the remove is unavailable.");
-                        });
+        getConfigMap(jobContext).ifPresent(configMap -> configMap.getData().remove(key));
     }
 
     public void flush(KubernetesJobAutoScalerContext jobContext) {
         Optional<ConfigMap> configMapOpt = cache.get(jobContext.getJobKey());
         if (configMapOpt == null || configMapOpt.isEmpty()) {
             LOG.debug("The configMap isn't updated, so skip the flush.");
+            // Do not flush if there are no updates.
```
1996fanrui, line 77: I'm thinking whether any updating or removeing can be determined, please give me a few times to test it.

1996fanrui: I write 2 tests:
1. one is `testRemoveAndFlushEveryTime`, it's similar to current PR code
2. one is `testQueryAndFlushIfNeeded`, it's similar to master code

- `testRemoveAndFlushEveryTime` will remove and flush every time, and it will call kubernetes every time.
- `testQueryAndFlushIfNeeded` will query from cache, and value is null, so remove and flush isn't necessary. So it doens't call kubernetes.

My question is: the extra call is expected? And I think this PR is fine if it doesn't introuduce any extra opration.

These 2 tests can be added to ConfigMapStoreTest, and they can run directly.


```    
    @Test
    void testRemoveAndFlushEveryTime() {
        KubernetesJobAutoScalerContext ctx = createContext("cr1", kubernetesClient);
        var key = "key";
        var value = "value";

        var configMapStore = new ConfigMapStore(kubernetesClient);
        configMapStore.putSerializedState(ctx, key, value);
        assertEquals(2, mockWebServer.getRequestCount());
        configMapStore.flush(ctx);
        assertEquals(3, mockWebServer.getRequestCount());

        // Get from cache
        assertThat(configMapStore.getSerializedState(ctx, key)).hasValue(value);
        assertEquals(3, mockWebServer.getRequestCount());

        // Get from kubernetes
        configMapStore = new ConfigMapStore(kubernetesClient);
        assertThat(configMapStore.getSerializedState(ctx, key)).hasValue(value);
        assertEquals(4, mockWebServer.getRequestCount());

        // Remove the key
        configMapStore.removeSerializedState(ctx, key);
        assertEquals(4, mockWebServer.getRequestCount());

        configMapStore.flush(ctx);
        assertEquals(5, mockWebServer.getRequestCount());

        // Every flush will call kubernetes.
        configMapStore.flush(ctx);
        assertEquals(6, mockWebServer.getRequestCount());
        configMapStore.flush(ctx);
        assertEquals(7, mockWebServer.getRequestCount());
        configMapStore.flush(ctx);
        assertEquals(8, mockWebServer.getRequestCount());
    }


    @Test
    void testQueryAndFlushIfNeeded() {
        KubernetesJobAutoScalerContext ctx = createContext("cr1", kubernetesClient);
        var key = "key";
        var value = "value";

        var configMapStore = new ConfigMapStore(kubernetesClient);
        configMapStore.putSerializedState(ctx, key, value);
        assertEquals(2, mockWebServer.getRequestCount());
        configMapStore.flush(ctx);
        assertEquals(3, mockWebServer.getRequestCount());

        // Get from cache
        assertThat(configMapStore.getSerializedState(ctx, key)).hasValue(value);
        assertEquals(3, mockWebServer.getRequestCount());

        // Get from kubernetes
        configMapStore = new ConfigMapStore(kubernetesClient);
        assertThat(configMapStore.getSerializedState(ctx, key)).hasValue(value);
        assertEquals(4, mockWebServer.getRequestCount());

        // Remove the key
        configMapStore.removeSerializedState(ctx, key);
        assertEquals(4, mockWebServer.getRequestCount());

        configMapStore.flush(ctx);
        assertEquals(5, mockWebServer.getRequestCount());

        // get is null, so don't flush
        configMapStore.getSerializedState(ctx, key);
        assertEquals(5, mockWebServer.getRequestCount());
        configMapStore.getSerializedState(ctx, key);
        assertEquals(5, mockWebServer.getRequestCount());
        configMapStore.getSerializedState(ctx, key);
        assertEquals(5, mockWebServer.getRequestCount());
    }
```

1996fanrui (changes requested):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/ConfigMapStore.java
1996fanrui: We may need to added a extra cache to store the state that last updated. When the flush is called, then compare them. And call kubernetes if they are not same.

And it's better to add a test to check the flush doesn't call kubernetes if state isn't updated.

gyfora: In this case the flush only calls Kubernetes because there is still data in the state store that wasn't removed. Maybe we should add a `clearAll` method to the state store interface and simply call that with the contract that subsequent `clearAll()` and `flush()` actions should be idempotent and ideally should not call Kubernetes again and again

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
```diff
@@ -121,37 +120,10 @@ public void cleanup(KEY jobKey) {
...
-            }
-        }
-
-        if (needFlush) {
-            stateStore.flush(ctx);
-        }
+        stateStore.removeParallelismOverrides(ctx);
+        stateStore.removeCollectedMetrics(ctx);
+        stateStore.removeScalingHistory(ctx);
+        stateStore.flush(ctx);
```
gyfora, line 126: can we add a new `clearAll` or `removeAll` method that would guarantee that all state is cleaned and then we can test for subsequent flushing not hitting k8s api

gyfora (changes requested):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/ConfigMapStore.java
mxm: Thanks for adding the tests. You are right that the retrieval will always go to Kubernetes. The current implementation has some bugs with the bookkeeping to minimize the number of API calls. It does make more requests than necessary. 

I've rewrote the implementation to minimize the number of times we go to Kubernetes. We only retrieve once now on lookup. If the ConfigMap does not exist, we will only create the ConfigMap once we flush back any changes.

mxm:
Please have a look at the new implementation. Also see: https://github.com/apache/flink-kubernetes-operator/pull/710#discussion_r1391538017

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/ConfigMapStore.java
```diff
@@ -44,13 +43,44 @@ public class ConfigMapStore {
...
     private final KubernetesClient kubernetesClient;
 
-    // The cache for each resourceId may be in three states:
-    // 1. The resourceId doesn't exist : ConfigMap isn't loaded from kubernetes, or it's deleted
-    // 2  Exists, Optional.empty() : The ConfigMap doesn't exist in Kubernetes
-    // 3. Exists, Not Empty : We have loaded the ConfigMap from kubernetes, it may not be the same
-    // if not flushed already
-    private final ConcurrentHashMap<ResourceID, Optional<ConfigMap>> cache =
-            new ConcurrentHashMap<>();
+    static class ConfigMapState {
```
gyfora, line 46: Could we please move this class declaration to the end of the enclosing class?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/ConfigMapStore.java
```diff
@@ -63,29 +93,29 @@ protected void putSerializedState(
...
-                                    "The configMap isn't created, so the remove is unavailable.");
-                        });
+        getConfigMap(jobContext).removeKey(key);
     }
 
     public void flush(KubernetesJobAutoScalerContext jobContext) {
-        Optional<ConfigMap> configMapOpt = cache.get(jobContext.getJobKey());
-        if (configMapOpt == null || configMapOpt.isEmpty()) {
+        ConfigMapState configMapState = cache.get(jobContext.getJobKey());
+        if (configMapState == null || (configMapState.flushed && configMapState.exists)) {
```
1996fanrui, line 105: It's better to extract a method `isFlushNeeded` for `ConfigMapState` instead of call `configMapState.flushed && configMapState.exists` directly.

1996fanrui (changes requested):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/ConfigMapStore.java
```diff
@@ -44,13 +43,44 @@ public class ConfigMapStore {
...
+        private boolean flushed = true;
+        private boolean exists = true;
+
+        @VisibleForTesting ConfigMap configMap;
+
+        public Map<String, String> getData() {
+            return Collections.unmodifiableMap(configMap.getData());
+        }
+
+        public void clear() {
```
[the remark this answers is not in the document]
gyfora: Maybe we should actually use this instead of calling the separate clear methods. Makes it much "cleaner"

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/ConfigMapStore.java
mxm: Sure, moved.

mxm: Actually now a top-level class.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/ConfigMapStore.java
```diff
@@ -44,13 +43,44 @@ public class ConfigMapStore {
...
-    // 2  Exists, Optional.empty() : The ConfigMap doesn't exist in Kubernetes
-    // 3. Exists, Not Empty : We have loaded the ConfigMap from kubernetes, it may not be the same
-    // if not flushed already
-    private final ConcurrentHashMap<ResourceID, Optional<ConfigMap>> cache =
-            new ConcurrentHashMap<>();
+    static class ConfigMapState {
+        private boolean flushed = true;
+        private boolean exists = true;
+
+        @VisibleForTesting ConfigMap configMap;
```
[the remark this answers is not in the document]
mxm: The updated code should reflect your request.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/ConfigMapStore.java
mxm: This should be nicer now.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/ConfigMapStore.java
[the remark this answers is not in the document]
mxm: Added a method to clear all state.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
mxm: Done.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/state/ConfigMapView.java
```diff
@@ -0,0 +1,111 @@
...
+    public ConfigMapView(
+            Supplier<ConfigMap> configMapRetriever,
+            Supplier<ConfigMap> configMapBuilder,
+            Function<ConfigMap, Resource<ConfigMap>> resourceRetriever) {
+        var existingConfigMap = configMapRetriever.get();
+        if (existingConfigMap != null) {
+            refreshConfigMap(existingConfigMap);
+        } else {
+            this.configMap = configMapBuilder.get();
+            this.state = State.NEEDS_CREATE;
```
gyfora, line 54: I am not super familiar with the `Resource<T>` class but couldn't we use it also to get the `ConfigMap`? Then we could simply use that and don't need the `configMapRetriever`

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/state/ConfigMapView.java
```diff
@@ -0,0 +1,111 @@
...
+        }
+        this.resourceRetriever = resourceRetriever;
```
gyfora, line 56: I think we could further simplify this by keeping only:
```
Resource<ConfigMap> resource:
ConfigMap configMap;
State state;
```


And pass those directly in the constructor / through a static create method instead of the Suppliers/Functions which feel a bit unnecessary

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/state/ConfigMapView.java
mxm: Lookup is performed by name and namespace. Update / Create uses the ConfigMap object which potentially involves version checks. That's how the code currently works. I didn't want to mess with the APIs to avoid introducing regressions.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/state/ConfigMapView.java
mxm: The current state is the following btw:

```
Function<ConfigMap, Resource<ConfigMap>> resource:
ConfigMap configMap;
State state;
```

That occupies less space than what you are suggesting. Resource is only a pointer to an existing method.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/state/ConfigMapStore.java
```diff
@@ -44,13 +43,13 @@ public class ConfigMapStore {
...
-    // 3. Exists, Not Empty : We have loaded the ConfigMap from kubernetes, it may not be the same
-    // if not flushed already
-    private final ConcurrentHashMap<ResourceID, Optional<ConfigMap>> cache =
-            new ConcurrentHashMap<>();
+    // The cache for each resourceId may be in four states:
+    // 1. No cache entry: ConfigMap isn't loaded from kubernetes, or it's deleted.
+    // 2. Cache entry, not created : The ConfigMap doesn't exist in Kubernetes.
+    // 3. Cache entry, not flushed : The ConfigMap exists in Kubernetes, but it is not updated yet.
+    // 4. Cache entry, flushed and created : We have loaded the ConfigMap from kubernetes, and it's
+    // up-to-date.
```
1996fanrui, line 51: Would you mind updating these comments based on the latest `ConfigMapView.State`? For example, using `NEEDS_CREATE` instead of `not created`.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/ConfigMapStore.java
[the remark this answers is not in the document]
1996fanrui: Great work, Max! 👍

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/state/ConfigMapStore.java
mxm: Sure. Take a look at https://github.com/apache/flink-kubernetes-operator/pull/710/commits/e6568d958a141e77be54af6c456212faa69869b8. Does that reflect what you had in mind?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/state/ConfigMapView.java
mxm: Updated.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/autoscaler/state/KubernetesAutoScalerStateStoreTest.java
```diff
@@ -330,4 +331,42 @@ public void testDiscardInvalidHistory() throws Exception {
                                 ctx, KubernetesAutoScalerStateStore.SCALING_HISTORY_KEY))
                 .isEmpty();
     }
+
+    @Test
+    public void testDiscardAllState() {
```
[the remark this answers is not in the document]
mxm: Good catch! I've added tests for both classes to test the actual class: https://github.com/apache/flink-kubernetes-operator/pull/710/commits/d36180265b753c991ca86dc5e8d7e6347aa45ec2

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/state/ConfigMapStore.java
```diff
@@ -99,49 +93,30 @@ public void removeInfoFromCache(ResourceID resourceID) {
...
+                    }
+                    // Otherwise retrieve if it exists
+                    return getConfigMapFromKubernetes(jobContext);
+                });
     }
 
     @VisibleForTesting
-    protected Optional<ConfigMap> getConfigMapFromKubernetes(
-            KubernetesJobAutoScalerContext jobContext) {
+    ConfigMapView getConfigMapFromKubernetes(KubernetesJobAutoScalerContext jobContext) {
```
1996fanrui, line 115:
```suggestion
    ConfigMapView createConfigMapView(KubernetesJobAutoScalerContext jobContext) {
```

nit

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/state/ConfigMapView.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    private State state;
+
+    private ConfigMap configMap;
+
+    private final Function<ConfigMap, Resource<ConfigMap>> resourceRetriever;
+
+    public ConfigMapView(
+            ConfigMap configMap, Function<ConfigMap, Resource<ConfigMap>> resourceRetriever) {
```
1996fanrui, line 52:
```suggestion
            ConfigMap newConfigMap, Function<ConfigMap, Resource<ConfigMap>> resourceRetriever) {
```

How about rename the configMap to `newConfigMap` or `emptyConfigMap`? It's clear to let us know it's only used when `resourceRetriever.get()` is null.

1996fanrui (commented):
Thanks @mxm  for the updated, I just left last round comments, please take a look~

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/state/ConfigMapStore.java
```diff
@@ -99,49 +93,30 @@ public void removeInfoFromCache(ResourceID resourceID) {
...
-                            return Optional.of(getOrCreateConfigMapFromKubernetes(jobContext));
-                        })
-                .get()
-                .getData();
+                jobContext.getJobKey(),
+                (id, configMapView) -> {
+                    // If in the cache and valid simply return
+                    if (configMapView != null) {
+                        return configMapView;
+                    }
```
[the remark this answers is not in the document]
mxm: That method was duplicated anyhow, removed.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/state/ConfigMapView.java
mxm: Renamed to `configMapSkeleton`.

1996fanrui (approved):

1996fanrui:
Thanks @mxm for the cleaning up and simplifying, great work!

LGTM for now.

gyfora (approved):

gyfora:
🚢

mxm:
Small follow-up which was meant to be included in this PR: https://github.com/apache/flink-kubernetes-operator/pull/714

PR #711, opened 2023-11-14
[FLINK-30593][autoscaler] Determine restart time on the fly fo Autoscaler

afedulov:
This PR does not yet contain tests since I would like to first reach consensus on the general approach.

Currently the autoscaler uses a preconfigured restart time for the job. This PR adds the ability to dynamically adjust this based on the observed restart times for scale operations

## Brief change log

  - Adds `ScalingTracking` for tracking job-scoped scaling data.
  - Adds autoscaler properties for enabling the observed restart times usage in the rescaling logic(opt in).

**High level approach:**
The autoscaler adds a `ScalingTracking` object into the config map in the following format:
```
 scalingTracking: |                                                                                                                                                         
 ---
    scalingRecords: 
      "2023-11-13T13:36:35.240245Z": 
         endTime: null
           targetParallelism: 
             e44d8435128fbb8c0406b9ae78407aec: 2                                                                                                                                           
```
It contains a map of applied scaling decisions together with the target topology and the `endTime`, sorted by the time when the scaling was applied (map's key). When the job transitions into the RUNNING state, the latest record is fetched, the current parallelism is compared with the target one and the `endTime` is set to now (only if was not previously set).
  
## Verifying this change

Manual. Testing implementation will follow after the general approach gets approved.

## Documentation

  - If yes, how is the feature documented? (not applicable / **docs** / **JavaDocs** / not documented)

### DIFF
--- a/docs/layouts/shortcodes/generated/auto_scaler_configuration.html
+++ b/docs/layouts/shortcodes/generated/auto_scaler_configuration.html
@@ -81,16 +81,16 @@
             <td>Expected restart time to be used until the operator can determine it reliably from history.</td>
         </tr>
         <tr>
-            <td><h5>job.autoscaler.restart.time.tracked.enabled</h5></td>
+            <td><h5>job.autoscaler.restart.time-tracking.enabled</h5></td>
             <td style="word-wrap: break-word;">false</td>
             <td>Boolean</td>
-            <td>Whether to use the actually observed rescaling restart times instead of the fixed 'job.autoscaler.restart.time' configuration. If set to true, the maximum restart duration over a number of samples will be used. The value of 'job.autoscaler.restart.time' will act as an upper bound.</td>
+            <td>Whether to use the actual observed rescaling restart times instead of the fixed 'job.autoscaler.restart.time' configuration. If set to true, the maximum restart duration over a number of samples will be used. The value of 'job.autoscaler.restart.time' will act as an upper bound.</td>
         </tr>
         <tr>
-            <td><h5>job.autoscaler.restart.time.tracked.limit</h5></td>
+            <td><h5>job.autoscaler.restart.time-tracking.limit</h5></td>
             <td style="word-wrap: break-word;">15 min</td>
             <td>Duration</td>
-            <td>Maximum cap for the calculated restart time when 'job.autoscaler.restart.time.tracked.enabled' is set to true.</td>
+            <td>Maximum cap for the observed restart time when 'job.autoscaler.restart.time-tracking.enabled' is set to true.</td>
         </tr>
         <tr>
             <td><h5>job.autoscaler.scale-down.max-factor</h5></td>
--- a/docs/layouts/shortcodes/generated/auto_scaler_configuration.html
+++ b/docs/layouts/shortcodes/generated/auto_scaler_configuration.html
@@ -81,6 +81,18 @@
             <td>Expected restart time to be used until the operator can determine it reliably from history.</td>
         </tr>
         <tr>
+            <td><h5>job.autoscaler.restart.time.tracked.enabled</h5></td>
+            <td style="word-wrap: break-word;">false</td>
+            <td>Boolean</td>
+            <td>Whether to use the actually observed rescaling restart times instead of the fixed 'job.autoscaler.restart.time' configuration. If set to true, the maximum restart duration over a number of samples will be used. The value of 'job.autoscaler.restart.time' will act as an upper bound.</td>
+        </tr>
+        <tr>
+            <td><h5>job.autoscaler.restart.time.tracked.limit</h5></td>
+            <td style="word-wrap: break-word;">15 min</td>
+            <td>Duration</td>
+            <td>Maximum cap for the calculated restart time when 'job.autoscaler.restart.time.tracked.enabled' is set to true.</td>
+        </tr>
+        <tr>
             <td><h5>job.autoscaler.scale-down.max-factor</h5></td>
             <td style="word-wrap: break-word;">0.6</td>
             <td>Double</td>
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
@@ -171,7 +171,7 @@ public class JobAutoScalerImpl<KEY, Context extends JobAutoScalerContext<KEY>>
         var now = clock.instant();
         // Scaling tracking data contains previous restart times that are taken into account
         var scalingTracking = getTrimmedScalingTracking(stateStore, ctx, now);
-        var restartTime = scalingTracking.getMaxRestartTimeSecondsOrDefault(ctx.getConfiguration());
+        var restartTime = scalingTracking.getMaxRestartTimeOrDefault(ctx.getConfiguration());
         var evaluatedMetrics =
                 evaluator.evaluate(ctx.getConfiguration(), collectedMetrics, restartTime);
         LOG.debug("Evaluated metrics: {}", evaluatedMetrics);
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
@@ -171,10 +171,9 @@ public class JobAutoScalerImpl<KEY, Context extends JobAutoScalerContext<KEY>>
         var now = clock.instant();
         // Scaling tracking data contains previous restart times that are taken into account
         var scalingTracking = getTrimmedScalingTracking(stateStore, ctx, now);
-        var restartTimeSec =
-                scalingTracking.getMaxRestartTimeSecondsOrDefault(ctx.getConfiguration());
+        var restartTime = scalingTracking.getMaxRestartTimeSecondsOrDefault(ctx.getConfiguration());
         var evaluatedMetrics =
-                evaluator.evaluate(ctx.getConfiguration(), collectedMetrics, restartTimeSec);
+                evaluator.evaluate(ctx.getConfiguration(), collectedMetrics, restartTime);
         LOG.debug("Evaluated metrics: {}", evaluatedMetrics);
         lastEvaluatedMetrics.put(ctx.getJobKey(), evaluatedMetrics);
 
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
@@ -42,6 +42,8 @@ import java.util.concurrent.ConcurrentHashMap;
 import static org.apache.flink.autoscaler.config.AutoScalerOptions.AUTOSCALER_ENABLED;
 import static org.apache.flink.autoscaler.metrics.AutoscalerFlinkMetrics.initRecommendedParallelism;
 import static org.apache.flink.autoscaler.metrics.AutoscalerFlinkMetrics.resetRecommendedParallelism;
+import static org.apache.flink.autoscaler.metrics.ScalingHistoryUtils.getTrimmedScalingHistory;
+import static org.apache.flink.autoscaler.metrics.ScalingHistoryUtils.getTrimmedScalingTracking;
 
 /** The default implementation of {@link JobAutoScaler}. */
 public class JobAutoScalerImpl<KEY, Context extends JobAutoScalerContext<KEY>>
@@ -159,19 +161,26 @@ public class JobAutoScalerImpl<KEY, Context extends JobAutoScalerContext<KEY>>
             throws Exception {
 
         var collectedMetrics = metricsCollector.updateMetrics(ctx, stateStore);
+        var jobTopology = collectedMetrics.getJobTopology();
 
         if (collectedMetrics.getMetricHistory().isEmpty()) {
             return;
         }
         LOG.debug("Collected metrics: {}", collectedMetrics);
 
-        var evaluatedMetrics = evaluator.evaluate(ctx.getConfiguration(), collectedMetrics);
+        var now = clock.instant();
+        // Scaling tracking data contains previous restart times that are taken into account
+        var scalingTracking = getTrimmedScalingTracking(stateStore, ctx, now);
+        var restartTimeSec =
+                scalingTracking.getMaxRestartTimeSecondsOrDefault(ctx.getConfiguration());
+        var evaluatedMetrics =
+                evaluator.evaluate(ctx.getConfiguration(), collectedMetrics, restartTimeSec);
         LOG.debug("Evaluated metrics: {}", evaluatedMetrics);
         lastEvaluatedMetrics.put(ctx.getJobKey(), evaluatedMetrics);
 
         initRecommendedParallelism(evaluatedMetrics);
         autoscalerMetrics.registerScalingMetrics(
-                collectedMetrics.getJobTopology().getVerticesInTopologicalOrder(),
+                jobTopology.getVerticesInTopologicalOrder(),
                 () -> lastEvaluatedMetrics.get(ctx.getJobKey()));
 
         if (!collectedMetrics.isFullyCollected()) {
@@ -180,7 +189,19 @@ public class JobAutoScalerImpl<KEY, Context extends JobAutoScalerContext<KEY>>
             return;
         }
 
-        var parallelismChanged = scalingExecutor.scaleResource(ctx, evaluatedMetrics);
+        var scalingHistory = getTrimmedScalingHistory(stateStore, ctx, now);
+        // A scaling tracking without an end time gets created whenever a scaling decision is
+        // applied. Here, when the job transitions to RUNNING, we record the time for it.
+        if (ctx.getJobStatus() == JobStatus.RUNNING) {
+            if (scalingTracking.setEndTimeIfTrackedAndParallelismMatches(
+                    now, jobTopology, scalingHistory)) {
+                stateStore.storeScalingTracking(ctx, scalingTracking);
+            }
+        }
+
+        var parallelismChanged =
+                scalingExecutor.scaleResource(
+                        ctx, evaluatedMetrics, scalingHistory, scalingTracking, now);
 
         if (parallelismChanged) {
             autoscalerMetrics.incrementScaling();
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingRecord.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingRecord.java
@@ -1,0 +1,35 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.autoscaler;
+
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+import java.time.Instant;
+
+/**
+ * Class for tracking scaling details, including time it took for the job to transition to the
+ * target parallelism.
+ */
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+public class ScalingRecord {
+    private Instant endTime;
+}
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingTracking.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingTracking.java
@@ -125,12 +125,12 @@ public class ScalingTracking {
     }
 
     /**
-     * Retrieves the maximum restart time in seconds based on the provided configuration and scaling
-     * records. Defaults to the RESTART_TIME from configuration if the PREFER_TRACKED_RESTART_TIME
-     * option is set to false, or if there are no tracking records available. Otherwise, the maximum
-     * observed restart time is capped by the MAX_RESTART_TIME.
+     * Retrieves the maximum restart time based on the provided configuration and scaling records.
+     * Defaults to the RESTART_TIME from configuration if the PREFER_TRACKED_RESTART_TIME option is
+     * set to false, or if there are no tracking records available. Otherwise, the maximum observed
+     * restart time is capped by the MAX_RESTART_TIME.
      */
-    public Duration getMaxRestartTimeSecondsOrDefault(Configuration conf) {
+    public Duration getMaxRestartTimeOrDefault(Configuration conf) {
         long maxRestartTime = -1;
         if (conf.get(AutoScalerOptions.PREFER_TRACKED_RESTART_TIME)) {
             for (Map.Entry<Instant, ScalingRecord> entry : scalingRecords.entrySet()) {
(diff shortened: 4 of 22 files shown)


gyfora (changes requested):
I like the overall approach but I would recommend the following changes:
 1. Remove the newly introduced configs, this should be automatic and always on
 2. Track the start/end times for the restart in memory and only record the `observed_restart_time` in the autoscaler state store. This way we add minimal extra state that is easy to implement.
 3. Instead of computing restart time from a fixed number of samples, use a simple moving average: `observed_restart_time = (prev_observed + new_observed) / 2`
 4. During autoscaler logic use: `restart_time = min(conf_restart_time, observed_restart_time)`

mxm:
Commenting on some of the requests: 
> 1. Remove the newly introduced configs, this should be automatic and always on

I think it is fair to have a ON/OFF switch. It should be on by default but we want to keep the ability to roll back to the old behavior.

> 2. Track the start/end times for the restart in memory and only record the `observed_restart_time` in the autoscaler state store. This way we add minimal extra state that is easy to implement.

We already have the start time of the last scaling in memory via the scaling history. We can then keep note of the end time once we detect the scaling is over. That leaves a little bit of error in case of downtime of the operator which will produce a long rescaling time. I think that should be fine though, since we cap at the max configured rescale time. 
 
> 3. Instead of computing restart time from a fixed number of samples, use a simple moving average: `observed_restart_time = (prev_observed + new_observed) / 2`

I think we can do an exponentially weighted average.

> 4. During autoscaler logic use: `restart_time = min(conf_restart_time, observed_restart_time)`

+1

mxm:
>Are we talking about having a field in the ScalingExecutor? Because fetching scalingHistory won't be sufficient - we need some indication that a new rescaling was applied by the time we see the transition into RUNNING with the expected parallelism. This "flag" then needs to be cleaned.

I meant a flag in the ConfigMap. I guess something like `startTime: endTime` would be sufficient? The endTime would initially be null.

>General question: it feels like we are very focused on optimizing the size of this particular configmap. Can't we create a separate configmap, if this is a concern? 

In the past we ran into the 1MB size limit for large pipelines with hundreds of vertices. That's why we don't want to increase the state size too much. We also added compression because of this. Of course we can add a new configmap. We have just not come around to do it and it might not be a good idea to put too much load on etcd. While it would be nice to have a separate config map for the scaling metrics and the scaling history, but it is much simpler to do it in one.

afedulov:
>I meant a flag in the ConfigMap. I guess something like startTime: endTime would be sufficient? The endTime would initially be null.

Got it, this is how it actually works currently (modulo additional target vertices parallelism tracking that I will attempt to fetch from the history instead) - I thought it was more about the configmap vs in-memory discussion.

> In the past we ran into the 1MB size limit for large pipelines with hundreds of vertices. 

I see, let's I'll try to keep the size to the minimum then.

mxm:
> Got it, this is how it actually works currently (modulo additional target vertices parallelism tracking that I will attempt to fetch from the history instead) - I thought it was more about the configmap vs in-memory discussion.

Yes, I saw that. I think Gyula is right though, that it's enough to use in-memory state to start tracking a scaling execution. When the scaling is completed, we then persist the duration in the ConfigMap. The reason is that keeping track of the start time when we begin tracking would only be useful if we could use that state to recover, e.g. in case of downtime. However, that's only useful when the job does not complete rescaling during the downtime. If it completes before, we don't know how long a scaling took when we come back up. For simplicity, I think it makes sense to just track the scaling start time in memory and then persist the actual rescale time in the ConfigMap.

afedulov:
Sounds reasonable. If we are optimizing state size at this level, how about we also remove `RECOMMENDED_PARALLELISM` and `PARALLELISM` from the history metrics, since they are duplicated as `ScalingSummary` top level fields?

afedulov:
@gyfora I'd like to propose to use this format:
 ```
scalingTracking: |                                                                                                                             
 ---                                                                                                                                            
scalingRecords:
      "2023-11-14T20:20:39.250503Z": 
            endTime: "2023-11-14T20:22:04.360972Z"
       "2023-11-14T20:22:04.360972Z": 
            endTime: "2023-11-14T20:23:39.022220Z"
```
Some reasoning:
- It allows to associate tracked restart time with scaling history, which could be useful for debugging. Without the start timestamp, this won't be possible.
- It lays out a data structure that is open for extension. This makes it trivial to add additional fields if we ever need some other scaling- and not vertex-scoped data which we currently do not anticipate in the future (the decision to move it into a different config map in such case is a separate issue).
- The overhead is pretty minimal. 

@mxm, I removed the redundant tracking of the target parallelism from the `ScalingRecords`. The current implementation completely relies on the `scalingHistory` for this. I had change the `scaleResource` signature to avoid fetching the tracking and the history twice (we need both in the JobAutoScalerImpl now).

**Comment by afedulov** [comment]
@mxm @gyfora the PR is ready for review
@gyfora me and Max briefly discussed offline and came to the conclusion that starting with evaluating the maximum restart time capped by the RESTART_TIME setting is probably good enough for the first step. It has the benefit of giving the most "conservative" evaluation and we can add the moving average after some baseline testing. What do you think?

File: flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingRecord.java [hunk 1]
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Class for tracking scaling details, including time it took for the job to transition to the
+ * target parallelism.
+ */
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+public class ScalingRecord {
+    private Instant endTime;
+}
```
mxm, line 35: Could we just store Instant directly and remove this wrapper class?

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingTracking.java
```diff
@@ -0,0 +1,168 @@
...
+    private static boolean targetParallelismMatchesActual(
+            Map<JobVertexID, Integer> targetParallelisms,
+            Map<JobVertexID, Integer> actualParallelisms) {
+        return targetParallelisms.entrySet().stream()
+                .allMatch(
+                        entry -> {
+                            var vertexID = entry.getKey();
+                            var targetParallelism = entry.getValue();
+                            var actualParallelism = actualParallelisms.getOrDefault(vertexID, -1);
+                            return actualParallelism.equals(targetParallelism);
```
mxm, line 123: This doesn't factor in any excluded vertices through the excluded vertices setting.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingTracking.java
```diff
@@ -0,0 +1,168 @@
...
+        if (conf.get(AutoScalerOptions.PREFER_TRACKED_RESTART_TIME)) {
+            for (Map.Entry<Instant, ScalingRecord> entry : scalingRecords.entrySet()) {
+                var startTime = entry.getKey();
+                var endTime = entry.getValue().getEndTime();
+                if (endTime != null) {
+                    var restartTime = Duration.between(startTime, endTime).toSeconds();
+                    maxRestartTime = Math.max(restartTime, maxRestartTime);
+                }
+            }
+        }
```
mxm, line 144: Just wondering, if we store the max restart time, isn't it enough to store a single value? This would simplify the code.

mxm (commented):
Thanks Alex!

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingRecord.java
afedulov: My idea was to have structures in place that would be easily extensible. If we are sure we do not not ever need to store anything else here we can store the instant directly. 
For example, if we had a wrapper on top of the history's jobvertex-scoped maps object, if would have been much easier to add the job-scoped restart time on top.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
```diff
@@ -159,19 +161,24 @@ private void runScalingLogic(Context ctx, AutoscalerFlinkMetrics autoscalerMetri
...
             return;
         }
         LOG.debug("Collected metrics: {}", collectedMetrics);
 
-        var evaluatedMetrics = evaluator.evaluate(ctx.getConfiguration(), collectedMetrics);
+        var now = clock.instant();
+        // Scaling tracking data contains previous restart times that are taken into account
+        var scalingTracking = getTrimmedScalingTracking(stateStore, ctx, now);
+        var evaluatedMetrics =
+                evaluator.evaluate(ctx.getConfiguration(), collectedMetrics, scalingTracking);
```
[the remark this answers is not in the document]
afedulov: I don't think this works with the current metrics scoping since it would lead to duplicating the restart time per vertex and we are striving to minimizing the size of the config map. Also, instead of just checking one tracking entry, should we instead iterate over all records of this metric across all vertices and take the maximum over that? Or just trust one observation? If we'll just use one observation, why would we need this data in every vertex and if we know it is supposed to be the same for all vertices, why store it at the vertex level?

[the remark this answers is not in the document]
afedulov: How do you propose to make the association between the rescaling decision and the collected metrics record? My understanding is that `updateMetrics` writes collected metrics irrespective from the rescalings, so we need to somehow decide which history entry to associate the restart time with. To be honest this feels like misusing the data structure not for its original purposes and making code harder to interpret.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingTracking.java
afedulov: Not sure I am missing something, but I thought only vertices with are scaled are entered into the scaling history. Since this is what we build our `targetParallelisms` based on (see `getTargetParallelismOfScaledVertices`) and iterate over  `targetParallelisms`, not the `actualParallelism`, I assumed this is already dealt with.

mxm: I think we only exclude vertices when we generate the `ScalingSummary`. I just want to make sure this logic works, when there are excluded vertices because those would not be scaled and could be set to a different parallelism for other reasons. Just something to double-check.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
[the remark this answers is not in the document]
mxm: Sorry, small correction: I meant `evaluatedMetrics`, not `collectedMetrics`.

We don't need to associate the history entry with any of the metrics used for scaling. We would insert into the evaluated metrics map whatever the current determined rescale time is. This is similar to the MAX_PARALLELISM metric.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingTracking.java
mxm: If we just store the max rescale time, it is sufficient to use a single value. But, trimming the history won't actually work correctly then. We need to keep all values for this to work correctly. So makes sense as-is.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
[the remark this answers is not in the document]
afedulov: The way the scalingHistory is currently structured, this would currently lead to duplicating the restart time unnecessarily per job vertex.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingTracking.java
mxm: Looks like this is already taken care of because we do not emit a ScalingSummary for excluded vertices.

gyfora (changes requested):
I think if we are going to use ` Math.min(maxRestartTime, restartTimeFromConfig)` we should increase the default restart time value in the config from 5 minutes to at least 15 minutes (maybe even higher like 20-30)

With 5 minutes almost all prod jobs with many pods /slower envs or large state will probably take longer (let's say 10-15 minutes) and they will be cut back to 5 minutes leading to an underestimation and potentially multiple scaleups.

What do you think?
cc @mxm

mxm:
What you describe is already the case. Any deployment which currently takes longer than the configured rescale time, will not scale accurately. By adding the feature in its current state, we are conservatively addressing those deployment which come back up much quicker. In our environment, typical rescale time is 1 minute or less.

I think your request requires adding a new configuration `MAX_RESTART_TIME` because we want to (1) have a good default when we haven't yet estimated the rescale time (e.g. first scaling) via the existing setting, and (2) cap the determined rescale time for safety reasons (e.g. unexpected cluster downtime).

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
```diff
@@ -154,6 +154,15 @@ private static ConfigOptions.OptionBuilder autoScalerConfig(String key) {
                                     + RESTART_TIME.key()
                                     + "' will act as an upper bound.");
 
+    public static final ConfigOption<Duration> MAX_RESTART_TIME =
+            autoScalerConfig("restart.time.max")
+                    .durationType()
+                    .defaultValue(Duration.ofMinutes(30))
```
mxm, line 160: We have internal overrides for all configuration defaults, but I don't think this overly pessimistic value is a good default for the general public. I would set this to not more than 15 minutes, even that is still conservative.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingTracking.java
```diff
@@ -142,9 +142,11 @@ public double getMaxRestartTimeSecondsOrDefault(Configuration conf) {
                 }
             }
         }
+        long restartTimeFromConfig = conf.get(AutoScalerOptions.RESTART_TIME).toSeconds();
+        long maxRestartTimeFromConfig = conf.get(AutoScalerOptions.MAX_RESTART_TIME).toSeconds();
         return maxRestartTime == -1
                 ? restartTimeFromConfig
```
mxm, line 148: I think we should always cap the RESTART_TIME by the configured MAX_RESTART_TIME (if configured).

mxm (commented):
Thanks for following up so quickly! 🙏

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingTracking.java
afedulov: @gyfora I'd like to hear your thoughts on this

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
afedulov: Done.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingTracking.java
afedulov: Synced with Max offline - we decided to rename the option to `TRACKED_RESTART_TIME_LIMIT` to make the scope clear.

gyfora: sounds good!

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
```diff
@@ -142,6 +142,27 @@ private static ConfigOptions.OptionBuilder autoScalerConfig(String key) {
                     .withDescription(
                             "Expected restart time to be used until the operator can determine it reliably from history.");
 
+    public static final ConfigOption<Boolean> PREFER_TRACKED_RESTART_TIME =
+            autoScalerConfig("restart.time.tracked.enabled")
+                    .booleanType()
+                    .defaultValue(false)
+                    .withDescription(
+                            "Whether to use the actually observed rescaling restart times instead of the fixed '"
```
mxm, line 150:
```suggestion
                            "Whether to use the actual observed rescaling restart times instead of the fixed '"
```

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
```diff
@@ -142,6 +142,27 @@ private static ConfigOptions.OptionBuilder autoScalerConfig(String key) {
...
+                                    + "samples will be used. The value of '"
+                                    + RESTART_TIME.key()
+                                    + "' will act as an upper bound.");
+
+    public static final ConfigOption<Duration> TRACKED_RESTART_TIME_LIMIT =
+            autoScalerConfig("restart.time.tracked.limit")
+                    .durationType()
+                    .defaultValue(Duration.ofMinutes(15))
+                    .withDescription(
+                            "Maximum cap for the calculated restart time when '"
```
mxm, line 162:
```suggestion
                            "Maximum cap for the observed restart time when '"
```

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingTracking.java
```diff
@@ -0,0 +1,171 @@
...
+                        });
+    }
+
+    /**
+     * Retrieves the maximum restart time in seconds based on the provided configuration and scaling
+     * records. Defaults to the RESTART_TIME from configuration if the PREFER_TRACKED_RESTART_TIME
+     * option is set to false, or if there are no tracking records available. Otherwise, the maximum
+     * observed restart time is capped by the MAX_RESTART_TIME.
+     */
+    public double getMaxRestartTimeSecondsOrDefault(Configuration conf) {
```
mxm, line 133: I would prefer if we passed `Duration` until we perform any calculation.

mxm (approved):
LGTM. Thank you Alex!

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingTracking.java
afedulov: https://github.com/apache/flink-kubernetes-operator/pull/711/commits/93ae804e1466fc75009ca10714a65ad984efbd14

mxm:
Thanks!

docs/layouts/shortcodes/generated/auto_scaler_configuration.html
```diff
@@ -80,6 +80,18 @@
...
             <td>Expected restart time to be used until the operator can determine it reliably from history.</td>
         </tr>
+        <tr>
+            <td><h5>job.autoscaler.restart.time.tracked.enabled</h5></td>
+            <td style="word-wrap: break-word;">false</td>
+            <td>Boolean</td>
+            <td>Whether to use the actually observed rescaling restart times instead of the fixed 'job.autoscaler.restart.time' configuration. If set to true, the maximum restart duration over a number of samples will be used. The value of 'job.autoscaler.restart.time' will act as an upper bound.</td>
+        </tr>
+        <tr>
+            <td><h5>job.autoscaler.restart.time.tracked.limit</h5></td>
```
gyfora, line 90: Sorry for the late comment, we could consider changing this to `job.autoscaler.restart.time-tracking.enabled` and `job.autoscaler.restart.time-tracking.limit` so that we avoid having another config match a prefix (and therefore make it non yaml compliant) 

Not a huge thing but if we want to support yaml configs like in flink we will have to fix it eventually

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingTracking.java
```diff
@@ -0,0 +1,171 @@
...
+                            var scalingTimestamp = entry.getKey();
+                            if (value.getEndTime() == null) {
+                                var targetParallelism =
+                                        getTargetParallelismOfScaledVertices(
+                                                scalingTimestamp, scalingHistory);
+                                var actualParallelism = jobTopology.getParallelisms();
+
+                                if (targetParallelismMatchesActual(
+                                        targetParallelism, actualParallelism)) {
+                                    value.setEndTime(now);
```
gyfora, line 91: Should we maybe log this on debug? so we have an overview if we want to debug this?

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingTracking.java
```diff
@@ -0,0 +1,171 @@
...
+
+/** Stores rescaling related information for the job. */
+@Experimental
+@Data
+@NoArgsConstructor
+@Builder
+public class ScalingTracking {
+
+    /** Details related to recent rescaling operations. */
+    private final TreeMap<Instant, ScalingRecord> scalingRecords = new TreeMap<>();
```
gyfora, line 48: I thought we were going to keep a single record and exponential moving avg

gyfora (commented):
Sorry I did not have time to add some of these comments earlier. I wonder why we decided to keep a history instead of the exponential moving average. That would eliminate the bookkeeping

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingTracking.java
```diff
@@ -0,0 +1,171 @@
...
+     * the number of records to the specified maximum count.
+     *
+     * @param keptTimeSpan Duration for how long recent records are to be kept.
+     * @param keptNumRecords The maximum number of recent records to keep.
+     */
+    public void removeOldRecords(Instant now, Duration keptTimeSpan, int keptNumRecords) {
+        var cutoffTime = now.minus(keptTimeSpan);
+
+        // Remove records older than the cutoff time
+        scalingRecords.headMap(cutoffTime).clear();
```
gyfora, line 164: Wouldn't this clear the history if we don't scale for 24 hours? then we fall back to the config ?

gyfora: We should keep at least 1 in the history. But given that scalings do not happen that often the history will always only have 1-2 records only. So EMA may be more robust. cc @mxm

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingTracking.java
afedulov: I guess [this](https://github.com/apache/flink-kubernetes-operator/pull/711#issuecomment-1815381974) got buried in the notifications:
```
@gyfora me and Max briefly discussed offline and came to the conclusion that starting with
 evaluating the maximum restart time capped by the RESTART_TIME setting is probably 
good enough for the first step. It has the benefit of giving the most "conservative" 
evaluation and we can add the moving average after some baseline testing. What do you think?
```

gyfora: I saw this but this doesn't mention anything about history etc and refers to an offline discussion :) 
Combined with the other comment related to the trimming issue (losing the restart info after 24h) I think the exponential moving avg is a simpler and slightly more robust initial approach

gyfora: `estimatedEMA = estimatedEMA * x + newMeasurmenet * (1-x)`
we could start with `x=0.5` which is pretty aggressive smoothing but should be fine give we don't have many scalings

afedulov: General question - we are missing the job-based data structure that keeps track of the past rescaling details. Should be need to add something in the future, with the current structure it is as simple as adding data fields to the ScalingRecord. I am OK with removing the map, but the question is - are we sure we won't require something similar in the future anyways?

afedulov: I guess it could be argued that we can always send previous stats as metrics, but why do we then keep vertex-based scalingHistory?

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingTracking.java
mxm: +1

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingTracking.java
mxm: We agreed offline to keep at least one observation to avoid having to recalibrate the restart time when the observation history expires. EMA is a good alternative as well but for now we chose to take the max of the last observed restart times.

mxm:
The docs need to be regenerated after the last config change.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
```diff
@@ -142,6 +142,27 @@ private static ConfigOptions.OptionBuilder autoScalerConfig(String key) {
                     .withDescription(
                             "Expected restart time to be used until the operator can determine it reliably from history.");
 
+    public static final ConfigOption<Boolean> PREFER_TRACKED_RESTART_TIME =
+            autoScalerConfig("restart.time-tracking.enabled")
+                    .booleanType()
+                    .defaultValue(false)
```
mxm, line 148: This should probably be enabled by default. We can still change this default though.

mxm:
Great work! Thanks @afedulov!

afedulov:
Thanks @mxm and @gyfora !

PR #712, opened 2023-11-14
[Flink 31966] Flink Kubernetes operator lacks TLS support

tagarr:

If the FlinkDeployment CR is configured to use ssl, the operator is unable to properly reconcile the deployment as the underlying RestClient the operator uses tries to load the certificates defined in the config.

## Brief change log

- Change helm chart to provide values for a cert-manager created jks secret and its password
- Create env vars that define the location of the truststore (optionally keystore) on disk
- If the deployments config isRestSSLEnabled, modify the config to point to the operators certs on creation of rest clients
- Catch runtime exceptions on submitting flink applications or flink sessions if they are due to invalid ssl config

## Verifying this change

- Added a SecureFlinkServiceTest to check that restclients are created when certs are defined and fail otherwise
- Added flink tls examples and updated the README

## Documentation


tagarr:
@gaborgsomogyi @gyfora I'd love to get his in for the 1.7.0 release. Any feedback would be much appreciated

gaborgsomogyi:
Thanks for the efforts! I've gone through the code and the direction looks good in general.
There are some points where we can simplify the code without changing the functionality:
* It's not required to make the certs mount point configurable. A good how it can be done is `/opt/flink/artifacts`. This can spare quite some code lines. Of course keeping it optional still makes sense.
* The template gives the possibility not to hardcode keys inside but just define the root yaml node. During YAML generation all the underneath subtree is just copied over. A good example is `livenessProbe`. This allows us to support any custom parameters without touching the template again.

tagarr:
Hi @gaborgsomogyi made changes according to your review

helm/flink-kubernetes-operator/templates/flink-operator.yaml
```diff
@@ -130,6 +141,10 @@ spec:
             - name: flink-artifacts-volume
               mountPath: /opt/flink/artifacts
             {{- end }}
+          {{- if .Values.tls.create }}
+            - name: flink-operator-cert-secret
+              mountPath: /opt/flink/artifacts/operator-cert
```
gaborgsomogyi, line 146: AFAIK when there is an overlap in the mounts then one of them is going to be hidden (Either `/opt/flink/artifacts` or `/opt/flink/artifacts/operator-cert`). Maybe `/opt/flink/certs` would be a better choice.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
```diff
@@ -195,8 +199,25 @@ public void submitApplicationCluster(
...
             validateHaMetadataExists(conf);
         }
+        try {
+            deployApplicationCluster(jobSpec, removeOperatorConfigs(conf));
+        } catch (RuntimeException e) {
+            LOG.warn("Caught Exception " + e.getMessage());
+            if (!isValidRuntimeException(conf, e)) {
+                throw e;
+            }
+        }
```
gyfora, line 209: I am very confused by this logic, can we keep the previous logic and just let the exceptions fall through? What's the point of logging them? Deploy errors should result in errors

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
```diff
@@ -1075,4 +1106,59 @@ protected void updateStatusAfterClusterDeletion(FlinkDeploymentStatus status) {
...
+                            }
+                            conf.removeConfig(SecurityOptions.SSL_TRUSTSTORE);
+                            conf.removeConfig(SecurityOptions.SSL_TRUSTSTORE_PASSWORD);
+                            conf.removeConfig(SecurityOptions.SSL_KEYSTORE);
+                            conf.removeConfig(SecurityOptions.SSL_KEYSTORE_PASSWORD);
+                        });
+        return conf;
+    }
+
+    private boolean isValidRuntimeException(Configuration conf, RuntimeException e) {
```
gyfora, line 1150: Please see my previous comment, I think we should get rid of this logic completely.

helm/flink-kubernetes-operator/templates/flink-operator.yaml
gyfora, line 146: I don't think we can mount under `/opt/flink/artifacts` , this should not overlap and be something like `/opt/flink/certs` instead

helm/flink-kubernetes-operator/templates/flink-operator.yaml
```diff
@@ -110,6 +110,17 @@ spec:
...
               value: {{ .Values.jvmArgs.operator }}
+            {{- if .Values.tls.create }}
+            - name: OPERATOR_KEYSTORE_PATH
+              value: /opt/flink/artifacts/operator-cert/keystore.jks
+            - name: OPERATOR_TRUSTSTORE_PATH
+              value: /opt/flink/artifacts/operator-cert/truststore.jks
+            - name: OPERATOR_KEYSTORE_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  {{- toYaml .Values.tls.certPasswordSecretRef | nindent 18 }}
```
gyfora, line 121: can we call this property `.Values.tls.secretKeyRef` ? that way it's obvious what is expected to go there

helm/flink-kubernetes-operator/templates/flink-operator.yaml
```diff
@@ -229,6 +244,12 @@ spec:
             - key: keystore.p12
               path: keystore.p12
         {{- end }}
+        {{- if .Values.tls.create }}
+        - name: flink-operator-cert-secret
+          secret:
+            secretName: {{ .Values.tls.certSecret }}
```
gyfora, line 250: this property could be called `.Values.tls.secretName` to be consistent

gyfora (changes requested):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
```diff
@@ -1075,4 +1106,59 @@ protected void updateStatusAfterClusterDeletion(FlinkDeploymentStatus status) {
...
+        // that would be given to the deployment. When SSL is enabled it will throw
+        // a ClusterRetrieveException as the operator does not have the certs where they
+        // would be mounted on the client
+        if (SecurityOptions.isRestSSLEnabled(conf)
+                && trustStorePath.isPresent()
+                && Files.exists(Paths.get(trustStorePath.get()))
+                && e.getCause() instanceof ClusterRetrieveException) {
+            return true;
+        }
+        return false;
```
gaborgsomogyi, line 1162: I don't understand this functionality at all🤔 How I can imagine this feature from high level perspective is the following:
* One configs certs for the operator
* One configs either unsecure or secure connection per job
* Operator picks up the job config
* If unsecure then opens an unsecure REST client
* If secure then opens an secure REST client
* If any connection fails then we shouldn't catch exceptions

If you can elaborate it would be helpful.

gaborgsomogyi:
Sorry for the late response we've quite some tasks to do.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
tagarr: I'll try and explain the problem. If the tls config for the flinkdeployment is different to the operator then on creating (deploying) a FlinkDeployment, a rest client is always returned see https://github.com/apache/flink/blob/c5808b04fdce9ca0b705b6cc7a64666ab6426875/flink-kubernetes/src/main/java/org/apache/flink/kubernetes/KubernetesClusterDescriptor.java#L232. This has the tls configuration for the deployment and may not match that of the operator, so will throw an exception even though the rest client is not and will never be used. This exception is then propagated into the CR and looks like its failed. However after a few reconciles it will correct itself. I thought catching this erroneous exception would prevent users thinking they have an issue with their deployment

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
tagarr: Explained above why I'm catching exception. This check here is to see if the exception is expected i.e. ssl is on, the operator has a truststore and the exception is that of trying to create a restClient. If we remove this, I will have to document that the tls config for the deployment needs to be exactly the same as the operator i.e. mounted in the same place and using the same store password.

tagarr:
So if you prefer not catching the exception I can just add some info in the example readme about ensuring ssl location and password are consistent or you may see exceptions being thrown in the operator

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
gyfora: It’s normal to get exceptions on invalid configs . It is the expected behavior , we have handling for these errors I believe

tagarr:
@gyfora @gaborgsomogyi made required changes

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
```diff
@@ -199,6 +202,10 @@ public void submitApplicationCluster(
         deployApplicationCluster(jobSpec, removeOperatorConfigs(conf));
     }
 
+    public void submitSessionCluster(Configuration conf) throws Exception {
```
gaborgsomogyi, line 205: Override is missing.

gaborgsomogyi (commented):
Thanks for the efforts! Looks good and when the tiny thing fixed it's good to go.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
tagarr: Great spot, thanks

PR #715, opened 2023-11-24
[FLINK-30593] Add tests for ScalingTracking#removeOldRecords()

afedulov:
This is a trivial addition of tests for removal of old records in `ScalingTracking`

### DIFF
--- a/flink-autoscaler/src/test/java/org/apache/flink/autoscaler/ScalingTrackingTest.java
+++ b/flink-autoscaler/src/test/java/org/apache/flink/autoscaler/ScalingTrackingTest.java
@@ -105,16 +105,6 @@ class ScalingTrackingTest {
         assertThat(result).isEqualTo(configuredMaxRestartTime);
     }
 
-    private void setUpScalingRecords(Duration secondRescaleDuration) {
-        scalingTracking.addScalingRecord(
-                Instant.parse("2023-11-15T16:00:00.00Z"),
-                new ScalingRecord(Instant.parse("2023-11-15T16:03:00.00Z")));
-        var secondRecordStart = Instant.parse("2023-11-15T16:20:00.00Z");
-        scalingTracking.addScalingRecord(
-                secondRecordStart,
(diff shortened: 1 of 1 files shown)


flink-autoscaler/src/test/java/org/apache/flink/autoscaler/ScalingTrackingTest.java
```diff
@@ -52,6 +52,7 @@ void setUp() {
         conf.set(AutoScalerOptions.PREFER_TRACKED_RESTART_TIME, true);
     }
 
+    // ------------- getMaxRestartTimeOrDefault -------------
```
mxm, line 55: What does this comment mean?

mxm: Why is it formatted that way?

afedulov: The test was growing large and I added sections for easier navigation. The name indicates the method that is the focus for the tests in the respective section. Formatting is arbitrary. We could do something like this, in line with how code blocks are sometimes demarcated in Flink:
```
    // ------------------------------------------------------------------------
    //  Tests for getMaxRestartTimeOrDefault()
    // ------------------------------------------------------------------------
```

afedulov:
@mxm is this the right ticket?  
[FLINK-33572] Minimize ConfigMap API operations for autoscaler state

flink-autoscaler/src/test/java/org/apache/flink/autoscaler/ScalingTrackingTest.java
mxm: I would advise against these kind of comments. Similarly to JavaDocs, they tend to add more confusion than actually helping new people. Plus, they become outdated quickly.

afedulov: removed...

PR #706, opened 2023-11-10
[FLINK-33522] Use created savepoint even if job cancellation fails

mxm:
Under certain circumstances, savepoint creation can succeed but the job fails afterwards. One example is when there are messages being distributed by the source coordinator to finished tasks. This is possibly a Flink bug although it's not clear yet how to resolve the issue.

After the savepoint succeeded Flink fails the job like this:

```
Source (1/2) (cd4d56ddb71c0e763cc400bcfe2fd8ac_4081cf0163fcce7fe6af0cf07ad2d43c_0_0) switched from RUNNING to FAILED on host-taskmanager-1-1 @ ip(dataPort=36519). 
```

```
An OperatorEvent from an OperatorCoordinator to a task was lost. Triggering task failover to ensure consistency. Event: 'AddSplitEvents[[[B@722a23fa]]', targetTask: Source (1/2) - execution #0
Caused by:
org.apache.flink.runtime.operators.coordination.TaskNotRunningException: Task is not running, but in state FINISHED
   at org.apache.flink.runtime.taskmanager.Task.deliverOperatorEvent(Task.java:1502)
   at org.apache.flink.runtime.taskexecutor.TaskExecutor.sendOperatorEventToTask
```

Inside the operator this is processed as:

```
java.util.concurrent.CompletionException: org.apache.flink.runtime.scheduler.stopwithsavepoint.StopWithSavepointStoppingException: A savepoint has been created at: s3://..., but the corresponding job 1b1a3061194c62ded6e2fe823b61b2ea failed during stopping. The savepoint is consistent, but might have uncommitted transactions. If you want to commit the transaction please restart a job from this savepoint. 

          java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:395) 
          java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2022) 
          org.apache.flink.kubernetes.operator.service.AbstractFlinkService.cancelJob(AbstractFlinkService.java:319) 
          org.apache.flink.kubernetes.operator.service.NativeFlinkService.cancelJob(NativeFlinkService.java:121) 
          org.apache.flink.kubernetes.operator.reconciler.deployment.ApplicationReconciler.cancelJob(ApplicationReconciler.java:223) 
          org.apache.flink.kubernetes.operator.reconciler.deployment.AbstractJobReconciler.reconcileSpecChange(AbstractJobReconciler.java:122) 
[...]

PR #716, opened 2023-11-29
[FLINK-33522] Be ware of SerializedThrowable when checking for StopWithSavepointStoppingException

mxm:
Turns out that the previous detection code in #706 may not always fire correctly due to an encapsulated serialized throwable. This minor change fixes that.


1996fanrui (approved):
Thanks @mxm for the fix, LGTM.

PR #717, opened 2023-11-30
[FLINK-33701][autoscaler] Fix the doc about restart.time-tracking.enabled

1996fanrui:
The restart.time-tracking.limit as the upper bound instead of restart.time.

image


1996fanrui (commented):
Hey @afedulov , would you mind helping take a look in your free time? thanks~

PR #720, opened 2023-11-30
[FLINK-33710] Prevent NOOP spec updates from the autoscaler parallelism override map

mxm:
The operator supports two modes to apply autoscaler changes:

1. Use the internal Flink config {{pipeline.jobvertex-parallelism-overrides}} 
2. Make use of Flink's Rescale API 

For (1), a string has to be generated for the Flink config with the actual overrides. This string has to be deterministic for a given map. But it is not.

Consider the following observed log:

```
  >>> Event  | Info    | SPECCHANGED     | SCALE change(s) detected (Diff: FlinkDeploymentSpec[flinkConfiguration.pipeline.jobvertex-parallelism-overrides : 92542d1280187bd464274368a5f86977:3,9f979ed859083299d29f281832cb5be0:1,84881d7bda0dc3d44026e37403420039:1,1652184ffd0522859c7840a24936847c:1 -> 9f979ed859083299d29f281832cb5be0:1,84881d7bda0dc3d44026e37403420039:1,92542d1280187bd464274368a5f86977:3,1652184ffd0522859c7840a24936847c:1]), starting reconciliation. 
```

The overrides are identical but the order is different which triggers a redeploy. This does not seem to happen often but some deterministic string generation (e.g. sorting by key) is required to prevent any NOOP updates.

### DIFF
--- a/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/autoscaler/KubernetesScalingRealizerTest.java
+++ b/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/autoscaler/KubernetesScalingRealizerTest.java
@@ -1,0 +1,53 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.kubernetes.operator.autoscaler;
+
+import org.apache.flink.configuration.PipelineOptions;
+
+import org.junit.Test;
+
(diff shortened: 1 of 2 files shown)


flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/autoscaler/KubernetesScalingRealizerTest.java
```diff
@@ -0,0 +1,53 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.kubernetes.operator.autoscaler;
+
+import org.apache.flink.configuration.PipelineOptions;
+
+import org.junit.Test;
```
1996fanrui, line 22:
```suggestion
import org.junit.jupiter.api.Test;
```

nit: Would you mind using the junit5? I see flink project is upgrading it, and new test is better to use it.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/autoscaler/KubernetesScalingRealizerTest.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
```
1996fanrui, line 28: How about using `org.junit.jupiter.api.Assertions`?

1996fanrui (approved):
Hey @mxm , thanks for the fix!

The change looks good to me, I just left 2 minor comments about junit version.

1996fanrui (approved):

mxm:
I was thinking to push one more commit here to address existing deployments.

PR #721, opened 2023-12-01
[FLINK-33710] Prevent triggering cluster upgrades for permutations of the same overrides

mxm:
Previous fix in #720 made the parallelism override string deterministic, but it will likely result in existing deployments to trigger a one-off unneeded spec update.

To prevent this and to update only once an actual scaling occurs, we need to compare the existing overrides with the new ones and check if they are identical. In this case, we restore the current override string with its own permutation.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
@@ -20,6 +20,8 @@ package org.apache.flink.kubernetes.operator.reconciler.deployment;
 import org.apache.flink.annotation.VisibleForTesting;
 import org.apache.flink.autoscaler.JobAutoScaler;
 import org.apache.flink.configuration.Configuration;
+import org.apache.flink.configuration.ConfigurationUtils;
+import org.apache.flink.configuration.PipelineOptions;
 import org.apache.flink.kubernetes.configuration.KubernetesConfigOptions;
 import org.apache.flink.kubernetes.operator.api.AbstractFlinkResource;
 import org.apache.flink.kubernetes.operator.api.FlinkDeployment;
@@ -52,6 +54,8 @@ import io.javaoperatorsdk.operator.api.reconciler.DeleteControl;
 import io.javaoperatorsdk.operator.processing.event.ResourceID;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nullable;
 
 import java.time.Clock;
 import java.time.Duration;
@@ -137,7 +141,13 @@ public abstract class AbstractFlinkResourceReconciler<
                 cr.getStatus().getReconciliationStatus().deserializeLastReconciledSpec();
         SPEC currentDeploySpec = cr.getSpec();
 
-        applyAutoscaler(ctx);
+        applyAutoscaler(
+                ctx,
+                lastReconciledSpec != null
+                        ? lastReconciledSpec
+                                .getFlinkConfiguration()
+                                .get(PipelineOptions.PARALLELISM_OVERRIDES.key())
+                        : null);
 
         var reconciliationState = reconciliationStatus.getState();
         var specDiff =
@@ -180,13 +190,28 @@ public abstract class AbstractFlinkResourceReconciler<
         }
     }
 
-    private void applyAutoscaler(FlinkResourceContext<CR> ctx) throws Exception {
+    private void applyAutoscaler(FlinkResourceContext<CR> ctx, @Nullable String existingOverrides)
+            throws Exception {
         var autoScalerCtx = ctx.getJobAutoScalerContext();
         boolean autoscalerEnabled =
                 ctx.getResource().getSpec().getJob() != null
                         && ctx.getObserveConfig().getBoolean(AUTOSCALER_ENABLED);
         autoScalerCtx.getConfiguration().set(AUTOSCALER_ENABLED, autoscalerEnabled);
+
         autoscaler.scale(autoScalerCtx);
+
+        // Check that the overrides actually changed and not merely the String representation
+        var flinkConfig = ctx.getResource().getSpec().getFlinkConfiguration();
+        var newOverrides = flinkConfig.get(PipelineOptions.PARALLELISM_OVERRIDES.key());
+        if (existingOverrides != null && newOverrides != null) {
+            var existingOverridesMap =
+                    ConfigurationUtils.convertValue(existingOverrides, Map.class);
+            var newOverridesMap = ConfigurationUtils.convertValue(newOverrides, Map.class);
+            if (newOverridesMap.equals(existingOverridesMap)) {
+                // Existing overrides equal current ones, avoid changing the spec
+                flinkConfig.put(PipelineOptions.PARALLELISM_OVERRIDES.key(), existingOverrides);
+            }
+        }
     }
 
     private void triggerSpecChangeEvent(CR cr, DiffResult<SPEC> specDiff, KubernetesClient client) {
--- a/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconcilerTest.java
+++ b/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconcilerTest.java
@@ -931,6 +931,56 @@ public class ApplicationReconcilerTest extends OperatorTestBase {
                         .get(PipelineOptions.PARALLELISM_OVERRIDES));
     }
 
+    @Test
+    public void testNoSpecChangeForEffectivelyIdenticalParallelismOverrides() throws Exception {
+        var ctxFactory =
+                new TestingFlinkResourceContextFactory(
+                        configManager, operatorMetricGroup, flinkService, eventRecorder);
+        appReconciler =
+                new ApplicationReconciler(eventRecorder, statusRecorder, new NoopJobAutoscaler<>());
+
+        var deployment = TestUtils.buildApplicationCluster();
+
+        var v1 = new JobVertexID(0, 0);
+        var v2 = new JobVertexID(0, 1);
+        var existingOverrides = String.format("%s:1,%s:2", v1, v2);
+
+        // Apply initial overrides
+        deployment
+                .getSpec()
+                .getFlinkConfiguration()
+                .put(PipelineOptions.PARALLELISM_OVERRIDES.key(), existingOverrides);
+
+        // Reconcile overrides
+        appReconciler.reconcile(ctxFactory.getResourceContext(deployment, context));
+        verifyAndSetRunningJobsToStatus(deployment, flinkService.listJobs());
+        assertEquals(
+                ReconciliationState.DEPLOYED,
+                deployment.getStatus().getReconciliationStatus().getState());
+        assertEquals("RUNNING", deployment.getStatus().getJobStatus().getState());
+
+        // Reverse the order of the same overrides
+        var existingOverridesDifferentPermutation = String.format("%s:2,%s:1", v2, v1);
+        deployment
+                .getSpec()
+                .getFlinkConfiguration()
+                .put(
+                        PipelineOptions.PARALLELISM_OVERRIDES.key(),
+                        existingOverridesDifferentPermutation);
+
+        // Ensure that the permutation in the override string does not trigger a deploy
+        appReconciler.reconcile(ctxFactory.getResourceContext(deployment, context));
+        assertEquals(
+                ReconciliationState.DEPLOYED,
+                deployment.getStatus().getReconciliationStatus().getState());
+        assertEquals(
+                existingOverrides,
+                ctxFactory
+                        .getResourceContext(deployment, context)
+                        .getObserveConfig()
+                        .getString(PipelineOptions.PARALLELISM_OVERRIDES.key(), ""));
+    }
+
     @ParameterizedTest
     @MethodSource("org.apache.flink.kubernetes.operator.TestUtils#flinkVersions")
     public void verifyJobIdNotResetDuringLastStateRecovery(FlinkVersion flinkVersion) {


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
```diff
@@ -180,13 +190,28 @@ public void reconcile(FlinkResourceContext<CR> ctx) throws Exception {
...
-    private void applyAutoscaler(FlinkResourceContext<CR> ctx) throws Exception {
+    private void applyAutoscaler(FlinkResourceContext<CR> ctx, @Nullable String existingOverrides)
+            throws Exception {
         var autoScalerCtx = ctx.getJobAutoScalerContext();
         boolean autoscalerEnabled =
                 ctx.getResource().getSpec().getJob() != null
                         && ctx.getObserveConfig().getBoolean(AUTOSCALER_ENABLED);
         autoScalerCtx.getConfiguration().set(AUTOSCALER_ENABLED, autoscalerEnabled);
+
         autoscaler.scale(autoScalerCtx);
```
afedulov, line 201: Maybe add a comment explaining why this call to `scale` does not lead to the unnecessary update if we only patch the new config after it. I personally do not directly understand why this is the case.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
```diff
@@ -180,13 +190,28 @@ public void reconcile(FlinkResourceContext<CR> ctx) throws Exception {
...
+
+        // Check that the overrides actually changed and not merely the String representation
+        var flinkConfig = ctx.getResource().getSpec().getFlinkConfiguration();
+        var newOverrides = flinkConfig.get(PipelineOptions.PARALLELISM_OVERRIDES.key());
```
gyfora, line 205: It’s not safe to access configs by key directly . We can use the ctx getDeployConfig and getObserveConfig to get the configs and then we can even avoid passing the extra parameter to the method

mxm: I'm curious, why is it not safe?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
```diff
@@ -137,7 +141,13 @@ public void reconcile(FlinkResourceContext<CR> ctx) throws Exception {
                 cr.getStatus().getReconciliationStatus().deserializeLastReconciledSpec();
         SPEC currentDeploySpec = cr.getSpec();
 
-        applyAutoscaler(ctx);
+        applyAutoscaler(
+                ctx,
+                lastReconciledSpec != null
```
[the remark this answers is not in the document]
mxm: Sure, let me look into that.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
```diff
@@ -180,13 +190,28 @@ public void reconcile(FlinkResourceContext<CR> ctx) throws Exception {
...
                 ctx.getResource().getSpec().getJob() != null
                         && ctx.getObserveConfig().getBoolean(AUTOSCALER_ENABLED);
         autoScalerCtx.getConfiguration().set(AUTOSCALER_ENABLED, autoscalerEnabled);
+
         autoscaler.scale(autoScalerCtx);
+
+        // Check that the overrides actually changed and not merely the String representation
+        var flinkConfig = ctx.getResource().getSpec().getFlinkConfiguration();
+        var newOverrides = flinkConfig.get(PipelineOptions.PARALLELISM_OVERRIDES.key());
+        if (existingOverrides != null && newOverrides != null) {
```
1996fanrui, line 206:
```suggestion
        if (existingOverrides != null && newOverrides != null && !existingOverrides.equals(newOverrides)) {
```

When `existingOverrides.equals(newOverrides)`, we don't need to execute this logic. So we can add one condition here.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
1996fanrui: If I understand correct. The root cause is the logic related to `boolean specChanged` in `AbstractFlinkResourceReconciler#reconcile`. The `specChanged` will be true, when the ordering of the parallelism overrides is changed. And operator will scale job.

Right?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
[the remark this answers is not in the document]
mxm: @gyfora Please check out the newest commit.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
mxm: The rewritten version using Gyula's suggestions doesn't require this anymore.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
afedulov: Makes sense, thanks.

1996fanrui (approved):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
```diff
@@ -186,7 +187,41 @@ private void applyAutoscaler(FlinkResourceContext<CR> ctx) throws Exception {
...
+
         autoscaler.scale(autoScalerCtx);
+        putBackOldParallelismOverridesIfNewOnesAreMerelyAPermutation(ctx);
+    }
+
+    private static <
+                    CR extends AbstractFlinkResource<SPEC, STATUS>,
+                    SPEC extends AbstractFlinkSpec,
+                    STATUS extends CommonStatus<SPEC>>
+            void putBackOldParallelismOverridesIfNewOnesAreMerelyAPermutation(
```
gyfora, line 199: Could we move this logic to the `ScalingRealizer`? https://github.com/apache/flink-kubernetes-operator/commit/158cbe29169cbfb7fa7ad676fb0273fd7ef6d25e adds the logic there and this issue comes from that change. I feel these 2 changes logically belong together and it's weird that we break something in once place and fix it in another while it could be simply next to each other.

It will likely also make the whole logic a bit simpler

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
```diff
@@ -186,7 +187,41 @@ private void applyAutoscaler(FlinkResourceContext<CR> ctx) throws Exception {
...
         autoScalerCtx.getConfiguration().set(AUTOSCALER_ENABLED, autoscalerEnabled);
+
         autoscaler.scale(autoScalerCtx);
+        putBackOldParallelismOverridesIfNewOnesAreMerelyAPermutation(ctx);
+    }
+
+    private static <
+                    CR extends AbstractFlinkResource<SPEC, STATUS>,
+                    SPEC extends AbstractFlinkSpec,
+                    STATUS extends CommonStatus<SPEC>>
```
gyfora, line 198: Can we rename this to `resetParallelismOverridesIfUnchanged` ? The current naming is unusually verbose for the codebase. It may be better to add a javadoc comment instead. But this may be irrelevant if you check my other comment. We don't need to replace it if we don't set it in the first place, but for that we need to move the logic

gyfora (changes requested):
Thanks @mxm , I think we should move this to the `KubernetesScalingRealizer` to remove this from the reconciler. It doesn't really belong here and moving it to the realiser where the problem originates will simplify the logic further.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
mxm: It's not possible to add the change there because we do not have access to the current overrides.

mxm: Also https://github.com/apache/flink-kubernetes-operator/commit/158cbe29169cbfb7fa7ad676fb0273fd7ef6d25e does not introduce the problem. It fixes it at the cost of restarting once instead of every time we redeploy the operator.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
```diff
@@ -180,13 +190,28 @@ public void reconcile(FlinkResourceContext<CR> ctx) throws Exception {
...
+            throws Exception {
         var autoScalerCtx = ctx.getJobAutoScalerContext();
         boolean autoscalerEnabled =
                 ctx.getResource().getSpec().getJob() != null
                         && ctx.getObserveConfig().getBoolean(AUTOSCALER_ENABLED);
         autoScalerCtx.getConfiguration().set(AUTOSCALER_ENABLED, autoscalerEnabled);
+
         autoscaler.scale(autoScalerCtx);
+
+        // Check that the overrides actually changed and not merely the String representation
```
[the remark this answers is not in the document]
mxm: This is logic has been moved. I think it is easier to understand now.

mxm:
PTAL @gyfora

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/autoscaler/KubernetesScalingRealizerTest.java
```diff
@@ -30,12 +31,23 @@
...
-
+    public void testAutoscalerOverridesStringDoesNotChangeUnlessOverridesChange() {
         KubernetesJobAutoScalerContext ctx =
                 TestingKubernetesAutoscalerUtils.createContext("test", null);
+        FlinkDeployment resource = (FlinkDeployment) ctx.getResource();
+
+        // Create resource with existing parallelism overrides
+        resource.getSpec()
+                .getFlinkConfiguration()
+                .put(PipelineOptions.PARALLELISM_OVERRIDES.key(), "a:1,b:2");
```
1996fanrui, line 42: Could we test `"a:1,b:2"` and `"b:2,a:1"`?

Also, the `assertThat` have  "a:1,b:2"`  as well, it's  better to extract  an field.

gyfora, line 42: Would it make sense to reverse this to `"b:2,a1" ` so that the test actually fails without the change?

1996fanrui: It's HashMap instead of TreeMap or SortedMap, so the ordering cannot be determined, right?

If so, I prefer to test both of `"a:1,b:2"` and `"b:2,a:1"`, wdyt?

mxm: There is no need to test any permutations since the overrides are deterministically set to `a:1,b:2` in the beginning. Then, new overrides are supplied via a LinkedHashMap which produces a deterministic `b:2,a:1`. Then the test verifies that the config string is not changed.

gyfora (approved):

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/autoscaler/KubernetesScalingRealizerTest.java
1996fanrui: Thanks for your clarification!

I prefer to test both of `"a:1,b:2"` and `"b:2,a:1"` for 2 reasons:

- We always use the black box testing, it means we don't care about some detailed logic or `Data Structure` inside of test method. We just test its feature.
- In the most of cases, the traversal order is determined when we using the `LinkedHashMap`, and the serialization result is often `b:2,a:1`, but I don't think the serialization result is fixed. That's means when 2 `LinkedHashMap` are equal, but the serialization results may not be equal. And that's why we need this PR, right?
  - So I don't think our test should make an assumption: the serialization result is fixed.
  - Also, the test is better to cover more cases. (For example, the `LinkedHashMap` is changed to `TreeMap` in the future.)


BTW, current change is fine, and this comment isn't necessary for me, so I approved it. If you think it's not needed, feel free to merge it. Thanks~

mxm: Thanks for elaborating further on this. Testing is somewhat of an art. There is also an argument for striking a balance between a specific test and one which tests all possible code paths (maybe even ones which don't yet exist).

Black box testing is one way of testing. White box testing is also a valid approach. It depends on the situation and it is also a matter of taste and how confident one feels that the test covers the most important scenarios.

I completely understand your thought, and I've changed the test method to address this concern.

1996fanrui: Thank you for the feedback! ❤️

1996fanrui (approved):

PR #724, opened 2023-12-06
[FLINK-33763] Support savepoint redeployment through a nonce

gyfora:

A common request is to support a streamlined, user friendly way of redeploying from a target savepoint.
Previously this was only possible by deleting the CR and recreating it with initialSavepointPath. A big downside of this approach is a loss of savepoint/checkpoint history in the status that some platforms may need, resulting in non-cleaned up save points etc.

We  introduce a `savepointRedeployNonce` field in the job spec similar to other action trigger nonces.
If the nonce changes to a new non null value the job will be redeployed from the path specified in the initialSavepointPath (or empty state If the path is empty)

As redeployment requires the deletion of HA metadata and previous checkpoint info, rollbacks are not supported after redeployments (changing the nonce)

The PR also introduces an improvement to the restartNonce and savepointRedeployNonce to ignore null changes (when the user nulls out the nonce). This is more practical and inline with the behaviour of the savepoint/checkpoint triggering logic

## Brief change log

  - *Add savepointRedeployNonce to `spec.job`*
  - *Implement redeployment logic*
  - *Tests*
  - *Doc updates*

## Verifying this change

Unit tests and manual validation on local and remote envs.

## Documentation

### DIFF
--- a/docs/content/docs/custom-resource/job-management.md
+++ b/docs/content/docs/custom-resource/job-management.md
@@ -288,12 +288,33 @@ Rollback is currently only supported for `FlinkDeployments`.
 
 ## Manual Recovery
 
-There are cases when manual intervention is required from the user to recover a Flink application deployment.
+There are cases when manual intervention is required from the user to recover a Flink application deployment or to restore to a user specified state.
 
 In most of these situations the main reason for this is that the deployment got into a state where the operator cannot determine the health of the application or the latest checkpoint information to be used for recovery.
 While these cases are not common, we need to be prepared to handle them.
 
-Fortunately almost any issue can be recovered by the user manually by using the following steps:
+Users have two options to restore a job from a target savepoint / checkpoint
+
+### Redeploy using the savepointRedeployNonce
+
+It is possible to redeploy a `FlinkDeployment` or `FlinkSessionJob` resource from a target savepoint by using the combination of `savepointRedeployNonce` and `initialSavepointPath` in the job spec:
+
+```yaml
+ job:
+   initialSavepointPath: file://redeploy-target-savepoint
+   savepointRedeployNonce: null -> 1
+```
+
+When changing the `savepointRedeployNonce` the operator will redeploy the job to the savepoint defined in the `initialSavepointPath` or empty state if the path is null. 
+
+{{< hint warning >}}
+Rollbacks are not supported after redeployments.
+{{< /hint >}}
+
+### Delete and recreate the custom resource 
+ 
+Alternatively you can completely delete and recreate the custom resources to solve almost any issues. This will fully reset the status information to start from a clean slate.
+However, this also means that savepoint history is lost and the operator won't clean up past periodic savepoints taken before the deletion.
 
  1. Locate the latest checkpoint/savepoint metafile in your configured checkpoint/savepoint directory.
  2. Delete the `FlinkDeployment` resource for your application
@@ -303,10 +324,3 @@ Fortunately almost any issue can be recovered by the user manually by using the 
 
 These steps ensure that the operator will start completely fresh from the user defined savepoint path and can hopefully fully recover.
 Keep an eye on your job to see what could have cause the problem in the first place.
-
-{{< hint info >}}
-The main idea behind the recovery process is that the user needs to manually override the target checkpoint/savepoint location because it is not known to the operator.
-The only way to do this is to delete the previous deployment resource fully, and recreate it with `initialSavepointPath` set.
-
-The `initialSavepointPath` setting only takes effect on the first deployment and the operator takes over checkpoint management after that.
-{{< /hint >}}
--- a/docs/content/docs/custom-resource/job-management.md
+++ b/docs/content/docs/custom-resource/job-management.md
@@ -306,7 +306,7 @@ It is possible to redeploy a `FlinkDeployment` or `FlinkSessionJob` resource fro
    savepointRedeployNonce: 1
 ```
 
-When changing the `savepointRedeployNonce` the operator will redeploy the job to the savepoint defined in the `initialSavepointPath` or empty state if the path is null. 
+When changing the `savepointRedeployNonce` the operator will redeploy the job to the savepoint defined in the `initialSavepointPath`. The savepoint path must not be empty. 
 
 {{< hint warning >}}
 Rollbacks are not supported after redeployments.
--- a/docs/content/docs/custom-resource/job-management.md
+++ b/docs/content/docs/custom-resource/job-management.md
@@ -302,7 +302,8 @@ It is possible to redeploy a `FlinkDeployment` or `FlinkSessionJob` resource fro
 ```yaml
  job:
    initialSavepointPath: file://redeploy-target-savepoint
-   savepointRedeployNonce: null -> 1
+   # If not set previously, set to 1, otherwise increment, e.g. 2
+   savepointRedeployNonce: 1
 ```
 
 When changing the `savepointRedeployNonce` the operator will redeploy the job to the savepoint defined in the `initialSavepointPath` or empty state if the path is null. 
--- a/docs/content/docs/custom-resource/reference.md
+++ b/docs/content/docs/custom-resource/reference.md
@@ -121,11 +121,11 @@ This page serves as a full reference for FlinkDeployment custom resource definit
 | args | java.lang.String[] | Arguments for the Flink job main class. |
 | state | org.apache.flink.kubernetes.operator.api.spec.JobState | Desired state for the job. |
 | savepointTriggerNonce | java.lang.Long | Nonce used to manually trigger savepoint for the running job. In order to trigger a savepoint, change the number to a different non-null value. |
-| initialSavepointPath | java.lang.String | Savepoint path used by the job the first time it is deployed. Upgrades/redeployments will not be affected. |
+| initialSavepointPath | java.lang.String | Savepoint path used by the job the first time it is deployed or during savepoint redeployments (triggered by changing the savepointRedeployNonce). |
 | checkpointTriggerNonce | java.lang.Long | Nonce used to manually trigger checkpoint for the running job. In order to trigger a checkpoint, change the number to a different non-null value. |
 | upgradeMode | org.apache.flink.kubernetes.operator.api.spec.UpgradeMode | Upgrade mode of the Flink job. |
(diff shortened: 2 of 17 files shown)


docs/content/docs/custom-resource/reference.md
```diff
@@ -125,6 +125,7 @@ This page serves as a full reference for FlinkDeployment custom resource definit
 | checkpointTriggerNonce | java.lang.Long | Nonce used to manually trigger checkpoint for the running job. In order to trigger a checkpoint, change the number to anything other than the current value. |
 | upgradeMode | org.apache.flink.kubernetes.operator.api.spec.UpgradeMode | Upgrade mode of the Flink job. |
 | allowNonRestoredState | java.lang.Boolean | Allow checkpoint state that cannot be mapped to any job vertex in tasks. |
+| savepointRedeployNonce | java.lang.Long | Nonce used to trigger a full redeployment of the job from the savepoint path specified in initialSavepointPath or from empty state if initialSavepointPath is null. Rollback is not possible after redeployment. |
```
1996fanrui, line 128: Should we mention `In order to trigger a XXX, change the number to anything other than the current value.`?

Other `Nonce` parameters have already mention it, or we can mention it before all parameters part(In a common place, and specific parameter doesn't need to do it.).

gyfora: good point, will update this.

docs/content/docs/custom-resource/job-management.md
```diff
@@ -288,12 +288,33 @@ Rollback is currently only supported for `FlinkDeployments`.
...
+Users have two options to restore a job from a target savepoint / checkpoint
+
+### Redeploy using the savepointRedeployNonce
+
+It is possible to redeploy a `FlinkDeployment` or `FlinkSessionJob` resource from a target savepoint by using the combination of `savepointRedeployNonce` and `initialSavepointPath` in the job spec:
+
+```yaml
+ job:
+   initialSavepointPath: file://redeploy-target-savepoint
+   savepointRedeployNonce: null -> 1
```
mxm, line 305: I think this is irritating. How about adding a comment? 
```suggestion
   # If not set previously, set to 1, otherwise increment, e.g. 2
   savepointRedeployNonce: 1
```

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/diff/DiffType.java
```diff
@@ -28,7 +28,9 @@ public enum DiffType {
     /** Scalable spec change. */
     SCALE,
     /** Upgradable spec change. */
-    UPGRADE;
+    UPGRADE,
+    /** Full redeploy from new state. */
+    SAVEPOINT_REDEPLOY;
```
mxm, line 33: Why do we need a new state? Isn't that also an `UPGRADE`? I suppose we could also special case inside the upgrade path when the nonce has been set. We also use `UPGRADE` for normal restarts.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractJobReconciler.java
```diff
@@ -306,6 +319,40 @@ protected void resubmitJob(FlinkResourceContext<CR> ctx, boolean requireHaMetada
...
+    private void redeployWithSavepoint(
+            FlinkResourceContext<CR> ctx,
+            Configuration deployConfig,
+            CR resource,
+            STATUS status,
+            SPEC currentDeploySpec,
+            JobState desiredJobState)
+            throws Exception {
+        LOG.info("Redeploying from savepoint");
+        cancelJob(ctx, UpgradeMode.STATELESS);
```
mxm, line 331: So we ignore the upgrade policy here and discard the state? What if the new savepoint path is corrupted or doesn't exist?

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/diff/DiffType.java
gyfora: This allows us to distinguish this more cleanly in a single place instead of having to add some new logic to compare with the lastReconciled spec etc.

This is not really an upgrade in how it was understood before. It doesn't use / preserve state and also does not respect he upgradeMode. Special casing under UPGRADE would probably be more confusing to users and make it harder to test implement.

docs/content/docs/custom-resource/job-management.md
gyfora: We use the `a -> b` notation in logs where we display spec changes. I think it's concise and intuitive. Definitely would not call it "irritating" :)

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/diff/DiffType.java
gyfora: Also this enum is not part of the CRD/status anywhere so it's basically an internal implementation detail to make the logic nice and clean

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractJobReconciler.java
gyfora: Users can always take a manual savepoint / set the state to suspended before performing a redeployment. But again redeploy is quite specific for cases when you actually have some unrecoverable error with your job during / after upgrade.

docs/content/docs/custom-resource/job-management.md
mxm: If I were a new user, I would copy this as-is. I'm suspecting it will fail. My suggestion would work fine.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractJobReconciler.java
```diff
@@ -306,6 +319,40 @@ protected void resubmitJob(FlinkResourceContext<CR> ctx, boolean requireHaMetada
...
+        var savepointOpt =
+                Optional.ofNullable(currentDeploySpec.getJob().getInitialSavepointPath());
+        currentDeploySpec
+                .getJob()
+                .setUpgradeMode(
+                        savepointOpt.isPresent() ? UpgradeMode.SAVEPOINT : UpgradeMode.STATELESS);
```
mxm, line 337: I feel like we should test the savepoint path is present that before we tear down the job. If it is not there, we can't proceed.

docs/content/docs/custom-resource/job-management.md
gyfora: alright, I will update it to avoid unexpected copy-paste behaviour

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractJobReconciler.java
gyfora: if the `initialSavepointPath` is not present the job will be restarted from empty state. At least that's how I implemented and documented it. We could discuss whether this should be allowed or not. The stateless upgradeMode serves this purpose to some extent but then the user has to change the upgradeMode back and forth

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractJobReconciler.java
gyfora: I will add some extra docs

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/diff/DiffResult.java
```diff
@@ -101,14 +101,7 @@ public String toString() {
...
     private static DiffType getSpechChangeType(List<Diff<?>> diffs) {
-        var type = DiffType.IGNORE;
-        for (var diff : diffs) {
-            type = DiffType.max(type, diff.getType());
-            if (type == DiffType.UPGRADE) {
-                return type;
-            }
-        }
-        return type;
+        return diffs.stream().map(Diff::getType).reduce(DiffType::max).orElse(DiffType.IGNORE);
```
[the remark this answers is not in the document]
gyfora: Yes the logic didn't really change here. This is a common way to represent an ordinal structure between the upgrade modes.

REDEPLOY > UPGRADE > SCALE > IGNORE , basically the higher one completely subsumes the other for instance a SCALE change can be executed as a redeploy/upgrade but not vice versa. I will add a note to the enum that makes sense

docs/content/docs/custom-resource/job-management.md
mxm: Please only update if you agree, not intending to block the PR on this.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractJobReconciler.java
mxm: As I said, I would not expect the savepoint loading to go through if the savepoint path is set.

mxm: Recommending to use this feature with an empty savepointpath to clear the state, without having to change the upgradeMode to stateless, seems beyond the scope of this change. I don't think we should allow this.

gyfora: makes sense, we can throw a validation error in that case

mxm (approved):
LGTM! I would prefer to not allow empty values for `initialSavepointPath` when the `savepointRedeployNonce` is used.

gyfora:
@mxm I have updated the PR based on the discussion, please take a look when you get a chance.

mxm (approved):

PR #725, opened 2023-12-07
[FLINK-33770] Promote deprecated autoscaler keys to fallback keys

mxm:
We moved all autoscaler configuration from `kubernetes.operator.job.autoscaler.` to `job.autoscaler.`

With the latest release, the logs are full with logs like this:

```
level:  WARN
logger:  org.apache.flink.configuration.Configuration
message:  Config uses deprecated configuration key 'kubernetes.operator.job.autoscaler.target.utilization' instead of proper key 'job.autoscaler.target.utilization'
```

The reason is that the configuration is loaded for every reconciliation.

This configuration is already widely adopted across hundreds of pipelines. I propose to remove the deprecation from the config keys and make them "fallback" keys instead which removes the deprecation warning.


gyfora (approved):
looks good

1996fanrui (approved):
Thanks for the improvemnt, LGTM.

PR #727, opened 2023-12-08
[FLINK-33454] [Operator] Add IngressTls to support TLS within the managed Ingress, also add Label Passthrough

ryanvanhuuksloot:
The managed ingress `IngressSpec` doesn't support TLS. It means that the ingress can't be used with secrets to secure the hosts. We can extend the `IngressSpec` to have `tls: List<IngressTlsSpec>`. In theory we shouldn't need a List spec here since the template only supports one host but I thought it was better to follow the Kubernetes spec since the change is pretty simple.

The managed ingress also doesn't have the ability to pass labels through to the Ingress. It would be nice to support this. It could be in a separate PR if we'd like but is very trivial.

Open to `IngressTLSSpec` if folks think that is better. I didn't like the double capital S beside each other.

## Brief change log
Added an `IngressTlsSpec` to support tls entries in the `IngressSpec`
Added a labels entry to the `IngressSpec` to support label passthrough to the Ingress similar to annotations.

## Verifying this change
- I did not add tests for the labels as we don't have explicit tests for the annotations. I wasn't sure if this was intentional. I'd be happy to add tests for the labels.

- Added tests for IngressTls from the IngressUtils including
1. Empty tls
2. empty hosts with single secretName, single tls entry
3. empty secretName with single hosts, single tls entry
4. single hosts and single secretName, single tls entry
5. multiple hosts and single secretName, single tls entry
6. multiple tls entries

I still need to manually validate the ingress spec with a job.

## Documentation

### DIFF
--- a/docs/content/docs/custom-resource/reference.md
+++ b/docs/content/docs/custom-resource/reference.md
@@ -96,6 +96,8 @@ This page serves as a full reference for FlinkDeployment custom resource definit
 | template | java.lang.String | Ingress template for the JobManager service. |
 | className | java.lang.String | Ingress className for the Flink deployment. |
 | annotations | java.util.Map<java.lang.String,java.lang.String> | Ingress annotations. |
+| labels | java.util.Map<java.lang.String,java.lang.String> | Ingress labels. |
+| tls | java.util.List<io.fabric8.kubernetes.api.model.networking.v1.IngressTLS> | Ingress tls. |
 
 ### JobManagerSpec
 **Class**: org.apache.flink.kubernetes.operator.api.spec.JobManagerSpec
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/IngressUtils.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/IngressUtils.java
@@ -33,6 +33,7 @@ import io.fabric8.kubernetes.api.model.networking.v1.HTTPIngressRuleValueBuilder
 import io.fabric8.kubernetes.api.model.networking.v1.IngressBuilder;
 import io.fabric8.kubernetes.api.model.networking.v1.IngressRule;
 import io.fabric8.kubernetes.api.model.networking.v1.IngressRuleBuilder;
+import io.fabric8.kubernetes.api.model.networking.v1beta1.IngressTLS;
 import io.fabric8.kubernetes.client.KubernetesClient;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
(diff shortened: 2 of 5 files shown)


flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/spec/IngressTlsSpec.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.List;
+
+/** Ingress spec. */
+@Experimental
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+@Builder
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class IngressTlsSpec {
```
gyfora, line 37: Why don't we use ` io.fabric8.kubernetes.api.model.networking.v1.IngressTLS` here directly? We can still convert back to the beta

docs/content/docs/custom-resource/reference.md
```diff
@@ -96,6 +96,8 @@ This page serves as a full reference for FlinkDeployment custom resource definit
 | template | java.lang.String | Ingress template for the JobManager service. |
 | className | java.lang.String | Ingress className for the Flink deployment. |
 | annotations | java.util.Map<java.lang.String,java.lang.String> | Ingress annotations. |
+| labels | java.util.Map<java.lang.String,java.lang.String> | Ingress labels. |
+| tls | java.util.List<io.fabric8.kubernetes.api.model.networking.v1.IngressTLS> | Ingress tls. |
```
ryanvanhuuksloot, line 100: Like you said in your review - I think it is fine to only show v1 here.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/IngressUtils.java
```diff
@@ -93,24 +96,45 @@ private static HasMetadata getIngress(
...
                     .withNamespace(objectMeta.getNamespace())
                     .endMetadata()
                     .withNewSpec()
                     .withIngressClassName(spec.getIngress().getClassName())
+                    .withTls(spec.getIngress().getTls())
                     .withRules(getIngressRule(objectMeta, spec, effectiveConfig))
                     .endSpec()
                     .build();
         } else {
+            List<IngressTLS> ingressTLS =
```
ryanvanhuuksloot, line 111: Pulled it outside of the return because otherwise it is even more impossible to read.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/spec/IngressTlsSpec.java
ryanvanhuuksloot: No particular reason - didn't think about it. Swapped to the kubernetes spec.

gyfora (approved):

PR #728, opened 2023-12-11
[FLINK-33795] Add a new config to forbid autoscale execution in the configured excluded periods

flashJd:

Hi, I'm new here, I use flink operator's autoscaling function in my project, this PR add a new config to forbid autoscale execution in the configured excluded periods so that we keep flink job unaffected by autoscaling's job restart behavior during this periods

## Brief change log

  - *A new config added*

## Verifying this change

  - *New tests `testForbidPeriods` and `testMetricCollectionDuringForbidden` added*

## Documentation

### DIFF
--- a/flink-autoscaler/pom.xml
+++ b/flink-autoscaler/pom.xml
@@ -69,7 +69,7 @@ under the License.
         <dependency>
             <groupId>org.quartz-scheduler</groupId>
             <artifactId>quartz</artifactId>
-            <version>2.3.2</version>
+            <version>${quartz.version}</version>
         </dependency>
 
         <dependency>
--- a/flink-autoscaler/pom.xml
+++ b/flink-autoscaler/pom.xml
@@ -67,6 +67,12 @@ under the License.
         </dependency>
 
         <dependency>
+            <groupId>org.quartz-scheduler</groupId>
+            <artifactId>quartz</artifactId>
+            <version>2.3.2</version>
+        </dependency>
+
+        <dependency>
             <groupId>org.projectlombok</groupId>
             <artifactId>lombok</artifactId>
             <version>${lombok.version}</version>
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
@@ -24,6 +24,8 @@ import org.apache.flink.autoscaler.metrics.EvaluatedMetrics;
 import org.apache.flink.autoscaler.metrics.EvaluatedScalingMetric;
 import org.apache.flink.autoscaler.metrics.ScalingMetric;
 import org.apache.flink.autoscaler.state.AutoScalerStateStore;
+import org.apache.flink.autoscaler.utils.CalendarUtils;
+import org.apache.flink.configuration.Configuration;
 import org.apache.flink.runtime.jobgraph.JobVertexID;
 
 import org.slf4j.Logger;
@@ -101,11 +103,7 @@ public class ScalingExecutor<KEY, Context extends JobAutoScalerContext<KEY>> {
 
         updateRecommendedParallelism(evaluatedMetrics.getVertexMetrics(), scalingSummaries);
 
-        var scaleEnabled = conf.get(SCALING_ENABLED);
-        autoScalerEventHandler.handleScalingEvent(
-                context, scalingSummaries, scaleEnabled, conf.get(SCALING_EVENT_INTERVAL));
-
-        if (!scaleEnabled) {
+        if (blockScalingExecution(context, scalingSummaries, conf, now)) {
             return false;
         }
 
@@ -263,5 +261,25 @@ public class ScalingExecutor<KEY, Context extends JobAutoScalerContext<KEY>> {
                     }
                 });
         return overrides;
+    }
+
+    private boolean blockScalingExecution(
+            Context context,
+            Map<JobVertexID, ScalingSummary> scalingSummaries,
+            Configuration conf,
+            Instant now) {
+        var scaleEnabled = conf.get(SCALING_ENABLED);
+        var isExcluded = CalendarUtils.inExcludedPeriods(conf, now);
+        autoScalerEventHandler.handleScalingEvent(
+                context,
+                scalingSummaries,
+                scaleEnabled,
+                isExcluded,
+                conf.get(SCALING_EVENT_INTERVAL));
+
+        if (!scaleEnabled || isExcluded) {
+            return true;
+        }
+        return false;
     }
 }
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
@@ -38,8 +38,12 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.SortedMap;
 
+import static org.apache.flink.autoscaler.config.AutoScalerOptions.EXCLUDED_PERIODS;
 import static org.apache.flink.autoscaler.config.AutoScalerOptions.SCALING_ENABLED;
 import static org.apache.flink.autoscaler.config.AutoScalerOptions.SCALING_EVENT_INTERVAL;
+import static org.apache.flink.autoscaler.event.AutoScalerEventHandler.SCALING_EXECUTION_DISABLED_REASON;
+import static org.apache.flink.autoscaler.event.AutoScalerEventHandler.SCALING_SUMMARY_HEADER_SCALING_EXECUTION_DISABLED;
+import static org.apache.flink.autoscaler.event.AutoScalerEventHandler.SCALING_SUMMARY_HEADER_SCALING_EXECUTION_ENABLED;
 import static org.apache.flink.autoscaler.metrics.ScalingHistoryUtils.addToScalingHistoryAndStore;
 import static org.apache.flink.autoscaler.metrics.ScalingMetric.SCALE_DOWN_RATE_THRESHOLD;
 import static org.apache.flink.autoscaler.metrics.ScalingMetric.SCALE_UP_RATE_THRESHOLD;
@@ -103,7 +107,7 @@ public class ScalingExecutor<KEY, Context extends JobAutoScalerContext<KEY>> {
 
         updateRecommendedParallelism(evaluatedMetrics.getVertexMetrics(), scalingSummaries);
 
-        if (blockScalingExecution(context, scalingSummaries, conf, now)) {
+        if (checkIfBlockedAndTriggerScalingEvent(context, scalingSummaries, conf, now)) {
             return false;
         }
 
@@ -263,20 +267,34 @@ public class ScalingExecutor<KEY, Context extends JobAutoScalerContext<KEY>> {
         return overrides;
     }
 
-    private boolean blockScalingExecution(
+    private boolean checkIfBlockedAndTriggerScalingEvent(
             Context context,
             Map<JobVertexID, ScalingSummary> scalingSummaries,
             Configuration conf,
             Instant now) {
         var scaleEnabled = conf.get(SCALING_ENABLED);
         var isExcluded = CalendarUtils.inExcludedPeriods(conf, now);
+        String message;
+        if (!scaleEnabled) {
+            message =
+                    SCALING_SUMMARY_HEADER_SCALING_EXECUTION_DISABLED
+                            + String.format(
+                                    SCALING_EXECUTION_DISABLED_REASON,
+                                    SCALING_ENABLED.key(),
+                                    false);
+        } else if (isExcluded) {
+            message =
+                    SCALING_SUMMARY_HEADER_SCALING_EXECUTION_DISABLED
+                            + String.format(
+                                    SCALING_EXECUTION_DISABLED_REASON,
+                                    EXCLUDED_PERIODS.key(),
+                                    conf.get(EXCLUDED_PERIODS));
+        } else {
+            message = SCALING_SUMMARY_HEADER_SCALING_EXECUTION_ENABLED;
+        }
         autoScalerEventHandler.handleScalingEvent(
-                context,
-                scalingSummaries,
-                scaleEnabled,
-                isExcluded,
-                conf.get(SCALING_EVENT_INTERVAL));
+                context, scalingSummaries, message, conf.get(SCALING_EVENT_INTERVAL));
 
-      return !scaleEnabled || isExcluded;
+        return !scaleEnabled || isExcluded;
     }
 }
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
@@ -277,9 +277,6 @@ public class ScalingExecutor<KEY, Context extends JobAutoScalerContext<KEY>> {
                 isExcluded,
                 conf.get(SCALING_EVENT_INTERVAL));
 
-        if (!scaleEnabled || isExcluded) {
-            return true;
-        }
-        return false;
+      return !scaleEnabled || isExcluded;
     }
 }
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java
@@ -150,7 +150,9 @@ public abstract class ScalingMetricCollector<KEY, Context extends JobAutoScalerC
             LOG.info("Metric window not full until {}", readable(windowFullTime));
         } else {
             if (isExcluded) {
-                LOG.info("autoscaling now in excluded period");
+                LOG.info(
+                        "Autoscaling on halt based on exclusion rule {}",
+                        conf.get(AutoScalerOptions.EXCLUDED_PERIODS));
             } else {
                 collectedMetrics.setFullyCollected(true);
             }
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java
@@ -29,6 +29,7 @@ import org.apache.flink.autoscaler.metrics.ScalingMetric;
 import org.apache.flink.autoscaler.metrics.ScalingMetrics;
 import org.apache.flink.autoscaler.state.AutoScalerStateStore;
 import org.apache.flink.autoscaler.topology.JobTopology;
+import org.apache.flink.autoscaler.utils.AutoScalerUtils;
 import org.apache.flink.client.program.rest.RestClusterClient;
 import org.apache.flink.configuration.Configuration;
 import org.apache.flink.runtime.execution.ExecutionState;
@@ -115,6 +116,7 @@ public abstract class ScalingMetricCollector<KEY, Context extends JobAutoScalerC
         var topology = getJobTopology(ctx, stateStore, jobDetailsInfo);
         var stableTime = jobUpdateTs.plus(conf.get(AutoScalerOptions.STABILIZATION_INTERVAL));
         final boolean isStabilizing = now.isBefore(stableTime);
+        final boolean isExcluded = AutoScalerUtils.inExcludedPeriods(conf, now);
 
         // Calculate timestamp when the metric windows is full
         var metricWindowSize = getMetricWindowSize(conf);
@@ -147,7 +149,11 @@ public abstract class ScalingMetricCollector<KEY, Context extends JobAutoScalerC
         if (now.isBefore(windowFullTime)) {
             LOG.info("Metric window not full until {}", readable(windowFullTime));
         } else {
-            collectedMetrics.setFullyCollected(true);
+            if (isExcluded) {
+                LOG.info("autoscaling now in excluded period");
+            } else {
+                collectedMetrics.setFullyCollected(true);
+            }
             // Trim metrics outside the metric window from metrics history
             metricHistory.headMap(now.minus(metricWindowSize)).clear();
         }
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java
@@ -29,7 +29,7 @@ import org.apache.flink.autoscaler.metrics.ScalingMetric;
 import org.apache.flink.autoscaler.metrics.ScalingMetrics;
 import org.apache.flink.autoscaler.state.AutoScalerStateStore;
 import org.apache.flink.autoscaler.topology.JobTopology;
-import org.apache.flink.autoscaler.utils.AutoScalerUtils;
+import org.apache.flink.autoscaler.utils.CalendarUtils;
 import org.apache.flink.client.program.rest.RestClusterClient;
 import org.apache.flink.configuration.Configuration;
 import org.apache.flink.runtime.execution.ExecutionState;
@@ -116,7 +116,7 @@ public abstract class ScalingMetricCollector<KEY, Context extends JobAutoScalerC
         var topology = getJobTopology(ctx, stateStore, jobDetailsInfo);
         var stableTime = jobUpdateTs.plus(conf.get(AutoScalerOptions.STABILIZATION_INTERVAL));
         final boolean isStabilizing = now.isBefore(stableTime);
-        final boolean isExcluded = AutoScalerUtils.inExcludedPeriods(conf, now);
+        final boolean isExcluded = CalendarUtils.inExcludedPeriods(conf, now);
 
         // Calculate timestamp when the metric windows is full
         var metricWindowSize = getMetricWindowSize(conf);
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java
@@ -29,7 +29,6 @@ import org.apache.flink.autoscaler.metrics.ScalingMetric;
 import org.apache.flink.autoscaler.metrics.ScalingMetrics;
 import org.apache.flink.autoscaler.state.AutoScalerStateStore;
 import org.apache.flink.autoscaler.topology.JobTopology;
-import org.apache.flink.autoscaler.utils.CalendarUtils;
 import org.apache.flink.client.program.rest.RestClusterClient;
 import org.apache.flink.configuration.Configuration;
 import org.apache.flink.runtime.execution.ExecutionState;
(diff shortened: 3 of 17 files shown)


flink-autoscaler/src/test/java/org/apache/flink/autoscaler/MetricsCollectionAndEvaluationTest.java
```diff
@@ -599,6 +602,73 @@ public void testMetricCollectionDuringStabilization() throws Exception {
         assertEquals(2, stateStore.getCollectedMetrics(context).size());
     }
 
+    @Test
+    public void testMetricCollectionDuringForbidden() throws Exception {
```
gyfora, line 606: I think this test should be part of the `JobAutoscalerImplTest` and then it can be simplified a lot.

flink-autoscaler/src/test/java/org/apache/flink/autoscaler/utils/AutoScalerUtilsTest.java
```diff
@@ -56,4 +60,30 @@ public void testVertexExclusion() {
                 Set.of(v1.toString(), v2.toString(), v3.toString()),
                 new HashSet<>(conf.get(AutoScalerOptions.VERTEX_EXCLUDE_IDS)));
     }
+
+    @Test
+    public void testForbidPeriods() {
```
gyfora, line 65: Should be `Forbidden`

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
```diff
@@ -96,6 +97,13 @@ public void scale(Context ctx) throws Exception {
                 return;
             }
 
+            if (!AutoScalerUtils.verifyForbidPeriods(ctx.getConfiguration())) {
```
gyfora, line 100: Config validation should be in the validator module, not here.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java
```diff
@@ -114,6 +115,7 @@ public CollectedMetricHistory updateMetrics(
         var topology = getJobTopology(ctx, stateStore, jobDetailsInfo);
         var stableTime = jobUpdateTs.plus(conf.get(AutoScalerOptions.STABILIZATION_INTERVAL));
         final boolean isStabilizing = now.isBefore(stableTime);
+        final boolean isForbidding = AutoScalerUtils.inForbidPeriod(conf, now);
```
gyfora, line 118: Everywhere the word `Forbid` should be replaced with `Forbidden` or `Blocked` to be correct.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
```diff
@@ -72,6 +72,15 @@ private static ConfigOptions.OptionBuilder autoScalerConfig(String key) {
...
                             "Stabilization period in which no new scaling will be executed");
 
+    public static final ConfigOption<List<String>> FORBID_PERIOD =
+            autoScalerConfig("forbid.periods")
+                    .stringType()
+                    .asList()
+                    .defaultValues()
+                    .withDeprecatedKeys(deprecatedOperatorConfigKey("forbid.periods"))
+                    .withDescription(
+                            "A (semicolon-separated) list of certain times of the day during which autoscaling is forbidden, 10:00:00-11:00:00;21:30:00-22:30:00 for example");
```
gyfora, line 82: Where did you get this syntax from? What are the potential alternatives?

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
```diff
@@ -72,6 +72,15 @@ private static ConfigOptions.OptionBuilder autoScalerConfig(String key) {
                     .withDescription(
                             "Stabilization period in which no new scaling will be executed");
 
+    public static final ConfigOption<List<String>> FORBID_PERIOD =
+            autoScalerConfig("forbid.periods")
+                    .stringType()
+                    .asList()
+                    .defaultValues()
+                    .withDeprecatedKeys(deprecatedOperatorConfigKey("forbid.periods"))
```
gyfora, line 80: deprecated key should not be added

gyfora (changes requested):
I think we should have a small discussion somewhere (here or the ML) about this feature. Why did you select this specific config syntax? Any alternatives?

cc @mxm

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
mxm: This was also my first thought seeing this. Ideally, we would use some available standard for time spans. I think this also has to include the time zone.

What about different times at different days of the week / month?

mxm (commented):
Thanks @flashJd! This is a useful feature. Let's discuss on the mailing list on the syntax for the exclude periods, as @gyfora suggested.

flashJd:
> I think we should have a small discussion somewhere (here or the ML) about this feature. Why did you select this specific config syntax? Any alternatives?
> 
> cc @mxm

flashJd:
> I think we should have a small discussion somewhere (here or the ML) about this feature. Why did you select this specific config syntax? Any alternatives?
> 
> cc @mxm



> Thanks @flashJd! This is a useful feature. Let's discuss on the mailing list on the syntax for the exclude periods, as @gyfora suggested.

Thanks for the review, the specific config syntax is just a simple implementation for config forbidden periods in a day, it sounds reasonable  `we would use some available standard for time spans` and `What about different times at different days of the week / month?`
As I just subscribed to the flink dev mailing list, I can't receive the jira email associated with this pr which is sent yesterday and reply to it

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
flashJd: I've seached several time scheduler framework, `Quartz Scheduler` seems to be feasible.
1. It supports `certain time of day` `certain days of the week` `certain days of the month` , etc.
https://github.com/quartz-scheduler/quartz/blob/main/docs/introduction.adoc#job-scheduling
2. Here is the implementation: 
https://github.com/quartz-scheduler/quartz/blob/a5c4d27e963f51097f9b2777489d310a88897ca4/quartz/src/main/java/org/quartz/Calendar.java#L21-L41
https://github.com/quartz-scheduler/quartz/blob/a5c4d27e963f51097f9b2777489d310a88897ca4/quartz/src/main/java/org/quartz/impl/calendar/CronCalendar.java#L10-L26

mxm: Sounds good, we already use Quartz for `periodic.checkpoint.interval`.

flashJd: I checked the pr https://github.com/apache/flink-kubernetes-operator/pull/656 which introduced `class CronExpression` from log4j-core and log4j-core copied from Quartz.

1.  So can we set the `excluded periods` config to be a cron expression string and use it's `isSatisfiedBy(final Date date)` method to judge whether in excluded periods
2.  Since standardized cron string doesn't contain a timeZone, should we just use the default timeZone or some other way to pass it in?

mxm: Yes, if the cron-based notation allows time ranges that would be perfect. We could just use UTC and document this for now.

flink-autoscaler/src/test/java/org/apache/flink/autoscaler/utils/AutoScalerUtilsTest.java
flashJd: renamed to `testExcludedPeriods()`

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
flashJd: removed to `DefaultValidator`

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java
flashJd: fixed

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
flashJd: deleted

flink-autoscaler/src/test/java/org/apache/flink/autoscaler/MetricsCollectionAndEvaluationTest.java
flashJd: I want to test this config along with config STABILIZATION_INTERVAL and METRICS_WINDOW, so put it here

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
flashJd: Quartz's  DailyCalendar along with CronCalendar allows  `daily excluded, days of the week/month excluded, etc.`
I've updated the config and implemented it, looking forward for the review.

flink-autoscaler/pom.xml
```diff
@@ -66,6 +66,12 @@ under the License.
             <artifactId>jackson-databind</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>org.quartz-scheduler</groupId>
+            <artifactId>quartz</artifactId>
+            <version>2.3.2</version>
```
1996fanrui, line 72: Would you mind extract a property for this version? Such as : `quartz.version`.

flashJd: sure, already extracted

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/utils/AutoScalerUtils.java
```diff
@@ -94,4 +99,108 @@ public static boolean excludeVerticesFromScaling(
         conf.set(AutoScalerOptions.VERTEX_EXCLUDE_IDS, new ArrayList<>(excludedIds));
         return anyAdded;
     }
+
+    /** Quartz doesn't have the invertTimeRange flag so rewrite this method. */
+    static boolean isTimeIncluded(CronCalendar cron, long timeInMillis) {
```
mxm, line 104: Can we move these methods to a new class, e.g. `DateUtils` or `CronUtils`?

mxm (approved):
LGTM. Just a couple minor comments.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/utils/AutoScalerUtils.java
flashJd: Removed to `CalendarUtils` and corresponding tests removed to `CalendarUtilsTest`

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/utils/AutoScalerUtils.java
```diff
@@ -94,4 +99,106 @@ public static boolean excludeVerticesFromScaling(
...
+        } catch (Exception e) {
+            return Optional.empty();
+        }
+    }
+
+    static Optional<CronCalendar> interpretAsCron(String subExpression) {
+        try {
+            return Optional.of(new CronCalendar(subExpression));
+        } catch (Exception e) {
+            return Optional.empty();
```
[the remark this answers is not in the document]
flashJd: Incorrect expression config will first be validated in `DefaultValidator` and report exception there, I've added extra tests in `DefaultValidatorTest`

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java
```diff
@@ -143,7 +145,11 @@ public CollectedMetricHistory updateMetrics(
         if (now.isBefore(windowFullTime)) {
             LOG.info("Metric window not full until {}", windowFullTime);
         } else {
-            collectedMetrics.setFullyCollected(true);
+            if (isExcluded) {
+                LOG.info("autoscaling now in excluded period");
```
mxm, line 149:
```suggestion
                LOG.info("Autoscaling on halt based on exclusion rule {}",
                    conf.get(AutoScalerOptions. EXCLUDED_PERIODS));
```

I think this could be helpful for debugging.

mxm:
Thanks @flashJd! Changes look great.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java
flashJd: Agreed. Adjustment has been made

flashJd:
@mxm can you approval the workflow again, I've fixed the ci failure, thanks.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java
```diff
@@ -147,7 +149,13 @@ public CollectedMetricHistory updateMetrics(
...
             LOG.info("Metric window not full until {}", readable(windowFullTime));
         } else {
-            collectedMetrics.setFullyCollected(true);
+            if (isExcluded) {
+                LOG.info(
+                        "Autoscaling on halt based on exclusion rule {}",
+                        conf.get(AutoScalerOptions.EXCLUDED_PERIODS));
+            } else {
+                collectedMetrics.setFullyCollected(true);
+            }
```
gyfora, line 158: Enforcing this through collectedMetrics.setFullyCollected(true); seems a bit hacky and susceptible to breaking if we treat the collected metrics differently.

Maybe we should move this check to the `ScalingExecutor`

gyfora: what do you think @mxm ?

mxm: Although the current logic works fine for blocking the scaling during the configured phases, I agree that the logic would be better suited for ScalingExecutor. There are other features, e.g. recommended parallelism, which will stop working if the metric collection phase hasn't completed. We only want to block executing the scaling. All other logic should run normally.

flashJd: @gyfora @mxm you are right, it's better to place the excludedPeriods blocking logic in ScalingExecutor. As the meaning of excludedPeriods is very similar with config `scaling.enabled`, i combines the logic in `ScalingExecutor`. Now excludedPeriods blocking works well with recommended parallelism

flashJd:
@gyfora @mxm by the way, I'd like to discuss the logic of scaling effectiveness evaluation with you. 

1. Now it's controlled by two config `scaling.effectiveness.detection.enabled` and `scaling.effectiveness.threshold` and we evaluate the effectiveness under the condition `last scaling is scale up` and only refer to the last scaling effectiveness. 
2. Image the following scenario: scale up double parallism first, then scale down to 0.8 parallism, then scale up double scale down 0.8, the effectiveness detection will be invalid in this scenario, even scale up is ineffecive, we'll continue scale up
3. Maybe we can add a new config like `scaling.effectiveness.history.reference.num` and set a default value, then we can evaluate based on the last `scaling.effectiveness.history.reference.num` scale up summaries.

Looking forward to your reply.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
```diff
@@ -264,4 +262,24 @@ private static Map<String, String> getVertexParallelismOverrides(
...
+                context,
+                scalingSummaries,
+                scaleEnabled,
+                isExcluded,
+                conf.get(SCALING_EVENT_INTERVAL));
+
+        if (!scaleEnabled || isExcluded) {
+            return true;
+        }
+        return false;
```
mxm, line 283:
```suggestion
        return !scaleEnabled || isExcluded;

**Review by mxm — COMMENTED** [review]
LGTM. Any additional thoughts @gyfora?

**Comment by mxm** [comment]
> @gyfora @mxm by the way, I'd like to discuss the logic of scaling effectiveness evaluation with you.
> 
> 1. Now it's controlled by two config `scaling.effectiveness.detection.enabled` and `scaling.effectiveness.threshold` and we evaluate the effectiveness under the condition `last scaling is scale up` and only refer to the last scaling effectiveness.
> 2. Image the following scenario: scale up double parallism first, then scale down to 0.8 parallism, then scale up double scale down 0.8, the effectiveness detection will be invalid in this scenario, even scale up is ineffecive, we'll continue scale up
> 3. Maybe we can add a new config like `scaling.effectiveness.history.reference.num` and set a default value, then we can evaluate based on the last `scaling.effectiveness.history.reference.num` scale up summaries.
> 
> Looking forward to your reply.

What yo you describe is an edge case we hadn't considered. We were more concerned about a continuous increase in parallelism. If there is any scale down we are currently assuming that the algorithm is not completely broken.

File: flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java (hunk 12 above)
Comment (flashJd) [reply flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java]: fixed

File: flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java [hunk 13]
```diff
@@ -264,4 +262,21 @@ private static Map<String, String> getVertexParallelismOverrides(
...
+        var scaleEnabled = conf.get(SCALING_ENABLED);
+        var isExcluded = CalendarUtils.inExcludedPeriods(conf, now);
+        autoScalerEventHandler.handleScalingEvent(
+                context,
+                scalingSummaries,
+                scaleEnabled,
+                isExcluded,
+                conf.get(SCALING_EVENT_INTERVAL));
+
+      return !scaleEnabled || isExcluded;
```
gyfora, line 280: I don't completely understand why we had to move this into a method. It could simply be:
```
       var scaleEnabled = conf.get(SCALING_ENABLED) && !CalendarUtils.inExcludedPeriods(conf, now);
        autoScalerEventHandler.handleScalingEvent(
                context,
                scalingSummaries,
                scaleEnabled,
                conf.get(SCALING_EVENT_INTERVAL));
```
Then we don't need any changes to the autoscaler event handler, we could simply log a line here that we are in an excluded period instead of changing code in multiple places.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/event/AutoScalerEventHandler.java
```diff
@@ -83,12 +92,27 @@ default void handleScalingEvent(
...
+                            false));
+        } else if (isExcluded) {
+            sb.append(
+                    String.format(
+                            SCALING_SUMMARY_HEADER_SCALING_EXECUTION_DISABLED,
+                            EXCLUDED_PERIODS.key(),
+                            config.get(EXCLUDED_PERIODS)));
+        } else {
+            sb.append(SCALING_SUMMARY_HEADER_SCALING_EXECUTION_ENABLED);
+        }
```
gyfora, line 114: as I wrote in the other comment, I prefer not to make any changes in the event handler

gyfora (changes requested):

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
flashJd: I've considered just print a log but finally chose adjust autoscaler event handler a bit:

1. Print a log will cause many annoying duplicated logs if exludedPeriods set is relatively wide
2. Now that we have the interval based event trigger method, why not use it

@mxm what do you think

gyfora: I am not a big fan of scattering features across many interfaces without a very good reason. It makes the event handler and the whole logic more complex

gyfora: The event handler really has nothing to do with this feature (and many other things that happen in the scaling executor)

mxm: I was contemplating to comment on this as well, but after looking at the code, there is some benefit to hinting whether scaling is disabled or blocked. By treating blocked scaling as "disabled", we yield less information to the user. I think it will be beneficial to learn WHY the scaling is disabled, i.e. scaling disabled or scaling blocked.

>The event handler really has nothing to do with this feature (and many other things that happen in the scaling executor)

Not sure about that. Event handlers are supposed to process events and in the context of the autoscaler, a blocked scaling decision is a type of event we may present to the user. We are already leaking whether scaling is enabled or not to the event handler, why not also annotate why it is disabled?

gyfora: If scaling is blocked or disabled the event that we send out is a recommendation. Wouldn't it make sense to have uniform events for recommendation regardless of the cause? The user always knows why it's only a recommendation, either because the autoscaler is disabled completely or if in a blocking period

mxm: I don't feel strongly about including the cause. We can also just send out scaling events only for actual scaling operations. What @flashJd did is actually more in line with the current code which does include meta information like "scalingEnabled" in the event message. If we decide to skip those, for consistency, I would also remove the "scalingEnabled" flag and only ever emit scaling events when scaling is enabled and not blocked.

gyfora: In the current logic we send 2 types of events on a parallelism change:
 1. Scaling Recommendation (when disabled) -> `Recommended parallelism change: {parallelisms}`
 2. Scaling Event (when enabled) -> `Scaling Vertices: {parallelisms}`
 
The PR basically adds a 3rd type of event which looks like a recommendation but it doesn't say recommendation it prints:

`Scaling execution disabled by config: configkey:false {parallelism:map}`

I don't think that this is fully consistent with the current logic. We intended to keep the recommendations, but we changed the code so that it's not a recommendation anymore but looks different.

gyfora: Ah sorry I did not realise that you also changed the event message for the basic recommendations... my bad

gyfora: Sorry, I misunderstood the code, I am fine with this :)

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/event/AutoScalerEventHandler.java
gyfora: I think it would make the code much cleaner to simply submit a string to the event handler instead of the 2 booleans.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
```diff
@@ -264,4 +262,21 @@ private static Map<String, String> getVertexParallelismOverrides(
                 });
         return overrides;
     }
+
+    private boolean blockScalingExecution(
```
gyfora, line 266: We should rename this to `triggerScalingEvent` or `checkIfBlockedAndTriggerScalingEvent` or add javadocs or something that suggests that the scaling event is triggered inside

gyfora (changes requested):
Based on the discussion, I agree to keep the changed recommendation event message. 

But I would still prefer to add 2 minor changes (according to my comments):
 - Pass the recommendation message in a generic way from the executor so that it's more extensible in the future
 - Rename the ` blockScalingExecution` method or at least add javadocs

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
flashJd: Renamed to `checkIfBlockedAndTriggerScalingEvent`

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/event/AutoScalerEventHandler.java
flashJd: Modify the parameter of method `handleScalingEvent`  to replace the scalingEnabled and scalingBlocked flags via a generic String

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
flashJd: Thanks for the review. A small question, what do you think of the frequent log printing, `Stabilizing until ...` and `Metric window not full until ...`, is it worth to put some frequent log to autoscale event handler.

mxm: I think it is a good idea to send events for those **once** per stabilization / metric collection period. We have an event de-duplication feature to avoid sending them more often.

mxm: Let's do that in a separate PR.

gyfora (approved):

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
```diff
@@ -72,6 +72,21 @@ private static ConfigOptions.OptionBuilder autoScalerConfig(String key) {
                     .withDescription(
                             "Stabilization period in which no new scaling will be executed");
 
+    public static final ConfigOption<List<String>> EXCLUDED_PERIODS =
+            autoScalerConfig("excluded.periods")
+                    .stringType()
+                    .asList()
+                    .defaultValues()
```
1996fanrui, line 79:
```suggestion
                    .noDefaultValue()
```

flashJd: Fixed, can you approve the workflow again, thanks.

flashJd: @1996fanrui change the defaultValue to `noDefaultValue` causes NPE， we have to add many annoying
 notNull judgments, as config `vertex.exclude.ids` also set to defaultValues(), I referred to it, what do you think.

1996fanrui: It's fine for me.

1996fanrui (approved):

PR #729, opened 2023-12-11
[FLINK-33770] Migrate legacy autoscaler config keys

mxm:
The previous PR in #725 made the legacy autoscaler config keys "fallback" keys to prevent logging a deprecation WARN message on every reconciliation loop in the operator. Turns out, fallback keys also log a warning.

This change moves to migrating all config keys in the legacy "kubernetes.operator." before any autoscaler ConfigOptions are used. This ensures no warnings will be logged but the older keys can still be used.  The new keys always have precedence over the old keys.


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkResourceContext.java
```diff
@@ -65,7 +66,9 @@ public KubernetesJobAutoScalerContext getJobAutoScalerContext() {
     }
 
     private KubernetesJobAutoScalerContext createJobAutoScalerContext() {
-        Configuration conf = new Configuration(getObserveConfig());
+        Configuration conf =
+                new KeyMigratingConfiguration(
+                        AutoScalerOptions.LEGACY_CONF_PREFIX, getObserveConfig());
```
1996fanrui, line 71: How about traversing the `getObserveConfig()` and put all keys and values to a new Configuration? We can removing the prefix directly during traversing.

If so, we don't need to introduce the `KeyMigratingConfiguration` class.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
```diff
@@ -46,107 +45,93 @@ private static ConfigOptions.OptionBuilder autoScalerConfig(String key) {
             autoScalerConfig("enabled")
                     .booleanType()
                     .defaultValue(false)
-                    .withFallbackKeys(oldOperatorConfigKey("enabled"))
```
1996fanrui, line 49: Do we need to remove it? I'm not sure should we keep it to let old users know it.

1996fanrui (commented):
Thanks @mxm  for the improvements~

I left 2 minor comments.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkResourceContext.java
mxm: Semantically, there is no difference. The reason I chose the subclass is that I didn't want to touch any other configs apart from the deprecated options. `Configuration` doesn't allow to access the configuration map directly. One has to extract the configuration as string which will runs through a conversion process.

I just did a quick conversion of the code (see latest commit). I don't prefer this approach but let me know if you find the code more readable.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
mxm: The core idea of this fix is to remove deprecated/fallback keys because they generate WARN messages in the logs for every key and every reconciliation loop. The keys are deprecated in the sense that they are not documented anymore but we can't remove them, at least not until much later.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
```diff
@@ -247,21 +222,41 @@ private static ConfigOptions.OptionBuilder autoScalerConfig(String key) {
...
+            String migratedKey = key.substring(LEGACY_CONF_PREFIX.length());
+            boolean keyDoesNotExist = config.getString(migratedKey, null) == null;
+            if (keyDoesNotExist) {
+                String migratedValue = Preconditions.checkNotNull(config.getString(key, null));
+                config.setString(migratedKey, migratedValue);
+            }
+            config.removeKey(key);
+        }
+        return config;
+    }
```
1996fanrui, line 261:
```suggestion
    public static Configuration migrateOldConfigKeys(Configuration config) {
        Map<String, String> oldConfig = Preconditions.checkNotNull(config).toMap();
        Configuration newConfig = new Configuration();

        for (Map.Entry<String, String> entry : oldConfig.entrySet()) {
            String key = entry.getKey();
            String value = entry.getValue();
            if (!key.startsWith(LEGACY_CONF_PREFIX)) {
                newConfig.setString(key, value);
                continue;
            }
            String migratedKey = key.substring(LEGACY_CONF_PREFIX.length());
            if (newConfig.getString(migratedKey, null) == null) {
                newConfig.setString(migratedKey, value);
            }
        }

        return newConfig;
    }
```

How about this? Only one for loop.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
1996fanrui: After migration, the log already can be removed, right?

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
mxm: The O(n) complexity of both approaches is the same. I still prefer the first commit in this PR.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
mxm: Yes, after migration the fallback keys are not needed anymore.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
```diff
@@ -247,21 +222,41 @@ private static ConfigOptions.OptionBuilder autoScalerConfig(String key) {
...
+        Set<String> toBeMigrated = new HashSet<>();
+        for (String key : config.keySet()) {
+            if (key.startsWith(LEGACY_CONF_PREFIX)) {
+                toBeMigrated.add(key);
+            }
+        }
+        for (String key : toBeMigrated) {
+            String migratedKey = key.substring(LEGACY_CONF_PREFIX.length());
+            boolean keyDoesNotExist = config.getString(migratedKey, null) == null;
+            if (keyDoesNotExist) {
```
gyfora, line 254: Can be simplified to `if(!config.containsKey(migratedKey))`

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
```diff
@@ -247,21 +222,41 @@ private static ConfigOptions.OptionBuilder autoScalerConfig(String key) {
...
+                String migratedValue = Preconditions.checkNotNull(config.getString(key, null));
+                config.setString(migratedKey, migratedValue);
+            }
+            config.removeKey(key);
```
gyfora, line 258: We don't really need to remove the old key if it's not marked as deprecated/fallback

mxm: I believe we should remove the key for clarify since it will be superseded by the non-legacy key.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
mxm: There are a lot of copy steps with this approach because `toMap()` converts every single value before it inserts it into the new map. `new Configuration(config)` directly passes on the internal map. 

With that in mind, I tried incorporating your code suggestions, i.e. do a single pass.

mxm:
Folks, any more comments? Thanks for your suggestions regarding the implementation. I had a good three attempts and I think it looks good now. The source of quirkiness is really `Configuration` which has some odd design choices.

1996fanrui (approved):
Thanks @mxm for the update!

LGTM.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
```diff
@@ -247,21 +221,36 @@ private static ConfigOptions.OptionBuilder autoScalerConfig(String key) {
...
+    /** Migrate config keys still prefixed with the old Kubernetes operator prefix. */
+    public static Configuration migrateOldConfigKeys(Configuration config) {
+        Preconditions.checkNotNull(config);
+        Set<String> allKeys = config.keySet();
+        config = new Configuration(config);
+        for (String key : allKeys) {
+            if (key.startsWith(LEGACY_CONF_PREFIX)) {
+                String migratedKey = key.substring(LEGACY_CONF_PREFIX.length());
+                if (!config.containsKey(migratedKey)) {
+                    String migratedValue = config.getString(key, null);
```
1996fanrui, line 248: Just a mark for removing the `@Depreated` for `getString(String key, String defaultValue)`, please ignore it for code review.

gyfora (approved):
Looks good. As long as you have validated the change on a local cluster at least, it's good to go from my side.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
gyfora: Actually I am not sure whether we will get the warn logs if we don't delete it

mxm: I changed the PR to not delete them as you suggested. The reason is that while testing I found that operating on the Configuration API is dangerous (which is why the first version of this PR did intentionally operated directly on the internal Map). When you delete a key via `Configuration#removeKey` which is a prefix of another key, it will delete both.

We won't get warnings because the deprecated config keys have been removed in this PR.

mxm:
While testing, I noticed that the abstraction in the PR doesn't fully hold because there is code in the reconciler which access autoscaler options and uses the observe config instead of the migrated config, e.g. [here](https://github.com/apache/flink-kubernetes-operator/blob/2136c7a3189753c45e167920bd509a687218a57f/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java#L190). I could fix this but I fear that the abstraction might become leaky again in the future.

I'm closing this PR because I want to avoid adding too much complexity to the config key handling. 

The current approach to use fallback instead of deprecated keys is fine. Fallback keys are logged at INFO, so we can simply set the logging for `Configuration` to WARN. See: https://github.com/apache/flink-kubernetes-operator/pull/734

PR #730, opened 2023-12-13
[FLINK-33812] Autoscale not working properly for standalone application deployment

jrthe42:

This pull request tries to fix a bug, that when the autoscaler decides to scale up/down a standalone application deployment, the taskmanager's replica number does not change accordingly.

## Brief change log
  - When calculating the number of tm replicas, give priority to `PipelineOptions.PARALLELISM_OVERRIDES`

## Verifying this change
Unit tests + manual validation

  - The public API, i.e., is any changes to the `CustomResourceDescriptors`:no)

## Documentation

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
@@ -68,6 +68,7 @@ import java.nio.file.Files;
 import java.time.Duration;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.Optional;
 
 import static org.apache.flink.configuration.DeploymentOptions.SHUTDOWN_ON_APPLICATION_FINISH;
 import static org.apache.flink.configuration.DeploymentOptions.SUBMIT_FAILED_JOB_ON_APPLICATION_ERROR;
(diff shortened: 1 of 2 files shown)


1996fanrui (approved):
Thanks @jrthe42 for the fix!

LGTM.

gyfora (approved):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
```diff
@@ -375,9 +376,24 @@ private int getParallelism() {
...
 
+        Optional<Integer> maxOverrideParallelism = getMaxParallelismFromOverrideConfig();
+        if (maxOverrideParallelism.isPresent() && maxOverrideParallelism.get() > 0) {
+            return maxOverrideParallelism.get();
+        }
+
         return spec.getJob().getParallelism();
     }
 
+    private Optional<Integer> getMaxParallelismFromOverrideConfig() {
```
mxm, line 387: If this method is used to determine the number of task slots, that won't work if slot sharing is disabled or if multiple slot sharing groups are used.

gyfora: You are completely right @mxm however this is a general limitation of the standalone mode itself.

PR #733, opened 2023-12-13
[FLINK-33632] Adding custom flink mutator

AncyRominus:

This PR is to provide pluggable mutators. Currently the operator allows for pluggable validators, we would like to extend this capability to provide custom mutation. Our organisation, for example, requires certain fields to be present on the deployment. Whilst we can do this using the podTemplate overrides it makes the CR quite unwieldy. Having a custom mutator would mean we could automatically add these values to the CR.

## Brief change log

Implemented custom mutator for the flink kuberenetes operator.
* Implemented the interface FlinkResourceMutator.
* FlinkResourceMutator interface contains - mutateDeployment and mutateSessionJob which is implemented by Default and custom mutator.
* MutatorUtils discoverMutators methods helps to identify all the mutators which are implemented using FlinkResourceMutator interface.
* The Flink-kubernetes-Webhook contains a FlinkOperatorWebhook class with a main method which is invoked for the webhoook validations. All the mutators (custom and default mutators) are discovered/identified over here. This is intern send to the AdmissionHandler.
* AdmissionHandler handles the mutator which does the actual mutate by invoking the mutate method.
* The Flink Mutator has the mutate method which does the mutation and also iterates through the default and custom mutator and performs the mutate function.

## Verifying this change

This change is already covered by existing tests, such as  AdmissionHandlerTests.

  - Dependencies (does it add or upgrade a dependency): (yes / no) No
  - The public API, i.e., is any changes to the `CustomResourceDescriptors`: (yes / no) :No
  - Core observer or reconciler logic that is regularly executed: (yes / no) :No

## Documentation

  - Does this pull request introduce a new feature? (yes / no) Yes
Docs

### DIFF
--- a/flink-kubernetes-operator/pom.xml
+++ b/flink-kubernetes-operator/pom.xml
@@ -208,18 +208,6 @@ under the License.
             <artifactId>junit-jupiter-params</artifactId>
             <scope>test</scope>
         </dependency>
-
-        <dependency>
-            <groupId>io.javaoperatorsdk</groupId>
-            <artifactId>kubernetes-webhooks-framework-core</artifactId>
-            <version>${operator.sdk.webhook-framework.version}</version>
-            <exclusions>
-                <exclusion>
-                    <groupId>*</groupId>
-                    <artifactId>*</artifactId>
-                </exclusion>
-            </exclusions>
-        </dependency>
     </dependencies>
 
     <build>
--- a/flink-kubernetes-operator/pom.xml
+++ b/flink-kubernetes-operator/pom.xml
@@ -208,6 +208,18 @@ under the License.
             <artifactId>junit-jupiter-params</artifactId>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>io.javaoperatorsdk</groupId>
+            <artifactId>kubernetes-webhooks-framework-core</artifactId>
+            <version>${operator.sdk.webhook-framework.version}</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>*</groupId>
+                    <artifactId>*</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
     </dependencies>
 
     <build>
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/MutatorUtils.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/MutatorUtils.java
@@ -1,0 +1,65 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.kubernetes.operator.utils;
+
+import org.apache.flink.configuration.ConfigConstants;
+import org.apache.flink.core.plugin.PluginUtils;
+import org.apache.flink.kubernetes.operator.config.FlinkConfigManager;
+import org.apache.flink.kubernetes.operator.mutator.DefaultFlinkMutator;
+import org.apache.flink.kubernetes.operator.mutator.FlinkResourceMutator;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.HashSet;
+import java.util.Set;
+
+/** Mutator utilities. */
+public final class MutatorUtils {
+
+    private static final Logger LOG = LoggerFactory.getLogger(MutatorUtils.class);
+
+    /**
+     * discovers mutators.
+     *
+     * @param configManager Flink Config manager
+     * @return Set of FlinkResourceMutator
+     */
+    public static Set<FlinkResourceMutator> discoverMutators(FlinkConfigManager configManager) {
+        var conf = configManager.getDefaultConfig();
+        Set<FlinkResourceMutator> flinkmutator = new HashSet<>();
+        DefaultFlinkMutator defaultFlinkMutator = new DefaultFlinkMutator();
+        defaultFlinkMutator.configure(conf);
+        flinkmutator.add(defaultFlinkMutator);
+        PluginUtils.createPluginManagerFromRootFolder(conf)
+                .load(FlinkResourceMutator.class)
+                .forEachRemaining(
+                        mutator -> {
+                            LOG.info(
+                                    "Discovered mutator from plugin directory[{}]: {}.",
+                                    System.getenv()
+                                            .getOrDefault(
+                                                    ConfigConstants.ENV_FLINK_PLUGINS_DIR,
+                                                    ConfigConstants.DEFAULT_FLINK_PLUGINS_DIRS),
+                                    mutator.getClass().getName());
+                            mutator.configure(conf);
+                            flinkmutator.add(mutator);
+                        });
+        return flinkmutator;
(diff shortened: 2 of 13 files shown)


tagarr:
Looks good. My only question would be to the committers.
1. Would we like the sessionjob changes (i.e. modify labels) be done in flink-kubernetes-webhook/src/main/java/org/apache/flink/kubernetes/operator/admission/mutator/FlinkMutator.java as it currently is or pushed down to the default mutator ?
2. Do we want to add sessionjob mutation for Update as well as for Create operations ?

tagarr:
@gyfora would love your opinion on this...

gyfora:
```
Brief change log

No change in logs
```

This section should ideally be filled out with the list of changes that were made to implement the given feature / bug fix :)

flink-kubernetes-webhook/pom.xml
```diff
@@ -127,6 +132,11 @@ under the License.
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <version>${surefire-plugin.version}</version>
```
gyfora, line 138: Why do we need this new plugin?

flink-kubernetes-operator/pom.xml
```diff
@@ -208,6 +208,18 @@ under the License.
...
+            <groupId>io.javaoperatorsdk</groupId>
+            <artifactId>kubernetes-webhooks-framework-core</artifactId>
+            <version>${operator.sdk.webhook-framework.version}</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>*</groupId>
+                    <artifactId>*</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
```
gyfora, line 222: Why do we need this new dependency inside the operator?

gyfora:
There is already a mutator implementation `FlinkMutator` I don't see that mentioned or touched anywhere here. Could we also make sure that it is covered by the new interfaces as well?

flink-kubernetes-webhook/pom.xml
AncyRominus: This plugin is required for the successful execution of the test AdmissionHandlerTest.testMutateHandler. Else it will throw <500 Internal Server Error> while mutating the SessionJob as it’s using the deployment from the informerManager which uses Kubernetes Client.

flink-kubernetes-operator/pom.xml
AncyRominus: This  dependency is for the Kubernetes-webhooks-framework. This can be removed as we don't have any strong relation with this dependency.

AncyRominus:
Thanks for the comments @tagarr and @gyfora !!! 
The modify labels are set in FlinkMutator.java and the other mutator changes (if any) be added to the default or the custom mutators. The default/custom mutator sessionJob changes are iterated and the mutateSessionJobs are processed.

Yes, we can add the sessionjob mutation for Update as well as for Create operations.

tagarr:
@gyfora We had considered adding mutation into the controller loop in the same manner that the validator is done, but we believe that mutation of the CR should only really be done on it’s creation or user modified updates. Changes within the reconcile could cause unknown issues and confuse the logic within the controller

flink-kubernetes-webhook/src/main/java/org/apache/flink/kubernetes/operator/admission/mutator/FlinkMutator.java
```diff
@@ -17,43 +17,88 @@
...
+    private FlinkSessionJob mutateSessionJob(HasMetadata resource) {
+        try {
+            var sessionJob = mapper.convertValue(resource, FlinkSessionJob.class);
+            var namespace = sessionJob.getMetadata().getNamespace();
+            var deploymentName = sessionJob.getSpec().getDeploymentName();
+            var key = Cache.namespaceKeyFunc(namespace, deploymentName);
+            var deployment =
+                    informerManager.getFlinkDepInformer(namespace).getStore().getByKey(key);
+
+            setSessionTargetLabel(sessionJob);
```
gyfora, line 75: I think if we are adding a mutator abstraction, this logic should be part of the default mutator for session jobs

gyfora: I mean `setSessionTargetLabel(sessionJob);`  should be in the default mutator

AncyRominus: Thanks @gyfora ! Added the logic for setting session target label as part of default mutator logic.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/MutatorUtils.java
```diff
@@ -0,0 +1,65 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(MutatorUtils.class);
+
+    /**
+     * discovers mutators.
+     *
+     * @param configManager Flink Config manager
+     * @return Set of FlinkResourceMutator
+     */
+    public static Set<FlinkResourceMutator> discoverMutators(FlinkConfigManager configManager) {
```
gyfora, line 43: We should have a test for this similar to the `ValidatorUtilsTest` alternatively it would be even better to refactor the code and eliminate the overlapping logic between discovering the validator and mutator and then we can have a single test

flink-kubernetes-webhook/src/test/java/org/apache/flink/kubernetes/operator/admission/AdmissionHandlerTest.java
```diff
@@ -136,6 +140,10 @@ public void testHandleValidateRequestWithAdmissionReview() throws IOException {
     public void testMutateHandler() throws Exception {
         final EmbeddedChannel embeddedChannel = new EmbeddedChannel(admissionHandler);
         var sessionJob = new FlinkSessionJob();
+        ObjectMeta objectMeta = new ObjectMeta();
```
gyfora, line 143: We should also add a test that has a custom mutator and make sure it is called.

gyfora (changes requested):
The code looks good but I think we need more test coverage overall.

flink-kubernetes-webhook/src/main/java/org/apache/flink/kubernetes/operator/admission/mutator/FlinkMutator.java
```diff
@@ -17,55 +17,82 @@
...
+            var namespace = sessionJob.getMetadata().getNamespace();
+            var deploymentName = sessionJob.getSpec().getDeploymentName();
+            var key = Cache.namespaceKeyFunc(namespace, deploymentName);
+            var deployment =
+                    informerManager.getFlinkDepInformer(namespace).getStore().getByKey(key);
+
+            for (FlinkResourceMutator mutator : mutators) {
+                FlinkSessionJob flinkSessionJob =
+                        mutator.mutateSessionJob(sessionJob, Optional.ofNullable(deployment));
+                sessionJob = flinkSessionJob;
```
gyfora, line 77: Simplify to : 
```
sessionJob =  mutator.mutateSessionJob(sessionJob, Optional.ofNullable(deployment));
```

flink-kubernetes-webhook/src/main/java/org/apache/flink/kubernetes/operator/admission/mutator/FlinkMutator.java
```diff
@@ -17,55 +17,82 @@
...
-                    CrdConstants.LABEL_TARGET_SESSION,
-                    flinkSessionJob.getSpec().getDeploymentName());
-            flinkSessionJob.getMetadata().setLabels(labels);
+    }
+
+    private FlinkDeployment mutateDeployment(HasMetadata resource) {
+        try {
+            var flinkDeployment = mapper.convertValue(resource, FlinkDeployment.class);
+            for (FlinkResourceMutator mutator : mutators) {
+                FlinkDeployment deployment = mutator.mutateDeployment(flinkDeployment);
```
gyfora, line 90: Simplify to:
```
flinkDeployment = mutator.mutateDeployment(flinkDeployment);
```

gyfora (changes requested):
Thank you, looks good. Added some lost minor cleanup comments :)

AncyRominus:
Thank you @gyfora ! I have incorporated the review comments. Please re-review

gyfora (approved):

PR #734, opened 2023-12-14
[FLINK-33770] Silence Configuration logging for INFO messages

mxm:
In #725 we coverted the deprecated keys fallback keys and added test coverage. This additional change silences logs for fallback keys by setting the configuration logging to WARN instead of the default INFO.

This avoids spamming the log with unneccessary messages on every reconciliation loop, while still reporting warnings and errors.


1996fanrui (approved):
This is indeed a simple and useful solution!

LGTM.

gyfora (approved):

mxm:
Thanks for the quick review!

PR #735, opened 2023-12-15
[FLINK-30593][autoscaler] Improve restart time tracking

afedulov:
This PR contains the following improvements to the restart tracking logic:
* Adds more debug logs
* Stores restart Duration directly instead of the endTime Instant
* Fixes a bug that makes restart duration tracking dependent on whether metrics are considered fully collected

### DIFF
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
@@ -179,12 +179,6 @@ public class JobAutoScalerImpl<KEY, Context extends JobAutoScalerContext<KEY>>
                 jobTopology.getVerticesInTopologicalOrder(),
                 () -> lastEvaluatedMetrics.get(ctx.getJobKey()));
 
-        if (!collectedMetrics.isFullyCollected()) {
-            // We have done an upfront evaluation, but we are not ready for scaling.
-            resetRecommendedParallelism(evaluatedMetrics.getVertexMetrics());
-            return;
-        }
-
         var scalingHistory = getTrimmedScalingHistory(stateStore, ctx, now);
         // A scaling tracking without an end time gets created whenever a scaling decision is
         // applied. Here, when the job transitions to RUNNING, we record the time for it.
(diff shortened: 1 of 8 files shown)


flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
```diff
@@ -182,22 +182,22 @@ private void runScalingLogic(Context ctx, AutoscalerFlinkMetrics autoscalerMetri
...
         var scalingHistory = getTrimmedScalingHistory(stateStore, ctx, now);
         // A scaling tracking without an end time gets created whenever a scaling decision is
         // applied. Here, when the job transitions to RUNNING, we record the time for it.
         if (ctx.getJobStatus() == JobStatus.RUNNING) {
-            if (scalingTracking.setEndTimeIfTrackedAndParallelismMatches(
+            if (scalingTracking.recordRestartDurationIfTrackedAndParallelismMatches(
                     now, jobTopology, scalingHistory)) {
                 stateStore.storeScalingTracking(ctx, scalingTracking);
             }
         }
```
mxm, line 193: We don't need the RUNNING job state check. This block can be reduced to:

```java
          if (scalingTracking.recordRestartDurationIfTrackedAndParallelismMatches(
                  now, jobTopology, scalingHistory)) {
              stateStore.storeScalingTracking(ctx, scalingTracking);
          }
```

The reason is that this method only gets called when the job is in running state (see line 99). Enforcing a RUNNING state has always been a precondition for executing autoscaling.

mxm (approved):
Nices fixes. LGTM. Just a minor suggestion.

mxm:
This needs a rebase. I'll run the tests afterwards.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
gyfora, line 193: can you please extract this logic into a method to keep the flow simpler?

gyfora (changes requested):

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
gyfora: It could be part of `stateStore.storeScalingTracking(ctx, scalingTracking);` even

mxm: I was going to recommend this as well. But these are only three lines of code (after removing the unneeded RUNNING condition) and the resulting method signature would be quite big. I don't think we need to block the PR on this refactoring.

afedulov:
@mxm I addressed the comments and rebased, could you please kick off the tests again?

mxm:
Sure, they should be running now.

mxm (approved):
Thanks Alex! Looks good. Are you ok with merging @gyfora?

gyfora (approved):
🚢

afedulov:
@mxm @gyfora 
Thanks!

PR #738, opened 2023-12-20
[FLINK-33917] fix hostname can't be null for certain edgecases

ensctom:
https://issues.apache.org/jira/browse/FLINK-33917

Fix hostname being null under certain edgecases. 

## Brief change log

  - refactor AbstractFlinkService#isJobManagerReady to use URL instead of URI
You can reproduce this locally

```
    @ParameterizedTest
    @ValueSource(
            strings = {"http://127.0.0.1:8081", "http://dev-test:8081", "http://dev-test.01:8081"})
    void testURLAddresses(String inputAddress) {

        assertDoesNotThrow(
                () -> {
                    final URL url = new URL(inputAddress);
                    new InetSocketAddress(url.getHost(), url.getPort());
                });
    }

    @ParameterizedTest
    @ValueSource(
            strings = {"http://127.0.0.1:8081", "http://dev-test:8081", "http://dev-test.01:8081"})
    void testURIAddresses(String inputAddress) {

        assertDoesNotThrow(
                () -> {
                    final URI uri= new URI(inputAddress);
                    new InetSocketAddress(uri.getHost(), uri.getPort());
                });
    }

```

The URL passes all test case but the URI fails on the test case `http://dev-test.01:8081`

## Verifying this change

*(example:)*
  - *Add new test cases to test different url form factors 
  - *Manually verified that the reconciliation loop still works as expected

  - Dependencies (does it add or upgrade a dependency): NO
  - The public API, i.e., is any changes to the `CustomResourceDescriptors`:  NO
  - Core observer or reconciler logic that is regularly executed:   YES

## Documentation

  - Does this pull request introduce a new feature? NO

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
@@ -113,9 +113,10 @@ import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.net.InetSocketAddress;
+import java.net.MalformedURLException;
 import java.net.Socket;
 import java.net.SocketAddress;
-import java.net.URI;
+import java.net.URL;
 import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
```diff
@@ -234,23 +235,29 @@ public JobID submitJobToSessionCluster(
...
-            socket.close();
             return true;
         } catch (IOException e) {
             return false;
         }
     }
 
+    protected SocketAddress getSocketAddress(RestClusterClient<String> clusterClient)
+            throws MalformedURLException {
+        final URL url = new URL(clusterClient.getWebInterfaceURL());
```
mxm, line 256: Flink uses URL to create this String. Hence using URL to extract the hostname makes sense. URI fails to interpret the host name correctly in certain URLs like `http://test.01-bla:8081` and treats it as a different part of an URI.

mxm (approved):

gyfora (approved):

PR #741, opened 2023-12-24
[FLINK-33450][autoscaler] Support the JDBCAutoScalerStateStore

1996fanrui:

Introduced the JDBCAutoScalerStateStore

## Brief change log

- [FLINK-33450][autoscaler] Extract the AbstractAutoScalerStateStoreTest and add the InMemoryAutoScalerStateStoreTest
  - JDBCAutoScalerStateStore can reuse the AbstractAutoScalerStateStoreTest as well 
- [FLINK-33450][autoscaler] Support the JDBCAutoScalerStateStore
  - Introduce a series of `DatabaseTest` in this commit as well, it's easy to test and finish the ITCase for derby, mysql and PostgreSQL
  - JDBCAutoScalerStateStore has cache as well, it's similar to KubernetesAutoScalerStateStore, core class:
  - StateType, JDBCStateInteractor, JDBCStore and JobStateView
- [FLINK-33450][autoscaler] StandaloneAutoscalerEntrypoint supports the JDBCAutoScalerStateStore
  - Introduced some options to load JDBCAutoScalerStateStore, such as:
  - autoscaler.standalone.state-store.type
  - autoscaler.standalone.state-store.jdbc.password
  - autoscaler.standalone.state-store.jdbc.url
  - autoscaler.standalone.state-store.jdbc.username
- [FLINK-33450][autoscaler] Document the JDBC Autoscaler State Store

## Verifying this change

- Added a series of unit tests and ITCases in `flink-autoscaler-plugin-jdbc` module
- Autoscaler Standalone uses mysql and PostgreSQL manually, it works well

  - Dependencies (does it add or upgrade a dependency): yes, added some dependencies to `flink-autoscaler-plugin-jdbc` module

## Documentation

  - Generated the configuration option docs
  - Improve the autoscaler standalone doc

### DIFF
--- a/docs/content/docs/custom-resource/autoscaler.md
+++ b/docs/content/docs/custom-resource/autoscaler.md
@@ -260,17 +260,46 @@ job.autoscaler.metrics.window : 3m
 > `ScalingReport` will show the recommended parallelism for each vertex.
 
 After the flink job starts, please start the StandaloneAutoscaler process by the
-following command.
+following command. Please download released autoscaler-standalone jar from 
+[here](https://repo.maven.apache.org/maven2/org/apache/flink/flink-autoscaler-standalone/) first.
 
 ```
 java -cp flink-autoscaler-standalone-{{< version >}}.jar \
 org.apache.flink.autoscaler.standalone.StandaloneAutoscalerEntrypoint \
---flinkClusterHost localhost \
---flinkClusterPort 8081
+--autoscaler.standalone.fetcher.flink-cluster.host localhost \
+--autoscaler.standalone.fetcher.flink-cluster.port 8081
 ```
 
-Updating the `flinkClusterHost` and `flinkClusterPort` based on your flink cluster.
-In general, the host and port are the same as Flink WebUI.
+Updating the `autoscaler.standalone.fetcher.flink-cluster.host` and `autoscaler.standalone.fetcher.flink-cluster.port` 
+based on your flink cluster. In general, the host and port are the same as Flink WebUI.
+
+### Using the JDBC Autoscaler State Store
+
+A driver dependency is required to connect to a specified database. Here are drivers currently supported, 
+please download JDBC driver and initialize database and table first.
+
+| Driver     | Group Id                   | Artifact Id            | JAR                                                                             | Schema                  |
+|:-----------|:---------------------------|:-----------------------|:--------------------------------------------------------------------------------|-------------------------|
+| MySQL      | `mysql`                    | `mysql-connector-java` | [Download](https://repo.maven.apache.org/maven2/mysql/mysql-connector-java/)    | [Table DDL](https://github.com/apache/flink-kubernetes-operator/blob/main/flink-autoscaler-plugin-jdbc/src/main/resources/schema/mysql_schema.sql)     |
+| PostgreSQL | `org.postgresql`           | `postgresql`           | [Download](https://jdbc.postgresql.org/download/)                               | [Table DDL](https://github.com/apache/flink-kubernetes-operator/blob/main/flink-autoscaler-plugin-jdbc/src/main/resources/schema/postgres_schema.sql)  |
+| Derby      | `org.apache.derby`         | `derby`                | [Download](http://db.apache.org/derby/derby_downloads.html)                     | [Table DDL](https://github.com/apache/flink-kubernetes-operator/blob/main/flink-autoscaler-plugin-jdbc/src/main/resources/schema/derby_schema.sql)     |
+
+```
+JDBC_DRIVER_JAR=./mysql-connector-java-8.0.30.jar
+# export the password of jdbc state store
+export STATE_STORE_JDBC_PWD=123456
+
+java -cp flink-autoscaler-standalone-{{< version >}}.jar:${JDBC_DRIVER_JAR} \
+org.apache.flink.autoscaler.standalone.StandaloneAutoscalerEntrypoint \
+--autoscaler.standalone.fetcher.flink-cluster.host localhost \
+--autoscaler.standalone.fetcher.flink-cluster.port 8081 \
+--autoscaler.standalone.state-store.type jdbc \
+--autoscaler.standalone.state-store.jdbc.url jdbc:mysql://localhost:3306/flink_autoscaler \
+--autoscaler.standalone.state-store.jdbc.username root
+```
+
+All supported options for autoscaler standalone can be viewed 
+[here]({{< ref "docs/operations/configuration#autoscaler-standalone-configuration" >}}).
 
 ### Extensibility of autoscaler standalone
 
--- a/flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/state/JdbcAutoScalerStateStore.java
+++ b/flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/state/JdbcAutoScalerStateStore.java
@@ -1,0 +1,253 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.autoscaler.jdbc.state;
+
+import org.apache.flink.annotation.Experimental;
+import org.apache.flink.annotation.VisibleForTesting;
+import org.apache.flink.autoscaler.JobAutoScalerContext;
+import org.apache.flink.autoscaler.ScalingSummary;
+import org.apache.flink.autoscaler.ScalingTracking;
+import org.apache.flink.autoscaler.metrics.CollectedMetrics;
+import org.apache.flink.autoscaler.state.AutoScalerStateStore;
+import org.apache.flink.autoscaler.utils.AutoScalerSerDeModule;
+import org.apache.flink.configuration.ConfigurationUtils;
+import org.apache.flink.runtime.jobgraph.JobVertexID;
+
+import com.fasterxml.jackson.core.JacksonException;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nonnull;
+
+import java.time.Instant;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import java.util.SortedMap;
+import java.util.TreeMap;
+
+import static org.apache.flink.autoscaler.jdbc.state.StateType.COLLECTED_METRICS;
+import static org.apache.flink.autoscaler.jdbc.state.StateType.PARALLELISM_OVERRIDES;
+import static org.apache.flink.autoscaler.jdbc.state.StateType.SCALING_HISTORY;
+import static org.apache.flink.autoscaler.jdbc.state.StateType.SCALING_TRACKING;
+
+/**
+ * The state store which persists its state in JDBC related database.
+ *
+ * @param <KEY> The job key.
+ * @param <Context> The job autoscaler context.
+ */
+@Experimental
+public class JdbcAutoScalerStateStore<KEY, Context extends JobAutoScalerContext<KEY>>
+        implements AutoScalerStateStore<KEY, Context> {
+
+    private static final Logger LOG = LoggerFactory.getLogger(JdbcAutoScalerStateStore.class);
+
+    private final JdbcStateStore jdbcStateStore;
+
+    protected static final ObjectMapper YAML_MAPPER =
+            new ObjectMapper()
+                    .registerModule(new JavaTimeModule())
+                    .registerModule(new AutoScalerSerDeModule());
+
+    public JdbcAutoScalerStateStore(JdbcStateStore jdbcStateStore) {
+        this.jdbcStateStore = jdbcStateStore;
+    }
+
+    @Override
+    public void storeScalingHistory(
+            Context jobContext, Map<JobVertexID, SortedMap<Instant, ScalingSummary>> scalingHistory)
+            throws Exception {
+        jdbcStateStore.putSerializedState(
+                getSerializeKey(jobContext),
+                SCALING_HISTORY,
+                serializeScalingHistory(scalingHistory));
+    }
+
+    @Nonnull
+    @Override
+    public Map<JobVertexID, SortedMap<Instant, ScalingSummary>> getScalingHistory(
+            Context jobContext) {
+        Optional<String> serializedScalingHistory =
+                jdbcStateStore.getSerializedState(getSerializeKey(jobContext), SCALING_HISTORY);
+        if (serializedScalingHistory.isEmpty()) {
+            return new HashMap<>();
+        }
+        try {
+            return deserializeScalingHistory(serializedScalingHistory.get());
+        } catch (JacksonException e) {
+            LOG.error(
+                    "Could not deserialize scaling history, possibly the format changed. Discarding...",
+                    e);
+            jdbcStateStore.removeSerializedState(getSerializeKey(jobContext), SCALING_HISTORY);
+            return new HashMap<>();
+        }
+    }
+
+    @Override
+    public void removeScalingHistory(Context jobContext) {
+        jdbcStateStore.removeSerializedState(getSerializeKey(jobContext), SCALING_HISTORY);
+    }
+
+    @Override
+    public void storeScalingTracking(Context jobContext, ScalingTracking scalingTrack)
+            throws Exception {
+        jdbcStateStore.putSerializedState(
+                getSerializeKey(jobContext),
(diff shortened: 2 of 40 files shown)


1996fanrui (commented):
Hi @gyfora @mxm , the `JDBCAutoScalerStateStore` is supported in this PR, and I have tested it manually with Autoscaler standalone, it works well.

Please take a look in your free time, thanks a lot~ ❤️

flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/state/JobStateView.java
```diff
@@ -0,0 +1,262 @@
...
+        /** State doesn't exist at database, and it's not used so far, so it's not needed. */
+        NOT_NEEDED(false, false, false),
+        /** State is only stored locally, not created in JDBC database yet. */
+        NEEDS_CREATE(true, false, true),
+        /** State exists in JDBC database but there are newer local changes. */
+        NEEDS_UPDATE(true, true, true),
+        /** State is stored locally and in database, and they are same. */
+        UP_TO_DATE(true, true, false),
+        /** State is stored in database, but it's deleted in local. */
+        NEEDS_DELETE(false, true, true);
```
mxm, line 61: This looks conceptually very similar to `ConfigMapView`. I wonder if we could have shared some of that logic.

mxm (approved):
LGTM!

flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/config/AutoscalerStandaloneOptions.java
```diff
@@ -59,4 +64,47 @@ private static ConfigOptions.OptionBuilder autoscalerStandaloneConfig(String key
                     .withDeprecatedKeys("flinkClusterPort")
                     .withDescription(
                             "The port of flink cluster when the flink-cluster fetcher is used.");
+
+    public static final ConfigOption<StateStoreType> STATE_STORE_TYPE =
+            autoscalerStandaloneConfig("state-store.type")
+                    .enumType(StateStoreType.class)
+                    .defaultValue(StateStoreType.MEMORY)
+                    .withDescription("The autoscaler state store type.");
```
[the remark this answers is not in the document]
1996fanrui: Emmm, I didn't list due to flink-kubernetes-operator doc will list automaticlly.

You can take a look the generated html, it has all enums of `StateStoreType`. 

And all old options have `Possible values:` as well.

图片

flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/state/JobStateView.java
1996fanrui: Yeah, I refer to `ConfigMapView` during developing. I didn't share logic due to some reasons:

1. The View and State are different
  - `ConfigMapView` stores `ConfigMap` for each job, so we create it directly.
  - `JobStateView` stores state for each stateType, it means that part of state types of current job are stored in the database, but the rest of the state types may have been created in the database.
  - That's why `JobStateView` have `NOT_NEEDED` and `NEEDS_DELETE`, but `ConfigMapView` doesn't.
2. Based on the first reason, the `ConfigMapView` accesses the whole ConfigMap, `JobStateView`  accesses each row of database. It's hard to share code.

mxm: Got it. The code complexity for adding a new state backend makes we wonder though, if we need to rethink some of the building blocks for state stores. It would be good to move more of the logic into the core, to avoid duplicating / writing similar logic. I was initially more in favor of moving this out of the core but we need to see if that might be too complicated in the long run. This is out of scope for this PR though.

flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/config/AutoscalerStandaloneOptions.java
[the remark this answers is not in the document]
mxm: Got it! Forgot about that.

flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/state/JobStateView.java
1996fanrui: > It would be good to move more of the logic into the core, to avoid duplicating / writing similar logic. 

Sounds make sense, after I analyze, some of logic are same, such as: 

- `ConfigMapStore` and `JDBCStore` can be abstracted to `StringStateStore` interface
  - They support `put`, `get` and `remove`
  - The parameters of `ConfigMapStore` are the (JobContext, String key, String value).
  - The parameters of `JDBCStore` are the (String jobKey, StateType stateType, String value).
  - We can define a interface `StringStateStore`, and the parameters are `(JobContext, StateType stateType, String value)`.

- `KubernetesAutoScalerStateStore` and `JDBCAutoScalerStateStore` can be abstracted to `AbstractAutoscalerStateStore`
  - They support serialize and compress `Original State` to String.
  - `AbstractAutoscalerStateStore` can reuse the serialize and compress logic
  - `KubernetesAutoScalerStateStore` support the limitation of stateValue
  - We can define a parameter for `AbstractAutoscalerStateStore`, the limitation is disabled by default, and `KubernetesAutoScalerStateStore` can enable it.

And I created FLINK-34065 to follow it.

1996fanrui: Hi @mxm @gyfora , in the beginning of design the state store, we use the state type id in the database table.

After I thinking, more and this suggestion from @mxm . It's better to share some code between `KubernetesAutoScalerStateStore` and `JDBCAutoScalerStateStore`. So I change the state type from id to `identifier` (From int to String).

Align it to `KubernetesAutoScalerStateStore`, using: `scalingHistory`, `scalingTracking`, `collectedMetrics` and `parallelismOverrides`. 

It has 2 advantages : 

- It's easy to share the `StateType` class
- When we use the `JDBCAutoScalerStateStore`, the identifier is easier to troubleshooting than id. For example, we needs to query  some states  from database manually. It's easy to understand.

We can ensure the `identifier` length isn't too long, it won't introduce too many storage cost.

I have update it in the latest hotfix commit, WDYT?

flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/state/JDBCStore.java
```diff
@@ -0,0 +1,92 @@
...
+package org.apache.flink.autoscaler.jdbc.state;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+
+/** The jdbc store. */
+public class JDBCStore {
```
gyfora, line 27: Should we call it `JdbcStateStore` instead?

flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/config/AutoscalerStandaloneOptions.java
```diff
@@ -59,4 +64,47 @@ private static ConfigOptions.OptionBuilder autoscalerStandaloneConfig(String key
...
+                    .noDefaultValue()
+                    .withDescription(
+                            Description.builder()
+                                    .text(
+                                            "The jdbc username of jdbc state store when %s has been set to %s.",
+                                            code(STATE_STORE_TYPE.key()), code(JDBC.toString()))
+                                    .build());
+
+    public static final ConfigOption<String> STATE_STORE_JDBC_PASSWORD =
+            autoscalerStandaloneConfig("state-store.jdbc.password")
```
gyfora, line 101: We need a way to set this from a secret / env. Otherwise this is very hard to be used in production properly

flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/state/JDBCAutoScalerStateStore.java
```diff
@@ -0,0 +1,286 @@
...
+import static org.apache.flink.autoscaler.jdbc.state.StateType.SCALING_TRACKING;
+
+/**
+ * The state store which persists its state in JDBC related database.
+ *
+ * @param <KEY> The job key.
+ * @param <Context> The job autoscaler context.
+ */
+@Experimental
+public class JDBCAutoScalerStateStore<KEY, Context extends JobAutoScalerContext<KEY>>
```
gyfora, line 67: Should be ` JdbcAutoScalerStateStore<`

flink-autoscaler-plugin-jdbc/src/test/java/org/apache/flink/autoscaler/jdbc/state/AbstractJDBCStateInteractorITCase.java
```diff
@@ -0,0 +1,93 @@
...
+
+import java.util.List;
+import java.util.Map;
+
+import static org.apache.flink.autoscaler.jdbc.state.StateType.COLLECTED_METRICS;
+import static org.apache.flink.autoscaler.jdbc.state.StateType.SCALING_HISTORY;
+import static org.assertj.core.api.Assertions.assertThat;
+
+/** The abstract IT case for {@link JDBCStateInteractor}. */
+abstract class AbstractJDBCStateInteractorITCase implements DatabaseTest {
```
gyfora, line 37: `JDBC` -> `Jdbc
`

flink-autoscaler-plugin-jdbc/src/test/java/org/apache/flink/autoscaler/jdbc/state/AbstractJDBCStoreITCase.java
```diff
@@ -0,0 +1,303 @@
...
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import static org.apache.flink.autoscaler.jdbc.state.StateType.COLLECTED_METRICS;
+import static org.apache.flink.autoscaler.jdbc.state.StateType.SCALING_HISTORY;
+import static org.apache.flink.autoscaler.jdbc.state.StateType.SCALING_TRACKING;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+/** The abstract IT case for {@link JDBCStore}. */
+abstract class AbstractJDBCStoreITCase implements DatabaseTest {
```
gyfora, line 45: JDBC -> Jdbc

flink-autoscaler-plugin-jdbc/src/test/java/org/apache/flink/autoscaler/jdbc/state/CountableJDBCStateInteractor.java
```diff
@@ -0,0 +1,81 @@
...
+
+import java.sql.Connection;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicLong;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/** Countable {@link JDBCStateInteractor}. */
+public class CountableJDBCStateInteractor extends JDBCStateInteractor {
```
gyfora, line 28: JDBC -> Jdbc

flink-autoscaler-plugin-jdbc/src/test/java/org/apache/flink/autoscaler/jdbc/state/JDBCAutoScalerStateStoreTest.java
```diff
@@ -0,0 +1,161 @@
...
+import static org.apache.flink.autoscaler.metrics.ScalingHistoryUtils.getTrimmedScalingHistory;
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Test for {@link JDBCAutoScalerStateStore}.
+ *
+ * <p>Note: {@link #testCompressionMigration} and {@link #testDiscardInvalidHistory} are referred
+ * from {@link KubernetesAutoScalerStateStoreTest}.
+ */
+class JDBCAutoScalerStateStoreTest
```
gyfora, line 53: JDBC -> Jdbc

gyfora (changes requested):
I added some minor comments @1996fanrui mostly for proper camel-casing the class names.

But more importantly we should allow a more flexible config of the password.

1996fanrui (commented):
Thanks @gyfora for the suggestion!

I have updated all `JDBC` to `Jdbc`. Do you have any specific solution about the password setting?

gyfora:
@1996fanrui would be great if Flink had environment variable resolution in the conf but an alternative would be an extra config called: `state-store.jdbc.password-env` where we could specify the env variable name to resolve. If password-env is defined then the password would be ignored or vice versa 

In Kubernetes world secrets can be easily mounted as env variables so this would make it nice and easy to configure

1996fanrui:
> @1996fanrui would be great if Flink had environment variable resolution in the conf but an alternative would be an extra config called: `state-store.jdbc.password-env` where we could specify the env variable name to resolve. If password-env is defined then the password would be ignored or vice versa
> 
> In Kubernetes world secrets can be easily mounted as env variables so this would make it nice and easy to configure

Thanks @gyfora for the quick feedback!

I have updated the `state-store.jdbc.password` to `state-store.jdbc.password-env-variable`, and updated the doc to guide user to export the `STATE_STORE_JDBC_PWD` env variable.

docs/content/docs/custom-resource/autoscaler.md
```diff
@@ -260,17 +260,46 @@ job.autoscaler.metrics.window : 3m
...
+# export the password of jdbc state store
+export STATE_STORE_JDBC_PWD=123456
+
+java -cp flink-autoscaler-standalone-{{< version >}}.jar:${JDBC_DRIVER_JAR} \
+org.apache.flink.autoscaler.standalone.StandaloneAutoscalerEntrypoint \
+--autoscaler.standalone.fetcher.flink-cluster.host localhost \
+--autoscaler.standalone.fetcher.flink-cluster.port 8081 \
+--autoscaler.standalone.state-store.type jdbc \
+--autoscaler.standalone.state-store.jdbc.url jdbc:mysql://localhost:3306/flink_autoscaler \
+--autoscaler.standalone.state-store.jdbc.username root
```
gyfora, line 298: should we also state the password-env-variable config here?

flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/state/JdbcAutoScalerStateStore.java
```diff
@@ -0,0 +1,286 @@
...
+
+    private static String serializeParallelismOverrides(Map<String, String> overrides) {
+        return ConfigurationUtils.convertValue(overrides, String.class);
+    }
+
+    private static Map<String, String> deserializeParallelismOverrides(String overrides) {
+        return ConfigurationUtils.convertValue(overrides, Map.class);
+    }
+
+    private static String compress(String original) throws IOException {
```
gyfora, line 262: Is there really a reason to compress the data in a DB? For configmaps this was done due to the very constrained size. But compressing makes it much harder to read for users

docs/content/docs/custom-resource/autoscaler.md
1996fanrui: In general, the environment variable name doesn't need to be changed. Users need to export the password using this environment variable.

So I didn't mention `password-env-variable` here. But in the beginning of this doc, it mentioned how to export password. WDYT?

```
# export the password of jdbc state store
export STATE_STORE_JDBC_PWD=123456
```

flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/state/JdbcAutoScalerStateStore.java
1996fanrui: No any strong reason, it refers the `KubernetesAutoScalerStateStore`.

I removed the the compression related logic for JdbcAutoScalerStateStore in the latest hotfix commit.

docs/content/docs/custom-resource/autoscaler.md
gyfora: You are right, sorry I got confused by the logic for second :) 

I think it would make sense to keep the config for the password there as well as an option for people to use instead of the env. The env variable is great for prod but for testing the password may be more convenient. Wdyt?

1996fanrui: Do you mean we provide the `autoscaler.standalone.state-store.jdbc.password-env-variable` and `autoscaler.standalone.state-store.jdbc.password` together?

- For prod, we recommend users to `export STATE_STORE_JDBC_PWD=xxx`.
- For testing, using `autoscaler.standalone.state-store.jdbc.password` is convenient.

Right?

I'm afraid 2 options let users get confused here. 

Also, if we provide 2 options, we need to consider the priority. Which one will be used if users set 2 options together.

1996fanrui: About testing, if we only test `JdbcAutoScalerStateStore`, the password is easy to be set.

JdbcAutoScalerStateStore will work well regardless of whether the Autoscaler standalone uses  or the Kubernetes-operator uses if `JdbcAutoScalerStateStore` is tested properly. 

So test `JdbcAutoScalerStateStore` in Autoscaler standalone isn't necessary, right?

If so, we don't need the `autoscaler.standalone.state-store.jdbc.password` for now.  And we can add it if we really needs it in the future.

WDYT?

gyfora: okay, sounds good

1996fanrui:
Hi @gyfora @mxm , do you have any other comments?

mxm:
LGTM

gyfora (approved):

1996fanrui:
Thanks for the patient review, merging~

PR #742, opened 2024-01-02
[FLINK-33965][autoscaler] Introducing the AutoscalerStandaloneOptions to simplify options for autoscaler standalone

1996fanrui:

Currently, all configurations of autoscaler standalone are maintained in string key.

When autoscaler standalone has a little options, it's easy to maintain. However, I found it's hard to maintain when we add more options.

During I developing the JDBC autoscaler state store and control loop supports multiple thread. It will introduce more options.

## Brief change log

Introducing the AutoscalerStandaloneOptions to manage all options of autoscaler standalone. And output the doc for it.

- [FLINK-33965][autoscaler] Introducing the AutoscalerStandaloneOptions to simplify options for autoscaler standalone
- [FLINK-33965][autoscaler] Generate the configuration doc for autoscaler standalone

## Verifying this change

## Documentation

  - Generated the docs

### DIFF
--- a/flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/config/AutoscalerStandaloneOptions.java
+++ b/flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/config/AutoscalerStandaloneOptions.java
@@ -1,0 +1,56 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.autoscaler.standalone.config;
+
+import org.apache.flink.configuration.ConfigOption;
+import org.apache.flink.configuration.ConfigOptions;
+
+import java.time.Duration;
+
+/** Config options related to the autoscaler standalone module. */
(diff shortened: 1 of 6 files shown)


1996fanrui (commented):
Happy new year @mxm @gyfora ~

Would you mind helping take a look this PR in your free time? Thanks~

flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/config/AutoscalerStandaloneOptions.java
```diff
@@ -0,0 +1,60 @@
...
+
+import org.apache.flink.configuration.ConfigOption;
+import org.apache.flink.configuration.ConfigOptions;
+
+import java.time.Duration;
+
+/** Config options related to the autoscaler standalone module. */
+public class AutoscalerStandaloneOptions {
+
+    public static final String AUTOSCALER_CONF_PREFIX = "autoscaler.standalone.";
```
mxm, line 28: Are we intentionally using a different namespace here? The regular one would be `job.autoscaler`. I guess autoscaling will still happen at job level, so that namespace would work as well?

mxm:
> Happy new year @mxm @gyfora ~

Happy New Year @1996fanrui! 🎆 🚀

flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/config/AutoscalerStandaloneOptions.java
1996fanrui: Thanks to @mxm for the quick review!

The original autoscaler options work at job level, these configs only work for autoscaler standalone. And it's not job level. Such as: 
- `autoscaler.standalone.control-loop.interval` is the interval of autoscaler standalone control loop. It's one option of `autoscaler.standalone` instead of job. 
- In FLINK-33814 (control loop supports multiple thread), I  will add `autoscaler.standalone.control-loop.parallelism`, it's similar with `kubernetes.operator.reconcile.parallelism`. 

That's why I remove the job prefix. Looking forward to your opinion!

mxm (approved):

flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/config/AutoscalerStandaloneOptions.java
mxm: That makes sense. Thanks for clarifying.

PR #744, opened 2024-01-03
[FLINK-33814][autoscaler] Autoscaler Standalone control loop supports multiple threads

1996fanrui:

When the job list has a lot of jobs, single thread isn't enough. So Autoscaler Standalone control loop supports multiple thread is very useful for massive production, it's similar to `kubernetes.operator.reconcile.parallelism`.

## Brief change log

Introduce `autoscaler.standalone.control-loop.parallelism`.

## Verifying this change

  - Added `StandaloneAutoscalerExecutorTest#testScalingParallelism`

## Documentation

### DIFF
--- a/flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/StandaloneAutoscalerExecutor.java
+++ b/flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/StandaloneAutoscalerExecutor.java
@@ -21,6 +21,8 @@ import org.apache.flink.annotation.VisibleForTesting;
 import org.apache.flink.autoscaler.JobAutoScaler;
 import org.apache.flink.autoscaler.JobAutoScalerContext;
 import org.apache.flink.autoscaler.event.AutoScalerEventHandler;
+import org.apache.flink.configuration.Configuration;
+import org.apache.flink.util.concurrent.ExecutorThreadFactory;
 
 import org.apache.flink.shaded.guava30.com.google.common.util.concurrent.ThreadFactoryBuilder;
 
@@ -31,9 +33,17 @@ import javax.annotation.Nonnull;
 
 import java.io.Closeable;
 import java.time.Duration;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
+
+import static org.apache.flink.autoscaler.standalone.config.AutoscalerStandaloneOptions.CONTROL_LOOP_INTERVAL;
+import static org.apache.flink.autoscaler.standalone.config.AutoscalerStandaloneOptions.CONTROL_LOOP_PARALLELISM;
 
 /** The executor of the standalone autoscaler. */
 public class StandaloneAutoscalerExecutor<KEY, Context extends JobAutoScalerContext<KEY>>
@@ -48,22 +58,35 @@ public class StandaloneAutoscalerExecutor<KEY, Context extends JobAutoScalerCont
     private final AutoScalerEventHandler<KEY, Context> eventHandler;
     private final JobAutoScaler<KEY, Context> autoScaler;
     private final ScheduledExecutorService scheduledExecutorService;
+    private final ExecutorService scalingThreadPool;
+
+    /**
+     * Maintain a set of job keys that during scaling, it should be updated at {@link
+     * #scheduledExecutorService} thread.
+     */
+    private final Set<KEY> scalingJobKeys;
 
     public StandaloneAutoscalerExecutor(
-            @Nonnull Duration scalingInterval,
+            @Nonnull Configuration conf,
             @Nonnull JobListFetcher<KEY, Context> jobListFetcher,
             @Nonnull AutoScalerEventHandler<KEY, Context> eventHandler,
             @Nonnull JobAutoScaler<KEY, Context> autoScaler) {
-        this.scalingInterval = scalingInterval;
+        this.scalingInterval = conf.get(CONTROL_LOOP_INTERVAL);
         this.jobListFetcher = jobListFetcher;
         this.eventHandler = eventHandler;
         this.autoScaler = autoScaler;
         this.scheduledExecutorService =
                 Executors.newSingleThreadScheduledExecutor(
                         new ThreadFactoryBuilder()
-                                .setNameFormat("StandaloneAutoscalerControlLoop")
+                                .setNameFormat("autoscaler-standalone-control-loop")
                                 .setDaemon(false)
                                 .build());
+
+        int parallelism = conf.get(CONTROL_LOOP_PARALLELISM);
+        this.scalingThreadPool =
+                Executors.newFixedThreadPool(
+                        parallelism, new ExecutorThreadFactory("autoscaler-standalone-scaling"));
+        this.scalingJobKeys = new HashSet<>();
     }
 
     public void start() {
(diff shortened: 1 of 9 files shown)


1996fanrui (commented):
Hi @mxm , please help take a look this PR in your free time, thanks a lot~

This PR supports `autoscaler.standalone.control-loop.parallelism` for Autoscaler Standalone.

flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/config/AutoscalerStandaloneOptions.java
```diff
@@ -38,6 +38,12 @@ private static ConfigOptions.OptionBuilder autoscalerStandaloneConfig(String key
                     .withDeprecatedKeys("scalingInterval")
                     .withDescription("The interval of autoscaler standalone control loop.");
 
+    public static final ConfigOption<Integer> CONTROL_LOOP_PARALLELISM =
+            autoscalerStandaloneConfig("control-loop.parallelism")
+                    .intType()
+                    .defaultValue(100)
```
mxm, line 44: FYI, the default for the Kubernetes operator is 200 but those two do not have to be aligned.

flink-autoscaler/src/test/java/org/apache/flink/autoscaler/event/TestingEventCollector.java
```diff
@@ -42,7 +42,7 @@ public class TestingEventCollector<KEY, Context extends JobAutoScalerContext<KEY
     public final Map<String, Event<KEY, Context>> eventMap = new ConcurrentHashMap<>();
 
     @Override
-    public void handleEvent(
+    public synchronized void handleEvent(
```
mxm, line 45: Is there a way to make this method thread-safe without synchronizing on its entirety?

mxm: Maybe using `Collections.synchronizedList(..)`?

flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/StandaloneAutoscalerExecutor.java
```diff
@@ -48,22 +59,34 @@ public class StandaloneAutoscalerExecutor<KEY, Context extends JobAutoScalerCont
...
                                 .build());
+
+        int parallelism = conf.get(CONTROL_LOOP_PARALLELISM);
+        this.scalingThreadPool =
+                new ThreadPoolExecutor(
+                        parallelism,
+                        parallelism,
+                        0L,
+                        TimeUnit.MILLISECONDS,
+                        new LinkedBlockingQueue<>(parallelism * 4),
```
mxm, line 87: Why `*4`? Shouldn't `parallelism` be enough for the work queue?

flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/StandaloneAutoscalerExecutor.java
```diff
@@ -75,29 +98,38 @@ public void start() {
...
-                            null);
-                }
+                futures.add(scalingThreadPool.submit(() -> scalingSingleJob(jobContext)));
+            }
+            for (Future<?> future : futures) {
+                future.get();
             }
         } catch (Throwable e) {
             LOG.error("Error while fetch job list.", e);
         }
```
mxm, line 118:
```suggestion
        } catch (Throwable e) {
            LOG.error("Error while executing autoscaling.", e);
        }
```

Note that none of the individual scheduled futures have error handling here. I guess that is ok because the contract is that the autoscaler handles all exceptions, so this is just a safe-guard to return a better error message.

mxm (approved):
LGTM @1996fanrui!

flink-autoscaler/src/test/java/org/apache/flink/autoscaler/event/TestingEventCollector.java
1996fanrui: Some callers use the poll methods of LinkedList, these methods from Queue, so I refactor it to `public final Queue<Event<KEY, Context>> events = new LinkedBlockingQueue<>();`.

flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/StandaloneAutoscalerExecutor.java
1996fanrui: From current code, `scalingThreadPool` execute the `scalingSingleJob` logic, and `scalingSingleJob` catches all exceptions. So I think all exceptions of control loop thread are from `fetch job list`.

WDYT?

flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/StandaloneAutoscalerExecutor.java
1996fanrui: First of all, why this `scalingThreadPool` doesn't use unbounded `LinkedBlockingQueue`?

My thought is if all threads are busy and `LinkedBlockingQueue` has a lot of tasks, we can execute `task` in control loop scheduler thread. So I choose the `new ThreadPoolExecutor.CallerRunsPolicy()`.

> Why *4?

It's not strict. My thought is scheduler thread only work when `LinkedBlockingQueue` has a lot of tasks. So parallelism * 3 or > 3 makes sense.

-------------------------------------------

Of course, if you think `scalingThreadPool` already has a lot of thread, and we don't need thejcontrol loop scheduler thread to execute any tasks. We can let control loop scheduler thread only submit tasks, and wait for all tasks are finished. It means we don't let control loop scheduler thread executes tasks.

WDYT?

1996fanrui (commented):

flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/StandaloneAutoscalerExecutor.java
mxm: If you are confident about this, how about moving the try/catch to be around the `.fetch()` method?

flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/StandaloneAutoscalerExecutor.java
mxm: What is the benefit of the queue? If we are waiting on all tasks to finish, then we do not need the queue.

I think the current logic of waiting for all tasks to finish is a bit fragile. For example, when we are executing a scaling decision, we may block for a long time on savepointing which will halt the next scaling for all pipelines. We could alternatively not block all but only the specific job for the next scaling.

1996fanrui: > What is the benefit of the queue? If we are waiting on all tasks to finish, then we do not need the queue.

Sounds make sense, if we want to waiting, the queue isn't useful.

> I think the current logic of waiting for all tasks to finish is a bit fragile. For example, when we are executing a scaling decision, we may block for a long time on savepointing which will halt the next scaling for all pipelines. We could alternatively not block all but only the specific job for the next scaling.

Sorry, I didn't notice the `scalingSingleJob` will cost too many times. Currently, the scaling action only call the rescale api, so the thread won't be wait the savepointing.

But I think your concern is reasonable. We define the `ScalingRealizer` interface, it means users can custom define the action. It might cost long time.

--------------------------------------------------------


Based on your suggestion, I think the `control loop scheduler thread` cannot wait all scaling. When the scaling of one job is slow, it shouldn't affect other jobs.

Also, if a new scaling round is started, and the last scaling of some of jobs aren't finished. Should we ignore this round for these jobs? Some reasons might cause this issues:

- The scaling action is slow
- It has too many jobs, and the thread number isn't enough. So the last round always cannot be finished when the new round is started.

So, I think we should ignore unfinished jobs in the new round.

WDYT?

flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/StandaloneAutoscalerExecutor.java
1996fanrui: Updated.

flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/StandaloneAutoscalerExecutor.java
mxm: >Based on your suggestion, I think the control loop scheduler thread cannot wait all scaling. When the scaling of one job is slow, it shouldn't affect other jobs.

Yes, I agree.

>Also, if a new scaling round is started, and the last scaling of some of jobs aren't finished. Should we ignore this round for these jobs?

Yes, that would be great. We just wait until they are ready. In the Kubernetes operator, there is currently no timeout for single jobs but we could add one in the future if this poses a problem.

mxm (approved):

1996fanrui:
Thanks @mxm for the quick review and great suggestions!

CI is green, merging~

PR #746, opened 2024-01-04
[hotfix][autoscaler] Fix the StandaloneAutoscalerExecutorTest.testScaling fails occasionally due to race condition

1996fanrui:
Sorry for this fix!

After #744 is merged, I rebased it. During developing other features, I found `StandaloneAutoscalerExecutorTest.testScaling fails occasionally due to race condition`.

Here is my fork repo, it fails with :

https://github.com/1996fanrui/flink-kubernetes-operator/actions/runs/7405774151/job/20149240482#step:5:22232

```
Error:  Failures: 
Error:    StandaloneAutoscalerExecutorTest.testScaling:93 
Actual and expected should have same size but actual size is:
  1
while expected size is:
  2
Actual was:
  [org.apache.flink.autoscaler.event.TestingEventCollector$Event@6b580b88,
    org.apache.flink.autoscaler.event.TestingEventCollector$Event@6d91790b]
Expected was:
  [9fe5243cd23e0727710fcd0b4d423970, 9eec17388e32cb02ec205ac9f638e2fe]
```

From this log, we know the reason is race condition. The Actual was has 2 events, but when we check the actual size, it was one.

## Why it fails occasionally?

https://github.com/apache/flink-kubernetes-operator/pull/744 supports multiple threads for autoscaler standalone, and the scaling control loop thread doesn't wait for the actual scaling of all jobs.

So I introduced the `CountDownLatch` to let test wait all scaling are finished. But the `eventHandler.handleEvent` is called inside of the `scalingSingleJob`. We should call `countDownLatch.countDown();` after `scalingSingleJob`.

### DIFF
--- a/flink-autoscaler-standalone/src/test/java/org/apache/flink/autoscaler/standalone/StandaloneAutoscalerExecutorTest.java
+++ b/flink-autoscaler-standalone/src/test/java/org/apache/flink/autoscaler/standalone/StandaloneAutoscalerExecutorTest.java
@@ -93,7 +93,7 @@ class StandaloneAutoscalerExecutorTest {
                 }) {
 
             autoscalerExecutor.scaling();
-            // Wait for all scalings are finished.
+            // Wait for all scalings to go finished.
             countDownLatch.await();
 
             assertThat(actualScaleContexts).isEqualTo(jobList);
--- a/flink-autoscaler-standalone/src/test/java/org/apache/flink/autoscaler/standalone/StandaloneAutoscalerExecutorTest.java
+++ b/flink-autoscaler-standalone/src/test/java/org/apache/flink/autoscaler/standalone/StandaloneAutoscalerExecutorTest.java
@@ -50,10 +50,10 @@ class StandaloneAutoscalerExecutorTest {
     @ParameterizedTest
     @ValueSource(booleans = {true, false})
     void testScaling(boolean throwExceptionWhileScale) throws Exception {
-        JobAutoScalerContext<JobID> jobContext1 = createJobAutoScalerContext();
-        JobAutoScalerContext<JobID> jobContext2 = createJobAutoScalerContext();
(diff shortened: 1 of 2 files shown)


1996fanrui (commented):
Hi @RocMarshal , would you mind helping take a look this PR in your free time? Thanks~

This PR is fixing the StandaloneAutoscalerExecutorTest.testScaling fails occasionally due to race condition.

flink-autoscaler-standalone/src/test/java/org/apache/flink/autoscaler/standalone/StandaloneAutoscalerExecutorTest.java
```diff
@@ -62,30 +62,38 @@ void testScaling(boolean throwExceptionWhileScale) throws Exception {
...
+                    }
+                }) {
 
-                            @Override
-                            public void cleanup(JobID jobKey) {
-                                fail("Should be called.");
-                            }
-                        })) {
             autoscalerExecutor.scaling();
+            // Wait for all scalings are finished.
```
RocMarshal, line 96:
```suggestion
            // Wait for all scalings to go finished.
```

RocMarshal (approved):
Hi, @1996fanrui Thanks for your quick catch & fix.

LGTM +1.

And If you have a bit of concern on the stability of the fix, you could use `@RepeatedTest` annotations from repeated test cases locally to test whether this fix is stable.

IIUC, this fix is already sufficient to fix this issue.

1996fanrui:
> And If you have a bit of concern on the stability of the fix, you could use @RepeatedTest annotations from repeated test cases locally to test whether this fix is stable.

Thanks @RocMarshal for your review and suggestion! After this PR, I have already used `@RepeatedTest ` to check this test. It works well for now.

PR #747, opened 2024-01-04
[hotfix][docs] Add the -am when generate the docs

1996fanrui:
## Purpose 

Currently, `mvn package -Pgenerate-docs -pl flink-kubernetes-docs -nsu -DskipTests` is the command of generating docs. It only build the ` flink-kubernetes-docs` module without dependency modules.

Sometimes, the dependency modules isn't up to date in the local repository, the generated doc will be wrong.

## How to fix it?

Adding `-am`, `-am` means building all dependency modules of `flink-kubernetes-docs` as well when generating doc.

Note: The command of flink project[1] has `-am` as well.

[1] https://github.com/apache/flink/blob/8e8e2d649e98071bc15ee768cf65da3e96f255b4/flink-docs/README.md?plain=1#L44


mxm (approved):
Nice change! Thanks.

1996fanrui (commented):
Thanks @mxm for the quick review! Merging~

PR #748, opened 2024-01-04
[FLINK-33993] Fix misleading events for scaling effectiveness detection

mxm:
When the ineffective scaling decision feature is turned off, events are regenerated which look like this:

```
Skipping further scale up after ineffective previous scale up for 65c763af14a952c064c400d516c25529
```

This is misleading because no action will be taken. It is fair to inform users about ineffective scale up even when the feature is disabled but a different message should be printed to convey that no action will be taken.


1996fanrui (approved):
Thanks @mxm for the good change!

LGTM

mxm:
Thanks for the quick review Rui!

PR #751, opened 2024-01-08
[FLINK-33771] Add cluster capacity awareness to autoscaler

mxm:
To avoid starvation of pipelines when the Kubernetes cluster runs out of resources, new scaling attempts should be stopped when the no more additional pods can be scheduled for rescaling.

While Flink's ResourceRequirement API can prevent some of these cases, it requires using Flink 1.18 and an entirely different Flink scheduler. Extensive testing still has to be done with the new scheduler and the rescaling behavior. We woud hand off control over the rescale time to Flink which uses various parameters to control the exact scaling behavior.

For the config-based parallelism overrides, we have pretty good heuristics in the operator to check in Kubernetes for the approximate number of free cluster resources, the max cluster scaleup for the Cluster Autoscaler, and the required scaling costs. Having cluster resource information will also allow to implement fairness between all the autoscaled pipelines.

This PR adds ClusterResourceManager which which provides a view over the allocatable resources within a Kubernetes cluster and allows to simulate scheduling pods with a defined number of required resources.

The goal is to provide a good indicator for whether resources needed for autoscaling are going to be available. This is achieved by pulling the node resource usage from the Kubernetes cluster at a regular configurable interval, after which we use this data to simulate adding / removing resources (pods). Note that this is merely a (pretty good) heuristic because the Kubernetes scheduler has the final saying. However, we prevent 99% of the scenarios after pipeline outages which can lead to massive scale up where all pipelines may be scaled up at the same time and exhaust the number of available resources.

The simulation can run on a fixed set of Kubernetes nodes. Additionally, if we detect that the cluster is using the Kubernetes Cluster Autoscaler, we will use this data to extrapolate the number of nodes to the maximum defined nodes in the autoscaler configuration.  We currently track CPU and memory. Ephemeral storage is missing because there is no easy way to get node statics on free storage.


1996fanrui (commented):
Hi @mxm  , thanks for the improvement!

Before reviewing, I have 2 question: 

## 1. Why does autoscaler need to aware the cluster capacity? 

It only works for flink version before 1.18 or without adaptive scheduler, right?


As I understand, the Adaptive Scheduler supports set lowerBound and upperBound for each task vertex. And autoscaler can set lowerBound=1 and upperBound = expectedParalleslism.

When kubernetes or yarn have enough resources, flink job will run task with expectedParalleslism
When the available resource < expectedParalleslism, flink job will run task with available resource.


For kubernetes or yarn side, we can enable the preemption strategy, it means high priority flink jobs will preempt the resources of low priority flink jobs. It will ensure the high priority flink jobs run with expectedParalleslism, and low priority flink jobs run with available resource when total resources are not enough.

## 2. Is there any problems when multiple jobs are scaling at the same time?

Assuming we have 2 jobs: jobA and jobB. The available resource is 10 CPU before their scaling. 

The JobA needs 10 CPU for this scaling, and the JobB needs 10 CPU for this scaling. During the resource checking phase, both of them think resource is enough. But during start job, the resource isn't enough for them. Because they need 20 CPUs.

Or is it possible that other services outside of `flink-kubernetes-operator` are also submitting jobs to the same resource pool? If yes, there will be similar problems. 

To put it simply: there are enough resources during the check phase, but insufficient resources when actually applying for resources.

Looking forward to your feedback, thanks~

mxm:
Hey @1996fanrui! I'll answer your questions in a bit. Just rebasing to the latest master changes.

mxm:
> ## 1. Why does autoscaler need to aware the cluster capacity?

The most pressing issue is to solve job starvation in a scenario where hundreds of pipelines scale up at the same time. A typical scenario when this happens is an outage after which all pipelines need to catch up again.

>It only works for flink version before 1.18 or without adaptive scheduler, right? 

It currently works regardless of the Flink version. Even with the resource requirements API, we wouldn't want to scale any pipelines unless their resource requirements can be met. Otherwise we risk rescaling too often or scaling to an undesirable parallelism configuration. That said, many users are still on Flink 1.16 for the foreseeable time.

> As I understand, the Adaptive Scheduler supports set lowerBound and upperBound for each task vertex. And autoscaler can set lowerBound=1 and upperBound = expectedParalleslism.

That is true. Setting the lower bound to 1 is kind of tricky because it means that the job can rescale to a lower expected parallelism while giving up the current parallelism. Ideally, the autoscaler controls the full lifecycle of the parallelism configuration.

> When kubernetes or yarn have enough resources, flink job will run task with expectedParalleslism When the available resource < expectedParalleslism, flink job will run task with available resource.

We may not want that though. We want to control the scaling because we have certain SLOs that we want to guarantee when we scale.

> ## 2. Is there any problems when multiple jobs are scaling at the same time?

The relevant code paths are thread-safe. Only one pipeline can run through the capacity check at a time. The actual check is very quick.

> Assuming we have 2 jobs: jobA and jobB. The available resource is 10 CPU before their scaling.
> 
> The JobA needs 10 CPU for this scaling, and the JobB needs 10 CPU for this scaling. During the resource checking phase, both of them think resource is enough. But during start job, the resource isn't enough for them. Because they need 20 CPUs.

That cannot happen because whichever pipeline scales first will reserve the available resources and they won't be available afterwards until they are freed again. There is a reservation system to ensure this.

> Or is it possible that other services outside of `flink-kubernetes-operator` are also submitting jobs to the same resource pool? If yes, there will be similar problems.

Yes, that can happen. It is a heuristic. New jobs are rarely started and don't typically coincident with an outage scenario. But we will detect the new resource usage as soon as we refresh the resource view at a configurable interval. We could also do it on every rescaling but I wanted to limit the number of API calls.

1996fanrui (commented):
Overall looks good to me, I left some minor comment, please tatke a look in your free time, thanks~

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
```diff
@@ -247,6 +254,69 @@ private boolean isJobUnderMemoryPressure(
...
+                        (int) entry.getValue().get(ScalingMetric.PARALLELISM).getCurrent());
+            }
+        }
+
+        double numTaskSlotsUsed = globalMetrics.get(ScalingMetric.NUM_TASK_SLOTS_USED).getCurrent();
+
+        final int numTaskSlotsAfterRescale;
+        if (oldParallelismSum == numTaskSlotsUsed) {
+            // Slot sharing activated
+            numTaskSlotsAfterRescale = newParallelisms.values().stream().reduce(0, Integer::sum);
```
1996fanrui, line 304: It could happen when resource is not enough and using adaptive scheduler, right?

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
```diff
@@ -247,6 +254,69 @@ private boolean isJobUnderMemoryPressure(
...
+        } else {
+            // Assuming slot sharing is not activated
+            numTaskSlotsAfterRescale = newParallelisms.values().stream().reduce(0, Integer::max);
```
1996fanrui, line 307: max and sum reversed?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/FlinkOperator.java
```diff
@@ -165,7 +167,12 @@ void registerDeploymentController() {
         var metricManager =
                 MetricManager.createFlinkDeploymentMetricManager(baseConfig, metricGroup);
         var statusRecorder = StatusRecorder.create(client, metricManager, listeners);
-        var autoscaler = AutoscalerFactory.create(client, eventRecorder);
+        Duration refreshClusterViewInterval =
```
gyfora, line 170: Can we use `var` for consistency?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/resources/ClusterResourceManager.java
```diff
@@ -0,0 +1,282 @@
...
+        }
+
+        if (shouldRefreshView(clusterResourceView, refreshInterval)) {
+            try {
+                clusterResourceView = createResourceView(kubernetesClient);
+            } catch (KubernetesClientException e) {
+                if (e.getCode() == 403) {
+                    LOG.warn(
+                            "No permission to retrieve node resource usage. Resource check disabled.");
+                    return true;
```
gyfora, line 101: Should we set a field and not try again and again?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/resources/ClusterResourceManager.java
```diff
@@ -0,0 +1,282 @@
...
+                }
+                throw e;
+            }
+        }
+
+        if (newInstances <= currentInstances) {
+            LOG.debug("Fewer or same amount of resources used after scaling.");
+            return true;
+        }
```
gyfora, line 110: This should be moved in front of everything

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/FlinkOperator.java
```diff
@@ -165,7 +167,12 @@ void registerDeploymentController() {
...
+                configManager
+                        .getDefaultConfig()
+                        .get(KubernetesOperatorConfigOptions.REFRESH_CLUSTER_RESOURCE_VIEW);
+        var clusterResourceManager = new ClusterResourceManager(refreshClusterViewInterval, client);
```
gyfora, line 174: We should add a method to create the `ClusterResourceManager` and we should probably have a config to disable it by default. This is only applicable in certain Kubernetes envs which are actually resource constrained only by node capacity and not by other quotas

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
gyfora: It could be `oldParallelism >= numTaskSlotsUsed`

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
```diff
@@ -247,6 +254,69 @@ private boolean isJobUnderMemoryPressure(
...
+        }
+
+        var numTotalTaskSlots = globalMetrics.get(ScalingMetric.NUM_TOTAL_TASK_SLOTS).getCurrent();
+        int currentNumTms = (int) globalMetrics.get(ScalingMetric.NUM_TASK_MANAGERS).getCurrent();
+
+        var numSlotsPerTm = numTotalTaskSlots / currentNumTms;
```
gyfora, line 313: We could also get this directly from the config, then we only need to query the slots used metrics.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/resources/ClusterResourceManager.java
```diff
@@ -0,0 +1,282 @@
...
+     * full-blown scheduler. We may return false-negatives, i.e. we may indicate scheduling is not
+     * possible when it actually is. This is still better than false positives where we suffer from
+     * downtime due to non-schedulable pods.
+     */
+    private static boolean trySchedule(
+            ClusterResourceView resourceView,
+            int currentInstances,
+            int newInstances,
+            double cpuPerInstance,
+            double memoryPerInstance) {
```
gyfora, line 131: Instead of having `current` and `newInstances` can we simply pass the diff / increment? That would make for a simpler logic and be more intuitive. Such as checking whether we can schedule 2 more instances

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/resources/ClusterResourceManager.java
```diff
@@ -0,0 +1,282 @@
...
+
+        resourceView.cancelPending();
+
+        for (int i = 0; i < currentInstances; i++) {
+            resourceView.release(cpuPerInstance, memoryPerInstance);
```
gyfora, line 136: If we only check for the increment in instances we wouldn't need the whole release logic right?

flink-autoscaler/src/test/java/org/apache/flink/autoscaler/RestApiMetricsCollectorTest.java
```diff
@@ -122,6 +126,56 @@ CompletableFuture<P> sendRequest(
...
+                Map<FlinkMetric, Metric> flinkMetricMetricMap =
+                        collector.queryJmMetrics(
+                                client,
+                                Map.of(
+                                        "numRegisteredTaskManagers", FlinkMetric.NUM_TASK_MANAGERS,
+                                        "taskSlotsTotal", FlinkMetric.NUM_TASK_SLOTS_TOTAL,
+                                        "taskSlotsAvailable",
+                                                FlinkMetric.NUM_TASK_SLOTS_AVAILABLE));
+                try {
+                    System.out.println(flinkMetricMetricMap);
```
gyfora, line 157: remove sys out

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/resources/ResourceCheck.java
```diff
@@ -0,0 +1,29 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.autoscaler.resources;
+
+/** An interface for checking the available capacity of the underlying resources. */
+public interface ResourceCheck {
+
+    /** Simulates scheduling the provided number of resources. */
```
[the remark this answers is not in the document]
mxm: Sure. Done.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/resources/ResourceCheck.java
```diff
@@ -0,0 +1,29 @@
...
+    boolean trySchedule(
+            int currentInstances,
+            int newInstances,
```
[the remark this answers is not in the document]
mxm: I think it does not matter if the pods are TaskManagers or not. In the concrete implementation, I use task managers though. Added a reference in the JavaDoc.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/resources/ResourceRequirements.java
```diff
@@ -0,0 +1,32 @@
...
+/** Interface to extract resource requirements. */
+public interface ResourceRequirements {
+
+    /** The required number of CPU for TaskManager nodes, if available, otherwise 0. */
+    default double getTaskManagerCpu() {
+        return 0;
+    }
+
+    /** The required number of CPU for TaskManager nodes, if available, otherwise 0. */
+    default double getTaskManagerMemory() {
```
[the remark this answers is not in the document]
mxm: Ok, using MemorySize. Removed interface and converted to use fields in the context. Drawback is that this affects all existing code which uses the constructor but I guess that is a fair tradeoff.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/RestApiMetricsCollector.java
```diff
@@ -103,6 +106,42 @@ protected Map<FlinkMetric, AggregatedMetric> queryAggregatedVertexMetrics(
...
     }
 
+    @Override
+    @SneakyThrows
+    protected Map<FlinkMetric, Metric> queryJmMetrics(Context ctx) {
+        Map<String, FlinkMetric> metrics =
+                Map.of(
+                        "numRegisteredTaskManagers", FlinkMetric.NUM_TASK_MANAGERS,
+                        "taskSlotsTotal", FlinkMetric.NUM_TASK_SLOTS_TOTAL,
+                        "taskSlotsAvailable", FlinkMetric.NUM_TASK_SLOTS_AVAILABLE);
```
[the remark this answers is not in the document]
mxm: They still use a predicate (`equals`). I think we can refactor this independently.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
mxm: This branch detects when slot sharing is activated, i.e. the max parallelism is identical to the number of task slots used.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
mxm: Yes! Good catch!

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/FlinkOperator.java
mxm: Sure. Done.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/resources/ClusterResourceManager.java
mxm: We can, although that would mean that fixing permission issues would require an operator restart. I don't think one API call in 5 minutes is anything to worry about.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/resources/ClusterResourceManager.java
mxm: That's what I also thought but moving this in front of everything will corrupt the resource view because we will never remove resources when we scale down.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/FlinkOperator.java
mxm: Yes, method created. Changed the default to be disabled. The feature can be extended as more contributors work on this.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
mxm: Sure, done.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/resources/ClusterResourceManager.java
mxm: I prefer not to because the scheduling is supposed to remove all instances from the internal resource view and then re-acquire them. Similarly to how the scaling works. Otherwise we end up with fragmentation issues and we would also be unable to recover from scenarios where we are overbooked. Providing more information will also allow us to  dynamically change the scheduling behavior.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/resources/ClusterResourceManager.java
mxm: We do need the release logic because otherwise we would never be able to remove resources from the internal view for scale downs. That means other pipelines might not get scheduled because the accounting is incorrect.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/RestApiMetricsCollector.java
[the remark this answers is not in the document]
mxm: Sure, but we will need taskSlotsAvailable and taskSlotsTotal to compute taskSlotsUsed. I've dropped the number of registered task managers.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/resources/ClusterResourceManager.java
mxm: True, right now this assumes a single operator per cluster.

Are you suggesting we should drop the precise bookkeeping and accept that we will yield false-negatives (block scaling) until we refresh our resource view? I guess we could do that.

gyfora: I'm a bit torn on this. With multiple operators and even without that when other compute resources also use the cluster the bookkeeping breaks down and produces false positives and false negatives anyways.

So it may be better to opt for a simpler approach and accept these limitations. What do you think?

mxm: I agree, this is tricky. If we removed the `release` function and used a diff instead of absolute number of instances, that would only be marginally simpler. While it is true that false-negatives and false-positives are possible, in a single operator scenario, we would accept many false-negatives which we could prevent. 

I understand your argument about multiple operators, but in its current state, even the simpler version without releasing resources, wouldn't work correctly for this scenario. The only proper solution would be to make the operator aware of the quotas for the namespace it manages, or share the resource allocations via some communication channel (e.g. config map).

mxm: Another safety measurement I had thought about, is to switch to a different mode when we are near max capacity. In this case, we would potentially go to the API for every scaling request. Of course with a rate limit.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/resources/ResourceCheck.java
```diff
@@ -0,0 +1,39 @@
...
+public interface ResourceCheck {
+
+    /**
+     * Simulates scheduling the provided number of TaskManager instances.
+     *
+     * @param currentInstances The current number of instances.
+     * @param newInstances The new number of instances.
+     * @param cpuPerInstance The number of CPU per instances.
+     * @param memoryPerInstance The amount of memory in bytes per instances.
+     * @return true if a scheduling configuration was found, false otherwise.
```
1996fanrui, line 32:
```suggestion
     * @param memoryPerInstance The total memory size per instances.
     * @return true if a scheduling resource is available, false otherwise.
```

How about it?

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
1996fanrui: I mean that the `Slot sharing` may be actived when `oldParallelismSum != numTaskSlotsUsed`.

For example, when resource is not enough, the numTaskSlotsUsed will be less than oldParallelismSum.

mxm: True, I think it should be fixed now.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/resources/ResourceCheck.java
mxm: That's why I prefer not to add any JavaDocs beyond a description. They become outdated very quickly. I applied the changes you suggested for `memoryPerInstance`. I kept the return parameter because it sounded correct to me.

1996fanrui: Thank you for the update ❤️

gyfora (approved):

PR #755, opened 2024-01-15
[FLINK-33803] Set observedGeneration at end of reconciliation

justin-chen:
https://issues.apache.org/jira/browse/FLINK-33803

This pull requests adds the `observedGeneration` field to the `FlinkDeployment` CRD status. This field is used to persist the generation that was last acted on by the operator. At the end of a successful reconciliation (`reconcile`, `RecoveryFailureException`, `DeploymentFailedException` ), the operator sets the `observedGeneration` field to the current  `generation` number (from the CRD metadata).

## Brief change log
  - Operator sets `observedGeneration` field within `FlinkDeploymentStatus` after a completed reconciliation

## Verifying this change

  - Test setting `observedGeneration` based on `generation` metadata taken from `lastReconciledSpec`

## Documentation

  - Does this pull request introduce a new feature? No
  - If yes, how is the feature documented? Docs

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
@@ -420,10 +420,7 @@ public class ReconciliationUtils {
                 // running
                 deployment.getSpec().getJob().setState(JobState.RUNNING);
             }
-            deployment
-                    .getMetadata()
-                    .setGeneration(
-                            lastReconciledSpecWithMeta.getMeta().getMetadata().getGeneration());
+            deployment.getMetadata().setGeneration(status.getObservedGeneration());
             return true;
         }
     }
@@ -469,16 +466,9 @@ public class ReconciliationUtils {
      * @return The spec generation for the upgrade.
      */
     public static Long getUpgradeTargetGeneration(AbstractFlinkResource<?, ?> resource) {
-        var lastSpecWithMeta =
-                resource.getStatus()
-                        .getReconciliationStatus()
-                        .deserializeLastReconciledSpecWithMeta();
+        var observedGeneration = resource.getStatus().getObservedGeneration();
 
-        if (lastSpecWithMeta.getMeta() == null) {
-            return -1L;
-        }
-
-        return lastSpecWithMeta.getMeta().getMetadata().getGeneration();
+        return observedGeneration == null ? -1L : observedGeneration;
     }
 
     /**
@@ -569,5 +559,6 @@ public class ReconciliationUtils {
 
         reconciliationStatus.setLastReconciledSpec(
                 SpecUtils.writeSpecWithMeta(lastSpecWithMeta.getSpec(), newMeta));
+        resource.getStatus().setObservedGeneration(resource.getMetadata().getGeneration());
     }
 }
--- a/docs/content/docs/custom-resource/reference.md
+++ b/docs/content/docs/custom-resource/reference.md
@@ -243,6 +243,7 @@ This page serves as a full reference for FlinkDeployment custom resource definit
 | ----------| ---- | ---- |
 | jobStatus | org.apache.flink.kubernetes.operator.api.status.JobStatus | Last observed status of the Flink job on Application/Session cluster. |
 | error | java.lang.String | Error information about the FlinkDeployment/FlinkSessionJob. |
+| observedGeneration | java.lang.Long | Last observed generation of the FlinkDeployment/FlinkSessionJob. |
 | lifecycleState | org.apache.flink.kubernetes.operator.api.lifecycle.ResourceLifecycleState | Lifecycle state of the Flink resource (including being rolled back, failed etc.). |
 | clusterInfo | java.util.Map<java.lang.String,java.lang.String> | Information from running clusters. |
 | jobManagerDeploymentStatus | org.apache.flink.kubernetes.operator.api.status.JobManagerDeploymentStatus | Last observed status of the JobManager deployment. |
@@ -270,6 +271,7 @@ This page serves as a full reference for FlinkDeployment custom resource definit
 | ----------| ---- | ---- |
 | jobStatus | org.apache.flink.kubernetes.operator.api.status.JobStatus | Last observed status of the Flink job on Application/Session cluster. |
 | error | java.lang.String | Error information about the FlinkDeployment/FlinkSessionJob. |
+| observedGeneration | java.lang.Long | Last observed generation of the FlinkDeployment/FlinkSessionJob. |
 | lifecycleState | org.apache.flink.kubernetes.operator.api.lifecycle.ResourceLifecycleState | Lifecycle state of the Flink resource (including being rolled back, failed etc.). |
 | reconciliationStatus | org.apache.flink.kubernetes.operator.api.status.FlinkSessionJobReconciliationStatus | Status of the last reconcile operation. |
 
--- a/flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/CommonStatus.java
+++ b/flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/CommonStatus.java
@@ -45,6 +45,9 @@ public abstract class CommonStatus<SPEC extends AbstractFlinkSpec> {
     /** Error information about the FlinkDeployment/FlinkSessionJob. */
     private String error;
 
+    /** Last observed generation of the FlinkDeployment/FlinkSessionJob. */
+    private Long observedGeneration;
+
     /** Lifecycle state of the Flink resource (including being rolled back, failed etc.). */
     @PrinterColumn(name = "Lifecycle State")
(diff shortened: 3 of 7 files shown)


flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/FlinkDeploymentStatus.java
```diff
@@ -55,4 +55,7 @@ public class FlinkDeploymentStatus extends CommonStatus<FlinkDeploymentSpec> {
 
     /** Information about the TaskManagers for the scale subresource. */
     private TaskManagerInfo taskManager;
+
+    /** Last observed generation of the FlinkDeployment. */
+    private Long observedGeneration;
```
mxm, line 60: We do have this information available already, via the `lastReconciledSpec` field. Not sure whether we need to expose this as a top-level field.

CC @gyfora

gyfora: Exactly as @mxm said. 

Also the field inside the last reconciled spec has a slightly different and more correct semantics. I think what you did here (update the observed generation) doesn't make too much practical sense as the operator may fail before even seeing the spec. 

I see the value of having a top level `observedGeneration` field but it should replace the current field hidden inside the lastReconciledSpec with the same semantics.

ryanvanhuuksloot: >We do have this information available already, via the lastReconciledSpec field. Not sure whether we need to expose this as a top-level field.

The idea is that the Kubernetes spec expects that this status field exists in this particular place. This type of PR allows for 3rd party tools to work well with the Flink Operator. (ie. our build tools internally now can communicate with the operator which is 🙏)

However, agreed with Gyula on how it should replace semantically what exists today.

justin-chen: Thanks for the guidance - I've added `observedGeneration` to `CommonStatus` class and set it alongside the `lastReconciledSpec`. I've left the `generation` within the `lastReconciledSpec` intact in case there are existing client dependencies on it. Let me know about any additional changes to be made

gyfora: @justin-chen I would prefer to replace the implementation and call it out in the release notes instead. That way we end up with a cleaner codebase.

Any concerns from your end @mxm ?

mxm: I would prefer to keep the field in lastReconciledSpec to avoid downstream breakage. At least for the next release.

gyfora: In that case we should set the new field next to where the old one is set so that the old logic is easy to remove later .

gyfora: And we should add a note of deprecation and a jira to track it. Although we don’t have any explicit guarantees here for the status especially for contents of json fields

justin-chen: > we should set the new field next to where the old one is set so that the old logic is easy to remove later

@gyfora Not sure I understand, is the intention to duplicate the `observedGeneration` such that it's available via both `status.observedGeneration` (we need it at this top level) and `status.reconciliationStatus.lastReconciledSpec.resource_metadata`?

Given the status currently looks like this (with my current PR):
```
status:
    ...
    observedGeneration: 123
    reconciliationStatus:
    lastReconciledSpec: '{
        "spec": {
        ...
        },
        "resource_metadata": {
            "apiVersion": "flink.apache.org/v1beta1",
            "metadata": {   // io.fabric8.kubernetes.api.model.ObjectMeta
                "generation": 123
            },
            "firstDeployment": false
        }
    }'
    lastStableSpec: '{...}'
    reconciliationTimestamp:
    state: DEPLOYED
```

gyfora: I think you misunderstand. Let's leave the `generation` field in the lastReconciledSpec where it is now for backward compatibility. 

However instead of getting it from there and setting it in the top level status, set it in the top level status in the same step where we set in the lastReconciledSpec. That way the next person doesn't have to rewrite the entire logic to remove the deprecated meta field from lastReconciledSpec.

mxm: In essence, the current code already reflected the desired functionality, we just need to move the code to set the top-level generation id to where we initially set it. I hope that's easily doable. If not, let us know.

justin-chen: 👍 I have updated the PR to set `status.observedGeneration` using the [same source of truth](https://github.com/apache/flink-kubernetes-operator/blob/main/flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/reconciler/ReconciliationMetadata.java#L44) as `status.reconciliationStatus.lastReconciledSpec.resource_metadata.generation`, such that we can later remove the latter field without the `observedGeneration` depending on it.

The new field is set in `updateStatusForSpecReconciliation`  method alongside the `lastReconciledSpec`.

mxm (approved):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
```diff
@@ -126,6 +126,9 @@ private static <SPEC extends AbstractFlinkSpec> void updateStatusForSpecReconcil
         status.setError(null);
         reconciliationStatus.setReconciliationTimestamp(clock.instant().toEpochMilli());
 
+        // Set observedGeneration
+        status.setObservedGeneration(target.getMetadata().getGeneration());
```
gyfora, line 130: I still don't think that this is consistent with the current content of the lastReconciledSpec metadata. It's not about setting it to the same value (source of truth) it's also about setting it at the exact same time.

Also we should remove internal useges of the lastReconciledSpec.meta.generation and replace it with this one. We have tests for the behaviour so that would validate it.

gyfora (changes requested):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
justin-chen: @gyfora I see, thanks for the review. I moved the setting of `observedGeneration` to be at the same time/place as `lastReconciledSpec.meta.generation`, and replaced the internal usage of the latter in two locations. Tests pass, however please advise if there is somewhere I may have overlooked.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/reconciler/ReconciliationMetadata.java
```diff
@@ -42,6 +42,7 @@ public class ReconciliationMetadata {
     public static ReconciliationMetadata from(AbstractFlinkResource<?, ?> resource) {
         ObjectMeta metadata = new ObjectMeta();
         metadata.setGeneration(resource.getMetadata().getGeneration());
+        resource.getStatus().setObservedGeneration(resource.getMetadata().getGeneration());
```
gyfora, line 45: I would move this code to:
`ReconciliationStatus#serializeAndSetLastReconciledSpec` and `ReconciliationUtils#updateReconciliationMetadata` that way it's a bit more intuitive when the setting of the generation happens. It's a bit unexpected that a constructor method for `ReconciliationMetadata` mutates the resources here.

I know that my suggestion will actually duplicate the setting of the observedGeneration in the status but I think it will be cleaner once we remove the now deprecated meta generation field :)

gyfora (commented):
The change looks good overall and thanks for updating the existing usage 👍 

I added a small comment to where we should set the generation so that it's a bit more intuitive when reading the code but that's an easy change.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/reconciler/ReconciliationMetadata.java
justin-chen: Done!

gyfora:
Please generate the docs via 'mvn clean install -DskipTests -Pgenerate-docs'

mxm:
This looks good. Any further comments @gyfora?

gyfora:
🚢

PR #759, opened 2024-01-22
[FLINK-34178][autoscaler] Fix the bug that observed scaling restart time is always great than `stabilization.interval`

1996fanrui:

The ScalingTracking of autoscaler is wrong, it's always greater than AutoScalerOptions#STABILIZATION_INTERVAL.

### Reason

When flink job isStabilizing, ScalingMetricCollector#updateMetrics will return a empty metric history. In the JobAutoScalerImpl#runScalingLogic method, if `collectedMetrics.getMetricHistory().isEmpty()` , we don't update the ScalingTracking.

The default value of AutoScalerOptions#STABILIZATION_INTERVAL is 5 min, so the restartTime is always greater than 5 min.

However, it's quick when we use rescale api.

## Brief change log

Update the scaling tracking even if the job is stabilizing.

## Verifying this change

Adding the test later.

## Documentation

### DIFF
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
@@ -160,14 +160,23 @@ public class JobAutoScalerImpl<KEY, Context extends JobAutoScalerContext<KEY>>
         var collectedMetrics = metricsCollector.updateMetrics(ctx, stateStore);
         var jobTopology = collectedMetrics.getJobTopology();
 
+        var now = clock.instant();
+        var scalingTracking = getTrimmedScalingTracking(stateStore, ctx, now);
+        var scalingHistory = getTrimmedScalingHistory(stateStore, ctx, now);
+        // A scaling tracking without an end time gets created whenever a scaling decision is
+        // applied. Here, we record the end time for it (runScalingLogic is only called when the job
+        // transitions back into the RUNNING state).
+        if (scalingTracking.recordRestartDurationIfTrackedAndParallelismMatches(
+                collectedMetrics.getJobRunningTs(), jobTopology, scalingHistory)) {
+            stateStore.storeScalingTracking(ctx, scalingTracking);
+        }
+
         if (collectedMetrics.getMetricHistory().isEmpty()) {
             return;
         }
         LOG.debug("Collected metrics: {}", collectedMetrics);
 
-        var now = clock.instant();
         // Scaling tracking data contains previous restart times that are taken into account
-        var scalingTracking = getTrimmedScalingTracking(stateStore, ctx, now);
         var restartTime = scalingTracking.getMaxRestartTimeOrDefault(ctx.getConfiguration());
         var evaluatedMetrics =
                 evaluator.evaluate(ctx.getConfiguration(), collectedMetrics, restartTime);
@@ -178,15 +187,6 @@ public class JobAutoScalerImpl<KEY, Context extends JobAutoScalerContext<KEY>>
         autoscalerMetrics.registerScalingMetrics(
                 jobTopology.getVerticesInTopologicalOrder(),
                 () -> lastEvaluatedMetrics.get(ctx.getJobKey()));
-
-        var scalingHistory = getTrimmedScalingHistory(stateStore, ctx, now);
-        // A scaling tracking without an end time gets created whenever a scaling decision is
-        // applied. Here, we record the end time for it (runScalingLogic is only called when the job
-        // transitions back into the RUNNING state).
-        if (scalingTracking.recordRestartDurationIfTrackedAndParallelismMatches(
-                now, jobTopology, scalingHistory)) {
-            stateStore.storeScalingTracking(ctx, scalingTracking);
-        }
 
         if (!collectedMetrics.isFullyCollected()) {
             // We have done an upfront evaluation, but we are not ready for scaling.
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java
@@ -149,10 +149,10 @@ public abstract class ScalingMetricCollector<KEY, Context extends JobAutoScalerC
         if (isStabilizing) {
             LOG.info("Stabilizing until {}", readable(stableTime));
             stateStore.storeCollectedMetrics(ctx, metricHistory);
-            return new CollectedMetricHistory(topology, Collections.emptySortedMap());
+            return new CollectedMetricHistory(topology, Collections.emptySortedMap(), jobRunningTs);
         }
 
-        var collectedMetrics = new CollectedMetricHistory(topology, metricHistory);
+        var collectedMetrics = new CollectedMetricHistory(topology, metricHistory, jobRunningTs);
         if (now.isBefore(windowFullTime)) {
             LOG.info("Metric window not full until {}", readable(windowFullTime));
         } else {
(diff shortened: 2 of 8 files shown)


flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
```diff
@@ -160,14 +160,23 @@ private void runScalingLogic(Context ctx, AutoscalerFlinkMetrics autoscalerMetri
         var collectedMetrics = metricsCollector.updateMetrics(ctx, stateStore);
         var jobTopology = collectedMetrics.getJobTopology();
 
+        var now = clock.instant();
```
1996fanrui, line 163: Hi @afedulov ,

We are using `now` as the `endTime` of ScalingRecord. The `kubernetes.operator.reconcile.interval` is 1 min by default and  `scaling` is called in each reconcile. 

So, the restartTime will be greater than `1 min` even if the scaling is very quick.

I'm not sure should we get the endTime from the `JobUpdateTs`? Currently, `stabilization.interval` checks the `JobUpdateTs`.

Looking forward to your feedback, thanks~

mxm: Reconciliation is triggered at least once per minute, but actual reconciliation time is more like 10-20 seconds. This is due cluster events also triggering the reconciliation loop. However, I think you are right, we should use the job update time and not the current time. Let's wait what Alex says.

afedulov: @mxm I just checked and unfortunately our assumption does not seem to hold true with that regard. The loop is not triggered when the job state changes to `RUNNING`. I tried setting a higher reconciliation interval and it directly is reflected in the recorded durations.

afedulov:
Hi @1996fanrui ,
thanks for catching this, I did not realize that "history" only ever spans for one rescale cycle. I just checked that your change works as expected and removes the erroneous lower bound of the `AutoScalerOptions#STABILIZATION_INTERVAL` in rescaling durations.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
mxm: >I just checked and, unfortunately, our assumption does not seem to hold true with that regard. The loop is not triggered when the job state changes to RUNNING. I tried setting a higher reconciliation interval and it directly is reflected in the recorded durations.

There is no strict guarantee how often we are being called, but in my tests I saw 10-20 seconds due to cluster events arriving. In any case, the main issue here is not the reconciliation interval but that we are using the current instant instead of the job update time. Let's address this as proposed by Rui and we should be good to go. This feature might then be robust enough to be enabled out of the box.

mxm:
> Hi @1996fanrui , thanks for catching this, I did not realize that "history" only ever spans for one rescale cycle. 

There is the metrics history and the scaling history. The former is only preserved for the current metric window (to tolerate restarts during the metric window). The latter is stored according to age / size.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java
```diff
@@ -181,9 +186,12 @@ private static Instant getWindowFullTime(
...
 
     @VisibleForTesting
-    protected Instant getJobUpdateTs(JobDetailsInfo jobDetailsInfo) {
-        return Instant.ofEpochMilli(
-                jobDetailsInfo.getTimestamps().values().stream().max(Long::compare).get());
+    protected Instant getJobSwitchToRunningTs(JobDetailsInfo jobDetailsInfo) {
+        final Map<JobStatus, Long> timestamps = jobDetailsInfo.getTimestamps();
+
+        final Long runningTs = timestamps.get(JobStatus.RUNNING);
+        checkState(runningTs != null, "Unable to find when the job was switched to RUNNING.");
```
1996fanrui, line 193: Checking `JobStatus.RUNNING` exists here. If it doesn't exist, throw exception directly and exit the current scaling.

mxm: Generally, this code will only ever be called when the job is in running mode, see: https://github.com/apache/flink-kubernetes-operator/blob/6be277cd0c7421f4822c49296198f4a34b2cd721/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java#L96

We can add this invariant here but it would only defend against a Byzantine behavior.

mxm: The old way of getting the "jobUpdateTs" was correct but maybe this one will defend us in case of refactors.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java
```diff
@@ -104,17 +106,18 @@ public CollectedMetricHistory updateMetrics(
 
         var jobDetailsInfo =
                 getJobDetailsInfo(ctx, conf.get(AutoScalerOptions.FLINK_CLIENT_TIMEOUT));
-        var jobUpdateTs = getJobUpdateTs(jobDetailsInfo);
+        var jobSwitchToRunningTs = getJobSwitchToRunningTs(jobDetailsInfo);
```
mxm, line 109: Can we keep the old name `jobUpdateTs`? Or rename to `jobRunningTs`? This change is a bit unrelated to the original issue in this PR.

mxm (approved):
Generally, looks good to me. Some minor suggestions.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java
1996fanrui: Yeah, I see this code only be called when the job is running. I updated it for 2 reasons:

1.  This job may be failed during scaling. And `jobDetailsInfo` is fetched by rest api, the latest status may be not RUNNING.
2. As you mentioned, it will more robust if we have some refactors in the future.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java
1996fanrui: > Can we keep the old name jobUpdateTs? Or rename to jobRunningTs? 

Renaming it to `jobRunningTs`.

> This change is a bit unrelated to the original issue in this PR.

That's why I split this PR into 2 commits. 

- The first commit is refactor, it gets the jobRunningTs from JobStatus.RUNNING instead of max timestamp.
- The second commit is core change of the original issue.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java
afedulov: >We can add this invariant here but it would only defend against a Byzantine behavior.

I don't quite agree - I find it it a bit one-sided to look at public classes and methods only from the perspective of the implicit context in which they are **currently** called. 
I believe the check is appropriate - nothing in the contract or comments  of `updateMetrics` states the scope in which it is supposed to be executed. The class is also open for extension - someone can later override methods partially without being clear about the implicit assumptions. An alternative is to seal everything and make this class only available to `JobAutoscalerImpl`.

afedulov (approved):
Thanks, @1996fanrui , LGTM.

afedulov (approved):
Thanks, @1996fanrui , LGTM.

1996fanrui:
Thanks for the review. I tested locally, it works fine, the tracking time is correct now. Merging~

PR #762, opened 2024-01-23
 [FLINK-34152] Tune heap memory of autoscaled jobs

mxm:
The current autoscaling algorithm adjusts the parallelism of the job task vertices according to the processing needs. By adjusting the parallelism, we systematically scale the amount of CPU for a task. At the same time, we also indirectly change the amount of memory tasks have at their dispense. However, there are some problems with this.

1.  Memory is overprovisioned: On scale up, we may add more memory than we actually need. On scale down, the memory / cpu ratio can still be off and too much memory is used.
2. Memory is underprovisioned: For stateful jobs, we risk running into OutOfMemoryErrors on scale down. Even before running out of memory, too little memory can have a negative impact on the effectiveness of the scaling.

We lack the capability to tune memory proportionally to the processing needs. In the same way that we measure CPU usage and size the tasks accordingly, we need to evaluate memory usage and adjust the heap memory size.

Design doc: https://docs.google.com/document/d/19GXHGL_FvN6WBgFvLeXpDABog2H_qqkw1_wrpamkFSc/edit


flink-autoscaler/src/main/java/org/apache/flink/autoscaler/state/AutoScalerStateStore.java
```diff
@@ -69,6 +70,13 @@ void storeParallelismOverrides(Context jobContext, Map<String, String> paralleli
 
     void removeParallelismOverrides(Context jobContext) throws Exception;
 
+    void storeConfigOverrides(Context jobContext, Configuration configOverrides) throws Exception;
```
1996fanrui, line 73: Could we rename it to `config` instead of `configOverrides`?

I guess we might store some other configs into it even if it isn't overrides related. WDYT?

1996fanrui (commented):
Hi @mxm , thanks for your contribution!

I have some questions for this tuning strategy, and I left some comments. Please take a look in your free time, thanks~

> ### 1. Establish a heap memory baseline
> 
> We observe the average heap memory usage (`heap_usage`) at task managers.

Is it possible the average heap memory usage  is low, but a little tms are high. (It happens when data skew.)

After the tuning, some tms cannot run well.

> ### 2. Calculate memory usage per record
> 
> The memory requirements per record can be estimated by calculating this ratio:
> 
> ```
> heap_memory_per_rec = sum(heap_usage) / sum(processing_rate)
> ```
> 
> This ratio is surprisingly constant based off looking at empirical data.

I'm curious about this assumption. In general, flink process data one by one, it only process one record at the same time. So the heap memory doesn't cache more data even if the processing rate get high.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/utils/MemoryTuningUtils.java
```diff
@@ -0,0 +1,132 @@
...
+                        evaluatedMetrics, scalingSummaries, numTaskSlotsUsed);
+        int newNumTms = (int) Math.ceil(numTaskSlotsAfterRescale / (double) taskSlotsPerTm);
+        LOG.info(
+                "Estimating {} task slots in use after rescale, spread across {} TaskManagers",
+                numTaskSlotsAfterRescale,
+                newNumTms);
+
+        MemorySize newHeapSize = new MemorySize((long) (usedTotalHeapSize / newNumTms));
+        // TM container memory can never grow beyond the user-specified max
+        Optional<MemorySize> maxMemory = context.getTaskManagerMemoryFromSpec();
```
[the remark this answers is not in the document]
mxm: > From the code, I understand that "Adjust memory" will only adjust the memory down to below the user-specified memory, but not above the user-specified memory, right?

That is correct.

> I'm not sure could we make this option name as generic as possible? Such as: `job.autoscaler.taskmanager-memory.scaling.enabled` and `job.autoscaler.taskmanager-memory.scaling.heap.min`.

Sure, we can change those. Maybe `job.autoscaler.memory.taskmanager.heap.min` would be even more generic? It would also allow to add a jobmanager section under memory.

> In the short term, we only turn down. In the long term, we might support turn up the memory. If the option name is more generic, it's easy to compatible. WDYT?

I think the current option prefix would also allow to support increasing in the future, e.g. by adding `*.heap.max` option.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/KubernetesJobAutoScalerContext.java
```diff
@@ -56,19 +57,47 @@ public KubernetesJobAutoScalerContext(
...
+    public Optional<MemorySize> getTaskManagerMemoryFromSpec() {
+        return getJobDeployment()
+                .map(
+                        flinkDeployment ->
+                                MemorySize.parse(
+                                        flinkDeployment
+                                                .getSpec()
+                                                .getTaskManager()
+                                                .getResource()
+                                                .getMemory()));
```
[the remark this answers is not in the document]
mxm: So the idea here is that the memory tuning works similarly to the config-based parallelism overrides, i.e. we do not change the original user spec but apply changes on top of the spec before the reconciliation loop runs. That way, we can always recover the original value.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/KubernetesJobAutoScalerContext.java
```diff
@@ -56,19 +57,47 @@ public KubernetesJobAutoScalerContext(
...
 
+    @Override
+    public Optional<Double> getTaskManagerCpu() {
+        return Optional.ofNullable(
+                getConfiguration().get(KubernetesConfigOptions.TASK_MANAGER_CPU));
+    }
+
+    @Override
+    public Optional<MemorySize> getTaskManagerMemory() {
+        return Optional.ofNullable(getConfiguration().get(TaskManagerOptions.TOTAL_PROCESS_MEMORY));
```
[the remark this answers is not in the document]
mxm: Sure thing.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
```diff
@@ -151,7 +154,18 @@ protected void applyParallelismOverrides(Context ctx) throws Exception {
...
+    }
+
+    @VisibleForTesting
+    void applyConfigOverrides(Context ctx) throws Exception {
+        Configuration configOverrides = stateStore.getConfigOverrides(ctx);
+        var tmHeapOverride = configOverrides.get(TaskManagerOptions.TASK_HEAP_MEMORY);
+
+        if (tmHeapOverride != null) {
+            LOG.info("Applying heap memory override: {}", tmHeapOverride);
+            scalingRealizer.realizeMemoryOverrides(ctx, tmHeapOverride);
```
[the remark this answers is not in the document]
mxm: Good catch, will rename.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/state/AutoScalerStateStore.java
mxm: The idea is that those are applied on top of the regular config, that's why I chose configOverrides over config.

gyfora:
@mxm would you mind adding a few example calculations when memory would be scaled up or down based on the processing rate and heap usage metrics?

I haven't looked at the PR logic in detail but it's a bit hard for me to mentally visualise the expectations just by the description alone. Some more illustrations / examples would go a long way :)

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/utils/MemoryTuningUtils.java
```diff
@@ -0,0 +1,132 @@
...
+        }
+        // Apply limits
+        newHeapSize =
+                new MemorySize(
+                        Math.min(
+                                maxMemory.get().getBytes(),
+                                Math.max(
+                                        config.get(AutoScalerOptions.MEMORY_TUNING_MIN_HEAP)
+                                                .getBytes(),
+                                        newHeapSize.getBytes())));
```
gyfora, line 86: is it correct to compare here against `maxMemory`? `getTaskManagerMemoryFromSpec` returns the total process memory size, and only a small part of that may be heap.

gyfora: Furthermore the user may have configured a specific heap size directly in the conf

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
```diff
@@ -131,6 +141,12 @@ public boolean scaleResource(
                 getVertexParallelismOverrides(
                         evaluatedMetrics.getVertexMetrics(), scalingSummaries));
 
+        if (tmHeapMemoryOpt.isPresent()) {
+            Configuration configOverrides = autoScalerStateStore.getConfigOverrides(context);
+            configOverrides.set(TaskManagerOptions.TASK_HEAP_MEMORY, tmHeapMemoryOpt.get());
```
gyfora, line 146: I wonder what will happen if this is too big and the user also defined other memory options like managed fraction etc. We may create a config that simply won't run / make sense

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/utils/MemoryTuningUtils.java
mxm: True, this boundary is too high. We need to use the orignal max heap memory here. If users have configured a specific heap size, that is intended to be overwritten. But we could also choose to disable this feature in this case.

mxm:
> Hi @mxm , thanks for your contribution!
> 
> I have some questions for this tuning strategy, and I left some comments. Please take a look in your free time, thanks~
> 
> > ### 1. Establish a heap memory baseline
> > We observe the average heap memory usage (`heap_usage`) at task managers.
> 
> Is it possible the average heap memory usage is low, but a little tms are high. (It happens when data skew.)

Certainly possible, e.g. for hot key scenarios. We could take the max of the average usage to alleviate that concern.

> 
> > ### 2. Calculate memory usage per record
> > The memory requirements per record can be estimated by calculating this ratio:
> > ```
> > heap_memory_per_rec = sum(heap_usage) / sum(processing_rate)
> > ```
> >     
> > This ratio is surprisingly constant based off looking at empirical data.
> 
> I'm curious about this assumption. In general, flink process data one by one, it only process one record at the same time. So the heap memory doesn't cache more data even if the processing rate get high.

From what I observed, heap memory correlated closely with the number of records processed. There is some amount of memory overhead with every record processed. Maybe the data I looked at was biased. I'm open to changing this formula.

mxm:
> @mxm would you mind adding a few example calculations when memory would be scaled up or down based on the processing rate and heap usage metrics?

An example:

```
num_task_managers = 10
tm_memory = 40 Gb
observed_max_heap_size = 30 Gb
observed_average_heap_usage = 15 Gb
```

This will configure the heap memory to 15 Gb and reduce the total memory by `num_task_managers*15=150 Gb`, if the expected true processing rate does not change. If it changes, the memory is updated as follows:

```
data_change_rate = sum(expected_processing_rate) / sum(current_processing_rate)
total_heap = num_task_managers * observed_average_heap_usage * data_change_rate
tm_memory = total_heap / num_task_managers_after_rescale
```

The total heap size of the job is sized according to actual usage and scaled proportionally to the max expected records. This will solve memory issues for downscaled jobs which suffer from too few memory because the memory size is not scaled proportionally to the actual memory usage, but proportionally to the CPU usage.

To use some example numbers:

Example 1:
```
num_task_managers_after_rescale = 5
data_change_rate = 75 rec/s / 100 rec/s
total_heap = 10 * 15Gb * 0.75 = 112.5 Gb
tm_memory = 112.5 Gb / 5 = 22.5 Gb
```

Example 2:
```
num_task_managers_after_rescale = 5
data_change_rate = 50 rec/s / 100 rec/s
total_heap = 10 * 15Gb * 0.5 = 75 Gb
tm_memory = 75 Gb / 5 = 15 Gb
```

You can see that the heap memory per TM in Example 2 actually stayed the same, despite the data processing needs halving. That is actually not that different from the scaling we currently have and I can see how the data_change_rate may seem redundant. The main reason I integrated it was to supply headroom to operators who consume more records but are not necessarily adding more task managers.

I'm open to simplifying and removing the `data_change_rate`, simply tuning the memory according to the used memory. From looking at metrics of actual deployments, I can observe that the sum(heap_used) / sum(num_records_processed) is a fairly constant value.

> I haven't looked at the PR logic in detail but it's a bit hard for me to mentally visualise the expectations just by the description alone. Some more illustrations / examples would go a long way :)

I'll create a doc.

mxm:
@gyfora @1996fanrui I created a doc to go over the rational and the ideas behind this PR: https://docs.google.com/document/d/19GXHGL_FvN6WBgFvLeXpDABog2H_qqkw1_wrpamkFSc/edit

The PR itself will have to be updated because I'm no longer planning to incorporate the data change rate, but the heap-based adjustments remain.

gyfora (changes requested):
The memory tuning logic looks good 👍 

I think we should unify the storing / application of parallelism and other overrides. In this PR we introduced a bunch of new methods for storing config overrides. But the next time we add other tuning logic such as ephemeralStorage or whatever that doesn't translate to a config we would have to introduce again a similar set of new methods everywhere.

Maybe we could have a dedicated class for the overrides applied by the autoscaler:

```
class AutoscalerOverrides{
Map<JobVertexId, Integer> parallelism;
Configuration configOverrides;
...other stuff later
}
```
This way we could have all the related logic / utilities for ser/de collected there.

This would make the storage / retrieval logic simpler and would be very easily extensible. Wdyt @mxm @1996fanrui ?

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
```diff
@@ -106,7 +107,13 @@ public void scale(Context ctx) throws Exception {
         } catch (Throwable e) {
             onError(ctx, autoscalerMetrics, e);
         } finally {
-            applyParallelismOverrides(ctx);
+            try {
+                applyParallelismOverrides(ctx);
+                applyConfigOverrides(ctx);
```
mateczagany, line 112: This seems to be causing some issues when both parallelism and memory configurations get changed and in-place scaling is enabled. I could not figure out why, but I'll try to look into it next week.

I see the following error coming from the reconciliation loop:
```
Could not create Kubernetes cluster "test-app-scaling"
...
Message: object is being deleted: deployments.apps \"test-app-scaling\" already exists.
```

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/utils/MemoryTuningUtils.java
```diff
@@ -0,0 +1,203 @@
...
+     * modifying the config directly, but we are emitting a new configuration which contains any
+     * overrides. This config is persisted separately and applied by the autoscaler. That way we can
+     * clear any applied overrides if auto-tuning is disabled.
+     */
+    public static Configuration tuneTaskManagerHeapMemory(
+            JobAutoScalerContext<?> context, EvaluatedMetrics evaluatedMetrics) {
+
+        // Please note that this config is the original configuration created from the user spec.
+        // It does not contain any already applied overrides.
+        var config = new UnmodifiableConfiguration(context.getConfiguration());
```
1996fanrui, line 60: I tried to debug it, I found it's not the original configuration.

I run the `autoscaling.yaml`, the original TM total memory is 2GB. After tuning, it's changed to about 1.85 GB. And the taskHeap is reduced to 512 MB.

I found the config and memSpecs is changed.

image

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/utils/MemoryTuningUtils.java
```diff
@@ -0,0 +1,203 @@
...
+                        new MemorySize(
+                                memSpecs.getTotalFlinkMemorySize().getBytes() + heapDiffBytes)));
+        tuningConfig.set(
+                TaskManagerOptions.JVM_OVERHEAD_FRACTION,
+                getFraction(memSpecs.getJvmOverheadSize(), totalMemory));
+
+        return tuningConfig;
+    }
+
+    private static boolean shouldTransferHeapToManagedMemory(
```
1996fanrui, line 137: As I discussed in the design doc, would you mind introducing a option to control it?

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/utils/MemoryTuningUtils.java
```diff
@@ -0,0 +1,203 @@
...
+import org.apache.flink.runtime.util.config.memory.ProcessMemoryUtils;
+import org.apache.flink.runtime.util.config.memory.taskmanager.TaskExecutorFlinkMemory;
+import org.apache.flink.runtime.util.config.memory.taskmanager.TaskExecutorFlinkMemoryUtils;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Arrays;
+
+import static org.apache.commons.lang3.math.Fraction.getFraction;
```
1996fanrui, line 40: This import can be removed. It's a bug of `google-java-format`.

https://github.com/google/google-java-format/issues/1038

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/utils/MemoryTuningUtils.java
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    private static float getFraction(MemorySize enumerator, MemorySize denominator) {
+        // Round to three decimal places
+        return (float)
+                (Math.round(enumerator.getBytes() / (double) denominator.getBytes() * 1000)
+                        / 1000.);
+    }
+
+    private static boolean leftGreaterThanRight(MemorySize greater, MemorySize smaller) {
```
1996fanrui, line 200: nit: This method can be removed.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
mxm: Oh right, good point. In-place scaling won't be possible with auto tuning because the Flink config can't be updated on the fly. We need to fall back to a full upgrade in this case.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/KubernetesScalingRealizer.java
```diff
@@ -43,6 +53,32 @@ public void realize(
...
+            flinkDeployment.getSpec().getFlinkConfiguration().putAll(configOverrides.toMap());
+
+            var totalMemoryOverride = MemoryTuningUtils.getTotalMemory(configOverrides, context);
+            if (totalMemoryOverride.compareTo(MemorySize.ZERO) > 0) {
+                Resource tmResource = flinkDeployment.getSpec().getTaskManager().getResource();
+                // Make sure to support the Kubernetes syntax here, which supports more formats than
+                // Flink's classes.
+                var currentMemory =
+                        new MemorySize(
+                                (Quantity.parse(tmResource.getMemory()).getNumericalAmount())
```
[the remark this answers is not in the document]
mxm: Thanks for raising this. I intentionally added this way of parsing because I thought it is compatible with Kubernetes. Flink's MemorySize is a subset of Kubernetes because it does not support base 2 notation (e.g. Gigibytes). I didn't realize Flink is base 2 by default. I also forgot about the M vs MB notation. I'm afraid this logic might be broken in other parts of the operator. I'll look into it.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerContext.java
```diff
@@ -56,15 +58,21 @@ public class JobAutoScalerContext<KEY> {
...
     @ToString.Exclude
     private final SupplierWithException<RestClusterClient<String>, Exception> restClientSupplier;
 
+    /** Retrieve the currently configured TaskManager CPU for this context. */
+    public Optional<Double> getTaskManagerCpu() {
+        // Not supported by default
+        return Optional.empty();
+    }
+
+    /** Retrieve the currently tuned TaskManager memory for this context. */
```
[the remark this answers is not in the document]
mxm: That wording is more precise. I agree.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
gyfora: If we apply the config overrides on the FlinkDeployment spec itself then this will be done automatically.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/utils/MemoryTuningUtils.java
mxm: Thanks Rui! I originally had another method on the context to retrieve the original config ("deploy" config). Let me see if we need that.

mxm:
> The memory tuning logic looks good 👍
> 
> I think we should unify the storing / application of parallelism and other overrides. In this PR we introduced a bunch of new methods for storing config overrides. But the next time we add other tuning logic such as ephemeralStorage or whatever that doesn't translate to a config we would have to introduce again a similar set of new methods everywhere.
> 
> Maybe we could have a dedicated class for the overrides applied by the autoscaler:
> 
> ```
> class AutoscalerOverrides{
> Map<JobVertexId, Integer> parallelism;
> Configuration configOverrides;
> ...other stuff later
> }
> ```
> 
> This way we could have all the related logic / utilities for ser/de collected there.
> 
> This would make the storage / retrieval logic simpler and would be very easily extensible. Wdyt @mxm @1996fanrui ?

I had thought about that but I was a bit concerned about touching the parallelism overrides because any breaking changes there can have a dramatic effect on any deployed autoscaled applications. The newly added Configuration at least will cover all configuration-related changes. Those would also include the parallelism overrides themselves. So I understand the idea to combine those two and potentially other types of overrides. Let me look into it.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/state/AutoScalerStateStore.java
mxm: I'm happy to rename but I can't imagine another use right now. Are you suggesting we might store other state in the Configuration which is not related to the configuration? I think that goes in the direction Gyula's comment (https://github.com/apache/flink-kubernetes-operator/pull/762#issuecomment-1926743545).

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
mxm: Actually, this is already handled by the current code. The in-place parallelism class doesn't apply config overrides. Need to look into this further.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/KubernetesScalingRealizer.java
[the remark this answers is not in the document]
mxm: Should be fixed now via https://github.com/apache/flink-kubernetes-operator/pull/762/commits/693e913122b197ed7dcc5d8d5b564f64d3b13168.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkResourceContext.java
```diff
@@ -65,7 +65,7 @@ public KubernetesJobAutoScalerContext getJobAutoScalerContext() {
     }
 
     private KubernetesJobAutoScalerContext createJobAutoScalerContext() {
-        Configuration conf = new Configuration(getObserveConfig());
+        Configuration conf = new Configuration(getDeployConfig(resource.getSpec()));
```
mxm, line 68: @gyfora @1996fanrui I changed this logic because I believe it is the right config to return in the context of the autoscaler. We want to run autoscaling based on the current user-provided configuration, not based on what is currently deployed. The former is the user intend, the latter is the result of a past reconciliation decision. This will also ensure that overrides are cleared immediately and not after completing the full reconciliation.

For the tuning, the current logic only works correctly if we operate on the "deploy config" because that allows us to correctly determine the original memory configuration which is required for dynamically increasing / decreasing total memory and managed memory without going past the original user-provided limits.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/utils/MemoryTuningUtils.java
mxm: This should be fixed after the latest commit. See the related change: https://github.com/apache/flink-kubernetes-operator/pull/762#discussion_r1478450912

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkResourceContext.java
gyfora: I think using the observe config is the right thing to do for the autoscaler in general. We are observing a flink job with its metrics and configuration and making scaling decisions.

The deploy config may contain yet undeployed settings .

I think this change only makes this feature more convenient to implement but I think the right thing to do would be to instead store the original memory limit in the autoscaler if that's important for the logic.

gyfora: Also for the standalone autoscaler you may only have the config that is actually running on the cluster (again that is equivalent to the observe config here)

mxm: Storing memory limits or user-specified configuration wouldn't be correct because those always come from the spec, but I can add a getter to the Autoscaler context to retrieve the memory limit from the spec. That's what I initially had.

gyfora: But you are right that if the user actually changes the memory limits in a new version of the spec then the deploy config would actually make more sense to use.

gyfora: Thinking through this again, I think you are right, that in some cases the deploy config is simply better. Probably the only clear solution is to provide both but we could also start with just the deploy config

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricEvaluator.java
```diff
@@ -314,13 +315,19 @@ protected static Map<ScalingMetric, EvaluatedScalingMetric> evaluateGlobalMetric
...
-                HEAP_USAGE,
+                HEAP_MAX_USAGE_RATIO,
                 new EvaluatedScalingMetric(
-                        lastHeapUsage, getAverageGlobalMetric(HEAP_USAGE, metricHistory)));
+                        lastHeapUsage,
+                        getAverageGlobalMetric(HEAP_MAX_USAGE_RATIO, metricHistory)));
+
+        var latestObservation = latest.getOrDefault(HEAP_AVERAGE_SIZE, Double.NaN);
+        double heapSizeAverage = getAverageGlobalMetric(HEAP_AVERAGE_SIZE, metricHistory);
+        LOG.debug("Latest: {}, Avg: {}", latestObservation, heapSizeAverage);
```
gyfora, line 329: Should we remove this?

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/utils/MemoryTuningUtils.java
```diff
@@ -0,0 +1,238 @@
...
+        CommonProcessMemorySpec<TaskExecutorFlinkMemory> memSpecs;
+        try {
+            memSpecs = FLINK_MEMORY_UTILS.memoryProcessSpecFromConfig(config);
+        } catch (IllegalConfigurationException e) {
+            LOG.warn("Current memory configuration is not valid. Aborting memory tuning.");
+            return EMPTY_CONFIG;
+        }
+
+        var maxHeapSize = memSpecs.getFlinkMemory().getJvmHeapMemorySize();
+        LOG.info("Current configured heap size: {}", maxHeapSize);
```
gyfora, line 78: should this be debug?

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
```diff
@@ -250,6 +251,31 @@ private static ConfigOptions.OptionBuilder autoScalerConfig(String key) {
...
+                            "The minimum amount of TaskManager memory, if memory tuning is enabled.");
+
+    public static final ConfigOption<Boolean> MEMORY_TUNING_TRANSFER_HEAP_TO_MANAGED =
+            autoScalerConfig("memory.tuning.heap.transfer-to-managed")
+                    .booleanType()
+                    .defaultValue(false)
+                    .withFallbackKeys(
+                            oldOperatorConfigKey("memory.tuning.heap.transfer-to-managed"))
+                    .withDescription(
+                            "If enabled, any reduction of heap memory will increase the managed memory for RocksDB. RocksDB needs to be enabled.");
```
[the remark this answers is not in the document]
mxm: Sounds good.

flink-autoscaler/src/test/java/org/apache/flink/autoscaler/utils/ResourceCheckUtilsTest.java
```diff
@@ -32,7 +32,7 @@
 class ResourceCheckUtilsTest {
 
     @Test
-    void testEstimateNumTaskSlotsAfterRescale() {
+    void estimateNumTaskSlotsAfterRescale() {
```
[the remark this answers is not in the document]
mxm: Was an issue with rebasing. I'll re-add.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/utils/MemoryTuningUtils.java
```diff
@@ -0,0 +1,238 @@
...
+                AutoScalerEventHandler.Type.Normal,
+                "Configuration recommendation",
+                String.format(
+                        "Memory tuning recommends the following configuration (automatic tuning is %s):\n%s",
+                        config.get(AutoScalerOptions.MEMORY_TUNING_ENABLED)
+                                ? "enabled"
+                                : "disabled",
+                        formatConfig(tuningConfig)),
+                "MemoryTuning",
+                null);
```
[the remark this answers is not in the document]
mxm: The idea is to add tuning recommendations even when we don't automatically apply updates. That was heavily requested by users. The tuning is only called on scaling decisions where we also emit a scaling event.

I agree that once we decouple the tuning from the scaling, we should only emit recommendations every N minutes.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricEvaluator.java
mxm: It's logged on debug, why not keep it?

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/utils/MemoryTuningUtils.java
mxm: Yes

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricEvaluator.java
gyfora: ""Latest: 100, Avg: 400" doesn't really mean anything. We can keep it but then we should add some more info the log

mxm: Removing.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
mateczagany: Sorry, it's not related to this change at all, and it's completely intermittent. I seem to be running into FLINK-32334 and the TaskManagers (thus the Kubernetes Deployment as well) are not properly deleted in native mode before starting the new deployment.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
```diff
@@ -250,6 +252,39 @@ private static ConfigOptions.OptionBuilder autoScalerConfig(String key) {
...
+                    .booleanType()
+                    .defaultValue(false)
+                    .withFallbackKeys(oldOperatorConfigKey("memory.tuning.enabled"))
+                    .withDescription(
+                            "If enabled, the initial amount of memory specified for TaskManagers will be reduced according to the observed needs.");
+
+    public static final ConfigOption<MemoryTuning.HeapTuningTarget> MEMORY_TUNING_HEAP_TARGET =
+            autoScalerConfig("memory.tuning.heap.target-usage")
+                    .enumType(MemoryTuning.HeapTuningTarget.class)
+                    .defaultValue(MemoryTuning.HeapTuningTarget.AVG)
```
gyfora, line 266: If Max is more conservative should that be the default?

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
```diff
@@ -250,6 +252,39 @@ private static ConfigOptions.OptionBuilder autoScalerConfig(String key) {
...
+    public static final ConfigOption<Boolean> MEMORY_TUNING_ENABLED =
+            autoScalerConfig("memory.tuning.enabled")
+                    .booleanType()
+                    .defaultValue(false)
+                    .withFallbackKeys(oldOperatorConfigKey("memory.tuning.enabled"))
+                    .withDescription(
+                            "If enabled, the initial amount of memory specified for TaskManagers will be reduced according to the observed needs.");
+
+    public static final ConfigOption<MemoryTuning.HeapTuningTarget> MEMORY_TUNING_HEAP_TARGET =
+            autoScalerConfig("memory.tuning.heap.target-usage")
```
gyfora, line 264: This config key sounds similar to target utilisation but works a completely different way, but I don't really have a better suggestion :)

gyfora (approved):

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
mxm: Updated to use MAX as the default. Also added extra overhead which is pretty important for the memory to be able to grow once reduced.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
```diff
@@ -250,6 +251,22 @@ private static ConfigOptions.OptionBuilder autoScalerConfig(String key) {
...
+                    .withDescription(
+                            "If enabled, the initial amount of memory specified for TaskManagers will be reduced according to the observed needs.");
+
+    public static final ConfigOption<MemorySize> MEMORY_TUNING_MIN_HEAP =
+            autoScalerConfig("memory.tuning.heap.min")
+                    .memoryType()
+                    .defaultValue(MemorySize.ofMebiBytes(2048L))
+                    .withFallbackKeys(oldOperatorConfigKey("memory.tuning.heap.min"))
+                    .withDescription(
+                            "The minimum amount of TaskManager memory, if memory tuning is enabled.");
```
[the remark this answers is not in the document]
1996fanrui: Hi @mxm , a gentle reminder: your change might missed this comment.

flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/state/JdbcAutoScalerStateStore.java
```diff
@@ -189,6 +191,28 @@ public void removeParallelismOverrides(Context jobContext) {
         jdbcStateStore.removeSerializedState(getSerializeKey(jobContext), PARALLELISM_OVERRIDES);
     }
 
+    @Override
+    public void storeConfigOverrides(Context jobContext, Configuration configOverrides) {
+        jdbcStateStore.putSerializedState(
+                getSerializeKey(jobContext),
+                CONFIG_OVERRIDES,
+                serializeParallelismOverrides(configOverrides.toMap()));
```
1996fanrui, line 199:
```suggestion
                serializeConfigOverrides(configOverrides.toMap()));
```

It may be a bug.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/utils/MemoryTuningUtils.java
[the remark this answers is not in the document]
1996fanrui: Sorry, I still don't know why we cannot add the interval here. Why cannot set interval before decoupling? Would you mind elaborating on it? Thanks a lot.

1996fanrui (approved):
Hi @mxm , thanks a lot for the update!

I still have 3 minor comments for this PR, please take a look in your free time, thanks~

https://github.com/apache/flink-kubernetes-operator/pull/762#discussion_r1464629621
https://github.com/apache/flink-kubernetes-operator/pull/762#discussion_r1482469765
https://github.com/apache/flink-kubernetes-operator/pull/762#discussion_r1490294787

flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/state/JdbcAutoScalerStateStore.java
mxm: It's not because the serialization is identical but it could become a bug, you are right. Thanks for spotting!

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/utils/MemoryTuningUtils.java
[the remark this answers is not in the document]
mxm: Sure, I'll add an interval here.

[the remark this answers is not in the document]
mxm: What would you like the interval to be?

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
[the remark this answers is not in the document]
mxm: Thanks! Not intentional! Addressing this now.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/utils/MemoryTuningUtils.java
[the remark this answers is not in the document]
mxm: I've reused AutoscalerOptions.SCALING_EVENT_INTERVAL for now.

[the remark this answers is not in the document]
1996fanrui: `AutoscalerOptions.SCALING_EVENT_INTERVAL` makes sense to me.

1996fanrui (approved):
Thanks for the quick update!

LGTM.

PR #763, opened 2024-01-30
[FLINK-34311] Do not change min resource requirements when rescaling with adaptive scheduler

gyfora:

Do not change the min parallelism requirements when rescaling pipelines with the adaptive scheduler in order to avoid unnecessary failures.

## Brief change log
 - Set min to previous min

## Verifying this change
Unit tests added

## Documentation

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/NativeFlinkService.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/NativeFlinkService.java
@@ -200,9 +200,15 @@ public class NativeFlinkService extends AbstractFlinkService {
                 var overrideStr = newOverrides.get(jobId);
 
                 if (overrideStr != null) {
-                    // We have an override for the vertex
-                    int p = Integer.parseInt(overrideStr);
-                    var newParallelism = new JobVertexResourceRequirements.Parallelism(p, p);
+                    // We set the parallelism upper bound to the target parallelism, anything higher
+                    // would defeat the purpose of scaling down
+                    int upperBound = Integer.parseInt(overrideStr);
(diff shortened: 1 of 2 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/NativeFlinkService.java
```diff
@@ -201,8 +201,11 @@ public ScalingResult scale(FlinkResourceContext<?> ctx, Configuration deployConf
 
                 if (overrideStr != null) {
                     // We have an override for the vertex
-                    int p = Integer.parseInt(overrideStr);
-                    var newParallelism = new JobVertexResourceRequirements.Parallelism(p, p);
+                    int upperBound = Integer.parseInt(overrideStr);
+                    // We keep the current parallelism lower bound unless we went below it
+                    int lowerBound = Math.min(upperBound, parallelism.getLowerBound());
```
mxm, line 206: Can we add a small comment why we do this?

mxm (approved):
LGTM

1996fanrui (approved):
LGTM

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/NativeFlinkService.java
1996fanrui: IIUC, Adaptive Scheduler will run the job without pause if we don't change the lowerBound.

Adaptive Scheduler can restart the job after the resource is ready.

gyfora: I will add some comments and do some extensive local testing before merging this

PR #765, opened 2024-01-31
[FLINK-33636][autoscaler] Support the JDBCAutoScalerEventHandler

1996fanrui:

-  [FLINK-33636][autoscaler] Support the JDBCAutoScalerEventHandler
- Design doc: https://docs.google.com/document/d/1wXs6lLvX5jw6KNsz1pP9rdNLGoeInWkd5lbZo-rQp4s/edit?usp=sharing

## Changes:

- [refactor] Moving the getParallelismHashCode from KubernetesAutoScalerEventHandler to AutoScalerEventHandler
- [FLINK-33636][autoscaler] Support the JDBCAutoScalerEventHandler
- [FLINK-33636][autoscaler] Autoscaler Standalone supports the JdbcAutoScalerEventHandler
- [FLINK-33636][autoscaler] Document the JDBC Autoscaler event handler

## Verifying this change

Added some test classes:

- AbstractJdbcAutoscalerEventHandlerITCase
- AbstractJdbcEventInteractorITCase
- AutoscalerEventHandlerFactoryTest

## Documentation

### DIFF
--- a/docs/content/docs/custom-resource/autoscaler.md
+++ b/docs/content/docs/custom-resource/autoscaler.md
@@ -273,7 +273,7 @@ org.apache.flink.autoscaler.standalone.StandaloneAutoscalerEntrypoint \
 Updating the `autoscaler.standalone.fetcher.flink-cluster.host` and `autoscaler.standalone.fetcher.flink-cluster.port` 
 based on your flink cluster. In general, the host and port are the same as Flink WebUI.
 
-### Using the JDBC Autoscaler State Store
+### Using the JDBC Autoscaler State Store & Event Handler
 
 A driver dependency is required to connect to a specified database. Here are drivers currently supported, 
 please download JDBC driver and initialize database and table first.
@@ -286,16 +286,17 @@ please download JDBC driver and initialize database and table first.
 
 ```
 JDBC_DRIVER_JAR=./mysql-connector-java-8.0.30.jar
-# export the password of jdbc state store
-export STATE_STORE_JDBC_PWD=123456
+# export the password of jdbc state store & jdbc event handler
+export JDBC_PWD=123456
 
 java -cp flink-autoscaler-standalone-{{< version >}}.jar:${JDBC_DRIVER_JAR} \
 org.apache.flink.autoscaler.standalone.StandaloneAutoscalerEntrypoint \
 --autoscaler.standalone.fetcher.flink-cluster.host localhost \
 --autoscaler.standalone.fetcher.flink-cluster.port 8081 \
 --autoscaler.standalone.state-store.type jdbc \
---autoscaler.standalone.state-store.jdbc.url jdbc:mysql://localhost:3306/flink_autoscaler \
---autoscaler.standalone.state-store.jdbc.username root
+--autoscaler.standalone.event-handler.type jdbc \
+--autoscaler.standalone.jdbc.url jdbc:mysql://localhost:3306/flink_autoscaler \
+--autoscaler.standalone.jdbc.username root
 ```
 
 All supported options for autoscaler standalone can be viewed 
--- a/flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/config/AutoscalerStandaloneOptions.java
+++ b/flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/config/AutoscalerStandaloneOptions.java
@@ -17,6 +17,7 @@
 
 package org.apache.flink.autoscaler.standalone.config;
 
+import org.apache.flink.autoscaler.standalone.AutoscalerEventHandlerFactory.EventHandlerType;
 import org.apache.flink.autoscaler.standalone.AutoscalerStateStoreFactory.StateStoreType;
 import org.apache.flink.configuration.ConfigOption;
 import org.apache.flink.configuration.ConfigOptions;
@@ -24,7 +25,6 @@ import org.apache.flink.configuration.description.Description;
 
 import java.time.Duration;
 
-import static org.apache.flink.autoscaler.standalone.AutoscalerStateStoreFactory.StateStoreType.JDBC;
 import static org.apache.flink.configuration.description.TextElement.code;
 
 /** Config options related to the autoscaler standalone module. */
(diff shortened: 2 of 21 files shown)


1996fanrui (commented):
Hi @gyfora @mxm , I see the `Kubernetes Operator 1.8.0` will be released soon.

I have finished the prod code development for the `JDBCAutoScalerEventHandler`, but the unit test and itcase are still developing. (Note: I have tested with MySQL manually. It works well.) Current PR only has prod code, it doesn't include any test code.

I'm not sure could we include it into 1.8.0? I will try to finish the unit test and itcase asap.

BTW, this Saturday is Chenese New Year. So I will be on vocation starting from this Friday, the vacation will last about 10 days.

1996fanrui:
@gyfora  @mxm , sorry to bother you again. During I supporting PostgreSQL and Derby, I found some defects for jdbc event handler and jdbc state store about the table schema.

In order to ensure the stability of the table structure, I'd like to polish the state store design first. And finish the event handler in the next version. Thank you.

mxm:
@1996fanrui How important is it for you to get this into 1.8.0? I think the 1.8.0 release could wait until then.

gyfora:
Sounds like @1996fanrui wants to have a peaceful vacation. I don't see any reason why we should rush this into the release or delay the release because of it.

1996fanrui:
> @1996fanrui How important is it for you to get this into 1.8.0? I think the 1.8.0 release could wait until then.

If I didn't take vacation, I might ask for a two-day delay in release. (With two days of delay plus the weekend, I believe I have enough time to complete this PR). But I will be on vacation for 10 days, I am worried about affecting the release progress. Sorry for that.

docs/content/docs/custom-resource/autoscaler.md
```diff
@@ -286,16 +286,20 @@ please download JDBC driver and initialize database and table first.
...
 org.apache.flink.autoscaler.standalone.StandaloneAutoscalerEntrypoint \
 --autoscaler.standalone.fetcher.flink-cluster.host localhost \
 --autoscaler.standalone.fetcher.flink-cluster.port 8081 \
 --autoscaler.standalone.state-store.type jdbc \
 --autoscaler.standalone.state-store.jdbc.url jdbc:mysql://localhost:3306/flink_autoscaler \
---autoscaler.standalone.state-store.jdbc.username root
+--autoscaler.standalone.state-store.jdbc.username root \
+--autoscaler.standalone.event-handler.type jdbc \
+--autoscaler.standalone.event-handler.jdbc.url jdbc:mysql://localhost:3306/flink_autoscaler \
+--autoscaler.standalone.event-handler.jdbc.username root
```
1996fanrui, line 302: These database related options are similar between jdbcStateStore and jdbcEventHandler. I'm not sure whether we should merge them or stay as is.

Merge them means we rename `autoscaler.standalone.state-store.jdbc.username` to `autoscaler.standalone.jdbc-plugin.username`. And both of jdbcStateStore and jdbcEventHandler will use `autoscaler.standalone.jdbc-plugin.username` and `autoscaler.standalone.jdbc-plugin.url`.

If  we define `autoscaler.standalone.event-handler.jdbc.url` and `autoscaler.standalone.state-store.jdbc.url`, it means jdbcStateStore and jdbcEventHandler can use different databases and different users.

WDYT? @gyfora  @mxm

1996fanrui: I'm still working on this PR now, I found it's better to include this PR in `1.8.0`. Because I believe  jdbcStateStore and jdbcEventHandler  always work together.

During I test the `jdbcEventHandler`, I found some defects of `jdbcStateStore`, such as: FLINK-34389, this comment https://github.com/apache/flink-kubernetes-operator/pull/765#discussion_r1480012807. 

It means if we includes `jdbcEventHandler` in  this version we can ensure they work well together. 

So I'd like to increase this priority, WDYT?

Note: the prod code is done(you can help review if you are free), and the progress of unit test and ITCase is about 40% and I will finish it asap..

1996fanrui:
Hi @gyfora @mxm , the development and test JdbcAutoscalerEventHandler are totally done,  it can be included in `1.8.0`.

I finished the unit test, itcase, and I tested it on my local manually. It works well.

docs/content/docs/custom-resource/autoscaler.md
gyfora: @1996fanrui I agree that we should probably merge the configs as they will be used together. Instead of calling it `jdbc-plugin` we could just call the properties:
```
autoscaler.standalone.jdbc.username
autoscaler.standalone.jdbc.url
...

```

flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/config/AutoscalerStandaloneOptions.java
```diff
@@ -109,4 +110,51 @@ private static ConfigOptions.OptionBuilder autoscalerStandaloneConfig(String key
                                                     + "export the password using this environment variable.",
                                             code(STATE_STORE_TYPE.key()), code(JDBC.toString()))
                                     .build());
+
+    public static final ConfigOption<EventHandlerType> EVENT_HANDLER_TYPE =
+            autoscalerStandaloneConfig("event-handler.type")
+                    .enumType(EventHandlerType.class)
+                    .defaultValue(EventHandlerType.LOGGING)
+                    .withDescription("The autoscaler event handler type.");
```
mxm, line 118: What if users want to configure both logging and storing via JDBC?

1996fanrui: We can configure them separately.

```
--autoscaler.standalone.state-store.type jdbc \
--autoscaler.standalone.event-handler.type jdbc \
```

After users choose JDBC for them, we can reuse all jdbc related options, such as:

```
autoscaler.standalone.jdbc.username
autoscaler.standalone.jdbc.url
```

docs/content/docs/custom-resource/autoscaler.md
1996fanrui: Thanks for the feedback! I have updated.

flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/config/AutoscalerStandaloneOptions.java
mxm: Got it. That makes sense. I meant what if I want to print both to the console and store in JDBC? Currently there can only be one event handler active at a time.

1996fanrui: Sorry, I misunderstand your question before, thanks for your clarification!

> I meant what if I want to print both to the console and store in JDBC? Currently there can only be one event handler active at a time.

You are right, currently there are only be one event handler or state store active at a time. I see kubernetes operator is similar, kubernetes operator only creates one event handler or state store.

If users want to use multiple event handlers, I can implement `CompositeEventHandler` in the future. Users can configure:

```
--autoscaler.standalone.event-handler.type composite
--autoscaler.standalone.event-handler.composite.types jdbc,logging
```

The `CompositeStateStore` is similar to `CompositeEventHandler`:

```
--autoscaler.standalone.state-store.type composite
--autoscaler.standalone.state-store.composite.types jdbc,memory
```

In the short term, it seems unnecessary. I can implement them in the future if some users feedback it's useful in some scenarios. What do you think about the solution and time plan?

mxm: That sounds good to me. We can add the composite handler if necessary in the future.

gyfora (approved):
looks good!

mxm (approved):

PR #769, opened 2024-02-01
[FLINK-34329][autoscaler] Fix the bug that scaling report parser doesn't support Locale

1996fanrui:

`AutoscalerEventUtils#parseVertexScalingReports` uses the `Double.parseDouble` to parse the `CurrentProcessCapacity`, `expectedProcessCapacity` and `targetDataRate`.

But they are formatted by `String.format("%.2f")`, it will use the local format by default. And the decimal point format in EU is `,` instead of `.`.  For example: `String.format(Locale.GERMAN, "%.2f", 812.58)` will return `812,58`.

## Brief change log

[FLINK-34329][autoscaler] Fix the bug that scaling report parser doesn't support Locale

- Step1: Using the NumberFormat to support Locale format because the event is formatted in String.format, it uses the Locale by default.
- Step2: Using the Double.parseDouble 
  - NumberFormat doesn't support Infinity and NaN
  - So we use the `Double.parseDouble` when the step1 doesn't work.

## Verifying this change

Locale as the parameters, and we tested some common Locales by default.

### DIFF
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/event/AutoscalerEventUtils.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/event/AutoscalerEventUtils.java
@@ -19,6 +19,10 @@ package org.apache.flink.autoscaler.event;
 
 import org.apache.flink.annotation.Experimental;
 
+import lombok.SneakyThrows;
+
+import java.text.NumberFormat;
+import java.text.ParseException;
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


mxm (approved):
Looks good to me. This only touches the parsing code for events and none of the generating code.

gyfora (approved):
Tested locally, works as expected :)

1996fanrui (commented):
Thank you for the quick review, merging~

PR #771, opened 2024-02-06
[FLINK-34389][autoscaler] JdbcAutoscalerStateStore explicitly writes update_time

1996fanrui:

JdbcAutoscalerStateStore explicitly writes update_time instead of relying on the database to update.

Some databases doesn't support update the timestamp column automatically. For example, Derby doesn't support update the update_time automatically when we update any data. It's hard to do a general test during I developing the test for JdbcAutoscalerEventHandler.

As the common&open source service, in order to support all databases well, it's better to handle it inside of the service.

In order to unify the design for JdbcAutoscalerEventHandler and JdbcAutoscalerStateStore, we update the design of JdbcAutoscalerStateStore in this PR.

## Brief change log

[FLINK-34389][autoscaler] JdbcAutoscalerStateStore explicitly writes update_time


gyfora (approved):

mxm (approved):

1996fanrui (commented):
Thanks for the review, merging~

PR #773, opened 2024-02-06
[FLINK-33771] Extract method to estimate task slots and add dedicated test

mxm:
While working on #762, I realized this change can be extracted and merged independently.


1996fanrui (approved):
Thanks for the refactor!

LGTM

PR #776, opened 2024-02-13
[FLINK-34441] Add guide to submit flink SQL scripts via the operator (using flink-sql-runner-example)

prakash-42:

This pull request adds documentation about flink-sql-runner-example in the Kubernetes operator documentation page. Majority of the content is copied over from the [flink-sql-runner-example's readme](https://github.com/apache/flink-kubernetes-operator/tree/release-1.7/examples/flink-sql-runner-example#flink-kubernetes-operator-sql-example).

## Brief change log

  - Added guide to submit flink SQL scripts via the kubernetes operator (using flink-sql-runner-example)

## Verifying this change

*This change is a trivial rework / code cleanup without any test coverage.*
I don't know if we have tests covering the docs project. I ran the project locally and made sure that my changes are being rendered properly.

## Documentation

### DIFF
--- a/docs/content/docs/custom-resource/overview.md
+++ b/docs/content/docs/custom-resource/overview.md
@@ -216,6 +216,38 @@ Alternatively, if you use helm to install flink-kubernetes-operator, it allows y
 
 - Last-state upgradeMode is currently not supported for FlinkSessionJobs
 
+## Flink SQL Jobs
+Another common way to deploy Flink jobs is by writing SQL scripts. We have put together [a Flink Java application](https://github.com/apache/flink-kubernetes-operator/tree/main/examples/flink-sql-runner-example#flink-kubernetes-operator-sql-example) that can read Flink SQL scripts and submit the queries using the table API. The user needs to package the job into a JAR and provide the SQL script's path as an argument to the job.
+
+You can clone the [flink-kubernetes-operator](https://github.com/apache/flink-kubernetes-operator.git) repo and follow these steps to deploy a Flink application that can run Flink SQL scripts. These steps assume that you have the Flink Kubernetes Operator installed and running in your environment.
+
+**Step 1**: Build Sql Runner maven project
+```bash
+cd examples/flink-sql-runner-example
+mvn clean package
+```
+
+**Step 2**: Add your SQL script files under the `sql-scripts` directory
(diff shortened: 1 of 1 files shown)


prakash-42:
@mxm Here is a PR for the discussion we were having at PR #596 . Please let me know if any updates/improvements are needed. Thank you!

docs/content/docs/custom-resource/overview.md
```diff
@@ -216,6 +216,38 @@ Alternatively, if you use helm to install flink-kubernetes-operator, it allows y
 
 - Last-state upgradeMode is currently not supported for FlinkSessionJobs
 
+## Flink SQL Jobs
```
mxm, line 219: Taking another look, this might not be the best place to add this information because it's an overview page and the steps are quite detailed. There is already a section mentioning the examples: https://nightlies.apache.org/flink/flink-kubernetes-operator-docs-main/docs/concepts/overview/#built-in-examples How about creating a dedicated Examples page?

mxm: I think we should avoid copying anything over from https://github.com/apache/flink-kubernetes-operator/tree/main/examples/. Instead, we can extract the above linked examples section to a dedicated page and add any extra information we find useful.

prakash-42: Thanks for the comments @mxm ! I agree that we should have a dedicated page for examples, instead of adding more in the overview page. However, I have a few follow-up concerns/questions:

1. If I don't copy over from the examples' readme, I'd still be needing to write content that's similar to whatever is present in the example's readme. I can try and make it short, while linking to the readme page for more details.

2. If I create a dedicated Examples page, then I think it shouldn't just have the Flink SQL example, but other examples as well. I will try and write a few more, but not sure if I have enough experience/knowledge to write about all the examples. Can we make this like a running doc where more examples will be added as the need arises?

3. One concern I have about linking to README pages is that I'm likely going to link to the examples in main branch, and that code may change with later releases in the operator. Would that possibly make the examples in the operator point to an incompatible readme? (I see that the existing links also point to the main branch, so maybe that's okay)

gyfora:
Closing this due to inactivity, please re-open if needed

PR #777, opened 2024-02-14
[FLINK-34438] Wait for Deployments to be deleted on cluster shutdown

mateczagany:

Waiting for TaskManager and JobManager pods to be removed might cause issues in a slow Kubernetes cluster. Upon upgrade, there is a slim possibility that the operator will attempt to create a new `Deployment` before the last one was removed. This was mitigated by [FLINK-32334](https://issues.apache.org/jira/browse/FLINK-32334), which worked well for `standalone`  mode.

I wanted to simply implement `getTmPodList` for `native` mode at first, but after some investigation it became clear that all Kubernetes resources created by the operator or the JobManager will be children of a `Deployment` with `blockOwnerDeletion: true`, so it made sense to instead change the logic to poll the `Deployment` resources.

I also removed the `Thread.sleep()` calls, and instead use the Fabric8 Informers API, which will use websockets to ease the stress on the Kubernetes cluster during cluster deletion.

## Brief change log

- Changed `AbstractFlinkService#waitUntilCondition` to use Informer API and check for `Deployment` resources.
- Added unit tests for the new method `AbstractFlinkService#waitForDeploymentToBeRemoved`.
- Added log when cluster shutdown timeout is exceeded

## Verifying this change

- Added new unit test
- Manual verification of timeout and 

## Documentation

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
@@ -129,8 +129,6 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.jar.JarOutputStream;
@@ -927,12 +925,10 @@ public abstract class AbstractFlinkService implements FlinkService {
     /** Wait until Deployment is removed, return false if timed out, otherwise return true. */
     @VisibleForTesting
     boolean waitForDeploymentToBeRemoved(String namespace, String deploymentName, long timeout) {
-        ScheduledExecutorService logger = Executors.newSingleThreadScheduledExecutor();
-        logger.scheduleWithFixedDelay(
-                () -> LOG.info("Waiting for Deployment {} to shut down...", deploymentName),
-                5,
-                5,
-                TimeUnit.SECONDS);
+        LOG.info(
+                "Waiting for Deployment {} to shut down with {} ms timeout...",
+                deploymentName,
+                timeout);
 
         try {
             kubernetesClient
@@ -945,8 +941,6 @@ public abstract class AbstractFlinkService implements FlinkService {
             LOG.info("Deployment {} successfully shut down", deploymentName);
         } catch (KubernetesClientTimeoutException e) {
             return false;
-        } finally {
-            logger.shutdown();
         }
         return true;
     }
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
@@ -101,8 +101,8 @@ import org.apache.flink.util.Preconditions;
 import io.fabric8.kubernetes.api.model.DeletionPropagation;
 import io.fabric8.kubernetes.api.model.ObjectMeta;
 import io.fabric8.kubernetes.api.model.PodList;
-import io.fabric8.kubernetes.api.model.Service;
 import io.fabric8.kubernetes.client.KubernetesClient;
+import io.fabric8.kubernetes.client.KubernetesClientTimeoutException;
 import org.apache.commons.lang3.ObjectUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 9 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
```diff
@@ -898,58 +901,54 @@ public JobDetailsInfo getJobDetailsInfo(JobID jobID, Configuration conf) throws
...
-            if (taskManagerRunning) {
-                PodList tmPodList = getTmPodList(namespace, clusterId);
+        for (var deploymentName : getDeploymentNames(namespace, clusterId)) {
+            long deploymentTimeout = timeoutAt - System.currentTimeMillis();
 
-                if (tmPodList.getItems().isEmpty()) {
-                    taskManagerRunning = false;
-                }
+            if (!waitForDeploymentToBeRemoved(namespace, deploymentName, deploymentTimeout)) {
+                LOG.error(
```
mateczagany, line 917: In case of a timeout, we just log the message as an error, but during upgrade this means that the operator will just proceed to create a new `Deployment`, then get `AlreadyExists` error, resulting in a deadlock.

Not sure what else can we do in this case.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
```diff
@@ -898,58 +901,54 @@ public JobDetailsInfo getJobDetailsInfo(JobID jobID, Configuration conf) throws
...
-            }
+    /** Wait until Deployment is removed, return false if timed out, otherwise return true. */
+    @VisibleForTesting
+    boolean waitForDeploymentToBeRemoved(String namespace, String deploymentName, long timeout) {
+        ScheduledExecutorService logger = Executors.newSingleThreadScheduledExecutor();
+        logger.scheduleWithFixedDelay(
+                () -> LOG.info("Waiting for Deployment {} to shut down...", deploymentName),
+                5,
+                5,
+                TimeUnit.SECONDS);
```
gyfora, line 935: This looks a bit strange / heavy to create always a new executor (thread), can we have a shared pool / or hook into the kubernetes client `waitUntilCondition` somehow? Or alternatively we could just remove the periodic logs and log before and after

gyfora (changes requested):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
mateczagany: You are right, I will remove it. I don't think it's worth to keep those logs, since it's easy to figure out what we're waiting for if we have before/after logs as in your second suggestion.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
numbnut: I fully agree, but don't have a solution. At least the default value for the cleanup timeout could be increased to a higher value. What do you think?

mateczagany: I don't think there could be any downside to setting the default cleanup timeout to e.g. 5 minutes. @gyfora WDYT? Should that be done in another PR or should I add that here, or do you think we should leave the timeout as is?

gyfora: If no trouble for you @mateczagany let's add the default timeout increase here. I am not a big fan of too many PRs it's just more work to review :)

gyfora: 5 minutes seems reasonable for prod envs

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
```diff
@@ -898,58 +899,50 @@ public JobDetailsInfo getJobDetailsInfo(JobID jobID, Configuration conf) throws
...
-                                .get();
-                if (service == null) {
-                    serviceRunning = false;
-                }
-            }
+    /** Wait until Deployment is removed, return false if timed out, otherwise return true. */
+    @VisibleForTesting
+    boolean waitForDeploymentToBeRemoved(String namespace, String deploymentName, long timeout) {
+        LOG.info(
+                "Waiting for Deployment {} to shut down with {} ms timeout...",
```
gyfora, line 929: Can we display this timeout log in seconds to be consistent?

gyfora (approved):
aside from the ms vs s logging, looks like a great improvement :) thanks @mateczagany

PR #781, opened 2024-02-22
[FLINK-34471] Tune network memory with memory tuning

mxm:
This PR is a follow-up to #778 where we skipped tuning network memory based on the actual usage and kept it constant instead. In this PR, we calculate the maximum amount of memory required for the network buffers of each TaskManager.

This is done based off the formula in https://nightlies.apache.org/flink/flink-docs-master/docs/deployment/memory/network_mem_tuning/#network-buffer-lifecycle.


1996fanrui (approved):
Thanks @mxm for the contribution!

LGTM.

PR #782, opened 2024-02-23
[FLINK-34502][autoscaler] Support calculating network memory for forward and rescale edge

1996fanrui:

This is follow up Jira of [FLINK-34471](https://issues.apache.org/jira/browse/FLINK-34471) #781 . [FLINK-34471](https://issues.apache.org/jira/browse/FLINK-34471) assuming all connections type are ALL_TO_ALL. This Jira will optimize it to save some network memory for forward and rescale connection.

Following is the optimization logic:

- When the connection type is FORWARD and the parallelism of 2 tasks are same. We calculate network memory based on FORWARD connection.
- When the connection type is FORWARD and the parallelism of 2 tasks are not same. We calculate network memory based on RESCALE connection.
- When the connection type is RESCALE. We calculate network memory based on RESCALE connection.

## Brief change log

[FLINK-34502][autoscaler] Support calculating network memory for forward and rescale edge

- Add the ship_strategy to JobTopology
- Improve the network memory calculating logic for forward and rescale connections.

## Verifying this change

- MemoryTuningTest#testCalculateNetworkSegmentNumber
- JobTopologyTest#testTopologyFromJson check the ship_strategy

## Documentation

### DIFF
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/topology/VertexInfo.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/topology/VertexInfo.java
@@ -22,7 +22,7 @@ import org.apache.flink.runtime.jobgraph.JobVertexID;
 
 import lombok.Data;
 
-import java.util.Set;
+import java.util.Map;
 
 /** Job vertex information. */
 @Data
@@ -30,9 +30,11 @@ public class VertexInfo {
 
     private final JobVertexID id;
 
-    private final Set<JobVertexID> inputs;
+    // All input vertices and the ship_strategy
+    private final Map<JobVertexID, String> inputs;
 
-    private Set<JobVertexID> outputs;
+    // All output vertices and the ship_strategy
+    private Map<JobVertexID, String> outputs;
 
     private final int parallelism;
 
@@ -46,7 +48,7 @@ public class VertexInfo {
 
     public VertexInfo(
             JobVertexID id,
-            Set<JobVertexID> inputs,
+            Map<JobVertexID, String> inputs,
             int parallelism,
             int maxParallelism,
             boolean finished,
@@ -63,7 +65,7 @@ public class VertexInfo {
     @VisibleForTesting
     public VertexInfo(
             JobVertexID id,
-            Set<JobVertexID> inputs,
+            Map<JobVertexID, String> inputs,
             int parallelism,
             int maxParallelism,
             IOMetrics ioMetrics) {
@@ -72,7 +74,7 @@ public class VertexInfo {
 
     @VisibleForTesting
     public VertexInfo(
-            JobVertexID id, Set<JobVertexID> inputs, int parallelism, int maxParallelism) {
+            JobVertexID id, Map<JobVertexID, String> inputs, int parallelism, int maxParallelism) {
         this(id, inputs, parallelism, maxParallelism, null);
     }
 
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/tuning/MemoryTuning.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/tuning/MemoryTuning.java
@@ -255,11 +255,13 @@ public class MemoryTuning {
         for (VertexInfo vertexInfo : jobTopology.getVertexInfos().values()) {
             // Add max amount of memory for each input gate
             for (Map.Entry<JobVertexID, String> inputEntry : vertexInfo.getInputs().entrySet()) {
+                final JobVertexID inputVertexId = inputEntry.getKey();
+                final String shipStrategy = inputEntry.getValue();
                 maxNetworkMemory +=
                         calculateNetworkSegmentNumber(
                                         updatedParallelisms.get(vertexInfo.getId()),
-                                        updatedParallelisms.get(inputEntry.getKey()),
-                                        inputEntry.getValue(),
+                                        updatedParallelisms.get(inputVertexId),
+                                        shipStrategy,
                                         buffersPerChannel,
                                         floatingBuffers)
                                 * memorySegmentBytes;
@@ -267,11 +269,13 @@ public class MemoryTuning {
             // Add max amount of memory for each output gate
             // Usually, there is just one output per task
             for (Map.Entry<JobVertexID, String> outputEntry : vertexInfo.getOutputs().entrySet()) {
+                final JobVertexID outputVertexId = outputEntry.getKey();
+                final String shipStrategy = outputEntry.getValue();
                 maxNetworkMemory +=
                         calculateNetworkSegmentNumber(
                                         updatedParallelisms.get(vertexInfo.getId()),
-                                        updatedParallelisms.get(outputEntry.getKey()),
-                                        outputEntry.getValue(),
+                                        updatedParallelisms.get(outputVertexId),
+                                        shipStrategy,
                                         buffersPerChannel,
                                         floatingBuffers)
                                 * memorySegmentBytes;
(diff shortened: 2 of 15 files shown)


flink-autoscaler/src/main/java/org/apache/flink/autoscaler/tuning/MemoryTuning.java
```diff
@@ -277,6 +286,32 @@ private static MemorySize adjustNetworkMemory(
...
+            String shipStrategy,
+            int buffersPerChannel,
+            int floatingBuffers) {
+        // TODO When the parallelism is changed via the rescale api, the FORWARD may be changed to
+        // RESCALE. This logic may needs to be updated after FLINK-33123.
+        if (currentVertexParallelism == otherVertexParallelism && "FORWARD".equals(shipStrategy)) {
+            return buffersPerChannel + floatingBuffers;
+        } else if ("FORWARD".equals(shipStrategy) || "RESCALE".equals(shipStrategy)) {
+            final int channelCount =
+                    (int) Math.ceil(1.0d * otherVertexParallelism / currentVertexParallelism);
```
mxm, line 308: I prefer the following to avoid adding another constant. 
```suggestion
                    (int) Math.ceil(otherVertexParallelism / (double) currentVertexParallelism);
```

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/tuning/MemoryTuning.java
```diff
@@ -277,6 +286,32 @@ private static MemorySize adjustNetworkMemory(
...
+            int currentVertexParallelism,
+            int otherVertexParallelism,
+            String shipStrategy,
+            int buffersPerChannel,
+            int floatingBuffers) {
+        // TODO When the parallelism is changed via the rescale api, the FORWARD may be changed to
+        // RESCALE. This logic may needs to be updated after FLINK-33123.
+        if (currentVertexParallelism == otherVertexParallelism && "FORWARD".equals(shipStrategy)) {
+            return buffersPerChannel + floatingBuffers;
+        } else if ("FORWARD".equals(shipStrategy) || "RESCALE".equals(shipStrategy)) {
```
mxm, line 306: The shipStragegy in the json plan comes from https://github.com/apache/flink/blob/6c8f3a0799c609d8076f782e5334e389e4d92eee/flink-streaming-java/src/main/java/org/apache/flink/streaming/api/graph/StreamingJobGraphGenerator.java#L1433. 

This is a bit concerning because any custom partitioner which has a `toString()` implementation which returns FORWARD, would run into this code branch. Although unlikely, that would break the network memory code.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/tuning/MemoryTuning.java
```diff
@@ -277,6 +286,32 @@ private static MemorySize adjustNetworkMemory(
...
+    /**
+     * Calculate how many network segment current vertex needs.
+     *
+     * @param currentVertexParallelism The parallelism of current vertex.
+     * @param otherVertexParallelism The parallelism of other vertex.
+     */
+    @VisibleForTesting
+    static int calculateNetworkSegmentNumber(
+            int currentVertexParallelism,
+            int otherVertexParallelism,
```
mxm, line 298:
```suggestion
            int upstreamVertexParallelism,
            int downstreamVertexParallelism,
```

mxm: Can we change the logic to use these consistently? I think that makes reading the code a lot easier. Currently, the parameters change based on the ship strategy.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/tuning/MemoryTuning.java
```diff
@@ -243,28 +244,36 @@ private static MemorySize adjustNetworkMemory(
...
             // Add max amount of memory for each input gate
-            for (JobVertexID input : vertexInfo.getInputs()) {
-                int inputParallelism = updatedParallelisms.get(input);
+            for (Map.Entry<JobVertexID, String> inputEntry : vertexInfo.getInputs().entrySet()) {
                 maxNetworkMemory +=
-                        (inputParallelism * buffersPerChannel + floatingBuffers)
+                        calculateNetworkSegmentNumber(
+                                        updatedParallelisms.get(vertexInfo.getId()),
+                                        updatedParallelisms.get(inputEntry.getKey()),
+                                        inputEntry.getValue(),
```
mxm, line 262: For readability, can we add a variable?

```suggestion
                                        shipStrategy,
```

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/topology/VertexInfo.java
```diff
@@ -22,17 +22,19 @@
...
 
     private final JobVertexID id;
 
-    private final Set<JobVertexID> inputs;
+    // All input vertices and the ship_strategy
+    private final Map<JobVertexID, String> inputs;
 
-    private Set<JobVertexID> outputs;
+    // All output vertices and the ship_strategy
+    private Map<JobVertexID, String> outputs;
```
mxm, line 37: This is a bit of a hack. Fine for now but I would prefer keeping the Set with a wrapper class containing JobVertexID and a field for the ship_strategy.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/tuning/MemoryTuning.java
```diff
@@ -243,28 +244,36 @@ private static MemorySize adjustNetworkMemory(
             Configuration config,
             MemoryBudget memBudget) {
 
-        final long buffersPerChannel =
+        final int buffersPerChannel =
                 config.get(NettyShuffleEnvironmentOptions.NETWORK_BUFFERS_PER_CHANNEL);
-        final long floatingBuffers =
+        final int floatingBuffers =
```
mxm, line 249: The reason this was long is that it's going to be converted to long every time when multiplying with the memory segment size.

mxm (commented):
Thanks @1996fanrui for the optimization!

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/tuning/MemoryTuning.java
1996fanrui: It comes from `org.apache.flink.streaming.runtime.partitioner.StreamPartitioner#toString` instead of `org.apache.flink.api.common.functions.Partitioner`.

IIUC, users implement `org.apache.flink.api.common.functions.Partitioner`. And `CustomPartitionerWrapper` wraps the `Partitioner`. The `CustomPartitionerWrapper#toString` is `CUSTOM`.

I have wrote a demo, and overwrite the toString of custom `Partitioner`. It still shows `CUSTOM`.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/tuning/MemoryTuning.java
1996fanrui: IIUC, if we using the `upstream` and `downstream`, we still need to know which one is currentVertex, which one is otherVertex. For example, rebalance type consider the `otherVertexParallelism` as the channelCount.

Please correct me if my understanding is wrong or I miss anything, thank you

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/tuning/MemoryTuning.java
1996fanrui: Currently, all types in `calculateNetworkSegmentNumber` is int, including: `currentVertexParallelism`, `otherVertexParallelism`, `buffersPerChannel` and `floatingBuffers`.

If we change `buffersPerChannel` and `floatingBuffers` to  long, and don't change `currentVertexParallelism`, `otherVertexParallelism`. It still needs to convert int to long.

How about keeping it as is? Inside of calculateNetworkSegmentNumber is int, we only need to convert int to long when ` * memorySegmentBytes`.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/topology/VertexInfo.java
1996fanrui: `Set<Class<JobVertexID, String>>` has a semantic difference compared to `Map<JobVertexID, String>`. It is possible that there are multiple ship_strategies with the same JobVertexID.

I don't have strong opinion for this. I'm opening for it.

1996fanrui (commented):

I have addressed or replied all your comments, please help take a look in your free time, thanks again.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/topology/VertexInfo.java
mxm: It would depend on the hashCode / equals implementation, but the same issue arrises with Map in case of duplicate JobVertexID, only that they would be silently overwritten. Let's keep it unchanged for now.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/tuning/MemoryTuning.java
mxm: Got it, that makes sense. So the toString() can't be overwritten by users to shadow the built-in ones.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/tuning/MemoryTuning.java
mxm: No objections.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/tuning/MemoryTuning.java
mxm: "Other" just seems unspecific. How about this naming:
```suggestion
            int currentVertexParallelism,
            int connectedVertexParallelism,
```

mxm (approved):
Thanks Rui! Changes look good to me.

PR #783, opened 2024-02-23
[FLINK-34504][autoscaler] Avoid the parallelism adjustment when the upstream shuffle type doesn't have keyBy

1996fanrui:

JobVertexScaler#scale has a optimization: Try to adjust the parallelism such that it divides the number of key groups without a remainder =>  data is evenly spread across subtasks.

It's only useful when the upstream shuffle type has keyBy. We should avoid this optimization when the upstream shuffle type doesn't have keyBy.

## Brief change log

[FLINK-34504][autoscaler] Avoid the parallelism adjustment when the upstream shuffle type doesn't have keyBy

## Verifying this change

- Improve `JobVertexScalerTest#testParallelismComputation`
- Add `JobVertexScalerTest#testParallelismComputationWithKeyedOperator`
- Add `ScalingExecutorTest#testAdjustByMaxParallelism`

## Documentation

### DIFF
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
@@ -289,7 +289,8 @@ public class JobVertexScaler<KEY, Context extends JobAutoScalerContext<KEY>> {
                 // underflow.
                 (int) Math.min(Math.ceil(scaleFactor * currentParallelism), Integer.MAX_VALUE);
 
-        // Cap parallelism at either number of key groups or parallelism limit
+        // Cap parallelism at either maxParallelism(number of key groups or source partitions) or
+        // parallelism upper limit
         final int upperBound = Math.min(maxParallelism, parallelismUpperLimit);
 
         // Apply min/max parallelism
@@ -299,16 +300,16 @@ public class JobVertexScaler<KEY, Context extends JobAutoScalerContext<KEY>> {
             return newParallelism;
         }
 
-        // When the shuffle type of vertex data source contains keyBy, we try to adjust the
-        // parallelism such that it divides the number of key groups without a remainder =>
-        // data is evenly spread across subtasks
+        // When the shuffle type of vertex inputs contains keyBy or vertex is a source, we try to
+        // adjust the parallelism such that it divides the number of key groups without a remainder
+        // => data is evenly spread across subtasks
         for (int p = newParallelism; p <= maxParallelism / 2 && p <= upperBound; p++) {
             if (maxParallelism % p == 0) {
                 return p;
             }
         }
 
-        // If key group adjustment fails, use originally computed parallelism
+        // If parallelism adjustment fails, use originally computed parallelism
         return newParallelism;
     }
 
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
@@ -22,6 +22,7 @@ import org.apache.flink.autoscaler.config.AutoScalerOptions;
 import org.apache.flink.autoscaler.event.AutoScalerEventHandler;
 import org.apache.flink.autoscaler.metrics.EvaluatedScalingMetric;
 import org.apache.flink.autoscaler.metrics.ScalingMetric;
+import org.apache.flink.autoscaler.topology.ShipStrategy;
 import org.apache.flink.autoscaler.utils.AutoScalerUtils;
 import org.apache.flink.configuration.Configuration;
 import org.apache.flink.runtime.jobgraph.JobVertexID;
@@ -48,6 +49,7 @@ import static org.apache.flink.autoscaler.metrics.ScalingMetric.EXPECTED_PROCESS
 import static org.apache.flink.autoscaler.metrics.ScalingMetric.MAX_PARALLELISM;
 import static org.apache.flink.autoscaler.metrics.ScalingMetric.PARALLELISM;
 import static org.apache.flink.autoscaler.metrics.ScalingMetric.TRUE_PROCESSING_RATE;
+import static org.apache.flink.autoscaler.topology.ShipStrategy.HASH;
 
 /** Component responsible for computing vertex parallelism based on the scaling metrics. */
 public class JobVertexScaler<KEY, Context extends JobAutoScalerContext<KEY>> {
@@ -71,7 +73,7 @@ public class JobVertexScaler<KEY, Context extends JobAutoScalerContext<KEY>> {
     public int computeScaleTargetParallelism(
             Context context,
             JobVertexID vertex,
-            boolean adjustByMaxParallelism,
+            Map<JobVertexID, ShipStrategy> inputs,
             Map<ScalingMetric, EvaluatedScalingMetric> evaluatedMetrics,
             SortedMap<Instant, ScalingSummary> history,
             Duration restartTime) {
@@ -122,11 +124,11 @@ public class JobVertexScaler<KEY, Context extends JobAutoScalerContext<KEY>> {
         int newParallelism =
                 scale(
                         currentParallelism,
+                        inputs,
                         (int) evaluatedMetrics.get(MAX_PARALLELISM).getCurrent(),
                         scaleFactor,
                         Math.min(currentParallelism, conf.getInteger(VERTEX_MIN_PARALLELISM)),
-                        Math.max(currentParallelism, conf.getInteger(VERTEX_MAX_PARALLELISM)),
-                        adjustByMaxParallelism);
+                        Math.max(currentParallelism, conf.getInteger(VERTEX_MAX_PARALLELISM)));
 
         if (newParallelism == currentParallelism
                 || blockScalingBasedOnPastActions(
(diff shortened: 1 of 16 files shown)


flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
```diff
@@ -280,11 +283,14 @@ protected static int scale(
...
 
-        // Try to adjust the parallelism such that it divides the number of key groups without a
-        // remainder => state is evenly spread across subtasks
-        for (int p = newParallelism; p <= numKeyGroups / 2 && p <= upperBound; p++) {
-            if (numKeyGroups % p == 0) {
-                return p;
+        // When the shuffle type of vertex data source contains keyBy, we try to adjust the
+        // parallelism such that it divides the number of key groups without a remainder =>
+        // data is evenly spread across subtasks
+        if (hasKeyBy) {
```
1996fanrui, line 289: Hi, @gyfora @mxm . During I test autoscaler, I found the parallelissm adjustment should not take effect for non-keyed operator.

It's better to include this optimization in 1.8.0. Would you mind helping take a look in your free time? thanks in advanced~

BTW, how about rename  `hasKeyBy` to `isKeyedVertex`?

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
```diff
@@ -225,10 +227,14 @@ Map<JobVertexID, ScalingSummary> computeScalingSummary(
                             } else {
                                 var currentParallelism =
                                         (int) metrics.get(ScalingMetric.PARALLELISM).getCurrent();
+
+                                final boolean hasKeyBy =
+                                        jobTopology.get(v).getInputs().containsValue("HASH");
```
gyfora, line 232: It's a bit strange to check string values like `HASH` can we make this an enum so that it's more explicit in the autoscaler code and easier to handle?

1996fanrui:
Hi @gyfora , thank you for the quick review!

I have addressed all you comments. For the string shipStrategy, I extracted a `ShipStrategy` enum to refactor it.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
```diff
@@ -225,10 +228,14 @@ Map<JobVertexID, ScalingSummary> computeScalingSummary(
                             } else {
                                 var currentParallelism =
                                         (int) metrics.get(ScalingMetric.PARALLELISM).getCurrent();
+
+                                final boolean hasKeyBy =
+                                        jobTopology.get(v).getInputs().containsValue(HASH);
```
gyfora, line 233: I assume this also means that sources are also not adjusted which can lead to partition imbalance for Kafka topics. We should either keep the adjustments for sources or remove the whole partition number querying/max parallelism logic that we have during collection. 

Overall I think for sources this makes sense and if we want to remove it that should be a separate ticket.

gyfora: This is related to https://issues.apache.org/jira/browse/FLINK-32119

1996fanrui: Thanks a lot for pointing it out. I forgot it before.

I have updated. The source can adjust parallelism for now. Also, I updated parameter names and the comment for `JobVertexScaler#scale` method, for exampe: changing the numKeyGroups to maxParallelism, and changing hasKeyBy to adjustByMaxParallelism.

Because the numKeyGroups and hasKeyBy aren't exactly for source vertex.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
```diff
@@ -229,13 +229,14 @@ Map<JobVertexID, ScalingSummary> computeScalingSummary(
...
+                                var inputs = jobTopology.get(v).getInputs();
+                                var adjustByMaxParallelism =
+                                        inputs.isEmpty() || inputs.containsValue(HASH);
                                 var newParallelism =
                                         jobVertexScaler.computeScaleTargetParallelism(
                                                 context,
                                                 v,
-                                                hasKeyBy,
+                                                adjustByMaxParallelism,
                                                 metrics,
```
gyfora, line 240: Can we please add a single unit test case to cover this core piece of the logic? Just to make sure we don't mess up sources in the future unintentionally

gyfora (changes requested):

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
1996fanrui: Good suggestion, added~

1996fanrui: Thank you very much for your review, especially on the weekend. ❤️

gyfora (approved):
Looks good! Nice improvement :)

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
```diff
@@ -245,50 +247,69 @@ private boolean detectIneffectiveScaleUp(
...
+     *
+     * @param currentParallelism The current parallelism.
+     * @param maxParallelism The max parallelism for job vertices. It's numKeyGroups by default, and
+     *     it's partition number for kafka source vertex.
+     * @param scaleFactor The scale factor.
+     * @param parallelismLowerLimit The parallelism lower limitation in autoscaler option.
+     * @param parallelismUpperLimit The parallelism upper limitation in autoscaler option.
+     * @param adjustByMaxParallelism True means we need to adjust parallelism according to the
+     *     maxParallelism to ensure the keyGroup or partition evenly.
+     */
```
mxm, line 261: I prefer adding just an explanation of what this method does. There is no additional information which is not already in the method signature. If we want to clarify, e.g. maxParallelism, we can add a dedicated comment for that.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
```diff
@@ -71,6 +71,7 @@ public JobVertexScaler(AutoScalerEventHandler<KEY, Context> autoScalerEventHandl
     public int computeScaleTargetParallelism(
             Context context,
             JobVertexID vertex,
+            boolean adjustByMaxParallelism,
```
mxm, line 74: Can we pass-through the ShipStrategy here? That way we decide not based on some boolean but on the relevant data for making the key group adjustment decision. Also makes unit testing more solid.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/topology/ShipStrategy.java
```diff
@@ -0,0 +1,55 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.autoscaler.topology;
+
+import javax.annotation.Nonnull;
+
+/** The ship strategy between 2 JobVertices. */
+public enum ShipStrategy {
```
mxm, line 23: Would be nice to reuse Flink's ShipStrategyType if possible.

mxm (approved):
Thanks Rui. The changes look good! We should keep an eye on the autoscaling behavior after this change because it's actually a breaking change. It might affect the stability of scaling decisions, but I think it should be fine, potentially even improving some situations.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
mxm: (I'm suggesting to remove the `@param` lines)

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/topology/ShipStrategy.java
mxm: Sounds good! Thanks for looking into it. Let's keep it as-is for now.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
mxm: Thanks!

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
mxm: Awesome, thanks!

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
```diff
@@ -248,25 +250,22 @@ private boolean detectIneffectiveScaleUp(
...
-     * @param adjustByMaxParallelism True means we need to adjust parallelism according to the
-     *     maxParallelism to ensure the keyGroup or partition evenly.
+     * <p>Also, in order to ensure the data is evenly spread across subtasks, we try to adjust the
+     * parallelism for source and keyed vertex such that it divides the maxParallelism without a
+     * remainder.
      */
     @VisibleForTesting
     protected static int scale(
             int currentParallelism,
+            Map<JobVertexID, ShipStrategy> inputs,
```
mxm, line 264: This could be reduced to just `Collection<ShipStrategy>`, we don't need the vertex.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
```diff
@@ -229,14 +228,11 @@ Map<JobVertexID, ScalingSummary> computeScalingSummary(
...
 
-                                var inputs = jobTopology.get(v).getInputs();
-                                var adjustByMaxParallelism =
-                                        inputs.isEmpty() || inputs.containsValue(HASH);
                                 var newParallelism =
                                         jobVertexScaler.computeScaleTargetParallelism(
                                                 context,
                                                 v,
-                                                adjustByMaxParallelism,
+                                                jobTopology.get(v).getInputs(),
```
mxm, line 235: getInputs().values() would suffice.

1996fanrui:
Thanks a lot @mxm for the quick review again, updated.

mxm (approved):

PR #786, opened 2024-02-28
[FLINK-34152] Add an option to scale memory when downscaling

mxm:
This adds an option to increase heap and managed memory when removing TaskManagers. The scaling is applied after adjusting the memory pools (heap, metaspace, network, managed) and only affects heap memory.

The reason for adding this functionality is that the likelihood of running into memory constrained scenarios when downscaling is increased after applying memory tuning. As a precaution, we temporarily increase the total cluster
memory proportionally to the removed TaskManagers up to the maximum allowed TaskManager memory.

### DIFF
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/tuning/MemoryTuning.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/tuning/MemoryTuning.java
@@ -169,6 +169,7 @@ public class MemoryTuning {
         // memory pools, there are no fractional variants for heap memory. Setting the absolute heap
         // memory options could cause invalid configuration states when users adapt the total amount
         // of memory. We also need to take care to remove any user-provided overrides for those.
+        tuningConfig.addRemoval(TaskManagerOptions.TOTAL_FLINK_MEMORY);
         tuningConfig.addRemoval(TaskManagerOptions.TASK_HEAP_MEMORY);
         // Set default to zero because we already account for heap via task heap.
         tuningConfig.addOverride(TaskManagerOptions.FRAMEWORK_HEAP_MEMORY, MemorySize.ZERO);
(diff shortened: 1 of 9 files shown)


flink-autoscaler/src/main/java/org/apache/flink/autoscaler/tuning/MemoryTuning.java
```diff
@@ -148,8 +149,11 @@ public static ConfigChanges tuneTaskManagerHeapMemory(
         final long flinkMemoryDiffBytes = heapDiffBytes + managedDiffBytes + networkDiffBytes;
 
         // Update total memory according to memory diffs
-        final MemorySize totalMemory =
+        MemorySize totalMemory =
                 new MemorySize(maxMemoryBySpec.getBytes() - memBudget.getRemaining());
+        totalMemory =
+                MemoryScaling.applyMemoryScaling(
+                        totalMemory, maxMemoryBySpec, context, scalingSummaries, evaluatedMetrics);
```
gyfora, line 156: The memory scaling config specifies that both managed and heap memory is scaled based on the TM scaling factor. But it seems like we only scale the heap as the managed memory size is computed before this. 

Am I missing something?

mxm: Good catch, the scaling is heap only. Managed memory is never adjusted, unless none is used at all (then it is set to zero) or the option to maximize managed memory is turned on. I've adjusted the docs.

gyfora (approved):

mxm:
@1996fanrui I'm merging but I'll take any comments from your side that you might have reviewing this later on.

1996fanrui:
> @1996fanrui I'm merging but I'll take any comments from your side that you might have reviewing this later on.

Thanks @mxm for the improvement and ping! Sorry for the late response here. I just finished my vacation, I will take a look these 2 days.

PR #787, opened 2024-03-01
feat: autoscaling decision parallelism improvement [FLINK-34563]

Yang-LI-CS:

Hi, it's a minor improvement based on my autoscaling experiments. The concept revolves around identifying the vertex with the highest level of parallelism and matching it to the maximum parallelism supported by our task manager.

The primary goal of this enhancement is to prevent any task slots from remaining unused after the Flink autoscaler performs a rescaling operation. I've already tested this modification in a custom build of the operator, excluding the memory tuning feature. However, I believe it could be beneficial, especially in scenarios where the memory tuning feature is not enabled.

## Brief change log

  - ScalingExecutor scaling decision optimisation

## Verifying this change

This change tested in our flink production environment

## Documentation

### DIFF
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
@@ -44,7 +44,9 @@ import java.time.Duration;
 import java.time.Instant;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
 import java.util.SortedMap;
 
 import static org.apache.flink.autoscaler.config.AutoScalerOptions.EXCLUDED_PERIODS;
@@ -129,6 +131,11 @@ public class ScalingExecutor<KEY, Context extends JobAutoScalerContext<KEY>> {
                         scalingSummaries,
                         autoScalerEventHandler);
 
+        // Increase the biggest vertex's parallelism to the current taskmanager count's
+        // max-parallelism
+        updateBiggestVertexParallelism(
+                context, evaluatedMetrics.getVertexMetrics(), scalingSummaries);
+
         if (scalingWouldExceedClusterResources(
                 configOverrides.applyOverrides(conf),
                 evaluatedMetrics,
@@ -151,6 +158,52 @@ public class ScalingExecutor<KEY, Context extends JobAutoScalerContext<KEY>> {
         autoScalerStateStore.storeConfigChanges(context, configOverrides);
 
         return true;
+    }
+
+    private void updateBiggestVertexParallelism(
+            Context context,
+            Map<JobVertexID, Map<ScalingMetric, EvaluatedScalingMetric>> evaluatedMetrics,
+            Map<JobVertexID, ScalingSummary> scalingSummaries) {
+        int taskSlotsPerTm = context.getConfiguration().get(TaskManagerOptions.NUM_TASK_SLOTS);
+
+        Map<Integer, Set<JobVertexID>> parallelismVertexMap = new HashMap<>();
+        int maxParallelism = 0;
+
+        for (Map.Entry<JobVertexID, ScalingSummary> entry : scalingSummaries.entrySet()) {
+            JobVertexID vertex = entry.getKey();
+            ScalingSummary summary = entry.getValue();
+            int newParallelism = summary.getNewParallelism();
+
+            // Update maxParallelism if a new maximum is found
+            if (newParallelism > maxParallelism) {
+                maxParallelism = newParallelism;
+            }
+
+            // Map newParallelism to JobVertexID
+            parallelismVertexMap.computeIfAbsent(newParallelism, k -> new HashSet<>()).add(vertex);
+        }
+
+        // After the loop, retrieve the JobVertexIDs with the maxParallelism value
+        Set<JobVertexID> verticesWithMaxParallelism =
+                parallelismVertexMap.getOrDefault(maxParallelism, new HashSet<>());
+
+        //  Compute the maximum parallelism for a given taskSlotsPerTm
+        var time = maxParallelism / taskSlotsPerTm;
+        if (maxParallelism > time * taskSlotsPerTm) {
+            maxParallelism = (time + 1) * taskSlotsPerTm;
+        }
+
+        final var finalMaxNewParallelism = maxParallelism;
+
+        scalingSummaries.forEach(
+                (vertexID, summary) -> {
+                    if (verticesWithMaxParallelism.contains(vertexID)) {
+                        // Here, you update the newParallelism for each matching JobVertexID
+                        summary.setNewParallelism(
+                                finalMaxNewParallelism); // Assuming setNewParallelism updates the
+                        // value
+                    }
+                });
     }
 
     private void updateRecommendedParallelism(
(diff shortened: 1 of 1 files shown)


flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
```diff
@@ -153,6 +160,52 @@ public boolean scaleResource(
...
+
+        final var finalMaxNewParallelism = maxParallelism;
+
+        scalingSummaries.forEach(
+                (vertexID, summary) -> {
+                    if (verticesWithMaxParallelism.contains(vertexID)) {
+                        // Here, you update the newParallelism for each matching JobVertexID
+                        summary.setNewParallelism(
+                                finalMaxNewParallelism); // Assuming setNewParallelism updates the
+                        // value
```
mxm, line 204:
```suggestion
```

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
```diff
@@ -129,6 +131,11 @@ public boolean scaleResource(
                         scalingSummaries,
                         autoScalerEventHandler);
 
+        // Increase the biggest vertex's parallelism to the current taskmanager count's
+        // max-parallelism
+        updateBiggestVertexParallelism(
```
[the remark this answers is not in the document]
mxm: There's also parallelism adjustment made to spread out Flink's KeyGroups evenly. The adjustment in this PR could interfere with this logic and lead to uneven spread of state / keys.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
```diff
@@ -153,6 +160,52 @@ public boolean scaleResource(
...
+        }
+
+        final var finalMaxNewParallelism = maxParallelism;
+
+        scalingSummaries.forEach(
+                (vertexID, summary) -> {
+                    if (verticesWithMaxParallelism.contains(vertexID)) {
+                        // Here, you update the newParallelism for each matching JobVertexID
+                        summary.setNewParallelism(
+                                finalMaxNewParallelism); // Assuming setNewParallelism updates the
```
mxm, line 203: I wonder, why only consider the vertex with the maximum parallelism? Why not +1 all the vertices?

mxm (commented):
Hi @Yang-LI-CS! Thank you for the PR. The general idea of using empty task slots is good, but the devil is in the details. First of all, we need to ensure that we apply the adjustment at the right time. Then, it needs to work with slot sharing disabled. Finally, there is already a parallelism correction based on the number of key groups which this feature should not interfere with.

Overall, these constraints still make this feature useful if the constraints are satisfied, but it introduces quite a bit of complexity. I would like to us to take a step back and ask what we will achieve from filling all the available slots. Can we expect TaskManager load to be better balanced or scaling decisions to be more stable?

Yang-LI-CS:
> Hi @Yang-LI-CS! Thank you for the PR. The general idea of using empty task slots is good, but the devil is in the details. First of all, we need to ensure that we apply the adjustment at the right time. Then, it needs to work with slot sharing disabled. Finally, there is already a parallelism correction based on the number of key groups which this feature should not interfere with.
> 
> Overall, these constraints still make this feature useful if the constraints are satisfied, but it introduces quite a bit of complexity. I would like to us to take a step back and ask what we will achieve from filling all the available slots. Can we expect TaskManager load to be better balanced or scaling decisions to be more stable?

Hi @mxm , thanks for the review, 

> it needs to work with slot sharing disabled. 

I am totally agreed with this, thanks for reminding me this and indeed my flink cluster does not use slot sharing

> here is already a parallelism correction based on the number of key groups which this feature should not interfere with

I'm going to review the code to understand this parallelism correction 🙏 

> I would like to us to take a step back and ask what we will achieve from filling all the available slots. Can we expect TaskManager load to be better balanced or scaling decisions to be more stable?




In my use case, the job graph comprises only 6 operators and allocates 6 task slots per task manager. Prior to implementing this improvement, setting the maximum parallelism to 18 resulted in frequent rescaling of my Flink job to various levels of parallelism for all vertex, such as 7, 8, 13, 15, and 16. However, with this enhancement, the Flink job rescales the biggest vertex only to parallelism levels of 6, 12, and 18. While it's true that other vertices may still experience rescaling to parallelism levels like 7, 13, or 15, the overall frequency of rescaling triggered by the Flink autoscaler has significantly decreased.

mxm:
>In my use case, the job graph comprises only 6 operators and allocates 6 task slots per task manager. Prior to implementing this improvement, setting the maximum parallelism to 18 resulted in frequent rescaling of my Flink job to various levels of parallelism for all vertex, such as 7, 8, 13, 15, and 16. However, with this enhancement, the Flink job rescales the biggest vertex only to parallelism levels of 6, 12, and 18. While it's true that other vertices may still experience rescaling to parallelism levels like 7, 13, or 15, the overall frequency of rescaling triggered by the Flink autoscaler has significantly decreased.

I agree that this improvement is beneficial especially for lower-parallelism jobs. I wonder whether it would make sense to align the parallelism with the number of task slots, i.e. have the parallelism always be a multiple of the number of task slots. This could result in more stable metrics because subtasks are equally distributed across the TaskManagers, which should stabilize the metrics for each associated job vertex (task).

For example, if the number of task slots is 6, like in your example, the minimum parallelism would be 6. The next parallelism 12, 18, 24,... That's essentially your idea but generalizing it across all vertices.

The only drawback is that, again, this needs to work with the key group alignment that we perform. Long term, it would probably be smarter to adjust the number of task slots such that they divide the number of key group without a remainder. We can start with adjusting according to multiples of the number of task slots configured whenever we do not perform the key group adjustments (e.g. no shuffle).

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
Yang-LI-CS: Good question, it's interesting, but since I have never tested this configuration in my test, I'll stay with adjustement only for the maximum in this PR :)

Yang-LI-CS:
Hello @mxm , I have updated my pr :

- With necessary adjustments to meet the constraints that render this feature effective.

"I would like to us to take a step back and ask what we will achieve from filling all the available slots. Can we expect TaskManager load to be better balanced or scaling decisions to be more stable?"

=> The improvement I've observed with this PR is the introduction of a fixed scaling step for the Flink cluster, which systematically enhances the scaling process. Excitingly, I'm even considering creating a benchmark table to record the new fixed step daily. This concept was inspired by a similar approach to Flink autoscaling I saw at Netflix ([link](https://www.youtube.com/watch?v=NV0jvA5ZDNc)). From my testing, the Flink cluster underwent fewer rescalings throughout the day after implementing this change.

gyfora:
I have some concerns about this change:
 1. It doesn't work with custom slot sharing configuration which is very common
 2. It provides almost no benefit with large taskmanager sizes / low number of task slots.
 3. It goes against some basic design philosophy in the autoscaler such that we do not scale vertices beyond their target capacity. It ties to @mxm 's question why the logic wouldn't apply to all vertices? 
 
 Taking that one step further why don't we scale all vertices to the same parallelism at that point? That would naturally cause more resource usage and less throughput. By the same logic I don't think we should scale even the largest ones further.

Yang-LI-CS:
> I have some concerns about this change:
> 
> 1. It doesn't work with custom slot sharing configuration which is very common
> 2. It provides almost no benefit with large taskmanager sizes / low number of task slots.
> 3. It goes against some basic design philosophy in the autoscaler such that we do not scale vertices beyond their target capacity. It ties to @mxm 's question why the logic wouldn't apply to all vertices?
> 
> Taking that one step further why don't we scale all vertices to the same parallelism at that point? That would naturally cause more resource usage and less throughput. By the same logic I don't think we should scale even the largest ones further.

@gyfora thanks for the comment! I see your point, indeed it's kind of against the basic design of autoscaler algorithm, I'll close this pr then 👍 , thanks @mxm also for your comments, looking forward to the next flink autoscaling release

mxm:
I think there are some really good ideas in this PR which we might consider in the future, but for the sake of simplicity it might be better to leave things as-is for now. Thanks for your time and effort @Yang-LI-CS!

PR #790, opened 2024-03-05
[FLINK-34566] Pass a FixedThreadPool to set reconciliation parallelism correctly

ysymi:

This pull request pass a fixed size thread pool to JOSDK, so that we can set a large reconciliation parallelism as we except .

## Brief change log

- set a fixed size thread pool as reconciliation executor in  JOSDK configuration

## Verifying this change

*org.apache.flink.kubernetes.operator.FlinkOperatorTest#testConfigurationPassedToJOSDK:*
  - *Extended check logic about reconciliation thread pool creation *

  - Core observer or reconciler logic that is regularly executed: ( yes)

## Documentation

### DIFF
--- a/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/FlinkOperatorTest.java
+++ b/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/FlinkOperatorTest.java
@@ -70,10 +70,21 @@ public class FlinkOperatorTest {
 
         var configService = testOperator.getOperator().getConfigurationService();
 
-        // Test parallelism being passed
+        // Test parallelism being passed expectedly
         var executorService = configService.getExecutorService();
         Assertions.assertInstanceOf(ThreadPoolExecutor.class, executorService);
         ThreadPoolExecutor threadPoolExecutor = (ThreadPoolExecutor) executorService;
+        for (int i = 0; i < testParallelism * 2; i++) {
+            threadPoolExecutor.execute(
+                    () -> {
+                        try {
+                            Thread.sleep(1000);
+                        } catch (InterruptedException e) {
+                            e.printStackTrace();
+                        }
+                    });
+        }
+        Assertions.assertEquals(threadPoolExecutor.getPoolSize(), testParallelism);
         Assertions.assertEquals(threadPoolExecutor.getMaximumPoolSize(), testParallelism);
 
         // Test label selector being passed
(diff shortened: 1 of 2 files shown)


flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/FlinkOperatorTest.java
```diff
@@ -70,10 +70,21 @@ public void testConfigurationPassedToJOSDK() {
...
+        for (int i = 0; i < testParallelism * 2; i++) {
+            threadPoolExecutor.execute(
+                    () -> {
+                        try {
+                            Thread.sleep(1000);
+                        } catch (InterruptedException e) {
+                            e.printStackTrace();
+                        }
+                    });
+        }
```
gyfora, line 86: what do we expect to test here?

ysymi: I want to test whether the number of threads in the executor can reach our expected maximumPoolSize when the executor receives enough tasks.
see assertion in line 87

ysymi: actually, I think code in lines 78 to 85 was ugly, any good suggestion?

gyfora: Maybe we could do a more functional test that will hold even if the service implementation changes. 

1. Set a smaller parallelism to not use soo many threads
2. For 2x the parallelism like you did we execute:
 - Increment atomic counter
 - Sleep for long (lets say 10sec)
 3. Assert the counter == 2x parallelism
 
 Wdyt? It would be good to test that the test actually fails before the change

ysymi: do you mean "increase counter and sleep for long in every task" ? if yes,  "Assert the counter == 2x parallelism" will always fail even if we set a fixed thread pool(what we want to change) 
I think check pool size (which equals really thread number）is functional enough.

gyfora: fine, by me for now

gyfora (approved):

PR #791, opened 2024-03-07
[FLINK-34524] Scale down JM deployment to 0 before deletion

gyfora:

We recently improved the JM deployment deletion mechanism, however it seems like task manager pod deletion can get stuck sometimes for a couple of minutes in native mode if we simply try to delete everything at once.

It speeds up the process and leads to cleaner shutdown if we scale down the JM deployment to 0 (shutting down the JM pods first) and then perform the deletion.

## Verifying this change

 - Manually verified in local Kube env
 - E2Es
 - Unit tests 
 
## Documentation

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
@@ -101,8 +101,12 @@ import org.apache.flink.util.Preconditions;
 import io.fabric8.kubernetes.api.model.DeletionPropagation;
 import io.fabric8.kubernetes.api.model.ObjectMeta;
 import io.fabric8.kubernetes.api.model.PodList;
+import io.fabric8.kubernetes.api.model.apps.Deployment;
 import io.fabric8.kubernetes.client.KubernetesClient;
-import io.fabric8.kubernetes.client.KubernetesClientTimeoutException;
+import io.fabric8.kubernetes.client.KubernetesClientException;
+import io.fabric8.kubernetes.client.dsl.Resource;
+import io.fabric8.kubernetes.client.dsl.Waitable;
+import lombok.SneakyThrows;
 import org.apache.commons.lang3.ObjectUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -112,6 +116,7 @@ import javax.annotation.Nullable;
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
+import java.net.HttpURLConnection;
 import java.net.InetSocketAddress;
 import java.net.MalformedURLException;
 import java.net.Socket;
@@ -119,6 +124,7 @@ import java.net.SocketAddress;
 import java.net.URL;
 import java.nio.file.Files;
 import java.nio.file.Paths;
+import java.time.Duration;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
@@ -127,6 +133,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.concurrent.Callable;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.TimeUnit;
@@ -399,7 +406,6 @@ public abstract class AbstractFlinkService implements FlinkService {
         // Unless we leave the jm around after savepoint, we should wait until it has finished
         // shutting down
         if (deleteClusterAfterSavepoint || upgradeMode != UpgradeMode.SAVEPOINT) {
-            waitForClusterShutdown(conf);
             deploymentStatus.setJobManagerDeploymentStatus(JobManagerDeploymentStatus.MISSING);
         }
     }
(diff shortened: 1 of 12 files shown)


gyfora:
cc @mateczagany , I would love to hear your thoughts on this as well.

I still need to finish up some tests but any feedback is welcome

mxm (commented):
I wonder, what is the difference between scaling the deployment to zero vs removing it? I would think both issue a SIGINT to the container. If that is the case, there should be the difference in shutdown behavior.

mxm:
Ah, got it! That makes sense.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
```diff
@@ -1134,4 +1141,28 @@ private Configuration getOperatorRestConfig(Configuration origConfig) throws IOE
                         });
         return conf;
     }
+
+    protected void scaleJmToZero(
+            EditReplacePatchable<Deployment> jmDeployment, String namespace, String clusterId) {
+        LOG.info("Scaling down JM deployment to 0 before deletion");
```
mateczagany, line 1147: Since we also have a possible timeout here, maybe it's worth adding the duration of the timeout in the log

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
```diff
@@ -1134,4 +1141,28 @@ private Configuration getOperatorRestConfig(Configuration origConfig) throws IOE
...
+        try {
+
+            // We use patching instead of scaling to avoid the need for new permissions
+            var patch = new DeploymentBuilder().editOrNewSpec().withReplicas(0).endSpec().build();
+            jmDeployment.patch(PatchContext.of(PatchType.JSON_MERGE), patch);
+            kubernetesClient
+                    .pods()
+                    .inNamespace(namespace)
+                    .withLabels(KubernetesUtils.getJobManagerSelectors(clusterId))
```
mateczagany, line 1156: `getJobManagerSelectors()` will not work for standalone clusters, as the `type` label will be different. This should be `StandaloneKubernetesUtils#getJobManagerSelectors` in that case.

mateczagany (changes requested):
Added some comments, but I really like this change, I think it makes a lot of sense to let the JMs shut down the TMs themselves in native mode.

But I wonder, how much difference will it make in standalone mode? Is it worth adding it there as well?

gyfora:
> Added some comments, but I really like this change, I think it makes a lot of sense to let the JMs shut down the TMs themselves in native mode.
> 
> But I wonder, how much difference will it make in standalone mode? Is it worth adding it there as well?

The logic can be the same in standalone as well, right now we are issuing the delete request for both TM and JM deployments directly one after the other. There is a good chance that the TM shuts down and sends an error signal to the JM before that terminates.

In standalone the alternative would be simply to wait for the JM deployment to be deleted before deleting the task managers.

I am working on some cleanup / improvements in this area to make this nicer

gyfora:
@mateczagany I actually reworked the deletion flow and moved some code around to remove some duplicated logic and the unnecessary decoupling of deletion vs waiting.

This leads to a much cleaner logic and avoids using the scale to zero for the standalone mode as well while keeping the benefits of this logic.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/NativeFlinkService.java
```diff
@@ -356,4 +353,32 @@ public boolean scalingCompleted(FlinkResourceContext<?> ctx) {
...
+            // We use patching instead of scaling to avoid the need for new permissions
+            var patch = new DeploymentBuilder().editOrNewSpec().withReplicas(0).endSpec().build();
+            jmDeployment.patch(PatchContext.of(PatchType.JSON_MERGE), patch);
+            kubernetesClient
+                    .pods()
+                    .inNamespace(namespace)
+                    .withLabels(KubernetesUtils.getJobManagerSelectors(clusterId))
+                    .waitUntilCondition(
+                            Objects::isNull,
+                            operatorConfig.getFlinkShutdownClusterTimeout().getSeconds(),
```
mateczagany, line 374: I think this should be `timeoutMillis`

mateczagany (approved):
Just one small comment, but overall this seems like a great improvement compared to the last solution

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/NativeFlinkService.java
gyfora: fixed, factored out the shared logic

gyfora:
The new and improved / consistent logging after the rework @mateczagany @mxm :

```
[INFO ][default/basic-checkpoint-ha-example] >>> Event  | Info    | CLEANUP         | Cleaning up FlinkDeployment
[INFO ][default/basic-checkpoint-ha-example] Cleaning up autoscaling meta data
[INFO ][default/basic-checkpoint-ha-example] Job is running, cancelling job.
[INFO ][default/basic-checkpoint-ha-example] Job successfully cancelled.
[INFO ][default/basic-checkpoint-ha-example] Deleting cluster with Foreground propagation
[INFO ][default/basic-checkpoint-ha-example] Scaling JobManager Deployment to zero with 300 seconds timeout...
[INFO ][default/basic-checkpoint-ha-example] Completed Scaling JobManager Deployment to zero
[INFO ][default/basic-checkpoint-ha-example] Deleting JobManager Deployment with 298 seconds timeout...
[INFO ][default/basic-checkpoint-ha-example] Completed Deleting JobManager Deployment
[INFO ][default/basic-checkpoint-ha-example] Deleting Kubernetes HA metadata

```

mxm:
@gyfora Nice! LGTM

mateczagany (approved):

PR #795, opened 2024-03-11
[FLINK-34152][autoscaler] Refactor the MemoryTuning to avoid the huge method

1996fanrui:

MemoryTuning#tuneTaskManagerMemory method is too long, the code line number is  135.

https://github.com/apache/flink-kubernetes-operator/blob/726e484c6a9b4121563829bc094b3eebeb8ddcf3/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/tuning/MemoryTuning.java#L80

## Brief change log

- [FLINK-34152][autoscaler] Refactor the MemoryTuning to avoid the huge method
  - Extracted some detailed implementations to the separate method.

After refactor, the code line number is changed from 135 to 53.

### DIFF
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/tuning/MemoryTuning.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/tuning/MemoryTuning.java
@@ -46,6 +46,8 @@ import org.apache.flink.runtime.util.config.memory.taskmanager.TaskExecutorFlink
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import javax.annotation.Nonnull;
+
 import java.math.BigDecimal;
 import java.math.RoundingMode;
 import java.util.Arrays;
@@ -78,6 +80,11 @@ public class MemoryTuning {
             JobTopology jobTopology,
             Map<JobVertexID, ScalingSummary> scalingSummaries,
             AutoScalerEventHandler eventHandler) {
+        MemorySize maxMemoryBySpec = context.getTaskManagerMemory().orElse(MemorySize.ZERO);
+        if (maxMemoryBySpec.compareTo(MemorySize.ZERO) <= 0) {
+            LOG.warn("Spec TaskManager memory size could not be determined.");
+            return EMPTY_CONFIG;
+        }
 
         // Please note that this config is the original configuration created from the user spec.
         // It does not contain any already applied overrides.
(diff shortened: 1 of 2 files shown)


flink-autoscaler/src/main/java/org/apache/flink/autoscaler/tuning/TuningSimpleMemorySpec.java
```diff
@@ -0,0 +1,58 @@
...
+import org.apache.flink.runtime.util.config.memory.CommonProcessMemorySpec;
+import org.apache.flink.runtime.util.config.memory.taskmanager.TaskExecutorFlinkMemory;
+
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+
+/** The task manager memory spec for memory tuning. */
+@AllArgsConstructor
+@Getter
+public class TuningSimpleMemorySpec {
```
1996fanrui, line 30: I'm not sure whether the current class name is good.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/tuning/MemoryTuning.java
```diff
@@ -92,24 +99,51 @@ public static ConfigChanges tuneTaskManagerMemory(
...
+    @Nonnull
+    private static TuningSimpleMemorySpec generateTunedMemorySpec(
+            CommonProcessMemorySpec<TaskExecutorFlinkMemory> memSpecs,
+            JobAutoScalerContext<?> context,
+            EvaluatedMetrics evaluatedMetrics,
+            JobTopology jobTopology,
+            Map<JobVertexID, ScalingSummary> scalingSummaries,
+            UnmodifiableConfiguration config,
+            TuningSimpleMemorySpec originalSimpleSpec,
+            MemoryBudget memBudget) {
```
mxm, line 146: I might be biased, but after this PR, the logic is much harder to understand. Slightly longer methods are not bad per-se. Using methods is actually tricky and the precise reason why the code is written the way it was written. I considered using methods but figured it would make the code harder to read because of the many dependencies which would go into the method signatures.

1996fanrui: Thanks for the clarification!

I'm worried that this method will become longer in the future and hard to read. I don't have strong opinion on this refactoring. Let me close this PR, thanks~

mxm: Thanks Rui! I'm not against refactoring the code. I just wasn't convinced this change made the code easier to understand. Thank you for listening to my concerns.

1996fanrui: Thank you, we can do it in the future if we think it's needed.

PR #796, opened 2024-03-11
[FLINK-34642] Add a test for zero lower utilization boundary

mxm:
Users recently reported that scale down as "broken". Turned out that `target.utilization` and `target.utilization.boundary` were set to the same value. This prevents scale down entirely, which is expected and desired.

This PR adds a test to verify this behavior to avoid breaking this in the future.


flink-autoscaler/src/test/java/org/apache/flink/autoscaler/ScalingExecutorTest.java
```diff
@@ -146,6 +150,55 @@ op1, evaluated(1, 70, 100),
...
+                                expectedParallelism,
+                                evaluated.getVertexMetrics().get(vertex)));
+        assertTrue(
+                ScalingExecutor.allVerticesWithinUtilizationTarget(
+                        evaluated.getVertexMetrics(), scalingSummary));
+
+        // Execute the full scaling path
+        HashMap<JobVertexID, SortedMap<Instant, ScalingSummary>> scalingHistory = new HashMap<>();
+        Instant now = Instant.now();
+        JobTopology jobTopology =
```
gyfora, line 189: you could use `var` on the left hand sides to make this shorter and consistent with the rest of the method

gyfora (approved):
Added a small code style comment but otherwise looks good

PR #799, opened 2024-03-19
[FLINK-34743][autoscaler] Memory tuning takes effect even if the parallelism isn't changed

1996fanrui:

Currently, the memory tuning related logic is only called when the parallelism is changed.

See ScalingExecutor#scaleResource to get more details.

It's better to let the memory tuning takes effect even if the parallelism isn't changed. For example, one flink job runs with desired parallelisms, but it wastes memory.

## Brief change log

[FLINK-34743][autoscaler] Memory tuning takes effect even if the parallelism isn't changed

- Do scaling when parallelism or memory configuration is changed.

## Verifying this change

Manually test done.

The unit test is writing.

## Documentation

### DIFF
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
@@ -214,11 +214,11 @@ public class JobAutoScalerImpl<KEY, Context extends JobAutoScalerContext<KEY>>
             return;
         }
 
-        var parallelismChanged =
+        var parallelismOrConfChanged =
                 scalingExecutor.scaleResource(
                         ctx, evaluatedMetrics, scalingHistory, scalingTracking, now, jobTopology);
 
-        if (parallelismChanged) {
+        if (parallelismOrConfChanged) {
             autoscalerMetrics.incrementScaling();
         } else {
             autoscalerMetrics.incrementBalanced();
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
@@ -105,20 +105,15 @@ public class ScalingExecutor<KEY, Context extends JobAutoScalerContext<KEY>> {
                 computeScalingSummary(
                         context, evaluatedMetrics, scalingHistory, restartTime, jobTopology);
 
+        var parallelismChanged = true;
         if (scalingSummaries.isEmpty()) {
             LOG.info("All job vertices are currently running at their target parallelism.");
-            return false;
+            parallelismChanged = false;
         }
 
         if (allVerticesWithinUtilizationTarget(
                 evaluatedMetrics.getVertexMetrics(), scalingSummaries)) {
-            return false;
-        }
-
-        updateRecommendedParallelism(evaluatedMetrics.getVertexMetrics(), scalingSummaries);
-
-        if (checkIfBlockedAndTriggerScalingEvent(context, scalingSummaries, conf, now)) {
-            return false;
+            parallelismChanged = false;
         }
 
         var configOverrides =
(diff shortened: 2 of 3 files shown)


mxm (commented):
This change requires more work. It may work well if memory tuning is disabled because it will provide continuous memory recommendations. However, if memory tuning is enabled, it will constantly restart the application with new spec updates. 

Memory tuning is currently coupled to autoscaling for a reason. We will have to have thresholds similarly to the ones the autoscaler has. Additionally, I think we want to introduce logic to minimize the number of scalings and make sure autoscaling and memory tuning work together collaboratively. For example, in most situations we may have to scale up for reconfiguring memory because of the downtime and the associated SLO metrics like autoscaler catch up time and restart time.

1996fanrui:
Thanks @mxm  for the quick review and valuable suggestion!

I originally planned to add a memory scaling threshold in a subsequent PR (It's better to be done in 1.9.0). If you think it's needed in this PR, let me do it recently.

How about reuse the `AutoScalerOptions#TARGET_UTILIZATION_BOUNDARY` option, it's 0.3 by default, it means we will ignore this tuning when the total memory saving less than 30%.

I prefer reuse option because it can lower the user's threshold for use. WDYT?

mxm:
This PR would introduce a breaking change to the autoscaling logic, effectively rendering autoscaling meaningless because we would constantly scale. Autoscaling wouldn't have a chance to realize its SLOs. Even with the thresholds, the autoscaling logic would be fundamentally changed because autoscaling loses control over the rescaling behavior.

For this feature to be mergable, it will either have to be disabled by default (opt-in via config), or be integrated with autoscaling, i.e. figure out a way to balance tuning / autoscaling decisions and feed back tuning decisions to the autoscaling algorithm to scale up whenever we redeploy for memory changes to avoid falling behind and preventing autoscaling to scale up after downtime due to memory reconfigurations.

We need to be extremely careful with these kind of changes. Soon hundreds of existing autoscaled jobs will be using autotuning. If we break the scaling behavior, there is a serious risks of impacting all of these pipelines. With more and more users, there is an additional responsibility to make sure autoscaling / autotuning does not cause major downtime for users.

1996fanrui:
> Autoscaling wouldn't have a chance to realize its SLOs. 

You are right. Autoscaler supports scaling parallelism and memory for now. As I understand, the downtime cannot be guaranteed even if users only use scaling parallelism. For example, flink jobs don't use the Adaptive Scheduler and the input rate is always changed, then flink jobs will be scaled frequently.

Fortunately, scaling parallelism consider the restart time than scaling memory, and then increase some parallelisms.

> For this feature to be mergable, it will either have to be disabled by default (opt-in via config)

IIUC, `job.autoscaler.memory.tuning.enabled` is disabled by default. It means the memory tuning is turned off by default even if this PR is merged, right?

> or be integrated with autoscaling, i.e. figure out a way to balance tuning / autoscaling decisions and feed back tuning decisions to the autoscaling algorithm to scale up whenever we redeploy for memory changes to avoid falling behind and preventing autoscaling to scale up after downtime due to memory reconfigurations.

The restartTime has been considered during `computeScalingSummary`, but we may ignore it due to the new parallelism is `WithinUtilizationTarget`. Do you mean we force adjust the parallelism to the new parallelism when scaling memory happens even if the new parallelism `WithinUtilizationTarget`?

> We need to be extremely careful with these kind of changes. Soon hundreds of existing autoscaled jobs will be using autotuning. If we break the scaling behavior, there is a serious risks of impacting all of these pipelines. With more and more users, there is an additional responsibility to make sure autoscaling / autotuning does not cause major downtime for users.

Yes, I totally agree with you. We need to control this risk to avoid effect users.

mxm:
> > Autoscaling wouldn't have a chance to realize its SLOs.
> 
> You are right. Autoscaler supports scaling parallelism and memory for now. As I understand, the downtime cannot be guaranteed even if users only use scaling parallelism. For example, flink jobs don't use the Adaptive Scheduler and the input rate is always changed, then flink jobs will be scaled frequently.

I agree that there are edge cases where the autoscaler cannot fulfill its service objectives. However, that doesn't mean we need to give up on them entirely. With restarts due to autotuning at any point in time, the autoscaling algorithm is inherently broken because downtime is never factored into the autoscaling decision. 

You mentioned the adaptive scheduler. Frankly, the use of the adaptive scheduler with autoscaling isn't fully developed. I would discourage users from using it with autoscaling at its current state.

> Fortunately, scaling parallelism consider the restart time than scaling memory, and then increase some parallelisms.

+1

> 
> > For this feature to be mergable, it will either have to be disabled by default (opt-in via config)
> 
> IIUC, `job.autoscaler.memory.tuning.enabled` is disabled by default. It means the memory tuning is turned off by default even if this PR is merged, right?

Autoscaling is also disabled by default. I think we want to make sure autoscaling and autotuning work together collaboratively.

> 
> > or be integrated with autoscaling, i.e. figure out a way to balance tuning / autoscaling decisions and feed back tuning decisions to the autoscaling algorithm to scale up whenever we redeploy for memory changes to avoid falling behind and preventing autoscaling to scale up after downtime due to memory reconfigurations.
> 
> The restartTime has been considered during `computeScalingSummary`, but we may ignore it due to the new parallelism is `WithinUtilizationTarget`. Do you mean we force adjust the parallelism to the new parallelism when scaling memory happens even if the new parallelism `WithinUtilizationTarget`?

True, the rescale time has been considered for the downscale / upscale processing capacity, but the current processing capacity doesn't factor in downtime. Unplanned restarts would reduce the processing capacity. If we know we are going to restart, the autoscaling algorithm should factor this in, e.g. by reducing the calculated processing capacity accordingly.

1996fanrui:
Thanks @mxm for the patient reply and clarification!

> You mentioned the adaptive scheduler. Frankly, the use of the adaptive scheduler with autoscaling isn't fully developed. I would discourage users from using it with autoscaling at its current state.

+1, I agree with you.

> the rescale time has been considered for the downscale / upscale processing capacity, but the current processing capacity doesn't factor in downtime. 

Sorry, I didn't notice that. I thought the restart time has been considered for any cases before. Thanks you for pointing it out, I will check later.

Also, do you have any other specific concerns if we allow `memory tuning takes effect even if the parallelism isn't changed`? I could try to solve all of them in this PR.

mxm:
> Also, do you have any other specific concerns if we allow `memory tuning takes effect even if the parallelism isn't changed`? I could try to solve all of them in this PR.

For decoupling autotuning, I would prefer if users would have to opt-in via a config for now, until we are sure both autotuning and autoscaling work together cooperatively.

1996fanrui:
> For decoupling autotuning, I would prefer if users would have to opt-in via a config for now, until we are sure both autotuning and autoscaling work together cooperatively.

Thanks @mxm for the patient reply, it makes sense to me. Let us do it later, I will close this PR first, and we can pick it up after we think it's the proper time to do.

PR #800, opened 2024-03-19
[FLINK-34744][autoscaler] Fix the issue that autoscaling-dynamic cannot run

1996fanrui:
autoscaling-dynamic cannot run on my Local

I guess yaml cannot hanlde the `\n` properly with multiple lines. 

It works after changing them to one line.

image


mxm (approved):
Thanks for the fix! This may be a Flink issue, not correctly escaping the inputs in the bash startup scripts.

1996fanrui (commented):

Merging~

PR #801, opened 2024-03-21
[FLINK-34906] Only scale when all tasks are running

1996fanrui:

Currently, the autoscaler will scale a job when the JobStatus is RUNNING. But the JobStatus will be RUNNING once job starts schedule, so it doesn't mean all tasks are running. Especially, when the resource isn't enough or job recovers from large state.

The autoscaler will throw exception and generate the AutoscalerError event when tasks are not ready. Also, we don't need to scale it when some tasks are not ready.

## Brief change log

- [FLINK-34906] Only scale when all tasks are running
  - Solution: we only scale job that all tasks are running(some of tasks may be finished). 

## Verifying this change

- Manually test is done
- Added the `JobStatusUtilsTest`

## Documentation

### DIFF
--- a/flink-autoscaler-standalone/src/test/java/org/apache/flink/autoscaler/standalone/flinkcluster/FlinkClusterJobListFetcherTest.java
+++ b/flink-autoscaler-standalone/src/test/java/org/apache/flink/autoscaler/standalone/flinkcluster/FlinkClusterJobListFetcherTest.java
@@ -21,10 +21,13 @@ import org.apache.flink.api.common.JobID;
 import org.apache.flink.api.common.JobStatus;
 import org.apache.flink.client.program.rest.RestClusterClient;
 import org.apache.flink.configuration.Configuration;
-import org.apache.flink.runtime.client.JobStatusMessage;
+import org.apache.flink.runtime.execution.ExecutionState;
 import org.apache.flink.runtime.highavailability.nonha.standalone.StandaloneClientHAServices;
+import org.apache.flink.runtime.messages.webmonitor.JobDetails;
+import org.apache.flink.runtime.messages.webmonitor.MultipleJobsDetails;
 import org.apache.flink.runtime.rest.messages.ConfigurationInfo;
 import org.apache.flink.runtime.rest.messages.JobMessageParameters;
+import org.apache.flink.runtime.rest.messages.JobsOverviewHeaders;
 import org.apache.flink.runtime.rest.messages.MessageHeaders;
 import org.apache.flink.runtime.rest.messages.MessageParameters;
 import org.apache.flink.runtime.rest.messages.RequestBody;
@@ -38,13 +41,12 @@ import org.junit.jupiter.api.Test;
 import javax.annotation.Nullable;
 
 import java.time.Duration;
-import java.time.Instant;
-import java.util.Collection;
-import java.util.List;
+import java.util.Arrays;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.stream.Collectors;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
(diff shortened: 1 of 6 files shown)


gyfora:
This issue only affects the standalone autoscaler as the kubernetes operator has this logic already in place for setting the RUNNING state. Can we somehow deduplicate this logic?

1996fanrui:
> This issue only affects the standalone autoscaler as the kubernetes operator has this logic already in place for setting the RUNNING state. Can we somehow deduplicate this logic?

Thanks @gyfora provide this information, I didn't notice kubernetes operator has this logic before. Let me look into it.

1996fanrui (commented):
Thanks @gyfora for the suggestion!

I have updated this PR and tested locally, it works well.

flink-autoscaler-standalone/src/test/java/org/apache/flink/autoscaler/standalone/flinkcluster/FlinkClusterJobListFetcherTest.java
```diff
@@ -231,6 +214,22 @@ CompletableFuture<P> sendRequest(M h, U p, R r) {
                             return (CompletableFuture<P>)
                                     CompletableFuture.completedFuture(
                                             configurationsOrException.left().get(jobID));
+                        } else if (h instanceof JobsOverviewHeaders) {
```
1996fanrui, line 217: Changing this file due to we start request `JobsOverviewHeaders` instead of `listJobs` in this PR.

mxm (commented):
This looks related to https://github.com/apache/flink-kubernetes-operator/pull/699 which took a different approach by ignoring certain exceptions during the stabilization phase and effectively postponing metric collection. 

>This issue only affects the standalone autoscaler as the kubernetes operator has this logic already in place for setting the RUNNING state. Can we somehow deduplicate this logic?

Is that really the case? AFAIK we only check for a RUNNING job state.

1996fanrui:
Thanks @mxm for the review and discussion!

> > This issue only affects the standalone autoscaler as the kubernetes operator has this logic already in place for setting the RUNNING state. Can we somehow deduplicate this logic?
> 
> Is that really the case? AFAIK we only check for a RUNNING job state.

`AbstractFlinkService#getEffectiveStatus` adjusts the `JobStatus.RUNNING` to `JobStatus.CREATED`, thanks @gyfora for helping find it. I didn't extract it as a common class due to @gyfora mentioned `autoscaler` may be moved to the separated repo, so it's better to copy related logic to `autoscaler standalone` module.

> This looks related to #699 which took a different approach by ignoring certain exceptions during the stabilization phase and effectively postponing metric collection.

The adjustment logic is introduced before #699 , it means the some of metrics may be not ready even if all tasks are running(I guess some metrics are generated after running). That's what exactly what #699  solved.

Why do we need to adjust the JobStatus?

- If some of tasks are not running, autoscaler doesn't need to call metric collection related logic.
- If `job.autoscaler.stabilization.interval` is set to small value by users, it's easy to throw metric not found exception.
- As I understand, `job.autoscaler.stabilization.interval` hopes to filter out unstable metrics when all tasks just start running. 
  - For example, job starts at `09:00:00`, and all tasks start running at `09:03:00`, and  `job.autoscaler.stabilization.interval` = 1 min.
  - We hopes the stabilization period is `09:03:00` to `09:04:00` instead of `09:00:00` to `09:01:00`, right?
  - All tasks starts since `09:03:00`, so the metric may be not stable from `09:03:00` to `09:04:00`.
  - Of course, this issue might needs FLINK-34907 as well.

Please correct me if anything is wrong, thanks a lot.

mxm:
Thanks Rui! The changes make sense to me. To Gyulas point, I think we should try to deduplicate the logic such that both Kubernetes autoscaler and standalone use the same code path.

1996fanrui:
> Thanks Rui! The changes make sense to me. To Gyulas point, I think we should try to deduplicate the logic such that both Kubernetes autoscaler and standalone use the same code path.

I could move `JobStatusUtils` from `flink-autoscaler-standalone` module to `flink-autoscaler` module, then both of `flink-autoscaler-standalone` and `flink-kubernetes-operator` module can use it.

Hey @gyfora , what do you think about it?

gyfora:
sounds good @1996fanrui

1996fanrui (commented):
Hey @gyfora @mxm , do you have any comments on this PR?

gyfora (approved):

1996fanrui:
Hi @mxm , I merge this PR first, feel free to ping me if you have any comment in the future, thanks~

PR #802, opened 2024-03-22
[hotfix] Correct the doc related to restart time tracking

1996fanrui:
## What is the purpose of the change

The doc is updated after `job.autoscaler.restart.time-tracking.enabled` is supported.

## Brief change log

[hotfix] Correct the doc related to restart time tracking

PR #803, opened 2024-03-25
[hotfix] Correct the doc related to restart time tracking

1996fanrui:
Backport #802  to 1.8, the doc can be updated timely

### DIFF
--- a/docs/content/docs/custom-resource/autoscaler.md
+++ b/docs/content/docs/custom-resource/autoscaler.md
@@ -162,12 +162,14 @@ Setting `job.autoscaler.target.utilization.boundary: "0.2"` means that we allow 
 ### Target catch-up duration and restart time
 
 When taking scaling decisions the operator need to account for the extra capacity required to catch up the backlog created during scaling operations.
-The amount of extra capacity is determined automatically by the following 2 configs:
+The amount of extra capacity is determined automatically by the following configs:
 
- - `job.autoscaler.restart.time` : Time it usually takes to restart the application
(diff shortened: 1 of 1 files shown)


docs/content/docs/custom-resource/autoscaler.md
```diff
@@ -162,12 +162,14 @@ Setting `job.autoscaler.target.utilization.boundary: "0.2"` means that we allow
 ### Target catch-up duration and restart time
 
 When taking scaling decisions the operator need to account for the extra capacity required to catch up the backlog created during scaling operations.
-The amount of extra capacity is determined automatically by the following 2 configs:
+The amount of extra capacity is determined automatically by the following 3 configs:
```
gyfora, line 165: maybe we can remove the number simply :
` determined automatically by the following configs`

gyfora (approved):
aside from a minor comment, looks good

docs/content/docs/custom-resource/autoscaler.md
gyfora: Oh this is only a backport, then nevermind :)

1996fanrui: > Oh this is only a backport, then nevermind :)

Sorry, I didn't notice this comment.

If so, the 1.8 doc could remove this number, and we don't need to update the doc for main branch.

mxm (approved):

1996fanrui:
Thanks for the review, merging~

PR #808, opened 2024-03-28
[FLINK-34957][autoscaler] Event handler records the exception stack trace when exception message is null

1996fanrui:

JDBC Autoscaler event handler doesn't allow the event message is null, but the message may be null when we handle the exception.

We consider the exception message as the event message, but the exception message may be null, such as: TimeoutException. (It has been shown in following picture.)

Also, recording a event without any message is meaningless. It doesn't have any benefit for troubleshooting.

## Brief change log

- [FLINK-34957][autoscaler] Event handler records the exception stack trace when exception message is null
  - Consider the exception message as the event message when exception message isn't null
  - The whole Exception as the event message if exception message is null.

## Verifying this change

Test manually, it works well. We can see the detailed exception from the `LoggingEventHandler`.

image

## Documentation

### DIFF
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/event/AutoScalerEventHandler.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/event/AutoScalerEventHandler.java
@@ -22,6 +22,9 @@ import org.apache.flink.autoscaler.JobAutoScalerContext;
 import org.apache.flink.autoscaler.ScalingSummary;
 import org.apache.flink.runtime.jobgraph.JobVertexID;
 
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.exception.ExceptionUtils;
+
 import javax.annotation.Nullable;
 
 import java.time.Duration;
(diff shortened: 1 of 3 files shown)


flink-autoscaler/src/main/java/org/apache/flink/autoscaler/event/AutoScalerEventHandler.java
```diff
@@ -63,6 +65,19 @@ void handleEvent(
...
 
+    /**
+     * Handle exception, and the exception event is warning type and don't deduplicate by default.
+     */
+    default void handleException(Context context, String reason, Throwable e) {
+        var message = e.getMessage();
+        if (message == null) {
+            var stream = new ByteArrayOutputStream();
+            e.printStackTrace(new PrintStream(stream));
+            message = stream.toString();
```
gyfora, line 76: Can be replaced with `ExceptionUtils.getStackTrace(e)` from apache commons  or ExceptionUtis.stringifyException from flink utils

gyfora: we should also set some reasonable (relatively small) limit on the string size here otherwise we might get an error when inserting the Kube event / db column

1996fanrui: Thanks @gyfora for the quick review and good suggestion!

> Can be replaced with ExceptionUtils.getStackTrace(e) from apache commons or ExceptionUtis.stringifyException from flink utils

Done~

> we should also set some reasonable (relatively small) limit on the string size here otherwise we might get an error when inserting the Kube event / db column

I saw the default value of `kubernetes.operator.exception.stacktrace.max.length` is 2048, so I use it here.

And I didn't extract an autoscaler option for it unless it's needed in the future.

gyfora (approved):

1996fanrui (commented):
CI is green, merging~

PR #811, opened 2024-04-02
[FLINK-35006] Use try with-resource for StandaloneAutoscalerExecutor

plugatarev:

*Use try with-resource for StandaloneAutoscalerExecutor*

## Documentation

### DIFF
--- a/flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/StandaloneAutoscalerEntrypoint.java
+++ b/flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/StandaloneAutoscalerEntrypoint.java
@@ -60,9 +60,11 @@ public class StandaloneAutoscalerEntrypoint {
 
         var autoScaler = createJobAutoscaler(eventHandler, stateStore);
 
-        var autoscalerExecutor =
-                new StandaloneAutoscalerExecutor<>(conf, jobListFetcher, eventHandler, autoScaler);
-        autoscalerExecutor.start();
+        try (var autoscalerExecutor =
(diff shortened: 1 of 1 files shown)


plugatarev:
Hi @1996fanrui , please help take a look in your free time, thanks~

flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/StandaloneAutoscalerEntrypoint.java
```diff
@@ -60,9 +60,11 @@ public static <KEY, Context extends JobAutoScalerContext<KEY>> void main(String[
...
         var autoScaler = createJobAutoscaler(eventHandler, stateStore);
 
-        var autoscalerExecutor =
-                new StandaloneAutoscalerExecutor<>(conf, jobListFetcher, eventHandler, autoScaler);
-        autoscalerExecutor.start();
+        try (var autoscalerExecutor =
+                new StandaloneAutoscalerExecutor<>(
+                        conf, jobListFetcher, eventHandler, autoScaler)) {
+            autoscalerExecutor.start();
+        }
```
1996fanrui, line 67: Hey @plugatarev , thanks for the PR!

IIUC, we cannot close `StandaloneAutoscalerExecutor`. We expect StandaloneAutoscalerExecutor scaling jobs periodically. It will stop work if we close executor.

plugatarev: Yes, you are right

PR #814, opened 2024-04-15
[FLINK-35105][autoscaler] Support setting default Autoscaler options at autoscaler standalone level

1996fanrui:

Currently, autoscaler standalone doesn't support set [autoscaler options](https://nightlies.apache.org/flink/flink-kubernetes-operator-docs-release-1.8/docs/operations/configuration/#autoscaler-configuration). We must set them at job level when we use autoscaler standalone. It's not convenient if platform administrator wanna change the default value for some autoscaler options, such as:

- job.autoscaler.enabled
- job.autoscaler.metrics.window
- etc

## Brief change log

- [FLINK-35105][autoscaler] Support setting default Autoscaler options at autoscaler standalone level
  - This Jira supports setting Autoscaler options at autoscaler standalone level, it's similar with flink kubernetes operator.
  - The  autoscaler options of autoscaler standalone will be as the base configuration, and the configuration at job-level can override the default value provided in the autoscaler standalone.

## Verifying this change

- Improved the `FlinkClusterJobListFetcherTest#testFetchJobListAndConfigurationInfo` to check baseConf is empty and not empty.

## Documentation

### DIFF
--- a/flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/JobListFetcher.java
+++ b/flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/JobListFetcher.java
@@ -19,6 +19,7 @@ package org.apache.flink.autoscaler.standalone;
 
 import org.apache.flink.annotation.Experimental;
 import org.apache.flink.autoscaler.JobAutoScalerContext;
+import org.apache.flink.configuration.Configuration;
 
 import java.util.Collection;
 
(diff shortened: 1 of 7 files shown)


flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/JobListFetcher.java
```diff
@@ -19,12 +19,19 @@
...
 public interface JobListFetcher<KEY, Context extends JobAutoScalerContext<KEY>> {
 
-    Collection<Context> fetch() throws Exception;
+    /**
+     * Fetch the job context.
+     *
+     * @param baseConf The basic configuration for standalone autoscaler. The basic configuration
+     *     can be overridden by the configuration at job-level.
+     */
+    Collection<Context> fetch(Configuration baseConf) throws Exception;
```
1996fanrui, line 36: This PR updated this `fetch` interface: adding a `baseConf` as the basic configuration.

Hi @gyfora , would you mind helping take a look this PR in your free time? thanks~

gyfora (approved):

PR #817, opened 2024-04-17
[FLINK-31860] FlinkDeployments never finalize when namespace is deleted

jiangzho:

This patch is to tackle the corner case where `cleanup` of CustomResource is constantly failing due to event publish failure, and therefore cause the `cleanup` hanging forever.

Operator attempts to publish events when reconcile changes or cleaning up for a CustomResource. This patch allows the reconcile & clean up logic to proceed, if and only if the event publishing fails as a result of 403 forbidden, which happens when the namespace is being deleted (in terminating state). k8s rejects events / resource creation in a terminating namespace.

In this way, at-least-once events delivery is still guaranteed in other cases. Event consumers need to account for the "namespace deleted" scenario, which all resources within are implicitly deleted.

## Brief change log

  - Fix event publish blocking reconcile & cleanup upon 403

## Verifying this change

  - Added additional scenario in EventUtilsTest and expect it to succeed

## Documentation


gyfora (approved):
Looks great @jiangzho  , have you validated this in a kubernetes environment manually? 
We are not adding an e2e here but it may make sense to add a follow up ticket and add it as well.

For the test failure, please run `mvn spotless:apply`

jiangzho:
Thanks @gyfora for quick turnaround !

The commit has been updated with style fix. This patch is validated via minikube setup.

PR #819, opened 2024-04-22
[FLINK-35183] MinorVersion metric for tracking applications

mbalassi:

* Adds new convenience metric for tracking minor version of running applications per namespace
* Changes current Flink version metric to report version as a dimension (both metrics are now dimensions)

## Verifying this change

* Updated existing test to cover this too

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/metrics/FlinkDeploymentMetrics.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/metrics/FlinkDeploymentMetrics.java
@@ -42,11 +42,17 @@ public class FlinkDeploymentMetrics implements CustomResourceMetrics<FlinkDeploy
     // map(namespace, map(version, set(deployment)))
     private final Map<String, Map<String, Set<String>>> deploymentFlinkVersions =
             new ConcurrentHashMap<>();
+    // map(namespace, map(version, set(deployment)))
+    private final Map<String, Map<String, Set<String>>> deploymentFlinkMinorVersions =
+            new ConcurrentHashMap<>();
     // map(namespace, map(deployment, cpu))
     private final Map<String, Map<String, Double>> deploymentCpuUsage = new ConcurrentHashMap<>();
     // map(namespace, map(deployment, memory))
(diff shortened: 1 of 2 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/metrics/FlinkDeploymentMetrics.java
```diff
@@ -94,6 +101,22 @@ public void onUpdate(FlinkDeployment flinkApp) {
                         })
                 .add(deploymentName);
 
+        // Minor version computed from the above
+        var subVersions = flinkVersion.split("\\.");
+        String minorVersion = MALFORMED_MINOR_VERSION;
```
gaborgsomogyi, line 106: Not sure why not use `UNKNOWN` just like the other case but I'm fine with that.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/metrics/FlinkDeploymentMetrics.java
```diff
@@ -165,13 +194,21 @@ private void initNamespaceStatusCounts(String ns) {
     private void initFlinkVersions(String ns, String flinkVersion) {
         parentMetricGroup
                 .createResourceNamespaceGroup(configuration, FlinkDeployment.class, ns)
-                .addGroup(FLINK_VERSION_GROUP_NAME)
-                .addGroup(flinkVersion)
+                .addGroup(FLINK_VERSION_GROUP_NAME, flinkVersion)
```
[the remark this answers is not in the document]
mbalassi: Good question. This is actually semantically different. We can consider a previous behavior a bug, this approach works well with modern metrics systems like Prometheus/Graphana.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/metrics/FlinkDeploymentMetrics.java
mbalassi: Fair, I am open to simply using `UNKNOWN`.

PR #821, opened 2024-04-29
[FLINK-35265] Implement FlinkStateSnapshot custom resource

mateczagany:

Implement FlinkStateSnapshot as according to FLIP-446. This PR does not include the e2e-tests and documentation.

## Brief change log

- Added `FlinkStateSnapshot` and all its dependent classes to `flink-kubernetes-operator-api`
- Deprecated several fields in `FlinkDeployment`/`FlinkSessionJob` as accepted in the FLIP
- Refactored several methods in `FlinkService` to extract the logic of saving snapshot path to other classes
- Added test in `FlinkConfigManager` class  to check if the CR `FlinkStateSnapshot` can be created on the current Kubernetes server during runtime. This is intended to be temporary to ensure a smooth upgrade process.
- Refactored metric- and status-related classes to be able to handle the new CR

## Verifying this change

- Added unit-tests for new features
- Manual testing
- Some custom E2E test cases here: https://github.com/mateczagany/flink-kubernetes-operator-testing/blob/main/flink-kubernetes-operator-tests/src/test/java/org/apache/flink/operator/testing/TestSnapshots.java

## Documentation

  - If yes, how is the feature documented? not documented, docs and examples will be provided in PR #854

## Other implementation details
- For periodic snapshots, the Operator will create new `FlinkStateSnapshot` CRs, and the snapshot will be taken when that resource is reconciled.
- For upgrade snapshots, the Operator will create a new `FlinkStateSnapshot` CR, marking it with `alreadyExists`.
- For checkpoints, the Operator will try to fetch its final path after the checkpoint has been marked as successful. If fetching the path is not successful, the checkpoint will be marked as `COMPLETED`, but with empty path.
- A new configuration (`kubernetes.operator.savepoint.dispose-on-delete`) was also added that was not specified in the FLIP.
- In contrast to the FLIP, `jobReference` will contain an optional field `namespace`. If null, this will default to the namespace of the snapshot.
- In contrast to the FLIP, `CheckpointSpec#type` is removed as the only supported manual checkpoint type from Flink 1.19 is `FULL`.

## Other PRs
- Metrics for snapshots will be implemented in in FLINK-35492.
- Configurable max history age/count and auto-cleanup will be implemented in FLINK-35493.


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
```diff
@@ -122,12 +127,20 @@ public void reconcile(FlinkResourceContext<CR> ctx) throws Exception {
...
+                    Optional.ofNullable(spec.getJob()).map(JobSpec::getInitialSavepointPath);
+            if (spec.getJob() != null && spec.getJob().getFlinkStateSnapshotReference() != null) {
+                var snapshotRef = spec.getJob().getFlinkStateSnapshotReference();
+                if (snapshotRef.getName() != null) {
+                    savepointPathOpt =
+                            Optional.of(
+                                    FlinkStateSnapshotUtils.getAndValidateFlinkStateSnapshotPath(
+                                            ctx.getKubernetesClient(), snapshotRef));
+                }
+            }
```
gyfora, line 141: At this point we should capture this in a method

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/FlinkStateSnapshotStatus.java
```diff
@@ -0,0 +1,66 @@
...
+@AllArgsConstructor
+@NoArgsConstructor
+@EqualsAndHashCode
+@ToString
+@SuperBuilder
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class FlinkStateSnapshotStatus implements Diffable<FlinkStateSnapshotStatus> {
+
+    /** Current state of the snapshot. */
+    @PrinterColumn(name = "Snapshot Status")
```
gyfora, line 44: Can we simply call it `State` ?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/CustomResourceDefinitionWatcherImpl.java
```diff
@@ -0,0 +1,52 @@
...
+import lombok.RequiredArgsConstructor;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.HashSet;
+import java.util.Set;
+
+/** Default implementation for CustomResourceDefinitionWatcher. */
+@RequiredArgsConstructor
+public class CustomResourceDefinitionWatcherImpl implements CustomResourceDefinitionWatcher {
```
gyfora, line 31: The ` CustomResourceDefinitionWatcher` interface and the way we use it feels a bit overly complicated for the simple purpose we want. My recommendation would be to remove the generic interface and replace it with some specific utility that checks the Snapshot CRD and based on this we simply set a boolean base config in the config manager. 

This way we can get if it's enabled from the operator config instead of having to pass and use the watcher from the controllers.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkStateSnapshotController.java
```diff
@@ -0,0 +1,207 @@
...
+                    validator.validateStateSnapshot(savepoint, ctx.getSecondaryResource());
+            if (validationError.isPresent()) {
+                eventRecorder.triggerSnapshotEvent(
+                        savepoint,
+                        EventRecorder.Type.Warning,
+                        EventRecorder.Reason.ValidationError,
+                        EventRecorder.Component.Operator,
+                        validationError.get(),
+                        ctx.getKubernetesClient());
+                return true;
```
gyfora, line 202: we always seem to return "true" is this correct for the validation error case?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkStateSnapshotController.java
```diff
@@ -0,0 +1,207 @@
...
+        if (!validateSavepoint(ctx)) {
+            statusRecorder.patchAndCacheStatus(flinkStateSnapshot, ctx.getKubernetesClient());
+            UpdateControl<FlinkStateSnapshot> updateControl = UpdateControl.noUpdate();
+            return updateControl.rescheduleAfter(
+                    ctx.getOperatorConfig().getReconcileInterval().toMillis());
+        }
+
+        // reconcile
+        try {
+            statusRecorder.patchAndCacheStatus(flinkStateSnapshot, ctx.getKubernetesClient());
```
gyfora, line 103: I think for FlinkStateSnapshot resources we should avoid using the custom status patching. We can just rely on the JOSDK built in Update control mechanism. 

We are using a custom patching logic for the deployments because we are executing destructive actions before which we must record the status in some cases. Here I believe it's not the case in general.

gyfora (changes requested):
I started reviewing this, added some initial comments.

In the PR description you mentioned that with the new mode the manual trigger nonce doesn't work anymore. Would it make sense to keep it working for backward compatibility (when the nonce changes simply create a snapshot resource)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkStateSnapshotController.java
mateczagany: StatusRecorder will also notify `FlinkResourceListener` instances, I am not sure if we should give up that functionality for the snapshot resource. But I agree that we should use the JOSDK update mechanism if possible, I will try to refactor it.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/crd/CustomResourceDefinitionWatcherImpl.java
mateczagany: I have pushed a commit for this, please take a look at it, I am personally not 100% satisfied with either solution. The new solution is a lot more simple, but I had to hardcode the check to `FlinkConfigManager`, and play with the constructors in a way such that the tests won't try to detect if the `FlinkStateSnapshot` CR is installed. 

I also kept `FlinkStateSnapshotUtils.shouldCreateSnapshotResource()` that checks both the operator config and Flink resource config to make it easy to opt-out on a per-job basis. This also helped me with the current tests being able to decide which snapshot method to use.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/spec/CheckpointSpec.java
```diff
@@ -0,0 +1,41 @@
...
+import lombok.EqualsAndHashCode;
+import lombok.NoArgsConstructor;
+
+/** Spec for checkpoint state snapshots. */
+@Experimental
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+@Builder
+@EqualsAndHashCode
```
gyfora, line 36: `@Data` already adds EqualsAndHashCode, you probably copied this from other classes (so maybe I messed up first :D) but it should not be necessary

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/spec/FlinkStateSnapshotSpec.java
```diff
@@ -0,0 +1,63 @@
...
+import lombok.ToString;
+import lombok.experimental.SuperBuilder;
+
+/** Spec that describes a FlinkStateSnapshot. */
+@Experimental
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+@EqualsAndHashCode
+@ToString
```
gyfora, line 37: Data covers equals and toString

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/CommonStatus.java
```diff
@@ -27,11 +27,13 @@
 import io.fabric8.crd.generator.annotation.PrinterColumn;
 import lombok.AllArgsConstructor;
 import lombok.Data;
+import lombok.EqualsAndHashCode;
 import lombok.NoArgsConstructor;
 import lombok.experimental.SuperBuilder;
 import org.apache.commons.lang3.StringUtils;
 
 /** Last observed common status of the Flink deployment/Flink SessionJob. */
+@EqualsAndHashCode
```
gyfora, line 36: Data should cover this, did you hit some issue without it?

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/FlinkStateSnapshotStatus.java
```diff
@@ -0,0 +1,66 @@
...
+import lombok.ToString;
+import lombok.experimental.SuperBuilder;
+
+/** Last observed status of the Flink state snapshot. */
+@Experimental
+@Data
+@AllArgsConstructor
+@NoArgsConstructor
+@EqualsAndHashCode
+@ToString
```
gyfora, line 38: see other comments

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
```diff
@@ -410,6 +430,14 @@ public static String operatorConfigKey(String key) {
...
                             "Max number of reconcile loops triggered within the rate limiter refresh period for each resource. Setting the limit <= 0 disables the limiter.");
 
+    @Documentation.Section(SECTION_DYNAMIC)
+    public static final ConfigOption<Boolean> OPERATOR_JOB_UPGRADE_SAVEPOINT_DISPOSE_ON_DELETE =
+            operatorConfig("job.upgrade.savepoint.dispose-on-delete")
+                    .booleanType()
+                    .defaultValue(false)
+                    .withDescription(
+                            "Savepoint data created by upgrade savepoints as FlinkStateSnapshot resources will be disposed of automatically when the Kubernetes resource gets deleted.");
+
```
gyfora, line 440: I think we should have a single dispose-on-delete config that generally applies to snapshot resources.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/snapshot/StateSnapshotObserver.java
```diff
@@ -0,0 +1,201 @@
...
+            return;
+        }
+
+        if (checkpointInfo.getError() != null) {
+            LOG.error("Checkpoint {} failed: {}", resourceName, checkpointInfo.getError());
+            snapshotFailed(ctx.getKubernetesClient(), resource, checkpointInfo.getError());
+        } else {
+            LOG.info(
+                    "Checkpoint {} was successful, querying final checkpoint path...",
+                    resourceName);
```
gyfora, line 147: I think this could be debug or removed

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/snapshot/StateSnapshotObserver.java
```diff
@@ -0,0 +1,201 @@
...
+        if (checkpointInfo.isPending()) {
+            LOG.debug(
+                    "Checkpoint for {} with ID {} is pending",
+                    resourceName,
+                    resource.getStatus().getTriggerId());
+            return;
+        }
+
+        if (checkpointInfo.getError() != null) {
+            LOG.error("Checkpoint {} failed: {}", resourceName, checkpointInfo.getError());
```
gyfora, line 142: Unnecessary as we trigger an event which logs this again

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/snapshot/StateSnapshotObserver.java
```diff
@@ -0,0 +1,201 @@
...
+        var resource = ctx.getResource();
+        var resourceName = resource.getMetadata().getName();
+
+        if (savepointInfo.isPending()) {
+            LOG.debug(
+                    "Savepoint '{}' with ID {} is pending",
+                    resourceName,
+                    resource.getStatus().getTriggerId());
+        } else if (savepointInfo.getError() != null) {
+            LOG.error("Savepoint {} failed: {}", resourceName, savepointInfo.getError());
```
gyfora, line 117: Unnecessary as there is an event for this already

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/snapshot/StateSnapshotObserver.java
```diff
@@ -0,0 +1,201 @@
...
+
+    private void observeSavepointState(FlinkStateSnapshotContext ctx) {
+        var resource = ctx.getResource();
+        var resourceName = resource.getMetadata().getName();
+        var triggerId = resource.getStatus().getTriggerId();
+
+        LOG.info("Observing savepoint state for resource {}...", resourceName);
+
+        if (StringUtils.isEmpty(triggerId)) {
+            LOG.debug("Trigger ID is not set for savepoint {} yet.", resourceName);
```
gyfora, line 66: we could remove this and move the Observing log after the if?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/snapshot/StateSnapshotObserver.java
```diff
@@ -0,0 +1,201 @@
...
+            observeSavepointState(ctx);
+        }
+    }
+
+    private void observeSavepointState(FlinkStateSnapshotContext ctx) {
+        var resource = ctx.getResource();
+        var resourceName = resource.getMetadata().getName();
+        var triggerId = resource.getStatus().getTriggerId();
+
+        LOG.info("Observing savepoint state for resource {}...", resourceName);
```
gyfora, line 63: This should be debug

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/CommonStatus.java
mateczagany: No clue how it got here, I could safely remove it.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
mateczagany: Good idea, I have named it `savepoint.dispose-on-delete`.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/snapshot/StateSnapshotObserver.java
```diff
@@ -0,0 +1,197 @@
...
+        LOG.debug("Observing savepoint state for resource {}...", resourceName);
+        var secondaryResource =
+                ctx.getSecondaryResource()
+                        .orElseThrow(
+                                () ->
+                                        new RuntimeException(
+                                                String.format(
+                                                        "Secondary resource %s for savepoint %s was not found",
+                                                        resource.getSpec().getJobReference(),
+                                                        resourceName)));
```
gyfora, line 76: If the target is not found anymore to me that sounds similar to the not running case, and probably should be marked abandoned. What is your thinking here?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
```diff
@@ -222,9 +235,12 @@ private void updateStatusBeforeFirstDeployment(
...
+            var snapshotRef = spec.getJob().getFlinkStateSnapshotReference();
             var initialSp = spec.getJob().getInitialSavepointPath();
 
-            if (initialSp != null) {
+            if (snapshotRef != null) {
+                initialUpgradeMode = UpgradeMode.SAVEPOINT;
+            } else if (initialSp != null) {
                 status.getJobStatus()
                         .getSavepointInfo()
                         .setLastSavepoint(Savepoint.of(initialSp, SnapshotTriggerType.UNKNOWN));
```
gyfora, line 246: shouldn't the last savepoint info be updated in case the snapshot ref is provided? This logic is there to handle failures before the initial startup so should apply to both

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractJobReconciler.java
```diff
@@ -266,19 +304,31 @@ protected void restoreJob(
...
+                savepointOpt = getLatestSavepointPathFromFlinkStateSnapshots(ctx);
+            } else {
+                savepointOpt =
+                        Optional.ofNullable(
+                                        ctx.getResource()
+                                                .getStatus()
+                                                .getJobStatus()
+                                                .getSavepointInfo()
+                                                .getLastSavepoint())
+                                .flatMap(s -> Optional.ofNullable(s.getLocation()));
```
gyfora, line 318: The problem with this logic is that the snapshotReference field allows us to specify the path directly. So if the user starts the job using that, in the code we don't update the lastSavepointInfo but we also won't find that path here. So we will simply restart from empty state in those cases which is not what should happen.

2 main ways to solve it: 
 1.  We keep the current lastSavepoint logic in the status and update it whenever a snapshot is completed. This is a bit problematic because we would have to update the status of the FlinkDeployment/SessionJob from the snapshot controller, which is a bit weird.
 2. Introduce a new field in the status to capture the savepoint reference used during a Savepoint deployment. I think this is the way to go because this is the functionality that is important to capture. When we submit with initialSavepointPath the first time / savepoint redeploy / savepoint upgrade, in the lastReconciledSpec we have upgradeMode: SAVEPOINT recorded and it is assumed that the lastSavepoint contains the savepoint used. Now we are getting rid of the lastSavepoint so we should put something more meaningful in place in the status to capture the actual upgrade related behaviour. One possibility here is simply to use the `job.snapshotReference` field in the lastReconciledSpec to capture the upgrade savepoint and clear it in other cases (same way of how we reuse the upgradeMode field to capture how we actually stopped the job in the lastReconciledSpec)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractJobReconciler.java
```diff
@@ -329,18 +459,26 @@ private void redeployWithSavepoint(
...
+        if (snapshotRef != null && snapshotRef.getName() != null) {
+            path =
+                    FlinkStateSnapshotUtils.getAndValidateFlinkStateSnapshotPath(
+                            ctx.getKubernetesClient(), snapshotRef);
+        }
+
         status.getJobStatus()
                 .getSavepointInfo()
-                .setLastSavepoint(Savepoint.of(savepoint, SnapshotTriggerType.UNKNOWN));
+                .setLastSavepoint(Savepoint.of(path, SnapshotTriggerType.UNKNOWN));
```
gyfora, line 474: See my other long comment

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
```diff
@@ -226,7 +230,36 @@ private void setJobIdIfNecessary(
...
+                        Savepoint sp =
+                                Savepoint.of(
+                                        location,
+                                        SnapshotTriggerType.UPGRADE,
+                                        SavepointFormatType.valueOf(savepointFormatType.name()));
+                        ctx.getResource()
+                                .getStatus()
+                                .getJobStatus()
+                                .getSavepointInfo()
+                                .updateLastSavepoint(sp);
```
gyfora, line 260: see my other long comment

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/snapshot/StateSnapshotReconciler.java
```diff
@@ -0,0 +1,235 @@
...
+        }
+
+        var secondaryResource =
+                ctx.getSecondaryResource()
+                        .orElseThrow(
+                                () ->
+                                        new RuntimeException(
+                                                String.format(
+                                                        "Secondary resource %s not found",
+                                                        source)));
```
gyfora, line 80: Should this just abandon the snapshot?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/snapshot/StateSnapshotReconciler.java
```diff
@@ -0,0 +1,235 @@
...
+        if (!ReconciliationUtils.isJobRunning(secondaryResource.getStatus())) {
+            LOG.warn(
+                    "Target job {} for savepoint {} is not running, cannot trigger snapshot.",
+                    secondaryResource.getMetadata().getName(),
+                    resource.getMetadata().getName());
+            return;
+        }
```
gyfora, line 87: we have already observed and abandoned the snapshot in this case right?

gyfora (changes requested):
I think we need some extra logic to handle savepoint job submissions and snapshot info during upgrades.

There is a bit of implicit logic currently in why the lastSavepoint is updated every time we submit a job from savepoint, and we need some new logic to capture that. (see my other long comment)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/snapshot/StateSnapshotObserver.java
mateczagany: Yeah, I agree this should be abandoned as well. I have also added an event that will be produced when the snapshot gets to an abandoned state.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/snapshot/StateSnapshotReconciler.java
mateczagany: This feels a bit weird, since after the observe phase, the code should not be able to run into this at all. I think it would help keep the code cleaner if we only abandon snapshots from the observer.

gyfora:
A small addition to the previous comments. The SnapshotObserver currently also observes and updates the lastSavepointInfo for terminal jobs. This is a key mechanism to be able to handle failures during stateful upgrades.

So we have to update that logic as well  so that instead of record the lastSavepoint status we use a new shared logic with the snapshot custom resource mechanism. This mechanism is covered by some tests currently, enabling the snapshot CR for those tests would probably help repro the problem. For example:

1. Introduce a failure after executing a stop-with-savepoint operation (before the snapshot CR was created)
2. This is the point where the observer would actually observe the savepoint/checkpoint info from the terminal job and update the status 
3. Assert that the upgrade is actually executed from the correct savepoint 

I believe with the current implementation the savpoint info would be lost

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractJobReconciler.java
gyfora: Actually I think the cleanest / simplest thing to do would be to add a new status field like: 
```
snapshotReference / upgradeSnapshotReference 
```
This should be updated during SAVEPOINT deployments (when we start from savepoint/checkpoint), after savepoint cancellation or when observing the last state of a terminal job. (basically every time we currently update the lastSavepoint for upgrade purposes
)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/snapshot/StateSnapshotReconciler.java
mateczagany: I have added a commit which will make sure that we abandon snapshots if referenced Flink resource is not found, or the job is not running, in both observe and reconcile phase.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractJobReconciler.java
mateczagany: Thank you for detailing this problem so well, it makes perfect sense, I have overlooked this part of the operator.

I think I could tackle this problem with a new status field as you have proposed called `upgradeSnapshotReference` in the job status, but this still leaves an important feature out which was covered before: Manual savepoints triggered by creating a new `FlinkStateSnapshot` resource, and periodic savepoints that create new `FlinkStateSnapshot` resources will not update this new field `upgradeSnapshotReference` in FlinkDeployment/FlinkSessionJob statuses.

I see 3 solutions:
- Update FlinkDeployment/FlinkSessionJob status from the snapshot controller, I agree that this is not an optimal pattern, and I would avoid it.
- List all `FlinkStateSnapshot` resources when re-deploying a Flink job and we check `upgradeSnapshotReference`. We could compare the time the snapshots were taken and use the most recent. This might lead to confusion because e.g. starting a suspended job might not use the savepoint we can see in the status.
- From the FlinkDeployment/FlinkSessionJob controllers check the status of `FlinkStateSnapshot` resources with the help of `InformerEventSource`, not sure how viable this is.

gyfora: @mateczagany I think the only good solution is to introduce a separate `upgradeSnapshotReference` that is only updated during the upgrade cycle (and during terminal job observe).  Any snapshot observed this way is logically speaking always the latest. 

A manual / periodic snapshot should never override this, doing that would be surely an error in the logic so this is absolutely not a problem. This way we will actually separate the upgrade snapshot handling from the savepoint taking / management which was kind of confused together until now.

In other words we don't need any of the "3 solutions" :)

mateczagany: Okay, I am definitely happy with this plan, I will push a commit soon:
- the new field will be updated on upgrade, stop-with-savepoint and when the job becomes terminal
- the field won't be set during manual and periodic snapshots (in contrast to how it worked so far)
- savepoint path in `AbstractJobReconciler#restoreJob()` will be set to this field
- `lastSavepoint` will still be used for periodic and manual savepoints if `FlinkStateSnapshot` is disabled. This will be deprecated.

mateczagany: I have added a commit that should cover these points, I have also resolved related conversations. The commit also adds support for `savepointTriggerNonce` and `checkpointTriggerNonce` with the new snapshot resources. I will also add unit tests for these later.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkStateSnapshotController.java
mateczagany: Done in commit 0315d1062dcf0b81e61af77be4a0c12e9330398b - I added some logic to decide if the status was modified, and notify listeners and `patchStatus` if it changed.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkStateSnapshotController.java
```diff
@@ -0,0 +1,202 @@
...
+            reconciler.reconcile(ctx);
+        } catch (Exception e) {
+            FlinkStateSnapshotUtils.snapshotFailed(
+                    josdkContext.getClient(), eventRecorder, flinkStateSnapshot, e.getMessage());
+            LOG.error("Failed to reconcile {}", flinkStateSnapshot, e);
```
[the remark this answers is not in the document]
mateczagany: Errors are now handled by updateErrorStatus since commit 0315d1062dcf0b81e61af77be4a0c12e9330398b

ferenc-csaky (commented):
I did some testing myself as well, I think this is getting close to a good and mergable shape! Did not checked through the code, but added 2 local comments. In general, did some testing with the `StateMachineExample` tutorial:

1. Deployed a fresh job
2. Created checkpoint via CRD (1 specific comment about `INCREMENTAL`)
3. Created savepoint via CRD
5. Restored job from savepoint

So far the main functionality looks good, I plan to go though the files in the next couple days.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
```diff
@@ -226,7 +229,27 @@ private void setJobIdIfNecessary(
     @Override
     protected void cancelJob(FlinkResourceContext<FlinkDeployment> ctx, UpgradeMode upgradeMode)
             throws Exception {
-        ctx.getFlinkService().cancelJob(ctx.getResource(), upgradeMode, ctx.getObserveConfig());
+        var conf = ctx.getObserveConfig() != null ? ctx.getObserveConfig() : new Configuration();
+        var savepointFormatType =
+                conf.get(KubernetesOperatorConfigOptions.OPERATOR_SAVEPOINT_FORMAT_TYPE);
+
+        var savepointOpt = ctx.getFlinkService().cancelJob(ctx.getResource(), upgradeMode, conf);
+        savepointOpt.ifPresent(
```
ferenc-csaky, line 237: The body of this optional could be moved to `AbstractJobReconciler` as a new protected method, so we only duplicate as minimal logic as we can.

ferenc-csaky (changes requested):
Second part of my review, hopefully I will have some time tomorrow to finish and check the rest of the classes.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
```diff
@@ -426,6 +438,14 @@ public static String operatorConfigKey(String key) {
                     .withDescription(
                             "Max allowed checkpoint age for initiating last-state upgrades on running jobs. If a checkpoint is not available within the desired age (and nothing in progress) a savepoint will be triggered.");
 
+    @Documentation.Section(SECTION_DYNAMIC)
+    public static final ConfigOption<Boolean> OPERATOR_JOB_SAVEPOINT_DISPOSE_ON_DELETE =
+            operatorConfig("savepoint.dispose-on-delete")
+                    .booleanType()
+                    .defaultValue(false)
+                    .withDescription(
+                            "Savepoint data for FlinkStateSnapshot resources created by the operator during upgrades and periodic savepoints will be disposed of automatically when the generated Kubernetes resource is deleted.");
```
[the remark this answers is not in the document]
mateczagany: I believe it's the correct way to say it (to dispose of something)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkStateSnapshotContext.java
```diff
@@ -0,0 +1,112 @@
...
+    /**
+     * @return Operator configuration for this resource.
+     */
+    public FlinkOperatorConfiguration getOperatorConfig() {
+        if (operatorConfig != null) {
+            return operatorConfig;
+        }
+        return operatorConfig =
+                configManager.getOperatorConfiguration(
+                        getResource().getMetadata().getNamespace(), null);
```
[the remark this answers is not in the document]
mateczagany: I have flipped the ifs for now, I will come back to check using Lombok lazy getter

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkStateSnapshotController.java
```diff
@@ -0,0 +1,216 @@
...
+
+        notifyListeners(ctx);
+        return getUpdateControl(ctx);
+    }
+
+    @Override
+    public DeleteControl cleanup(
+            FlinkStateSnapshot flinkStateSnapshot, Context<FlinkStateSnapshot> josdkContext) {
+        var ctx = ctxFactory.getFlinkStateSnapshotContext(flinkStateSnapshot, josdkContext);
+        DeleteControl deleteControl;
```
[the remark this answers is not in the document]
mateczagany: Thank you, it was a leftover before a refactor

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/SnapshotObserver.java
```diff
@@ -273,15 +280,19 @@ private void disposeSavepointQuietly(
         }
     }
 
-    private void observeLatestSavepoint(
+    private void observeLatestCheckpoint(
```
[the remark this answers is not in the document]
mateczagany: Yep, this will be the latest checkpoint retrieved via Flink REST API, I have updated the log

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/FlinkStateSnapshotState.java
```diff
@@ -0,0 +1,39 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.flink.kubernetes.operator.api.status;
+
+import org.apache.flink.annotation.Experimental;
+
+/** Describes current snapshot state. */
+@Experimental
+public enum FlinkStateSnapshotState {
```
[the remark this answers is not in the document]
mateczagany: Good idea, thank you!

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkStateSnapshotContext.java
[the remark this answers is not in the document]
mateczagany: Added lazy getter as well, thank you!

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/listener/AuditUtils.java
```diff
@@ -53,6 +62,16 @@ private static String format(@NonNull CommonStatus<?> status) {
                         : status.getError());
     }
 
+    private static String format(@NonNull FlinkStateSnapshotStatus status) {
+        if (StringUtils.isEmpty(status.getError())) {
+            return String.format(
+                    ">>> Status[Snapshot] | Info | %s | %s", status.getState(), status.getPath());
```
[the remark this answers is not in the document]
mateczagany: I have added some changes for this, thank you!

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
```diff
@@ -716,6 +695,48 @@ public CheckpointFetchResult fetchCheckpointInfo(
...
+            } else if (stats instanceof CheckpointStatistics.PendingCheckpointStatistics) {
+                return CheckpointStatsResult.pending();
+            } else {
+                throw new IllegalArgumentException(
+                        String.format(
+                                "Unknown checkpoint statistics result class: %s",
+                                stats.getClass().getSimpleName()));
+            }
+        } catch (Exception e) {
+            LOG.error("Exception while fetching checkpoint statistics", e);
```
mateczagany, line 735: We should handle cases where the checkpoint statistics are no longer stored on the web server, and we get the following error: 
```
Could not find checkpointing statistics for checkpoint 243.
```

anupamaggarwal:
Thanks for contributing this super useful feature @mateczagany 👍 !. 
I am still new to the operator codebase so pls feel free to ignore if something is not relevant, had some initial questions-


1) IIUC if a savepoint CR has a job reference (with an invalid job which does not exist/ is not running) we would set the status field to ABANDONED. Does it also make sense to forbid the snapshot CR creation altogether in these cases through the validation webhook ?
 
2) Would it be possible to add an example FlinkStateSnapshot CR to the examples folder ? (similar to [basic.yaml](https://nightlies.apache.org/flink/flink-kubernetes-operator-docs-main/docs/try-flink-kubernetes-operator/quick-start/#submitting-a-flink-job) referenced in the quickstart guide). Pls ignore if you were planning on covering this as part of the documentation PR.

3) Does it also make sense to include some of the new configs (for example `snapshot.resource.enabled`)  in `helm/*/conf/flink-conf.yaml` with the default values commented out? I was wondering if this would give more signal to first time users of this feature, but it seems we currently have a subset of operator properties exposed, so wasn't entirely sure.

mateczagany:
Thanks for your feedback @anupamaggarwal, really appreciate it!

1. I agree, and this is what I expected to happen. Except it does not, because I forgot to add the FlinkStateSnapshot resource to the `ValidatingWebhookConfiguration` resource in the Helm chart, I will push a commit soon.
2. This PR is already quite huge, so I have decided to do the examples/docs in a separate PR, there is currently a draft for it: https://github.com/apache/flink-kubernetes-operator/pull/854
3. You are right, adding the config there would improve its visibility to the end-users, so I will do that in a future commit, thank you!

flink-kubernetes-webhook/src/main/java/org/apache/flink/kubernetes/operator/admission/mutator/FlinkMutator.java
```diff
@@ -92,4 +96,16 @@ private FlinkDeployment mutateDeployment(HasMetadata resource) {
             throw new RuntimeException(e);
         }
     }
+
+    private FlinkStateSnapshot mutateStateSnapshot(HasMetadata resource) {
+        try {
+            var savepoint = mapper.convertValue(resource, FlinkStateSnapshot.class);
```
ferenc-csaky, line 102: nit: savepoint -> snapshot.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigManager.java
```diff
@@ -72,21 +74,26 @@ public class FlinkConfigManager {
 
     private volatile Configuration defaultConfig;
     private volatile FlinkOperatorConfiguration defaultOperatorConfiguration;
+    private final boolean snapshotCrdInstalled;
```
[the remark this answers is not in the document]
mateczagany: Added the note, thank you. I am going to also note here so others can see: 

This field will be determined in `FlinkOperator` when the operator starts, by checking if `FlinkStateSnapshot` CRD is installed on the Kubernetes cluster. 
When generating the default configuration for the operator, `kubernetes.operator.snapshot.resource.enabled` will be set to false in this class if `snapshotCrdInstalled` is false, to allow users to continue using the operator if they upgraded from a previous version but did not install the new `FlinkStateSnapshot` CRD.

This check and field is planned to be removed in the release after this FLIP gets released (so hopefully 1.11 if this PR is released in 1.10).

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkStateSnapshotUtils.java
```diff
@@ -0,0 +1,395 @@
...
+                                .withName(snapshotRef.getName())
+                                .get();
+            }
+        } else {
+            result =
+                    kubernetesClient
+                            .resources(FlinkStateSnapshot.class)
+                            .withName(snapshotRef.getName())
+                            .get();
+        }
```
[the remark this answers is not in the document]
mateczagany: Looks like late-night code, no idea how I left it like this :D thank you

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkStateSnapshotUtils.java
```diff
@@ -0,0 +1,382 @@
...
+    public static String getFlinkStateSnapshotName(
+            SnapshotType snapshotType,
+            SnapshotTriggerType triggerType,
+            AbstractFlinkResource<?, ?> referencedResource) {
+        return String.format(
+                "%s-%s-%s-%s",
+                referencedResource.getMetadata().getName(),
+                snapshotType.name().toLowerCase(),
+                triggerType.name().toLowerCase(),
+                UUID.randomUUID());
```
anupamaggarwal, line 233: Hi @mateczagany, wdyt about having a timestamp / epoch value instead of the UUID here? It might help users figure out the latest snapshot CR in case they do a
`kubectl get flinkstatesnapshots.flink.apache.org`  (and not have to pass in `--sort-by=.metadata.creationTimestamp` explicitly) ?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/snapshot/StateSnapshotReconciler.java
```diff
@@ -0,0 +1,230 @@
...
+        switch (state) {
+            case IN_PROGRESS:
+                LOG.info(
+                        "Cannot delete resource {} yet as savepoint is still in progress...",
+                        resourceName);
+                return DeleteControl.noFinalizerRemoval()
+                        .rescheduleAfter(ctx.getOperatorConfig().getReconcileInterval().toMillis());
+            case COMPLETED:
+                var flinkDeployment = getFlinkDeployment(ctx);
+                return handleSnapshotCleanup(resource, flinkDeployment, ctx);
```
anupamaggarwal, line 123: I was trying out the periodic snapshot feature today. Does it make sense to introduce a feature to auto-cleanup the  `flinkstatesnapshots.flink.apache.org` CRs and only retain upto some max retention limit (we could still retain the savepoint dir?) IIUC `disposeOnDelete` for periodic savepoints is set to false by default (so it should be relatively safe to do so). 

If the user sets the savepoint interval `kubernetes.operator.periodic.savepoint.interval` to a low value, and there are multiple jobs on the cluster it would lead to creation of many CRs. I am not familiar with K8s internals but could this increase the load on the API server?

mateczagany: Definitely! I would not feel comfortable releasing a new operator version that does not have auto-cleanup of FlinkStateSnapshot CRs. For the now-deprecated way of taking snapshots this could be configured using `kubernetes.operator.savepoint.history.max.age` and `kubernetes.operator.savepoint.history.max.count`.

You are right, `disposeOnDelete` is false by default, and it can be controlled by the config `kubernetes.operator.savepoint.dispose-on-delete`.

I am not sure of the exact approach I will take to handle cleanup, but it will be implemented under [FLINK-35493](https://issues.apache.org/jira/browse/FLINK-35493) in another PR. That's a feature I felt we could easily separate to another PR, as this one is already growing quite big :D 

With a low periodic interval, there won't be any more load on the API server and etcd than tracking snapshots in the Flink resource CR (as the operator does currently). Periodically cleaning them up might affect the load on API server/etcd depending on what solution we will go with. But that will be done under FLINK-35493.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkStateSnapshotUtils.java
mateczagany: Hi @anupamaggarwal you are right, that makes more sense than a random UUID. I have pushed a commit for this.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
mateczagany: This commit made it so we ignore errors when fetching checkpoint stats, since at that point we have determined that the checkpoint was successful anyways. We just set an empty path: https://github.com/apache/flink-kubernetes-operator/pull/821/commits/637b3053b9644b3d6cadd1fb6a05e1f5aab75fa6

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/snapshot/StateSnapshotReconciler.java
mateczagany: I have done some investigation, and I think I will be able to easily implement this in the FlinkDeployment and FlinkSessionJob controller observe phase, just like the operator currently cleans up savepoints. I could just add a new `InformerEventSource` so we would be able to list `FlinkStateSnapshot` resources as secondary resources when we reconcile FlinkDeployment/FlinkSessionJob resources without querying the Kubernetes API server.

Once this PR is merged, I will create a new PR for this functionality.

gyfora:
@anupamaggarwal @ferenc-csaky unless there are any concerns I would like to merge this. Afterwards we can focus on the follow up items and even get a chance to paralelnize this work

mateczagany:
Thank you @gyfora , I have done a lot of manual testing and I am quite happy with the current state of the PR. Once this is merged, I will start working on the other tickets in this FLIP, and create new PRs hopefully this week for the following:
- Docs
- E2E test case
- Snapshot history automatic cleanup
- FlinkStateSnapshot metrics integration

I will be unavailable for ~3 weeks starting from Aug 23, I hope to have the remaining PRs ready and in good state until then.

PR #824, opened 2024-05-06
[FLINK-34212][autoscaler] Autoscaler Standalone cleans up stopped jobs to prevent memory leaks

1996fanrui:

Autoscaler Standalone will fetch job list, and scale them. In general, autoscaler has some cache in memory.

Autoscaler Standalone should clean up them if some jobs cannot be fetched anymore.

## Brief change log

- [FLINK-34212][autoscaler] Autoscaler Standalone cleans up stopped jobs to prevent memory leaks
  - We assume one job is stopped when one job is fetched in the last control loop and it cannot be fetched in current control loop.

## Verifying this change

- Added `testCleanupAfterStopped` & `testCleanupForStoppedJobAfterScaling` in StandaloneAutoscalerExecutorTest class

## Documentation

### DIFF
--- a/flink-autoscaler-standalone/src/test/java/org/apache/flink/autoscaler/standalone/StandaloneAutoscalerExecutorTest.java
+++ b/flink-autoscaler-standalone/src/test/java/org/apache/flink/autoscaler/standalone/StandaloneAutoscalerExecutorTest.java
@@ -33,8 +33,10 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.atomic.AtomicLong;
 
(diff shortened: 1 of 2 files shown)


flink-autoscaler-standalone/src/test/java/org/apache/flink/autoscaler/standalone/StandaloneAutoscalerExecutorTest.java
```diff
@@ -227,6 +229,160 @@ public void cleanup(JobID jobID) {
...
+                            @Override
+                            public void cleanup(JobID jobID) {
+                                cleanupCounter.put(
+                                        jobID, cleanupCounter.getOrDefault(jobID, 0) + 1);
+                            }
+                        })) {
+
+            // Test for empty job list.
+            autoscalerExecutor.scaling();
+            Thread.sleep(100);
```
RocMarshal, line 265: Could we use the CompletableFuture to overwrite some lines like this one?

I made a rough draft.
[draft-patch.txt](https://github.com/apache/flink-kubernetes-operator/files/15256511/draft-patch.txt)

But I have to admit that the draft still is a bit of redundant. Because the draft use the `completableFuture` by introducing a useless `return value` of some methods.

RocMarshal (approved):
Hi, @1996fanrui Thanks for the contribution.
LGTM on the whole.
Just a few of optional comments.
PTAL~

flink-autoscaler-standalone/src/test/java/org/apache/flink/autoscaler/standalone/StandaloneAutoscalerExecutorTest.java
1996fanrui: Good suggestion!

Updated with your idea, and I changed some code for `StandaloneAutoscalerExecutorTest#testCleanupForStoppedJobAfterScaling`.

RocMarshal (approved):
Thanks @1996fanrui for the quick update.
LGTM +1. 👍

PR #825, opened 2024-05-08
[FLINK-35192] support jemalloc in image

chenyuzhi459:

According to [FLINK-35192](https://issues.apache.org/jira/browse/FLINK-35192), the current operator image uses glibc as the memory allocator, which may cause memory leaks due to memory fragmentation.

This pr introduces jemalloc as a memory allocator for the operator's image.

## Brief change log

Introducing the `DISABLE_JEMALLOC` parameter to control whether the operator's image uses jemalloc as the memory allocator

## Verifying this change

This change is already covered by existing  ci-tests

## Documentation

  - If yes, how is the feature documented?  not documented

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -22,6 +22,27 @@ args=("$@")
 
 cd /flink-kubernetes-operator || exit
 
+maybe_enable_jemalloc() {
+    if [ "${DISABLE_JEMALLOC:-false}" == "false" ]; then
+        JEMALLOC_PATH="/usr/lib/$(uname -m)-linux-gnu/libjemalloc.so"
+        JEMALLOC_FALLBACK="/usr/lib/x86_64-linux-gnu/libjemalloc.so"
+        if [ -f "$JEMALLOC_PATH" ]; then
+            export LD_PRELOAD=$LD_PRELOAD:$JEMALLOC_PATH
+        elif [ -f "$JEMALLOC_FALLBACK" ]; then
+            export LD_PRELOAD=$LD_PRELOAD:$JEMALLOC_FALLBACK
+        else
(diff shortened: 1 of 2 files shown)


docker-entrypoint.sh
```diff
@@ -22,6 +22,27 @@ args=("$@")
 
 cd /flink-kubernetes-operator || exit
 
+maybe_enable_jemalloc() {
+    if [ "${DISABLE_JEMALLOC:-false}" == "false" ]; then
```
gaborgsomogyi, line 26: Any specific reason why double equals used here and all other places single?

docker-entrypoint.sh
```diff
@@ -22,6 +22,27 @@ args=("$@")
...
+        JEMALLOC_PATH="/usr/lib/$(uname -m)-linux-gnu/libjemalloc.so"
+        JEMALLOC_FALLBACK="/usr/lib/x86_64-linux-gnu/libjemalloc.so"
```
gaborgsomogyi, line 28: Why do we fallback and not fail fast?

docker-entrypoint.sh
chenyuzhi459: Accidentally wrote an extra `=`, it has been fixed.

docker-entrypoint.sh
chenyuzhi459: Refer to the issue [FLINK-28057](https://issues.apache.org/jira/browse/FLINK-28057),

Maybe the user base image cannot use or install jemalloc，so will try fallback.

gaborgsomogyi (approved):

gaborgsomogyi:
Intended to merge this in a day when no new comments arrive.

PR #833, opened 2024-06-03
[FLINK-35489][Autoscaler] Ensure METASPACE size is computed before the heap

ashangit:

If the heap usage with overhead leads to remaining memory to be assigned to the HEAP no memory will be assigned to the METASPACE leading to JVM failure at startup
By setting this before the HEAP we ensure the JVM can start with appropriate amount of METASPACE memory

## Brief change log

*(for example:)*
  - *Compute METASPACE size to set before the HEAP size*

## Verifying this change

- Extended tests to validate that despite a big HEAP usage that would lead to the whole memory being used by it, the METASPACE has been well set with all appropriate memory

## Documentation

### DIFF
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/tuning/MemoryTuning.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/tuning/MemoryTuning.java
@@ -125,10 +125,12 @@ public class MemoryTuning {
                                 scalingSummaries, evaluatedMetrics.getVertexMetrics()),
                         config,
                         memBudget);
-        MemorySize newHeapSize =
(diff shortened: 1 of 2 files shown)


1996fanrui (approved):
Thanks @ashangit for the detailed analysis and fix!

LGTM, and I hope to wait for @mxm 's double check before merging.

ashangit:
Updated the PR to take in account this [comment](https://issues.apache.org/jira/browse/FLINK-35489?focusedCommentId=17853606&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17853606) in the jira ticket

mxm (approved):
Good fix @ashangit!

1996fanrui:
Thanks @ashangit for the fix and everyone for the review!

Merging~

PR #835, opened 2024-06-04
[hotfix][docs] Fix typo in Log Example Given on upgrade.md File

nacisimsek:
The ID in the savepoint folder name that is passed as a parameter, and the ID in the folder name that is given as the log example do NOT match. 

Expected ID in log example: `aec3dd08e76d`
Actual ID given in the log example: `2f40a9c8e4b9`


1996fanrui (approved):
Thanks for the fix! 

LGTM

PR #842, opened 2024-06-19
[FLINK-35649] Bump Flink version to 1.19.1 in k8s-operator

ferenc-csaky:

Bump Flink version to 1.19.1 in k8s-operator.

## Brief change log

- Bump Flink to 1.19.1 in `pom.xml`.
- Adapt `StandaloneKubernetesJobManagerFactoryTest#testTemplateSpec` test case, because of the newly introduced `user-artifact-volume` in FLINK-28915.
- Force `GlobalConfiguration` to legacy YAML in `KubernetesScalingRealizerTest#testApplyOverrides` test case, otherwise the standard YAML parsing introduced in 1.19 changes the config value conversion. Ref: FLINK-33364.
- Remove duplicated classes that are not necessary anymore.
- Excluded the JDK-specific `jackson-core` parts from `flink-kubernetes`, as they are shaded incorrectly at the moment and clashes with the ones coming via `jackson-bom` defined in this repo.

## Verifying this change

## Documentation


flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/autoscaler/KubernetesScalingRealizerTest.java
```diff
@@ -48,8 +48,8 @@ public void testApplyOverrides() {
                                 .get(PipelineOptions.PARALLELISM_OVERRIDES.key()))
                 .satisfiesAnyOf(
                         // Currently no enforced order inside the overrides string
-                        overrides -> assertThat(overrides).isEqualTo("a:1,b:2"),
-                        overrides -> assertThat(overrides).isEqualTo("b:2,a:1"));
+                        overrides -> assertThat(overrides).isEqualTo("{a: '1', b: '2'}"),
+                        overrides -> assertThat(overrides).isEqualTo("{b: '2', a: '1'}"));
```
gyfora, line 52: What is going on here? Did the format change somehow? I am a bit confused

ferenc-csaky: I was trying to give an explanation in the PR desc :) But the main thing is that now [`GlobalConfiguration` has a `standardYaml` field](https://github.com/apache/flink/blob/a2d42b745b1b337bc92d4cd780b7126b4f2f7121/flink-core/src/main/java/org/apache/flink/configuration/GlobalConfiguration.java#L73), which affects [`ConfigurationUtils#convertValue`](https://github.com/apache/flink/blob/a2d42b745b1b337bc92d4cd780b7126b4f2f7121/flink-core/src/main/java/org/apache/flink/configuration/ConfigurationUtils.java#L524), which is used to read back the `overrides` result in this test case. If `standardYaml` would be `false` the result string would be in the old format, but by default it is `true` and here we have a dummy `Configuration` obj.

gyfora: we may want to disable this by default as this can cause backward compatibility issues with the autoscaler stored override values

mateczagany: Just as a side note, someone on Flink Slack mentioned that they cannot use the configuration `pipeline.serialization-config` with the operator because that would require the new YAML parser.

I think it would make sense to add support for the new `config.yaml` file for Flink resources and possibly for the operator itself in the future.

ferenc-csaky: Agree with @mateczagany, I started already to play with this. Will create a ticket for that soon.

ferenc-csaky:
What should we do with the following overlapping classes that fails the CI:
```
Warning:  flink-autoscaler-1.10-SNAPSHOT.jar, flink-runtime-1.19.1.jar define 2 overlapping classes: 
Warning:    - org.apache.flink.runtime.rest.messages.job.JobDetailsInfo
Warning:    - org.apache.flink.runtime.rest.messages.job.JobDetailsInfo$JobVertexDetailsInfo
Warning:  flink-kubernetes-1.19.1.jar, jackson-core-2.15.0.jar define 10 overlapping classes: 
Warning:    - META-INF.versions.11.com.fasterxml.jackson.core.io.doubleparser.FastDoubleSwar
Warning:    - META-INF.versions.17.com.fasterxml.jackson.core.io.doubleparser.FastIntegerMath
Warning:    - META-INF.versions.19.com.fasterxml.jackson.core.io.doubleparser.FastIntegerMath
Warning:    - META-INF.versions.19.com.fasterxml.jackson.core.io.doubleparser.FastIntegerMath$UInt128
Warning:    - META-INF.versions.11.com.fasterxml.jackson.core.io.doubleparser.FastIntegerMath$UInt128
Warning:    - META-INF.versions.19.com.fasterxml.jackson.core.io.doubleparser.FastDoubleSwar
Warning:    - META-INF.versions.17.com.fasterxml.jackson.core.io.doubleparser.FastIntegerMath$UInt128
Warning:    - META-INF.versions.11.com.fasterxml.jackson.core.io.doubleparser.BigSignificand
Warning:    - META-INF.versions.11.com.fasterxml.jackson.core.io.doubleparser.FastIntegerMath
Warning:    - META-INF.versions.17.com.fasterxml.jackson.core.io.doubleparser.FastDoubleSwar
```

The `JobDetailsInfo.JobVertexDetailsInfo` in Flink 1.19 still not has the `slotSharingGroupId` field that is used here, so the duplicated classes has to stay.

The jackson related classes define the overlap between a core Flink artifact and the official jackson artifact.

mateczagany:
I think upgrading the Jackson version from 2.15.0 to 2.15.3 to match Flink's version would fix the overlapping classes issue. Just make sure to also re-generate the NOTICE file as well if you go with this route.

Flink 1.19 upgraded the Jackson version to 2.15.3 from 2.14.3 in [FLINK-33340](https://issues.apache.org/jira/browse/FLINK-33340), but since Jackson 2.15, jackson-core is now a multi-release jar which will contain the class files mentioned above, and they are not shaded correctly in `flink-kubernetes`.

ferenc-csaky:
Thanks for your thorough check @mateczagany, I ended up not touching the Jackson version, as that mentioned Jackson bump got reverted, so Flink is still on 2.14.2.

gyfora:
@mateczagany we decided with @ferenc-csaky to remove the overlapping check, it doesn't seem to add any value it's only an annoyance

mateczagany (approved):
+1 for removing the check

PR #843, opened 2024-06-19
[FLINK-35640] Deprecate Flink version 1.15

mateczagany:

Flink 1.15 is being deprecated in the next release. As can be seen in the conversation below, there are no codepaths related to this Flink version, the decision was to deprecate the version, and the operator should still be able to create Flink deployments.

## Brief change log

- Deprecated Flink 1.15
- Update tests to use newer versions where it made sense
- Update beam example to use Flink 1.16

## Verifying this change

## Documentation

### DIFF
--- a/docs/content.zh/docs/concepts/overview.md
+++ b/docs/content.zh/docs/concepts/overview.md
@@ -44,7 +44,7 @@ Flink Kubernetes Operator 旨在承担人工操作 Flink 部署的职责。 人�
   - 有状态和无状态应用程序升级
   - 保存点的触发和管理
   - 处理错误，回滚失败的升级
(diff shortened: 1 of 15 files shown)


gyfora:
Is there any practical benefit of removing this version? For <1.15 we actually cleaned up some codepaths with relation to dropping the support. But here no code cleanup happens so it "could" work but we simply break it for users.

Alternatively we could simply mark the version deprecated in the CRD but not actually block it through the validation as long as we don't have any changes related to this.

gyfora:
Sounds good @mateczagany and if we see some opportunities to clean up 1.15 related codepaths in the operator or some new feature would require it, we could simply remove it later at any point in time.

gyfora (approved):

PR #848, opened 2024-07-02
Support config yaml

ctrlaltdilj:
## Support Flink 1.19 config.yaml configuration

Flink 1.19 introduced new [config.yaml](https://nightlies.apache.org/flink/flink-docs-master/docs/deployment/config/#flink-configuration-file) format. There are some [dependencies](https://github.com/apache/flink/blob/release-1.19.1/flink-core/src/main/java/org/apache/flink/api/common/serialization/SerializerConfigImpl.java#L378) on this new format in 1.19 that prevent old flink-conf.yaml from being able to configure properties.

This PR, adds support to be able to set a config.yaml file which is able to be consumed by the operator as well as deployments. 

## Brief change log
- added support to be able to set config.yaml by replacing flink-conf.yaml in the `defaultConfiguration`

## Verifying this change

  - Build image
  - create a values.yaml similar to 

## Documentation

  - If yes, how is the feature documented? not documented, will add


ctrlaltdilj:
I do think this would be better supported via a flag on configuration on FlinkDeployment, as it enables operator update independently  from FlinkDeployments

I am happy to make that change and would prefer that. Any thoughts?

mateczagany:
Hi, first of all, thank you for the contribution, but for better observability please try to always create a JIRA before opening a PR. 
Unless trying to push a fix for a release candidate, please use the `main` branch for your PRs. Flink version is already 1.19.1 on the main branch.

ctrlaltdilj:
Thanks @mateczagany for all the info! I have requested a JIRA account, and will file a ticket

ctrlaltdilj:
@mateczagany 
Created following ticket: https://issues.apache.org/jira/browse/FLINK-35744

helm/flink-kubernetes-operator/values.yaml
```diff
@@ -154,6 +154,9 @@ defaultConfiguration:
   # If set to false, loads just the overrides as in (2).
   # This option has not effect, if create is equal to false.
   append: true
+  # If set to true, then will support YAML 1.2 syntax through use of `config.yaml` file.
+  # If set to false, will make use of deprecated flink-conf.yaml filen.
+  standardYaml: false
```
gyfora, line 159: Would be great if this would be "automatic" based on whether the user specified flink-conf.yaml vs conf.yaml. Do you think that's possible?

pom.xml
```diff
@@ -83,7 +83,7 @@ under the License.
         <lombok.version>1.18.30</lombok.version>
         <commons-lang3.version>3.12.0</commons-lang3.version>
         <commons-io.version>2.11.0</commons-io.version>
-        <flink.version>1.18.1</flink.version>
+        <flink.version>1.19.0</flink.version>
```
ljankows, line 86: Can we also change the version to 1.19.1 as it was already released?

gyfora:
Hey @ctrlaltdilj !
Are you still working on this PR?

myskaludek:
Hi , i made changes to helm template and in values set 
```yaml
defaultConfiguration:
  standardYaml: true
```

The operator dont start pod with errror
```
Exception in thread "main" org.apache.flink.configuration.IllegalConfigurationException: The Flink config file '/opt/flink/conf/flink-conf.yaml' (/opt/flink/conf/flink-conf.yaml) does not exist.
```

ctrlaltdilj:
Hi @hugovideira-ppb , yes, apologize got busy, I'll try to knock this out

ctrlaltdilj:
@hugovideira-ppb , I have updated the helm chart, if this approach looks good, can I proceed with some documentation changes and adding some tests?

ctrlaltdilj:
@gyfora does this approach look better?

ctrlaltdilj:
@gyfora @ljankows whenever you get a chance to take a look, if the approach makes sense, I can proceed with some documentation changes and adding some tests?

gyfora:
I think this will require some more testing and some e2es to make sure that this type of configuration on the operator side still works with Flink versions.

Can we extend the e2e logic to cover this somehow?

ctrlaltdilj:
@gyfora that should be possible, let me take a look

ctrlaltdilj:
@gyfora added, let me know what you think, or if you have any feedback

e2e-tests/test_dynamic_flink_conf.sh
```diff
@@ -0,0 +1,40 @@
...
+
+TIMEOUT=360
+
+operator_namespace=$(get_operator_pod_namespace)
+operator_pod=$(get_operator_pod_name)
+echo "Current operator pod is ${operator_pod}"
+create_namespace dynamic
+
+kubectl config set-context --current --namespace="${operator_namespace}"
+patch_flink_config '{"data": {"flink-conf.yaml": "kubernetes.operator.watched.namespaces: default,flink,dynamic"}}'
```
ctrlaltdilj, line 37: this test is used to make sure flink-conf.yaml still works

helm/flink-kubernetes-operator/values.yaml
ctrlaltdilj: this approach does take care of this

PR #850, opened 2024-07-04
[FLINK-35126] Rework default checkpoint progress check window

gyfora:

Currently the checkpoint progress health check window is configurable by Duration. This makes it hard to enable by default as the sensible interval depends on the checkpoint interval. 

At the same time the operator already contains logic for a minimum progress check interval computed from the checkpoint timeout , tolerable failures and checkpoint interval.

Furthermore for any job with checkpointing enabled this health check is very valuable to have enabled by default similar to the restart health check. This PR also proposes to enable this feature by default with the the minimum checkpoint check interval set.

## Brief change log

  - Enable checkpoint progress health check by default (for jobs with checkpointing configured)
  - Set the minimum based on the calculated that already enforces the lower bound 

## Verifying this change

## Documentation

### DIFF
--- a/docs/layouts/shortcodes/generated/dynamic_section.html
+++ b/docs/layouts/shortcodes/generated/dynamic_section.html
@@ -22,15 +22,15 @@
         </tr>
         <tr>
             <td><h5>kubernetes.operator.cluster.health-check.checkpoint-progress.enabled</h5></td>
-            <td style="word-wrap: break-word;">false</td>
+            <td style="word-wrap: break-word;">true</td>
             <td>Boolean</td>
             <td>Whether to enable checkpoint progress health check for clusters.</td>
         </tr>
         <tr>
             <td><h5>kubernetes.operator.cluster.health-check.checkpoint-progress.window</h5></td>
-            <td style="word-wrap: break-word;">5 min</td>
+            <td style="word-wrap: break-word;">(none)</td>
             <td>Duration</td>
-            <td>If no checkpoints are completed within the defined time window, the job is considered unhealthy. This must be bigger than checkpointing interval.</td>
+            <td>If no checkpoints are completed within the defined time window, the job is considered unhealthy. The minimum window size is `max(checkpointingInterval, checkpointTimeout) * (tolerableCheckpointFailures + 2)`, which also serves as the default value when checkpointing is enabled. For example with checkpoint interval 10 minutes and 0 tolerable failures, the default progress check window will be 20 minutes.</td>
         </tr>
         <tr>
             <td><h5>kubernetes.operator.cluster.health-check.enabled</h5></td>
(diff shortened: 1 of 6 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/ClusterHealthEvaluator.java
```diff
@@ -178,30 +178,46 @@ private boolean evaluateCheckpoints(
...
-            LOG.warn(
-                    "{} is not long enough. Default to max({} * {}, {} * {}): {}ms",
-                    OPERATOR_CLUSTER_HEALTH_CHECK_CHECKPOINT_PROGRESS_WINDOW.key(),
-                    CHECKPOINTING_INTERVAL.key(),
-                    TOLERABLE_FAILURE_NUMBER.key(),
-                    CHECKPOINTING_TIMEOUT.key(),
-                    TOLERABLE_FAILURE_NUMBER.key(),
-                    minCompletedCheckpointsCheckWindow);
-            completedCheckpointsCheckWindow = Duration.ofMillis(minCompletedCheckpointsCheckWindow);
+        var tolerationFailureNumber = checkpointConfig.getTolerableCheckpointFailureNumber() + 2;
```
mateczagany, line 188: Neither the docs nor the debug message mention adding 2 to the tolerable failures. I guess it makes sense given that this can be 0, but I think it's still worth a mention in the docs and possibly the debug message below to avoid confusion.

gyfora: Good point adding this to the docs

gaborgsomogyi (approved):

docs/layouts/shortcodes/generated/dynamic_section.html
```diff
@@ -22,15 +22,15 @@
...
             <td>Boolean</td>
             <td>Whether to enable checkpoint progress health check for clusters.</td>
         </tr>
         <tr>
             <td><h5>kubernetes.operator.cluster.health-check.checkpoint-progress.window</h5></td>
-            <td style="word-wrap: break-word;">5 min</td>
+            <td style="word-wrap: break-word;">(none)</td>
             <td>Duration</td>
-            <td>If no checkpoints are completed within the defined time window, the job is considered unhealthy. This must be bigger than checkpointing interval.</td>
+            <td>If no checkpoints are completed within the defined time window, the job is considered unhealthy. The minimum window size is `max(checkpointingInterval, checkpointTimeout) * (tolerableCheckpointFailures + 2)`, which also serves as the default value when checkpointing is enabled. For example with checkpoint interval 10 minutes and 0 tolerable failures, the default progress check window will be 20 minutes.</td>
```
gaborgsomogyi, line 33: Cool, that we added a description. This would be a hidden gem otherwise.

PR #852, opened 2024-07-14
[FLINK-35831] Rotate jobId for both savepoint and stateless deploys

gyfora:

Job Ids can and should be rotated on savepoint redeployments as well. It is only important to keep the jobId when we are relying on recovery from HA metadata during last-state restore.

## Brief change log

 - Change jobid rotation logic from stateless only to (not last-state)

## Verifying this change

Unit tests, existing e2es

## Documentation


gaborgsomogyi (approved):

gaborgsomogyi:
Restarted the failed job, the change looks good.

PR #853, opened 2024-07-17
[FLINK-35815][autoscaler] Introduce GlobalConfiguration.loadConfiguration load config

huyuanfeng2018:

 Introduce GlobalConfiguration.loadConfiguration load config

## Brief change log
 - Introduce GlobalConfiguration.loadConfiguration load config

## Verifying this change

## Documentation


huyuanfeng2018:
@1996fanrui 
Can you review it for me? Thank you very much.

flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/StandaloneAutoscalerEntrypoint.java
```diff
@@ -48,7 +49,9 @@ public class StandaloneAutoscalerEntrypoint {
 
     public static <KEY, Context extends JobAutoScalerContext<KEY>> void main(String[] args)
             throws Exception {
-        var conf = ParameterTool.fromArgs(args).getConfiguration();
+        var conf =
+                GlobalConfiguration.loadConfiguration(
+                        ParameterTool.fromArgs(args).getConfiguration());
```
1996fanrui, line 54: Would you mind extracting this part as a separate method and adding a unit test? 

Separate method is helpful for adding a unit test to check whether overwrite logic is fine.

1996fanrui (changes requested):
Thanks @huyuanfeng2018 for the contribution!

It's better to add doc to explain how to use this feature.

huyuanfeng2018:
> Thanks @huyuanfeng2018 for the contribution!
> 
> It's better to add doc to explain how to use this feature.

> Thanks @huyuanfeng2018 for the contribution!
> 
> It's better to add doc to explain how to use this feature.


Thank you for your review,

I added a new commit including

* Added documentation

* Added test cases

* Supports changing the config file directory through environment variables, and introduces a new class Envs to manage program environment variables, but currently there is only `ENV_CONF_OVERRIDE_DIR`, which seems a bit thin in the long run, But under this PR, I think it is enough

flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/StandaloneAutoscalerEntrypoint.java
```diff
@@ -94,4 +100,15 @@ JobAutoScaler<KEY, Context> createJobAutoscaler(
...
+
+    @VisibleForTesting
+    static Configuration loadConfiguration(Optional<String> confOverrideDir, String[] args) {
+        if (confOverrideDir.isPresent()) {
+            return GlobalConfiguration.loadConfiguration(
+                    confOverrideDir.get(), ParameterTool.fromArgs(args).getConfiguration());
+        } else {
+            return GlobalConfiguration.loadConfiguration(
+                    ParameterTool.fromArgs(args).getConfiguration());
+        }
```
1996fanrui, line 112: Why do we need `CONF_OVERRIDE_DIR`?  As I understand, `FLINK_CONF_DIR` is enough.

docs/content/docs/custom-resource/autoscaler.md
```diff
@@ -308,6 +308,13 @@ org.apache.flink.autoscaler.standalone.StandaloneAutoscalerEntrypoint \
 --autoscaler.standalone.jdbc.username root
 ```
 
+### Load StandaloneAutoscaler configuration through configuration file
+
+Can specify the running configuration of StandaloneAutoscaler in the configuration file. The config.yaml and
+flink-conf.yaml files in the environment variable `FLINK_CONF_DIR` directory will be loaded by default. You can also pass
+`export CONF_OVERRIDE_DIR=xxx` to specify the loaded configuration file directory.
+NOTE: main configuration will overwrite the configuration in the configuration file
```
1996fanrui, line 316: How about update the `main configuration` to `main parameters`? 

I saw most of doc saw the Java Main parameters. WDYT?

huyuanfeng2018: > How about update the `main configuration` to `main parameters`?
> 
> I saw most of doc saw the Java Main parameters. WDYT?

+1.

flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/StandaloneAutoscalerEntrypoint.java
huyuanfeng2018: > Why do we need `CONF_OVERRIDE_DIR`? As I understand, `FLINK_CONF_DIR` is enough.

https://github.com/apache/flink-kubernetes-operator/blob/cb90b10c93241f9994acdd11d59ce1ec25ccbc47/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/EnvUtils.java#L54C32-L54C53

I referred here

huyuanfeng2018: > > Why do we need `CONF_OVERRIDE_DIR`? As I understand, `FLINK_CONF_DIR` is enough.
> 
> https://github.com/apache/flink-kubernetes-operator/blob/cb90b10c93241f9994acdd11d59ce1ec25ccbc47/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/EnvUtils.java#L54C32-L54C53
> 
> I referred here

I thought carefully that adding an environment variable is indeed a bit redundant, because FLINK_CONF_DIR itself does not have a default configuration. for the opinion that FLINK_CONF_DIR is enough.

1996fanrui (approved):
Thanks for the update!

nuno-c-afonso:
Hi,

The title of the PR references a Jira that is for a KDS connector bug:
- https://issues.apache.org/jira/browse/FLINK-35815

Would it be possible to change it to `FLINK-35851` to point to the right Jira?

Thank you.

huyuanfeng2018:
> Hi,
> 
> The title of the PR references a Jira that is for a KDS connector bug:
> 
> * https://issues.apache.org/jira/browse/FLINK-35815
> 
> Would it be possible to change it to `FLINK-35851` to point to the right Jira?
> 
> Thank you.

ok

PR #856, opened 2024-07-26
[FLINK-35896][autoscaler] Handle exception events in RescaleApiScalingRealizer

huyuanfeng2018:
Throw exceptions that occur during the execution of RescaleApiScalingRealizer so that eventHandler can handle it

## Brief change log
  - Throws exceptions that occur with `RescaleApiScalingRealizer#realizeParallelismOverrides`

## Verifying this change

- A test method `testRealizeParallelismOverridesExceptions` has been added to JobAutoScalerImplTest, which will simulate an exception occurring in realizeParallelismOverrides and detect whether the EventHandler handles the exception.

## Documentation

### DIFF
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/realizer/ScalingRealizer.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/realizer/ScalingRealizer.java
@@ -41,6 +41,10 @@ public interface ScalingRealizer<KEY, Context extends JobAutoScalerContext<KEY>>
     void realizeParallelismOverrides(Context context, Map<String, String> parallelismOverrides)
             throws Exception;
 
-    /** Updates the TaskManager memory configuration. */
-    void realizeConfigOverrides(Context context, ConfigChanges configChanges);
+    /**
+     * Updates the TaskManager memory configuration.
+     *
+     * @throws Exception Error during realize config overrides.
+     */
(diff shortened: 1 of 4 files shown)


flink-autoscaler/src/main/java/org/apache/flink/autoscaler/realizer/ScalingRealizer.java
```diff
@@ -33,8 +33,13 @@
...
+    /**
+     * Update job's parallelism to parallelismOverrides.
+     *
+     * @throws Exception Error during realize parallelism overrides.
+     */
+    void realizeParallelismOverrides(Context context, Map<String, String> parallelismOverrides)
+            throws Exception;
 
     /** Updates the TaskManager memory configuration. */
     void realizeConfigOverrides(Context context, ConfigChanges configChanges);
```
1996fanrui, line 45: I'm thinking should these 2 methods keep same semantic?  Such as: assert ` throws Exception` together.

It's a little wired: one throws, one not.

WDYT?

huyuanfeng2018: +1

1996fanrui (approved):
Thanks @huyuanfeng2018 for the contribution!

LGTM.

Hey @gyfora , would you mind double checking this PR as well? It will record the Exception to event handler when realizeParallelismOverrides throws exception.

gyfora (approved):

PR #600, opened 2023-05-15
FLINK-32103 Update RBAC for OpenShift Deployment

jbusche:
## What is the purpose of the change

Noticing in the 1.5.0-rc1 and 1.5.0-rc2 that we're having a problem deploying on OpenShift.  The solution is to either manually update the flink-kubernetes-operator.v1.5.0 clusterrole, but a better way is to update the rbac.yaml.  In either case, a `- flinkdeployments/finalizers` is needed in the flink.apache.org apiGroup.

The following PR adds this finalizer for the deployment to succeed.

## Brief change log

  - *the `- flinkdeployments/finalizers` resource is added to the  flink.apache.org apiGroup*

## Verifying this change
This change is a trivial rework / code cleanup without any test coverage.

## Does this pull request potentially affect one of the following parts:

  - Dependencies (does it add or upgrade a dependency): No
  - The public API, i.e., is any changes to the `CustomResourceDescriptors`: No
  - Core observer or reconciler logic that is regularly executed: No

## Documentation

  - Does this pull request introduce a new feature? No
  - If yes, how is the feature documented? not applicable

PR #857, opened 2024-07-31
[FLINK-35937] RBAC cleanup

timsn:

In further research and testing with Kyverno I figured out that some apiGroups seem to be invalid and I removed them with this PR.

It seems that the "extensions" apiGroups does not exist on our recent cluster (Kubernetes 1.29.4). I'm not sure but it might be related to these deprecation notices: https://kubernetes.io/docs/reference/using-api/deprecation-guide/#deployment-v116

Same holds for the "finalizers" resources. They do not seem to exist anymore and lead to problems with our deployment. So I also removed them.

To complete the verb list I also added "deleteCollections" where applicable.

Ref: https://issues.apache.org/jira/projects/FLINK/issues/FLINK-35310

## Brief change log

- Remove "invalid" RBAC apiGroups and resouces
- Replace another occurence of wildcard ("*") verbs with the actual verbs

## Verifying this change

## Documentation

### DIFF
--- a/helm/flink-kubernetes-operator/templates/rbac.yaml
+++ b/helm/flink-kubernetes-operator/templates/rbac.yaml
@@ -50,8 +50,6 @@ rules:
       - apps
     resources:
       - deployments
-      - deployments/scale
-      - deployments/finalizers
       - replicasets
     verbs:
       - get
@@ -61,26 +59,20 @@ rules:
       - update
       - patch
       - delete
+      - deletecollection
   - apiGroups:
-      - extensions
+      - apps
     resources:
-      - deployments
-      - ingresses
+      - deployments/scale
     verbs:
       - get
-      - list
-      - watch
-      - create
       - update
       - patch
-      - delete
   - apiGroups:
       - flink.apache.org
     resources:
       - flinkdeployments
-      - flinkdeployments/finalizers
       - flinksessionjobs
-      - flinksessionjobs/finalizers
     verbs:
       - get
       - list
@@ -89,6 +81,7 @@ rules:
       - update
       - patch
       - delete
+      - deletecollection
   - apiGroups:
       - flink.apache.org
     resources:
@@ -110,6 +103,7 @@ rules:
       - update
       - patch
       - delete
+      - deletecollection
   - apiGroups:
       - coordination.k8s.io
     resources:
@@ -122,6 +116,7 @@ rules:
       - update
       - patch
       - delete
+      - deletecollection
 {{- end }}
 
 {{/*
@@ -142,11 +137,11 @@ rules:
       - update
       - patch
       - delete
+      - deletecollection
   - apiGroups:
       - apps
     resources:
       - deployments
-      - deployments/finalizers
     verbs:
       - get
       - list
@@ -155,6 +150,7 @@ rules:
       - update
       - patch
       - delete
+      - deletecollection
 {{- end }}
 
 ---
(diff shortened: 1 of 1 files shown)


helm/flink-kubernetes-operator/templates/rbac.yaml
```diff
@@ -61,26 +59,20 @@ rules:
       - update
       - patch
       - delete
+      - deletecollection
   - apiGroups:
-      - extensions
+      - apps
     resources:
-      - deployments
-      - ingresses
```
mateczagany, line 68: If we wish to drop support for extensions/v1beta1 API version of Ingress, I think we can also remove the code in `IngressUtils.java:266`. This will make the operator unable to create new Ingress resources for Kubernetes 1.19, but that version has been deprecated for a long time.

helm/flink-kubernetes-operator/templates/rbac.yaml
```diff
@@ -61,26 +59,20 @@ rules:
...
-      - watch
-      - create
       - update
       - patch
-      - delete
   - apiGroups:
       - flink.apache.org
     resources:
       - flinkdeployments
-      - flinkdeployments/finalizers
```
mateczagany, line 81: Won't removing finalizers here affect the fix provided in https://github.com/apache/flink-kubernetes-operator/pull/600/? The finalizers were added here specifically for OpenShift, I think it's related to the `OwnerReferencesPermissionEnforcement` admission plugin.

helm/flink-kubernetes-operator/templates/rbac.yaml
timsn: I would appreciate getting rid of the extensions part because this is making some trouble with our deployment on a recent managed Kubernetes cluster (1.29) in combination with Kyverno policy management. Kyverno is preventing the deployment because the extensions API does not seem to exist anymore.

helm/flink-kubernetes-operator/templates/rbac.yaml
timsn: I'm not the best to judge this. But at least from my point of view the "/finializers" resources are also making trouble with the deployment because it looks like the resources do not exist and (like mentioned above) Kyverno is blocking our deployment because it cannot add verbs to non existing resources. That's at least how I understand it.
Here is the output of available resources on our cluster for the flink operator:

```
API               Resource                 Verbs
flink.apache.org  flinkdeployments         delete,deletecollection,get,list,patch,create,update,watch
flink.apache.org  flinkdeployments/status  get,patch,update
flink.apache.org  flinkdeployments/scale   get,patch,update
flink.apache.org  flinksessionjobs         delete,deletecollection,get,list,patch,create,update,watch
flink.apache.org  flinksessionjobs/status  get,patch,update
```

helm/flink-kubernetes-operator/templates/rbac.yaml
gyfora: @timsn If this is something you specifically don't need you could simply modify your downstream helm chart. But as long as we have users who need this we cannot simply remove it from the base helm template. We would need a discussion on the ML before that

timsn: @gyfora thanks for your response. That's something I cannot judge because I know not enough about what the extensions are used for and if they are still in use. I just discovered that they lead to problems with our deployment process (Kyverno policy checks).

If this would brake something I would be totally fine to remove the change from this PR and do the changes downstream on my own.

helm/flink-kubernetes-operator/templates/rbac.yaml
```diff
@@ -89,6 +81,7 @@ rules:
       - update
       - patch
       - delete
+      - deletecollection
```
gyfora, line 84: Why are we adding the delete collection verb to many places? Is this used somewhere?

timsn: I just added them for "completeness". Because in the past the verbs just was a star ("*"), a `deletecollection` is implicitly a part of it. If not needed, I can also remove it from this PR.

PR #860, opened 2024-08-09
[FLINK-35493][snapshot] Add historical cleanup for FlinkStateSnapshot CRs

mateczagany:
## Brief change log

- *Added InformerEventSource for Flink resources to be able to list linked snapshots during reconciliation*
- *If FlinkStateSnapshot resources are enabled, clean them up by age/count in SnapshotObserver*
- *Add unit tests*

## Verifying this change

- Added unit tests
- Manual testing with checkpoint and snapshot resources, with age and count based cleanup policies

## Documentation

  - If yes, how is the feature documented? documentation and examples will be provided in PR #854

### DIFF
--- a/docs/layouts/shortcodes/generated/dynamic_section.html
+++ b/docs/layouts/shortcodes/generated/dynamic_section.html
@@ -9,6 +9,24 @@
     </thead>
     <tbody>
         <tr>
+            <td><h5>kubernetes.operator.checkpoint.cleanup.enabled</h5></td>
+            <td style="word-wrap: break-word;">true</td>
+            <td>Boolean</td>
+            <td>Whether to enable clean up of checkpoint FlinkStateSnapshot resources created by the operator automatically. This will only remove the FlinkStateSnapshot CR, and will not remove any data on the filesystem.</td>
+        </tr>
+        <tr>
+            <td><h5>kubernetes.operator.checkpoint.history.max.age</h5></td>
+            <td style="word-wrap: break-word;">1 d</td>
+            <td>Duration</td>
+            <td>Maximum age for checkpoint FlinkStateSnapshot resources to retain. Due to lazy clean-up, the most recent checkpoint may live longer than the max age.</td>
+        </tr>
+        <tr>
+            <td><h5>kubernetes.operator.checkpoint.history.max.count</h5></td>
+            <td style="word-wrap: break-word;">10</td>
+            <td>Integer</td>
+            <td>Maximum number of checkpoint FlinkStateSnapshot resources to retain.</td>
+        </tr>
+        <tr>
             <td><h5>kubernetes.operator.checkpoint.trigger.grace-period</h5></td>
             <td style="word-wrap: break-word;">1 min</td>
             <td>Duration</td>
@@ -156,7 +174,7 @@
             <td><h5>kubernetes.operator.savepoint.cleanup.enabled</h5></td>
             <td style="word-wrap: break-word;">true</td>
             <td>Boolean</td>
-            <td>Whether to enable clean up of savepoint history.</td>
+            <td>Whether to enable clean up of savepoint FlinkStateSnapshot resources. Savepoint state will be disposed of as well if the snapshot CR spec is configured as such. For automatic savepoints this can be configured via the kubernetes.operator.savepoint.dispose-on-delete config option.</td>
         </tr>
         <tr>
             <td><h5>kubernetes.operator.savepoint.dispose-on-delete</h5></td>
@@ -174,13 +192,13 @@
             <td><h5>kubernetes.operator.savepoint.history.max.age</h5></td>
             <td style="word-wrap: break-word;">1 d</td>
             <td>Duration</td>
-            <td>Maximum age for savepoint history entries to retain. Due to lazy clean-up, the most recent savepoint may live longer than the max age.</td>
+            <td>Maximum age for savepoint FlinkStateSnapshot resources to retain. Due to lazy clean-up, the most recent savepoint may live longer than the max age.</td>
         </tr>
         <tr>
             <td><h5>kubernetes.operator.savepoint.history.max.count</h5></td>
             <td style="word-wrap: break-word;">10</td>
             <td>Integer</td>
-            <td>Maximum number of savepoint history entries to retain.</td>
+            <td>Maximum number of savepoint FlinkStateSnapshot resources entries to retain.</td>
         </tr>
         <tr>
             <td><h5>kubernetes.operator.savepoint.trigger.grace-period</h5></td>
--- a/docs/layouts/shortcodes/generated/dynamic_section.html
+++ b/docs/layouts/shortcodes/generated/dynamic_section.html
@@ -9,24 +9,6 @@
     </thead>
     <tbody>
         <tr>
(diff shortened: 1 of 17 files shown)


mateczagany:
@anupamaggarwal @gyfora @ferenc-csaky 

We need this PR to be able to automatically clean up FlinkStateSnapshot resources created by the operator.

In the current state only periodic snapshots will be cleaned up, I was not sure if we want to also include upgrade snapshots. I would love to hear your inputs on this.

I also created new configs for checkpoints, because I felt like users might want to use different cleanup policies for checkpoints than savepoints.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/SnapshotObserver.java
```diff
@@ -220,63 +235,179 @@ private void observeTriggeredCheckpoint(FlinkResourceContext<CR> ctx, String job
...
+            FlinkOperatorConfiguration operatorConfig,
+            SnapshotType snapshotType) {
+        var snapshotList =
+                snapshots.stream()
+                        .filter(s -> s.getStatus() != null)
+                        .filter(s -> COMPLETED.equals(s.getStatus().getState()))
+                        .filter(
+                                s ->
+                                        SnapshotTriggerType.PERIODIC.equals(
+                                                FlinkStateSnapshotUtils.getSnapshotTriggerType(s)))
```
gyfora, line 279: Feels like filtering everything out may leave a lot of lingering upgrade / abandoned / unfinished savepoint CRs over time. How are we planning to address that? 

I think upgrade save points can be deleted as long as we have at least 1 right?

gyfora:
Just for clarification, if the snapshot resource is not installed / enabled, will the old cleanup mechanism still work in the legacy mode?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/SnapshotObserver.java
mateczagany: You are right. I will add a check to make sure that during cleanup the latest COMPLETED snapshot won't be deleted. Then it should be safe to add upgrade savepoints as well.

mateczagany:
I forgot to add a check when setting up informer event sources, the latest commit should take care of that. I have manually checked and the cleanup is still working the same way it has been if CRD is not installed.

mateczagany:
After talking with @gyfora I have made it so legacy savepoints will be cleaned up even if FlinkStateSnapshot is enabled for the job. This will use the number of completed FlinkStateSnapshot CRs and subtract that from the `maxCount`. This should help integrating jobs from older operator versions.

I have also changed the logic a bit so we extract `metadata.creationTimestamp` to determine the age of the snapshot as `triggerTimestamp` might not always be set (e.g. FlinkStateSnapshot controller did not start).

When this PR gets merged, I will also update #862 to test cleaning up legacy savepoints with FlinkStateSnapshot enabled.

docs/layouts/shortcodes/generated/dynamic_section.html
```diff
@@ -8,6 +8,24 @@
...
+            <td style="word-wrap: break-word;">1 d</td>
+            <td>Duration</td>
+            <td>Maximum age for checkpoint FlinkStateSnapshot resources to retain. Due to lazy clean-up, the most recent checkpoint may live longer than the max age.</td>
+        </tr>
+        <tr>
+            <td><h5>kubernetes.operator.checkpoint.history.max.count</h5></td>
+            <td style="word-wrap: break-word;">10</td>
+            <td>Integer</td>
+            <td>Maximum number of checkpoint FlinkStateSnapshot resources to retain.</td>
+        </tr>
```
gyfora, line 28: I wonder if we could get rid of these 3 configs by:
 1. Always enabling cleanup (it doesn't actually delete any data anyways)
 2. For max count use the configured execution.checkpointing.num-retained (and get rid of the age based deletion for checkpoints)
 
 I think this will cover most use-cases and if we really need another cleanup policy for some reason then we can re-introduce this in the future maybe?

mateczagany: That sounds good. I wasn't sure how fine-grained these config options should be, but if you think that we don't need these, I am happy to make this PR more simple and reduce the amount of config options. 

It will be very easy to add these config options back in the future if we deem them necessary anyways.

gyfora: Sounds good :) always better to start with simple / less configs for new features

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/SnapshotObserver.java
```diff
@@ -219,64 +240,205 @@ private void observeTriggeredCheckpoint(FlinkResourceContext<CR> ctx, String job
...
+
+        for (var snapshot : snapshotList) {
+            if (snapshot.equals(lastCompleteSnapshot)) {
+                continue;
+            }
+
+            var ts = EXTRACT_SNAPSHOT_TIME.apply(snapshot).toEpochMilli();
+            if (snapshotList.size() - result.size() > maxCount || ts < maxTms) {
+                result.add(snapshot);
+            }
```
gyfora, line 323: We should probably short circuit this if size becomes 1 otherwise we may delete all of them if too old

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/SnapshotObserver.java
```diff
@@ -219,64 +240,205 @@ private void observeTriggeredCheckpoint(FlinkResourceContext<CR> ctx, String job
...
+
+        var completedSavepointCrs =
+                allSecondarySnapshotResources.stream()
+                        .filter(
+                                s ->
+                                        s.getStatus() != null
+                                                && COMPLETED.equals(s.getStatus().getState()))
+                        .filter(s -> s.getSpec().isSavepoint())
+                        .count();
+        maxCount = Math.max(0, maxCount - completedSavepointCrs);
```
gyfora, line 355: Shouldn't we always keep at least 1?

gyfora (changes requested):
Not sure if I missed something but seems to me that in the new cleanup logic we can easily end up In situations where all snapshots are deleted which would be good to avoid

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/SnapshotObserver.java
mateczagany: Good catch. I have added a check and test for this.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/SnapshotObserver.java
mateczagany: I have added another `Math.max` to `getMaxCountForSnapshotType`. Here we can use 0, as we should be able to clear the list of legacy savepoints if we have enough completed FlinkStateSnapshot resources.

gyfora (approved):

PR #863, opened 2024-08-12
[FLINK-36031] [kubernetes-client] Migrate to Fabric8 interceptor API

SamBarker:

Migrate to the Fabric8 kubernetes-client interceptor API so that metrics are captured regardless of which HTTP client is used. 

## Brief change log

Migrate to Fabric8 interceptor API

## Verifying this change

  - Additional tests added to verify that specific interactions are tracked by the `KubernetesClientMetrics` in `KubernetesClientMetricsFabric8InterceptorTest`. While maintaining the existing `KubernetesClientMetricsTest`.

  - Upgrades to the latest Fabric8 kubernetes-client release.
  - Core observer or reconciler logic that is regularly executed:  no)

## Documentation


pom.xml
```diff
@@ -78,7 +78,7 @@ under the License.
         <operator.sdk.version>4.8.3</operator.sdk.version>
         <operator.sdk.webhook-framework.version>1.1.1</operator.sdk.webhook-framework.version>
 
-        <fabric8.version>6.11.0</fabric8.version>
+        <fabric8.version>6.13.2</fabric8.version>
```
csviri, line 81: Note that NOTICE file should be updated on dependency change, 
see: https://github.com/apache/flink-kubernetes-operator/blob/main/flink-kubernetes-operator/src/main/resources/META-INF/NOTICE

csviri (changes requested):

pom.xml
SamBarker: Thanks for the heads up, fixed in a94b9ee6e88d0925b80501223f57db22fc9ff2ef

SamBarker:
I can't reproduce the build failure with either JDK 17 or 21 locally from the previous workflow run. I do see a lot of errors logged locally as well as in the failed run so any pointers as to whats wrong with the build would be really appreciated.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/metrics/KubernetesClientMetrics.java
```diff
@@ -118,32 +139,91 @@ public KubernetesClientMetrics(
...
-        Request request = chain.request();
-        updateRequestMetrics(request);
-        Response response = null;
-        final long startTime = System.nanoTime();
-        try {
-            response = chain.proceed(request);
-            return response;
-        } finally {
-            updateResponseMetrics(response, startTime);
+    public AsyncBody.Consumer<List<ByteBuffer>> consumer(
```
mateczagany, line 146: The `consumer` method is not called when building websockets, thus the new implementation outputting different request count than before originally.

I think we could use `before` instead since we don't modify the consumer here anyways. That one will be called for websockets as well.

https://github.com/fabric8io/kubernetes-client/blob/37ee45b94659879718fae7e9b77c2357a9e10c28/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/http/StandardHttpClient.java#L264

mateczagany:
Also, if you run the command `mvn -B clean install javadoc:javadoc -Pgenerate-docs -DskipTests`, that will regenerate the Kubernetes CRDs for you. With the fabric8 upgrade, some nested classes changed in `PodTemplate` which is used in the FlinkDeployment CRD.

You won't be able to push this change without also including the CRD changes.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/metrics/KubernetesClientMetrics.java
SamBarker: Argh. The problem with before is that I'm the request.id we see might not be the actual one sent as another filter can modify it after the invocation of our before callback.

I'll have to have another poke around the latency tracking.

SamBarker: I've pushed up https://github.com/apache/flink-kubernetes-operator/pull/863/commits/c4c9edaeaa66830a24523e5a996a8c98abc0511a which (ab)uses request headers to tack a start time onto a request and read it back when the request completes. I've also added a  test which triggers websockets to ensure they are counted.

SamBarker:
@mateczagany, @gyfora, @morhidi or @csviri could one of you kindly approve the PR to run through the workflow, I think its good to go but would appreciate the the confirmation from the actions.

gyfora (approved):

PR #865, opened 2024-08-14
[FLINK-36039][autoscaler] Support clean historical event handler records in JDBC event handler

RocMarshal:

Support clean historical event handler records in JDBC event handler

## Brief change log

image

## Verifying this change

Tested locally.

  - Core observer or reconciler logic that is regularly executed: (yes / **no**)

## Documentation

  - If yes, how is the feature documented? (**not applicable** / docs / JavaDocs / not documented)

### DIFF
--- a/flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/event/JdbcAutoScalerEventHandler.java
+++ b/flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/event/JdbcAutoScalerEventHandler.java
@@ -22,14 +22,22 @@ import org.apache.flink.autoscaler.JobAutoScalerContext;
 import org.apache.flink.autoscaler.ScalingSummary;
 import org.apache.flink.autoscaler.event.AutoScalerEventHandler;
 import org.apache.flink.runtime.jobgraph.JobVertexID;
+import org.apache.flink.util.Preconditions;
+
+import org.apache.flink.shaded.guava31.com.google.common.util.concurrent.ThreadFactoryBuilder;
 
 import lombok.SneakyThrows;
+import lombok.extern.slf4j.Slf4j;
 
 import javax.annotation.Nullable;
 
 import java.time.Duration;
+import java.util.Date;
 import java.util.Map;
 import java.util.Objects;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
 
 /**
  * The event handler which persists its event in JDBC related database.
@@ -38,13 +46,34 @@ import java.util.Objects;
  * @param <Context> The job autoscaler context.
  */
 @Experimental
+@Slf4j
 public class JdbcAutoScalerEventHandler<KEY, Context extends JobAutoScalerContext<KEY>>
         implements AutoScalerEventHandler<KEY, Context> {
 
     private final JdbcEventInteractor jdbcEventInteractor;
+    private final Duration eventHandlerTtl;
+    @Nullable private final ScheduledExecutorService scheduledEventHandlerCleaner;
 
-    public JdbcAutoScalerEventHandler(JdbcEventInteractor jdbcEventInteractor) {
+    public JdbcAutoScalerEventHandler(
+            JdbcEventInteractor jdbcEventInteractor, Duration eventHandlerTtl) {
         this.jdbcEventInteractor = jdbcEventInteractor;
+        this.eventHandlerTtl = Preconditions.checkNotNull(eventHandlerTtl);
+
+        if (eventHandlerTtl.toMillis() <= 0) {
+            this.scheduledEventHandlerCleaner = null;
+            return;
+        }
+        this.scheduledEventHandlerCleaner =
+                Executors.newSingleThreadScheduledExecutor(
+                        new ThreadFactoryBuilder()
+                                .setNameFormat("jdbc-autoscaler-events-cleaner")
+                                .setDaemon(false)
+                                .build());
+        this.scheduledEventHandlerCleaner.scheduleAtFixedRate(
+                this::cleanExpiredEvents,
+                Duration.ofDays(1).toMillis(),
+                Duration.ofDays(1).toMillis(),
+                TimeUnit.MILLISECONDS);
     }
 
     @SneakyThrows
@@ -101,6 +130,41 @@ public class JdbcAutoScalerEventHandler<KEY, Context extends JobAutoScalerContex
                     AutoScalerEventHandler.scalingReport(scalingSummaries, message),
                     AutoScalerEventHandler.getParallelismHashCode(scalingSummaries),
                     interval);
+        }
+    }
+
+    @Override
+    public void close() {
+        if (Objects.nonNull(scheduledEventHandlerCleaner)
+                && !scheduledEventHandlerCleaner.isShutdown()) {
+            scheduledEventHandlerCleaner.shutdownNow();
+        }
+    }
+
+    private void cleanExpiredEvents() {
+
+        try {
+            JdbcEventInteractor.ExpiredEventsResult expiredResult =
+                    jdbcEventInteractor.queryExpiredEventsAndMaxId(eventHandlerTtl);
+            if (Objects.isNull(expiredResult)) {
+                log.warn("No expired event handlers queried at {}", new Date());
+                return;
+            }
+            var numToClean = expiredResult.expiredRecords;
+            var batch = 4098;
+            var restInterval = 10L;
+            while (numToClean > 0) {
+                jdbcEventInteractor.deleteExpiredEventsByMaxIdAndBatch(expiredResult.maxId, batch);
+                numToClean -= batch;
+                log.debug("Deleted expired event handler records by batch successfully.");
+                Thread.sleep(restInterval);
+            }
+            log.info(
+                    "Deleted expired {} event handler records successfully whose id is smaller than {}.",
+                    expiredResult.expiredRecords,
+                    expiredResult.maxId);
+        } catch (Exception e) {
+            log.error("Error in cleaning expired event handler records.", e);
         }
     }
 
--- a/flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/event/JdbcAutoScalerEventHandler.java
+++ b/flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/event/JdbcAutoScalerEventHandler.java
@@ -34,12 +34,14 @@ import javax.annotation.Nullable;
 
 import java.sql.Timestamp;
 import java.time.Duration;
-import java.util.Date;
+import java.time.Instant;
 import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
+
+import static org.apache.flink.autoscaler.utils.DateTimeUtils.readable;
 
 /**
  * The event handler which persists its event in JDBC related database.
@@ -145,34 +147,32 @@ public class JdbcAutoScalerEventHandler<KEY, Context extends JobAutoScalerContex
 
     @VisibleForTesting
     void cleanExpiredEvents() {
+        final var batchSize = 2000;
+        final var sleepMs = 1000;
+
         var date =
                 Timestamp.from(
                         jdbcEventInteractor
                                 .getCurrentInstant()
                                 .minusMillis(eventHandlerTtl.toMillis()));
         try {
-            Long minId = jdbcEventInteractor.queryMinEventIdByCreateTime(date);
-            if (Objects.isNull(minId)) {
-                log.warn("No expired event handlers queried at {}", new Date());
-                return;
-            }
-            int deleted = 0, batch = 4098;
-            var count = 0L;
-            boolean cleanable = true;
-            for (long startId = minId; deleted == batch || cleanable; startId += batch) {
-                var endId = startId + batch;
-                deleted =
-                        jdbcEventInteractor.deleteExpiredEventsByIdRangeAndDate(
-                                startId, endId, date);
-                count += deleted;
-                if (deleted != batch) {
-                    cleanable = jdbcEventInteractor.queryMinEventIdByCreateTime(date) != null;
+            var deletedTotalCount = 0L;
+            while (true) {
+                Long minId = jdbcEventInteractor.queryMinEventIdByCreateTime(date);
+                if (Objects.isNull(minId)) {
+                    log.info("No expired event handlers queried at {}", readable(Instant.now()));
+                    break;
                 }
-                log.debug(
-                        "Deleted expired {} event handler records by batch successfully.", deleted);
-                Thread.sleep(10L);
+
+                for (long startId = minId;
+                        batchSize
+                                == jdbcEventInteractor.deleteExpiredEventsByIdRangeAndDate(
+                                        startId, startId + batchSize, date);
+                        startId += batchSize) {
+                    Thread.sleep(sleepMs);
+                }
             }
-            log.info("Deleted expired {} event handler records successfully", count);
+            log.info("Deleted expired {} event handler records successfully", deletedTotalCount);
         } catch (Exception e) {
             log.error("Error in cleaning expired event handler records.", e);
         }
--- a/flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/event/JdbcAutoScalerEventHandler.java
+++ b/flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/event/JdbcAutoScalerEventHandler.java
@@ -18,6 +18,7 @@
 package org.apache.flink.autoscaler.jdbc.event;
 
 import org.apache.flink.annotation.Experimental;
+import org.apache.flink.annotation.VisibleForTesting;
 import org.apache.flink.autoscaler.JobAutoScalerContext;
 import org.apache.flink.autoscaler.ScalingSummary;
 import org.apache.flink.autoscaler.event.AutoScalerEventHandler;
@@ -67,7 +68,7 @@ public class JdbcAutoScalerEventHandler<KEY, Context extends JobAutoScalerContex
                 Executors.newSingleThreadScheduledExecutor(
                         new ThreadFactoryBuilder()
                                 .setNameFormat("jdbc-autoscaler-events-cleaner")
-                                .setDaemon(false)
+                                .setDaemon(true)
                                 .build());
         this.scheduledEventHandlerCleaner.scheduleAtFixedRate(
                 this::cleanExpiredEvents,
@@ -141,8 +142,8 @@ public class JdbcAutoScalerEventHandler<KEY, Context extends JobAutoScalerContex
         }
     }
 
-    private void cleanExpiredEvents() {
-
+    @VisibleForTesting
+    void cleanExpiredEvents() {
         try {
             JdbcEventInteractor.ExpiredEventsResult expiredResult =
                     jdbcEventInteractor.queryExpiredEventsAndMaxId(eventHandlerTtl);
@@ -150,18 +151,19 @@ public class JdbcAutoScalerEventHandler<KEY, Context extends JobAutoScalerContex
                 log.warn("No expired event handlers queried at {}", new Date());
                 return;
             }
-            var numToClean = expiredResult.expiredRecords;
             var batch = 4098;
-            var restInterval = 10L;
-            while (numToClean > 0) {
-                jdbcEventInteractor.deleteExpiredEventsByMaxIdAndBatch(expiredResult.maxId, batch);
-                numToClean -= batch;
+            for (var startId = expiredResult.minId;
+                    startId <= expiredResult.maxId;
+                    startId += batch) {
+                jdbcEventInteractor.deleteExpiredEventsByIdRange(
+                        startId, Math.min(startId + batch, expiredResult.maxId));
                 log.debug("Deleted expired event handler records by batch successfully.");
-                Thread.sleep(restInterval);
+                Thread.sleep(10L);
             }
             log.info(
-                    "Deleted expired {} event handler records successfully whose id is smaller than {}.",
-                    expiredResult.expiredRecords,
+                    "Deleted expired {} event handler records successfully whose id is between {} and {}.",
+                    expiredResult.getExpiredCount(),
+                    expiredResult.minId,
                     expiredResult.maxId);
         } catch (Exception e) {
             log.error("Error in cleaning expired event handler records.", e);
--- a/flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/event/JdbcAutoScalerEventHandler.java
+++ b/flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/event/JdbcAutoScalerEventHandler.java
@@ -160,19 +160,18 @@ public class JdbcAutoScalerEventHandler<KEY, Context extends JobAutoScalerContex
             while (true) {
                 Long minId = jdbcEventInteractor.queryMinEventIdByCreateTime(date);
                 if (Objects.isNull(minId)) {
-                    log.info("No expired event handlers queried at {}", readable(Instant.now()));
+                    log.info("Deleted expired {} event handler records successfully", deletedTotalCount);
                     break;
                 }
 
                 for (long startId = minId;
-                        batchSize
-                                == jdbcEventInteractor.deleteExpiredEventsByIdRangeAndDate(
-                                        startId, startId + batchSize, date);
+                        jdbcEventInteractor.deleteExpiredEventsByIdRangeAndDate(
+                                        startId, startId + batchSize, date)
+                                == batchSize;
                         startId += batchSize) {
                     Thread.sleep(sleepMs);
                 }
             }
-            log.info("Deleted expired {} event handler records successfully", deletedTotalCount);
         } catch (Exception e) {
             log.error("Error in cleaning expired event handler records.", e);
         }
--- a/flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/event/JdbcAutoScalerEventHandler.java
+++ b/flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/event/JdbcAutoScalerEventHandler.java
@@ -34,14 +34,11 @@ import javax.annotation.Nullable;
 
 import java.sql.Timestamp;
 import java.time.Duration;
-import java.time.Instant;
 import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
-
-import static org.apache.flink.autoscaler.utils.DateTimeUtils.readable;
 
 /**
  * The event handler which persists its event in JDBC related database.
@@ -160,7 +157,9 @@ public class JdbcAutoScalerEventHandler<KEY, Context extends JobAutoScalerContex
             while (true) {
                 Long minId = jdbcEventInteractor.queryMinEventIdByCreateTime(date);
                 if (Objects.isNull(minId)) {
-                    log.info("Deleted expired {} event handler records successfully", deletedTotalCount);
+                    log.info(
+                            "Deleted expired {} event handler records successfully",
+                            deletedTotalCount);
                     break;
                 }
 
--- a/flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/event/JdbcAutoScalerEventHandler.java
+++ b/flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/event/JdbcAutoScalerEventHandler.java
@@ -32,6 +32,7 @@ import lombok.extern.slf4j.Slf4j;
 
 import javax.annotation.Nullable;
 
+import java.sql.Timestamp;
 import java.time.Duration;
 import java.util.Date;
 import java.util.Map;
@@ -144,27 +145,34 @@ public class JdbcAutoScalerEventHandler<KEY, Context extends JobAutoScalerContex
 
     @VisibleForTesting
     void cleanExpiredEvents() {
+        var date =
+                Timestamp.from(
+                        jdbcEventInteractor
+                                .getCurrentInstant()
+                                .minusMillis(eventHandlerTtl.toMillis()));
         try {
-            JdbcEventInteractor.ExpiredEventsResult expiredResult =
-                    jdbcEventInteractor.queryExpiredEventsAndMaxId(eventHandlerTtl);
-            if (Objects.isNull(expiredResult)) {
+            Long minId = jdbcEventInteractor.queryMinEventIdByCreateTime(date);
+            if (Objects.isNull(minId)) {
                 log.warn("No expired event handlers queried at {}", new Date());
                 return;
             }
-            var batch = 4098;
-            for (var startId = expiredResult.minId;
-                    startId <= expiredResult.maxId;
-                    startId += batch) {
-                jdbcEventInteractor.deleteExpiredEventsByIdRange(
-                        startId, Math.min(startId + batch, expiredResult.maxId));
-                log.debug("Deleted expired event handler records by batch successfully.");
+            int deleted = 0, batch = 4098;
+            var count = 0L;
+            boolean cleanable = true;
+            for (long startId = minId; deleted == batch || cleanable; startId += batch) {
+                var endId = startId + batch;
+                deleted =
+                        jdbcEventInteractor.deleteExpiredEventsByIdRangeAndDate(
+                                startId, endId, date);
+                count += deleted;
+                if (deleted != batch) {
+                    cleanable = jdbcEventInteractor.queryMinEventIdByCreateTime(date) != null;
+                }
+                log.debug(
+                        "Deleted expired {} event handler records by batch successfully.", deleted);
                 Thread.sleep(10L);
             }
-            log.info(
-                    "Deleted expired {} event handler records successfully whose id is between {} and {}.",
-                    expiredResult.getExpiredCount(),
-                    expiredResult.minId,
-                    expiredResult.maxId);
+            log.info("Deleted expired {} event handler records successfully", count);
         } catch (Exception e) {
             log.error("Error in cleaning expired event handler records.", e);
         }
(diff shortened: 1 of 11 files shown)


flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/StandaloneAutoscalerExecutor.java
```diff
@@ -113,6 +113,8 @@ public void start() {
     public void close() {
         scheduledExecutorService.shutdownNow();
         scalingThreadPool.shutdownNow();
+        autoScaler.close();
+        eventHandler.close();
```
1996fanrui, line 117: IIUC, `eventHandler.close()` is called twice.

We don't need to introduce the close for `autoScaler`, because we could close eventHandler directly.

flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/config/AutoscalerStandaloneOptions.java
```diff
@@ -121,4 +121,11 @@ private static ConfigOptions.OptionBuilder autoscalerStandaloneConfig(String key
                                             code(EVENT_HANDLER_TYPE.key()),
                                             code("JDBC"))
                                     .build());
+
+    public static final ConfigOption<Duration> JDBC_EVENT_HANDLER_TTL =
+            autoscalerStandaloneConfig("jdbc-event-handler.ttl")
```
1996fanrui, line 126: how about `jdbc.event-handler.ttl`?

Other options are `jdbc.xxx`.

flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/AutoscalerEventHandlerFactory.java
```diff
@@ -73,6 +75,6 @@ AutoScalerEventHandler<KEY, Context> create(Configuration conf) throws Exception
             AutoScalerEventHandler<KEY, Context> createJdbcEventHandler(Configuration conf)
                     throws Exception {
         var conn = HikariJDBCUtil.getConnection(conf);
-        return new JdbcAutoScalerEventHandler<>(new JdbcEventInteractor(conn));
+        return new JdbcAutoScalerEventHandler<>(new JdbcEventInteractor(conn), Duration.ZERO);
```
1996fanrui, line 78: Why don't get `JDBC_EVENT_HANDLER_TTL` from conf?

1996fanrui (changes requested):

flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/event/JdbcEventInteractor.java
```diff
@@ -152,4 +154,49 @@ protected List<AutoScalerEvent> queryEvents(String jobKey, String reason) throws
...
+                return null;
+            }
+            var result = new ExpiredEventsResult(resultSet.getInt(1), resultSet.getLong(2));
+            resultSet.close();
+            return result;
+        }
+    }
+
+    public void deleteExpiredEventsByMaxIdAndBatch(long maxTargetId, int batch) throws Exception {
+        var query = "delete from t_flink_autoscaler_event_handler where id < ? limit ?";
```
1996fanrui, line 181:
```suggestion
        var query = "delete from t_flink_autoscaler_event_handler where id <= ? limit ?";
```

It should be `<=` ？

flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/event/JdbcEventInteractor.java
```diff
@@ -152,4 +154,49 @@ protected List<AutoScalerEvent> queryEvents(String jobKey, String reason) throws
     void setClock(@Nonnull Clock clock) {
         this.clock = Preconditions.checkNotNull(clock);
     }
+
+    @Nullable
+    ExpiredEventsResult queryExpiredEventsAndMaxId(Duration eventHandlerTtl) throws Exception {
```
1996fanrui, line 159: It's better to test the clean logic in AbstractJdbcAutoscalerEventHandlerITCase.

1996fanrui (changes requested):

flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/StandaloneAutoscalerExecutor.java
RocMarshal: >  IIUC, eventHandler.close() is called twice.

Yes.
But, There're  no semantic errors, it's just an extra call.

> We don't need to introduce the close for autoScaler, because we could close eventHandler directly.
Yes, In the initial edition, I also did it this way.

The reason why I done it as the current way is: 
In the current attributes of `StandaloneAutoscalerExecutor.java`, introducing a close method for `eventHandler` alone would meet the requirements and ensure good semantics. However, since `autoScaler` also references `eventHandler`, would it be better to close the objects that reference `eventHandler` before closing `eventHandler ` itself ?

Of course, The I'd like to update it based on the comments from you.

flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/config/AutoscalerStandaloneOptions.java
RocMarshal: thanks for the comments.
> jdbc.event-handler.ttl
Let's do it this way.
Consistent naming conventions are better!

flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/StandaloneAutoscalerExecutor.java
1996fanrui: >  In the current attributes of StandaloneAutoscalerExecutor.java, introducing a close method for eventHandler alone would meet the requirements and ensure good semantics. However, since autoScaler also references eventHandler, would it be better to close the objects that reference eventHandler before closing eventHandler itself ?


If so, should we introduce the `close` for `JobVertexScaler` and `ScalingExecutor` and close them as well? you can check the code, `eventHandler` is not only passed to `autoScaler`, but also to `JobVertexScaler` and `ScalingExecutor`.

flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/event/JdbcEventInteractor.java
RocMarshal: Nice idea. 
added.

flink-autoscaler-standalone/src/main/java/org/apache/flink/autoscaler/standalone/AutoscalerEventHandlerFactory.java
RocMarshal: sorry & thank for the catching.
updated.

flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/event/JdbcAutoScalerEventHandler.java
```diff
@@ -38,13 +47,34 @@
...
+
+        if (eventHandlerTtl.toMillis() <= 0) {
+            this.scheduledEventHandlerCleaner = null;
+            return;
+        }
+        this.scheduledEventHandlerCleaner =
+                Executors.newSingleThreadScheduledExecutor(
+                        new ThreadFactoryBuilder()
+                                .setNameFormat("jdbc-autoscaler-events-cleaner")
+                                .setDaemon(false)
```
1996fanrui, line 71: Why it's false here?

flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/event/JdbcEventInteractor.java
```diff
@@ -92,12 +94,23 @@ public void createEvent(
...
+
+    @VisibleForTesting
+    void createEvent(
+            String jobKey,
+            String reason,
+            AutoScalerEventHandler.Type type,
+            String message,
+            String eventKey,
+            Timestamp createTime)
+            throws Exception {
```
1996fanrui, line 108: Don't need introduce new method for testing, you could setClock to control the timestamp.

Check `AbstractJdbcAutoscalerEventHandlerITCase#testEventIntervalWithoutMessageKey`

flink-autoscaler-plugin-jdbc/src/test/java/org/apache/flink/autoscaler/jdbc/event/AbstractJdbcAutoscalerEventHandlerITCase.java
```diff
@@ -430,6 +447,75 @@ void testScalingEventWithParallelismChange() throws Exception {
...
+        eventHandler.cleanExpiredEvents();
+
+        // Check
+        try (ResultSet countResultSet =
+                        quickQuery("select count(1) from t_flink_autoscaler_event_handler");
+                ResultSet createTimeResultSet =
+                        quickQuery("select create_time from t_flink_autoscaler_event_handler")) {
+            assertThat(countResultSet).isNotNull();
+            assertThat(countResultSet.next()).isTrue();
+            assertThat(countResultSet.getInt(1)).isOne();
```
1996fanrui, line 475: Could we extract the unexpired record number into one parameter? It could cover more cases.

flink-autoscaler-plugin-jdbc/src/test/java/org/apache/flink/autoscaler/jdbc/event/AbstractJdbcAutoscalerEventHandlerITCase.java
```diff
@@ -430,6 +447,75 @@ void testScalingEventWithParallelismChange() throws Exception {
...
+
+            assertThat(createTimeResultSet).isNotNull();
+            assertThat(createTimeResultSet.next()).isTrue();
+            assertThat(createTimeResultSet.getTimestamp(1)).isEqualTo(createTimeOfUnExpiredRecord);
+        } catch (SQLException e) {
+            throw new RuntimeException(e);
```
1996fanrui, line 481: Don't understand why need this catch?

flink-autoscaler-plugin-jdbc/src/test/java/org/apache/flink/autoscaler/jdbc/event/AbstractJdbcAutoscalerEventHandlerITCase.java
```diff
@@ -430,6 +447,75 @@ void testScalingEventWithParallelismChange() throws Exception {
...
+        final Clock clock = Clock.systemDefaultZone();
+        initExpiredEventHandlerRecords(eventHandlerTtl, expiredRecordsNum, clock);
+        final Timestamp createTimeOfUnExpiredRecord = Timestamp.from(clock.instant());
+        jdbcEventInteractor.createEvent(
+                "jobKey1",
+                "reason",
+                AutoScalerEventHandler.Type.Normal,
+                "message",
+                "eventKey",
+                createTimeOfUnExpiredRecord);
```
1996fanrui, line 463: unexpired is a word, right?

flink-autoscaler-plugin-jdbc/src/test/java/org/apache/flink/autoscaler/jdbc/event/AbstractJdbcAutoscalerEventHandlerITCase.java
```diff
@@ -430,6 +447,75 @@ void testScalingEventWithParallelismChange() throws Exception {
...
     }
 
+    @MethodSource("getExpiredEventHandlersCaseMatrix")
+    @ParameterizedTest
+    void testCleanExpiredEvents(int expiredRecordsNum, Duration eventHandlerTtl) throws Exception {
+
+        final Clock clock = Clock.systemDefaultZone();
+        initExpiredEventHandlerRecords(eventHandlerTtl, expiredRecordsNum, clock);
+        final Timestamp createTimeOfUnExpiredRecord = Timestamp.from(clock.instant());
+        jdbcEventInteractor.createEvent(
```
1996fanrui, line 457: Why don't insert event via event handler? It's a little wired that creating event hanldler after inserting data. In addion, It could cover more cases.

flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/event/JdbcEventInteractor.java
```diff
@@ -152,4 +165,49 @@ protected List<AutoScalerEvent> queryEvents(String jobKey, String reason) throws
...
+        var date = Timestamp.from(clock.instant().minusMillis(eventHandlerTtl.toMillis()));
+        try (var pstmt = conn.prepareStatement(query)) {
+            pstmt.setObject(1, date);
+            pstmt.setObject(2, date);
+            ResultSet resultSet = pstmt.executeQuery();
+            if (!resultSet.next()) {
+                return null;
+            }
+            var result = new ExpiredEventsResult(resultSet.getInt(1), resultSet.getLong(2));
+            resultSet.close();
```
1996fanrui, line 186: Don't need it, copy from the comment of `java.sql.ResultSet#close`. Let us keep all queries are same style.

`Note: A ResultSet object is automatically closed by the Statement object that generated it when that Statement object is closed, re-executed, or is used to retrieve the next result from a sequence of multiple results.`

1996fanrui:
CI is failed, please take a look as well, thanks~

flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/event/JdbcAutoScalerEventHandler.java
1996fanrui: Why it's not true here?

1996fanrui: I'm curious what you think about this? In which cases should we use true and in which cases should we use false.

And in the current case, why do you use false? (I don't care if false is set manually or is the default value).

It is necessary for developers to know the meaning of every line of code they write and why.

RocMarshal: - IIUC, There is no explicit exit operation in the entire service, so the default value of false can be used here. It is not necessary to set it to true to automatically exit after waiting for other threads to finish.
- Explicit call for readability

1996fanrui: > There is no explicit exit operation in the entire service

Jdbc event handler doesn't know who call itself, it doesn't know Autoscaler Standalone, and doesn't know k8s operator.

RocMarshal (commented):
Thanks @1996fanrui for the quick review.
I updated it based on the comments. PTAL~

flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/event/JdbcAutoScalerEventHandler.java
```diff
@@ -104,6 +134,43 @@ public void handleScalingEvent(
...
+                    jdbcEventInteractor.queryExpiredEventsAndMaxId(eventHandlerTtl);
+            if (Objects.isNull(expiredResult)) {
+                log.warn("No expired event handlers queried at {}", new Date());
+                return;
+            }
+            var batch = 4098;
+            var startId = expiredResult.minId;
+            var endId = Math.min(startId + batch, expiredResult.maxId);
+            while (startId <= expiredResult.maxId) {
+                jdbcEventInteractor.deleteExpiredEventsByIdRange(startId, endId);
```
RocMarshal, line 158: The `limit` is not supported in postgres delete clause.
So, I try to use the current way, but I found there's no sequence related to the id of the corresponding table for auto incremental primary key in the postgresql ddl.

Should we add it if the current way was accepted?

flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/event/JdbcAutoScalerEventHandler.java
```diff
@@ -104,6 +134,43 @@ public void handleScalingEvent(
...
+                startId += batch;
+                endId = Math.min(endId + batch, expiredResult.maxId);
+                log.debug("Deleted expired event handler records by batch successfully.");
+                Thread.sleep(10L);
+            }
```
1996fanrui, line 163: It seems the loop is pretty easier than while loop, and we don't need to introduce a series of variables.

```suggestion
            var batch = 4098;
            for (var startId = expiredResult.minId; startId <= expiredResult.maxId; startId += batch) {
                jdbcEventInteractor.deleteExpiredEventsByIdRange(startId, Math.min(start + batch, expiredResult.maxId));
                log.debug("Deleted expired event handler records by batch successfully.");
                Thread.sleep(10L);
            }
```

flink-autoscaler-plugin-jdbc/src/test/java/org/apache/flink/autoscaler/jdbc/event/AbstractJdbcAutoscalerEventHandlerITCase.java
```diff
@@ -430,6 +446,70 @@ void testScalingEventWithParallelismChange() throws Exception {
...
+        }
+        jdbcEventInteractor.setClock(defaultClock);
+    }
+
+    private static Stream<Arguments> getExpiredEventHandlersCaseMatrix() {
+        return Stream.of(
+                Arguments.of(1024, Duration.ofMinutes(2), 1),
+                Arguments.of(1024 * 5, Duration.ofMinutes(2), 2),
+                Arguments.of(1024, Duration.ofMinutes(100), 3),
+                Arguments.of(1024 * 5, Duration.ofMinutes(100), 100));
```
1996fanrui, line 510: This test don't cover the case that `unexpiredRecordsNum` is 0.

flink-autoscaler-plugin-jdbc/src/test/java/org/apache/flink/autoscaler/jdbc/event/AbstractJdbcAutoscalerEventHandlerITCase.java
```diff
@@ -430,6 +446,70 @@ void testScalingEventWithParallelismChange() throws Exception {
...
+    @MethodSource("getExpiredEventHandlersCaseMatrix")
+    @ParameterizedTest
+    void testCleanExpiredEvents(
+            int expiredRecordsNum, Duration eventHandlerTtl, int unexpiredRecordsNum)
+            throws Exception {
+        try (Connection con = getConnection();
+                PreparedStatement ps =
+                        con.prepareStatement("delete from t_flink_autoscaler_event_handler")) {
+            ps.execute();
+        }
```
1996fanrui, line 458: Did you find the `t_flink_autoscaler_event_handler` table has data?

flink-autoscaler-plugin-jdbc/src/test/java/org/apache/flink/autoscaler/jdbc/event/AbstractJdbcAutoscalerEventHandlerITCase.java
```diff
@@ -430,6 +446,70 @@ void testScalingEventWithParallelismChange() throws Exception {
                         });
     }
 
+    @MethodSource("getExpiredEventHandlersCaseMatrix")
+    @ParameterizedTest
+    void testCleanExpiredEvents(
+            int expiredRecordsNum, Duration eventHandlerTtl, int unexpiredRecordsNum)
```
1996fanrui, line 452: It's a little wired that `JdbcAutoScalerEventHandler#eventHandlerTtl` is always 0, but it could clean expired records.

flink-autoscaler-plugin-jdbc/src/test/java/org/apache/flink/autoscaler/jdbc/event/AbstractJdbcAutoscalerEventHandlerITCase.java
RocMarshal: Here would be some data resulted by the parameterized test running due to the shared database instance.

flink-autoscaler-plugin-jdbc/src/test/java/org/apache/flink/autoscaler/jdbc/event/AbstractJdbcAutoscalerEventHandlerITCase.java
RocMarshal: Thanks for comments.
IIUC, Maybe it's accepted, because in the current case, we just want to test the method `cleanExpiredEvents` of JdbcAutoScalerEventHandler instead of the whole logic of the `JdbcAutoScalerEventHandler`.
If needed, I'd like to  do some change for the stable test cases.

flink-autoscaler-plugin-jdbc/src/test/java/org/apache/flink/autoscaler/jdbc/event/AbstractJdbcAutoscalerEventHandlerITCase.java
```diff
@@ -430,6 +446,74 @@ void testScalingEventWithParallelismChange() throws Exception {
...
+    private static Stream<Arguments> getExpiredEventHandlersCaseMatrix() {
+        return Stream.of(
+                Arguments.of(1024, Duration.ofMinutes(2), 1),
+                Arguments.of(1024, Duration.ofMinutes(2), 0),
+                Arguments.of(1024 * 5, Duration.ofMinutes(2), 2),
+                Arguments.of(1024 * 5, Duration.ofMinutes(2), 0),
+                Arguments.of(1024, Duration.ofMinutes(100), 3),
+                Arguments.of(1024, Duration.ofMinutes(100), 0),
+                Arguments.of(1024 * 5, Duration.ofMinutes(100), 100),
+                Arguments.of(1024 * 5, Duration.ofMinutes(100), 0));
```
1996fanrui, line 514: You could try run `unexpiredRecordsNum = 0` locally, all records cannot be cleaned.

In addition, it's needed to add the case that `expiredRecordsNum = 0.`

1996fanrui (changes requested):

flink-autoscaler-plugin-jdbc/src/test/java/org/apache/flink/autoscaler/jdbc/event/AbstractJdbcAutoscalerEventHandlerITCase.java
1996fanrui: Actually, PostgreSQLExtension and MySQLExtension have `afterEach`, it cleans all data.

But they missed `t_flink_autoscaler_event_handler` table. I submitted https://github.com/apache/flink-kubernetes-operator/pull/866 to fix it.

flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/event/JdbcEventInteractor.java
```diff
@@ -160,9 +160,9 @@ ExpiredEventsResult queryExpiredEventsAndMaxId(Duration eventHandlerTtl) throws
         var query =
                 "SELECT min(id) min_id, max(id) max_id "
                         + "FROM t_flink_autoscaler_event_handler "
-                        + "WHERE create_time < ? AND id < ("
+                        + "WHERE create_time < ? AND id <= ("
                         + "   SELECT id FROM t_flink_autoscaler_event_handler "
-                        + "   WHERE create_time >= ? ORDER BY id ASC LIMIT 1)";
+                        + "   WHERE create_time < ? ORDER BY id DESC LIMIT 1)";
```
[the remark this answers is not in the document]
RocMarshal: thx for the comments. 
Made some change for it. PTAL~ :)

flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/event/JdbcAutoScalerEventHandler.java
```diff
@@ -104,6 +135,49 @@ public void handleScalingEvent(
...
+        try {
+            Long minId = jdbcEventInteractor.queryMinEventIdByCreateTime(date);
+            if (Objects.isNull(minId)) {
+                log.warn("No expired event handlers queried at {}", new Date());
+                return;
+            }
+            int deleted = 0, batch = 4098;
+            var count = 0L;
+            boolean cleanable = true;
+            for (long startId = minId; deleted == batch || cleanable; startId += batch) {
```
1996fanrui, line 162: 1. IIUC, `deleted == batch` isn't needed. If they are not equal, you updated the cleanable, so we check cleanable is enough.
2. If 1 is acceptable, we don't need the `for loop`. `while(true) loop` is enough, and we can break the loop if `cleanable` is false.
3. If 1 is acceptable, `deleted` can be defined inside of `loop`.
4. If 2 is acceptable, we don't need `cleanable`. We can check `jdbcEventInteractor.queryMinEventIdByCreateTime(date) == null`, if it's true, break directly.

flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/event/JdbcAutoScalerEventHandler.java
```diff
@@ -104,6 +135,49 @@ public void handleScalingEvent(
...
+                                .getCurrentInstant()
+                                .minusMillis(eventHandlerTtl.toMillis()));
+        try {
+            Long minId = jdbcEventInteractor.queryMinEventIdByCreateTime(date);
+            if (Objects.isNull(minId)) {
+                log.warn("No expired event handlers queried at {}", new Date());
+                return;
+            }
+            int deleted = 0, batch = 4098;
+            var count = 0L;
```
1996fanrui, line 160:
```suggestion
            var deletedTotalCount = 0L;
```

flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/event/JdbcAutoScalerEventHandler.java
```diff
@@ -104,6 +135,49 @@ public void handleScalingEvent(
...
+                        jdbcEventInteractor
+                                .getCurrentInstant()
+                                .minusMillis(eventHandlerTtl.toMillis()));
+        try {
+            Long minId = jdbcEventInteractor.queryMinEventIdByCreateTime(date);
+            if (Objects.isNull(minId)) {
+                log.warn("No expired event handlers queried at {}", new Date());
+                return;
+            }
+            int deleted = 0, batch = 4098;
```
1996fanrui, line 159: How about renaming it to batchSize?

flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/event/JdbcAutoScalerEventHandler.java
```diff
@@ -104,6 +135,49 @@ public void handleScalingEvent(
...
+                deleted =
+                        jdbcEventInteractor.deleteExpiredEventsByIdRangeAndDate(
+                                startId, endId, date);
+                count += deleted;
+                if (deleted != batch) {
+                    cleanable = jdbcEventInteractor.queryMinEventIdByCreateTime(date) != null;
+                }
+                log.debug(
+                        "Deleted expired {} event handler records by batch successfully.", deleted);
+                Thread.sleep(10L);
```
1996fanrui, line 173: How about sleep 1s?

One minute could clean 4096 * 60 = 240 k, it's totally enough.

flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/event/JdbcEventInteractor.java
```diff
@@ -152,4 +153,29 @@ protected List<AutoScalerEvent> queryEvents(String jobKey, String reason) throws
     void setClock(@Nonnull Clock clock) {
         this.clock = Preconditions.checkNotNull(clock);
     }
+
+    @Nullable
+    Long queryMinEventIdByCreateTime(Timestamp timestamp) throws Exception {
+        var sql =
+                "SELECT id from t_flink_autoscaler_event_handler "
+                        + "           where id = (SELECT id FROM t_flink_autoscaler_event_handler order by id asc limit 1) "
+                        + "           and create_time < ?";
```
1996fanrui, line 162: `SELECT id, create_time FROM t_flink_autoscaler_event_handler order by id asc limit 1` is enough, we can check create_time in memory.

flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/event/JdbcAutoScalerEventHandler.java
```diff
@@ -104,6 +135,49 @@ public void handleScalingEvent(
...
+            var count = 0L;
+            boolean cleanable = true;
+            for (long startId = minId; deleted == batch || cleanable; startId += batch) {
+                var endId = startId + batch;
+                deleted =
+                        jdbcEventInteractor.deleteExpiredEventsByIdRangeAndDate(
+                                startId, endId, date);
+                count += deleted;
+                if (deleted != batch) {
+                    cleanable = jdbcEventInteractor.queryMinEventIdByCreateTime(date) != null;
```
1996fanrui, line 169: IIUC, if it's not null, we should update the `startId = jdbcEventInteractor.queryMinEventIdByCreateTime(date)`.

flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/event/JdbcAutoScalerEventHandler.java
```diff
@@ -104,6 +135,49 @@ public void handleScalingEvent(
...
+
+    @VisibleForTesting
+    void cleanExpiredEvents() {
+        var date =
+                Timestamp.from(
+                        jdbcEventInteractor
+                                .getCurrentInstant()
+                                .minusMillis(eventHandlerTtl.toMillis()));
+        try {
+            Long minId = jdbcEventInteractor.queryMinEventIdByCreateTime(date);
```
1996fanrui, line 154: We could define a nested loops:

- The outer loop check `Long minId = jdbcEventInteractor.queryMinEventIdByCreateTime(date)`, and `minId != null`
- The inner loop check `deleted == batchSize`.

flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/event/JdbcAutoScalerEventHandler.java
RocMarshal: Thank you very much for comments.

> IIUC, deleted == batch isn't needed. If they are not equal, you updated the cleanable, so we check cleanable is enough.

Yes, it does  overall. But here's a small difference in this condition.
The `deleted==batch` would reduce the sql query from the DB, 

`jdbcEventInteractor.queryMinEventIdByCreateTime(date)` This query function is more like searching for the starting point of an interval composed of consecutive IDs that meet the deletion criteria. the `deleted==batch` would drive the deletion action in the consecutive IDs.

Please correct me if I'm wrong.

flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/event/JdbcAutoScalerEventHandler.java
RocMarshal: nice catch

flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/event/JdbcAutoScalerEventHandler.java
1996fanrui: From the code, `jdbcEventInteractor.queryMinEventIdByCreateTime(date) != null;` is only called when `deleted!=batch`.

It means `jdbcEventInteractor.queryMinEventIdByCreateTime(date) != null;` won't be called when `deleted==batch`. And `for loop` will delete next round.

I don't understand why updating `deleted == batch || cleanable` to `cleanable` will query more times from Database.

flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/event/JdbcAutoScalerEventHandler.java
1996fanrui: And please add some tests to cover this case.

flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/event/JdbcEventInteractor.java
RocMarshal: thanks for the comment.

I conducted some tests based on the above logic, and this SQL statement has these two characteristics:
- It would run a few times  per triggering..
- Running on a large amount of data takes very little time. Even equivalent to this sentence `SELECT id FROM t_flink_autoscaler_event_handler order by id asc limit 1`

How about keeping it as now? of course, I'd like to updated it if needed.

flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/event/JdbcAutoScalerEventHandler.java
RocMarshal: pls https://github.com/apache/flink-kubernetes-operator/pull/865#discussion_r1728461523

flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/event/JdbcEventInteractor.java
1996fanrui: > How about keeping it as now? of course, I'd like to updated it if needed.

It's up to you, both of them are fine for me. 

This sql only query the first row, so the performance isn't my concern.

I leave this comment because sql looks weird at first glance.

1996fanrui (approved):

Hey @huyuanfeng2018 , as I know, your employer also needs this feature.  Would you mind reviewing this PR as well?  thank you very much.

flink-autoscaler-plugin-jdbc/src/main/java/org/apache/flink/autoscaler/jdbc/event/JdbcAutoScalerEventHandler.java
RocMarshal: added.

huyuanfeng2018 (approved):
Thanks to @RocMarshal for contribution and @1996fanrui for review, the code looks good to me!

PR #868, opened 2024-08-20
[FLINK-36110][snapshot] Store periodic snapshot trigger timestamps in memory

mateczagany:

SnapshotObserver uses deprecated fields as the latest timestamp of triggered snapshots. These fields are not updated when using FlinkStateSnapshot though. This new method of storing these timestamps is meant to fix this case.

## Brief change log

- Add new class to store/retrieve last timestamp a snapshot was triggered
- Updated relevant code to store/retrieve last timestamp in new class

## Verifying this change

- Added unit tests
- Manual testing

## Documentation

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/SnapshotTriggerTimestampStore.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/SnapshotTriggerTimestampStore.java
@@ -26,13 +26,13 @@ import org.apache.flink.kubernetes.operator.api.status.SnapshotInfo;
 import io.fabric8.kubernetes.api.model.HasMetadata;
 import lombok.RequiredArgsConstructor;
 
-import javax.annotation.Nullable;
-
 import java.time.Instant;
 import java.util.Comparator;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Supplier;
 
 import static org.apache.flink.kubernetes.operator.api.status.FlinkStateSnapshotStatus.State.COMPLETED;
 import static org.apache.flink.kubernetes.operator.api.status.SnapshotTriggerType.PERIODIC;
@@ -41,8 +41,10 @@ import static org.apache.flink.kubernetes.operator.reconciler.SnapshotType.SAVEP
 /** Class used to store latest timestamps of periodic checkpoint/savepoint. */
 @RequiredArgsConstructor
 public class SnapshotTriggerTimestampStore {
-    private final SnapshotType snapshotType;
-    private final ConcurrentHashMap<String, Instant> lastTriggered = new ConcurrentHashMap<>();
+    private final ConcurrentHashMap<String, Instant> checkpointsLastTriggeredCache =
+            new ConcurrentHashMap<>();
+    private final ConcurrentHashMap<String, Instant> savepointsLastTriggeredCache =
+            new ConcurrentHashMap<>();
 
     /**
      * Returns the time a periodic snapshot was last triggered for this resource. This is stored in
@@ -51,16 +53,20 @@ public class SnapshotTriggerTimestampStore {
      * and the latest triggered legacy snapshot.
      *
      * @param resource Flink resource
-     * @param snapshots related snapshot resources
+     * @param snapshotsSupplier supplies related snapshot resources
      * @return instant of last trigger
      */
     public Instant getLastPeriodicTriggerInstant(
-            AbstractFlinkResource<?, ?> resource, @Nullable Set<FlinkStateSnapshot> snapshots) {
-        if (lastTriggered.containsKey(resource.getMetadata().getUid())) {
-            return lastTriggered.get(resource.getMetadata().getUid());
+            AbstractFlinkResource<?, ?> resource,
+            SnapshotType snapshotType,
+            Supplier<Set<FlinkStateSnapshot>> snapshotsSupplier) {
+        var cache = getCacheForSnapshotType(snapshotType);
+        if (cache.containsKey(resource.getMetadata().getUid())) {
+            return cache.get(resource.getMetadata().getUid());
         }
 
-        return Optional.ofNullable(snapshots).orElse(Set.of()).stream()
+        var snapshots = Optional.ofNullable(snapshotsSupplier.get()).orElse(Set.of());
+        return snapshots.stream()
                 .filter(s -> s.getStatus() != null && COMPLETED.equals(s.getStatus().getState()))
                 .filter(s -> (snapshotType == SAVEPOINT) == s.getSpec().isSavepoint())
                 .filter(
@@ -74,7 +80,7 @@ public class SnapshotTriggerTimestampStore {
                 .max(Comparator.naturalOrder())
                 .orElseGet(
                         () -> {
-                            var legacyTs = getLegacyTimestamp(resource);
+                            var legacyTs = getLegacyTimestamp(resource, snapshotType);
                             var creationTs =
                                     Instant.parse(resource.getMetadata().getCreationTimestamp());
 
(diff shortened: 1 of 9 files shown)


flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconcilerTest.java
```diff
@@ -654,28 +653,6 @@ private void testSnapshotLegacy(FlinkDeployment deployment, SnapshotType snapsho
         assertTrue(isSnapshotInProgress.test(getJobStatus(snDeployment)));
         assertEquals(SnapshotStatus.PENDING, getLastSnapshotStatus(snDeployment, snapshotType));
         snDeployment.getSpec().getFlinkConfiguration().put(triggerSnapshotExpression.key(), "0");
-
```
mateczagany, line 657: I have deleted this test case as I could not find a good way to update the recorded last periodic timestamp without modifying access to the `SnapshotTriggerTimestampStore` fields. 

We already test the CRON ability by testing methods `SnapshotUtils#shouldTriggerAutomaticSnapshot` and `SnapshotUtils#interpretAsCron`.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/SnapshotTriggerTimestampStore.java
```diff
@@ -0,0 +1,114 @@
...
+                break;
+            case CHECKPOINT:
+                snapshotInfo = resource.getStatus().getJobStatus().getCheckpointInfo();
+                break;
+            default:
+                throw new IllegalArgumentException("Unsupported snapshot type: " + snapshotType);
+        }
+
+        return Instant.ofEpochMilli(snapshotInfo.getLastPeriodicTriggerTimestamp());
+    }
```
gyfora, line 113: Feels like we could get rid of the snapshotType field and pass the accessor logic in the constructor (and maybe instantiate with a static utility?)

gyfora: ah never mind I see that type field is used in other places too, we can keep it :)

mateczagany: Yep, once we remove these deprecated fields, we will be able to delete that field as well.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractJobReconciler.java
```diff
@@ -374,13 +382,27 @@ private boolean triggerSnapshotIfNeeded(FlinkResourceContext<CR> ctx, SnapshotTy
...
+        var timestampStore =
+                snapshotType == SAVEPOINT
+                        ? savepointTriggerTimestamps
+                        : checkpointTriggerTimestamps;
+        Set<FlinkStateSnapshot> snapshots = Set.of();
+        if (FlinkStateSnapshotUtils.isSnapshotResourceEnabled(ctx.getOperatorConfig(), conf)) {
+            snapshots = ctx.getJosdkContext().getSecondaryResources(FlinkStateSnapshot.class);
+        }
+
+        var lastTrigger = timestampStore.getLastPeriodicTriggerInstant(resource, snapshots);
```
gyfora, line 394: feels like in almost all cases we don't actually need to get the secondary resource. As long as we have the periodic timestamp already available. Can we pass a supplier or have some other lazy logic?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractJobReconciler.java
```diff
@@ -74,6 +77,11 @@ public abstract class AbstractJobReconciler<
 
     public static final String LAST_STATE_DUMMY_SP_PATH = "KUBERNETES_OPERATOR_LAST_STATE";
 
+    private final SnapshotTriggerTimestampStore checkpointTriggerTimestamps =
+            new SnapshotTriggerTimestampStore(CHECKPOINT);
+    private final SnapshotTriggerTimestampStore savepointTriggerTimestamps =
+            new SnapshotTriggerTimestampStore(SAVEPOINT);
```
gyfora, line 83: Feels like this should be a single object to wrap these two and hide the details

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/SnapshotTriggerTimestampStore.java
```diff
@@ -0,0 +1,133 @@
...
+                            var legacyTs = getLegacyTimestamp(resource, snapshotType);
+                            var creationTs =
+                                    Instant.parse(resource.getMetadata().getCreationTimestamp());
+
+                            if (legacyTs.compareTo(creationTs) > 0) {
+                                return legacyTs;
+                            } else {
+                                return creationTs;
+                            }
+                        });
```
gyfora, line 92: Should we put into the cache before returning?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/SnapshotTriggerTimestampStore.java
```diff
@@ -0,0 +1,133 @@
...
+import static org.apache.flink.kubernetes.operator.api.status.SnapshotTriggerType.PERIODIC;
+import static org.apache.flink.kubernetes.operator.reconciler.SnapshotType.SAVEPOINT;
+
+/** Class used to store latest timestamps of periodic checkpoint/savepoint. */
+@RequiredArgsConstructor
+public class SnapshotTriggerTimestampStore {
+    private final ConcurrentHashMap<String, Instant> checkpointsLastTriggeredCache =
+            new ConcurrentHashMap<>();
+    private final ConcurrentHashMap<String, Instant> savepointsLastTriggeredCache =
+            new ConcurrentHashMap<>();
```
gyfora, line 47: Do we want to ever clean this up? It's not gonna be a large size but I just wanted to point out that we "could" clean it up when a resource is deleted

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractJobReconciler.java
```diff
@@ -360,6 +367,25 @@ public boolean reconcileOtherChanges(FlinkResourceContext<CR> ctx) throws Except
...
+            FlinkResourceContext<CR> ctx) {
+        return () -> {
+            if (FlinkStateSnapshotUtils.isSnapshotResourceEnabled(
+                    ctx.getOperatorConfig(), ctx.getObserveConfig())) {
+                return ctx.getJosdkContext().getSecondaryResources(FlinkStateSnapshot.class);
+            } else {
+                return Set.of();
+            }
+        };
+    }
```
gyfora, line 387: Could this logic live in a static utility somewhere else? Maybe even in the timestamp storage or some other utility class

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/SnapshotTriggerTimestampStore.java
mateczagany: I like this idea, with this change we won't unnecessarily call the supplier if we don't trigger a periodic snapshot in AbstractJobReconciler.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/SnapshotTriggerTimestampStore.java
mateczagany: In my opinion this has such minimal memory footprint that we can safely ignore and not complicate the code more. Please let me know if you think otherwise.

gyfora: fine by me :)

gyfora (approved):

PR #873, opened 2024-08-27
[FLINK-36145] Rename flinkStateSnapshotReference to initialState

gyfora:

Rename JobSpec.flinkStateSnapshotReference to JobSpec.initialState. This should make less verbose and easier to memorize for users. 

flinkStateSnapshotReference field has not been released officially yet, so this change should not affect production users.

## Brief change log

- Rename in Java, Markdown files and examples
- Regenerate CRDs

## Verifying this change

- Unit tests

## Documentation

  - If yes, how is the feature documented? updated


gyfora:
Actually @tweise @mateczagany , I have a slight concern about this initialState reference field overall. It may make sense to simply keep the current `initialSavepointPath` field, because having a reference like field may allow users to "steal" the savepointpath of other users through the operator's permissions.

With the old way you either know the path or you don't.

PR #875, opened 2024-08-28
[FLINK-36018][autoscaler] Support lazy scale down to avoid frequent rescaling

1996fanrui:

When the source traffic decreases, the autoscaler scales down job in time to save resources. It causes that each job rescales more than 20 times a day.

## Brief change log

- [FLINK-36018][autoscaler] Support lazy scale down to avoid frequent rescaling
- Configuration Option:
  - Introduce job.autoscaler.scale-down.interval, the default value could be 1 hour.
  - Replace job.autoscaler.scale-up.grace-period with job.autoscaler.scale-down.interval
- Core code change:
  -  Introducing the `DelayedScaleDown` to store the firstTriggerTime for all vertices.
      - The scale down can be triggered when the (now -  firstTriggerTime  > scale-down.interval).
  - Updating the return value type of `JobVertexScaler#computeScaleTargetParallelism` from `int` to `ParallelismResult`
      - `ParallelismResult` introduced the `required` field, it means the parallelism change of current vertex is required or optional.
      - In general, scale up the required, and scale down is optional if it's within `scale-down.interval`.
      - If all parallelism changes are optional, we could ignore rescaling.

## Verifying this change

- Added some unit tests for `JobVertexScalerTest` and `ScalingExecutorTest`.
- Improved the test of `AbstractAutoScalerStateStoreTest`

  - The public API, i.e., is any changes to the `CustomResourceDescriptors`: yes, update the autoscaler option.

## Documentation

  - Does this pull request introduce a new feature?  yes
  - If yes, how is the feature documented? Added the option description.

### DIFF
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
@@ -45,7 +45,9 @@ import java.time.Duration;
 import java.time.Instant;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
 import java.util.SortedMap;
 
 import static org.apache.flink.autoscaler.config.AutoScalerOptions.EXCLUDED_PERIODS;
@@ -100,22 +102,23 @@ public class ScalingExecutor<KEY, Context extends JobAutoScalerContext<KEY>> {
             Map<JobVertexID, SortedMap<Instant, ScalingSummary>> scalingHistory,
             ScalingTracking scalingTracking,
             Instant now,
-            JobTopology jobTopology)
+            JobTopology jobTopology,
+            DelayedScaleDown delayedScaleDown)
             throws Exception {
         var conf = context.getConfiguration();
         var restartTime = scalingTracking.getMaxRestartTimeOrDefault(conf);
 
         var scalingSummaries =
                 computeScalingSummary(
-                        context, evaluatedMetrics, scalingHistory, restartTime, jobTopology);
+                        context,
+                        evaluatedMetrics,
+                        scalingHistory,
+                        restartTime,
+                        jobTopology,
+                        delayedScaleDown);
 
         if (scalingSummaries.isEmpty()) {
             LOG.info("All job vertices are currently running at their target parallelism.");
-            return false;
-        }
-
-        if (allVerticesWithinUtilizationTarget(
-                evaluatedMetrics.getVertexMetrics(), scalingSummaries)) {
             return false;
         }
 
@@ -156,6 +159,9 @@ public class ScalingExecutor<KEY, Context extends JobAutoScalerContext<KEY>> {
 
         autoScalerStateStore.storeConfigChanges(context, configOverrides);
 
+        // Try to clear all delayed scale down requests after scaling.
+        delayedScaleDown.clearAll();
+
         return true;
     }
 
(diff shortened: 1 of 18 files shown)


gyfora (changes requested):
Looks good, my only concern is related to the behaviour after a scale up. I feel like we should reset the `DelayedScaleDown. firstTriggerTimes` to the scale up time to avoid scaling down after a scale up

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
```diff
@@ -156,6 +159,9 @@ public boolean scaleResource(
 
         autoScalerStateStore.storeConfigChanges(context, configOverrides);
 
+        // Try to clear all delayed scale down requests after scaling.
+        delayedScaleDown.clearAll();
```
1996fanrui, line 163: Hi @gyfora , reply your comment[1] here.

IIUC, the previous logic still exists, if the rescale happens, `delayedScaleDown.clearAll` will be called (rescale includes scale up). It means the first trigger time is reset. 

Not sure whether your question is answered.

[1] https://github.com/apache/flink-kubernetes-operator/pull/875#discussion_r1738062495

1996fanrui: This logic isn't same with the previous logic.

## One demo of previous logic

- 9:00 scale up (be executed)
- 9:30 scale down (ignore)
- 9:59 scale down (ignore)
- 10:00 scale down (can be executed)

## New logic 

- 9:00 scale up (be executed, clear all first trigger time)
- 9:30 scale down (update the first trigger time to 9:30)
- 10:00 scale down (within interval from 9:30 to 10:30)
- 10:29 scale down (within interval from 9:30 to 10:30)
- 10:30 scale down (can be executed)

gyfora: So if there was a scale down before 9:00 and that set the first trigger time, that doesn't matter as the scale up resets it subsequently.

In that case we should probably rename the method ` detectImmediateScaleDown` to something like `applyScaleDownInterval`

1996fanrui:
Hey @gyfora , thank you for the quick review, all comments are addressed. ❤️

gyfora (approved):
Looks pretty good, let's give @mxm 1-2 days to comment on this :)

gyfora:
@1996fanrui I think we can go ahead with this :)

1996fanrui:
> @1996fanrui I think we can go ahead with this :) 

Thank you, let me merge it

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
```diff
@@ -102,13 +102,20 @@ private static ConfigOptions.OptionBuilder autoScalerConfig(String key) {
...
                             "Target vertex utilization boundary. Scaling won't be performed if the processing capacity is within [target_rate / (target_utilization - boundary), (target_rate / (target_utilization + boundary)]");
-    public static final ConfigOption<Duration> SCALE_UP_GRACE_PERIOD =
-            autoScalerConfig("scale-up.grace-period")
+
+    public static final ConfigOption<Duration> SCALE_DOWN_INTERVAL =
+            autoScalerConfig("scale-down.interval")
                     .durationType()
                     .defaultValue(Duration.ofHours(1))
-                    .withFallbackKeys(oldOperatorConfigKey("scale-up.grace-period"))
+                    .withDeprecatedKeys(autoScalerConfigKey("scale-up.grace-period"))
```
mxm, line 110: Thanks for adding this as a deprecated key for existing users!

mxm (commented):
Thanks @1996fanrui and @gyfora for this improvement! The changes look good to me. Thank you also for waiting for any comments from my side.

PR #879, opened 2024-09-10
[FLINK-36192][autocaler] Autocaler supports adjusting the parallelism of source vertex based on the number of partitions in Kafka or pulsars

huyuanfeng2018:
Autoscaler adjusts the parallelism of the corresponding vertex according to the number of partitions in Kafka or Pulsar, so that the parallelism is a divisor of the number of partitions.

## Brief change log
  - *Add a new `ScalingMetric.NUM_PARTITIONS` to record partition count of kafka or pulsar*
  - *adjusts `org.apache.flink.autoscaler.JobVertexScaler.scale` : This method also attempts to adjust the parallelism to ensure it aligns well with the number of partitions if a vertex has a known partition count*
  - *adjusts `org.apache.flink.autoscaler.JobVertexScaler.scale` Return exception information that occurs during the adjustment process*
  - *The `eventhandler` will handle events where the final degree of parallelism does not meet expectations due to the number of partitions or maxparallelism limitations.*

## Verifying this change
  - *Added test cases `org.apache.flink.autoscaler.JobVertexScalerTest#testNumPartitionsAdjustment` and `org.apache.flink.autoscaler.JobVertexScalerTest#testSendingScalingLimitedEvents`*

## Documentation


flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
```diff
@@ -191,16 +200,29 @@ public ParallelismChange computeScaleTargetParallelism(
         double cappedTargetCapacity = averageTrueProcessingRate * scaleFactor;
         LOG.debug("Capped target processing capacity for {} is {}", vertex, cappedTargetCapacity);
 
-        int newParallelism =
+        Tuple2<Integer, Optional<String>> newParallelism =
                 scale(
```
mxm, line 204: Instead of returning `Tuple2` here to return a message in addition to the new parallelism, can we pass the event handler to the scale method? We are already adding an additional parameter to the scale method. Eventually, we may want to reduce the number of parameters by grouping some of the parameters in wrapper objects, but no need to do this here.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
```diff
@@ -378,28 +405,70 @@ protected static int scale(
...
+        } else {
+
+            // When we know the numPartitions at a vertex,
+            // adjust the parallelism such that it divides the numPartitions without a remainder
+            // => Data is evenly distributed among subtasks
+            for (int p = newParallelism; p <= upperBound && p <= numPartitions; p++) {
+                if (numPartitions % p == 0) {
+                    return Tuple2.of(p, Optional.empty());
+                }
             }
```
mxm, line 439: This looks like code duplication to me. Only the for loop termination condition changed. We should be able to pass an argument to the for loop which we set based on the number of partitions.

We may even completely simplify like this:

```suggestion
        if (numPartitions <= 0) {
            // No partition information is available, assume numPartitions equals the number of key groups
            numPartitions = maxParallelism;
        }
       for (int p = newParallelism; p <= maxParallelism / 2 && p <= upperBound; p++) {
            if (maxParallelism % p == 0) {
                return Tuple2.of(p, Optional.empty());
            }
        }
        // If parallelism adjustment fails, use originally computed parallelism
        return Tuple2.of(newParallelism, Optional.empty());
```

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
```diff
@@ -378,28 +405,70 @@ protected static int scale(
...
+            }
+            // If a suitable degree of parallelism cannot be found, return parallelismLowerLimit
+            var message =
+                    String.format(
+                            SCALE_LIMITED_MESSAGE_FORMAT,
+                            vertex,
+                            newParallelism,
+                            parallelismLowerLimit,
+                            String.format("parallelismLowerLimit : %s", parallelismLowerLimit));
+            return Tuple2.of(parallelismLowerLimit, Optional.of(message));
```
mxm, line 470: I'm not sure about this behavior. Why return the lower limit, instead of the originally computed target parallelism? I think we should retain this logic: 

```java
           // If parallelism adjustment fails, use originally computed parallelism
           return newParallelism;
```

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
```diff
@@ -378,28 +405,70 @@ protected static int scale(
...
-        // If parallelism adjustment fails, use originally computed parallelism
-        return newParallelism;
+            // When the degree of parallelism after rounding up cannot be evenly divided by source
+            // PartitionCount, Try to find the smallest parallelism that can satisfy the current
+            // consumption rate.
+            for (int p = newParallelism; p > parallelismLowerLimit; p--) {
+                if (numPartitions / p > numPartitions / newParallelism) {
+                    if (numPartitions % p != 0) {
+                        p += 1;
+                    }
```
mxm, line 448: Why `p++` here? Maybe I'm overlooking something but `p` already divides `numPartitions` without a remainder.

mxm (commented):
Thanks @huyuanfeng2018 for this PR! A couple suggestions and a comment for my understanding.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
huyuanfeng2018: +1,
```suggestion
        int newParallelism =
            scale(
                vertex,
                currentParallelism,
                inputShipStrategies,
                (int) evaluatedMetrics.get(NUM_PARTITIONS).getCurrent(),
                (int) evaluatedMetrics.get(MAX_PARALLELISM).getCurrent(),
                scaleFactor,
                Math.min(currentParallelism, conf.getInteger(VERTEX_MIN_PARALLELISM)),
                Math.max(currentParallelism, conf.getInteger(VERTEX_MAX_PARALLELISM)),
                scalingLimitedMsg -> autoScalerEventHandler.handleEvent(
                    context,
                    AutoScalerEventHandler.Type.Warning,
                    SCALING_LIMITED,
                    scalingLimitedMsg,
                    SCALING_LIMITED + vertex + cappedTargetCapacity,
                    conf.get(SCALING_EVENT_INTERVAL)));
```
How about we pass in a function, This looks quite clear and does not require changing too many test cases.

huyuanfeng2018 (commented):
@mxm I explained some logic. You can review again when you have time. Thank you very much !

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
huyuanfeng2018: Resolved.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
huyuanfeng2018: > Right, I missed that. I was trying to generalize the two code blocks. How about the following?

Thansk, Fine with me , However, I do not recommend overriding the values ​​​​of MaxParallelism and UpperBound, so I added two new variables instead:

1. adjustableMaxParallelism( Indicates the MaxParallelism  in the adjustment process )
2. adjustableUpperBound (Indicates the UpperBound in the adjustment process)

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
huyuanfeng2018: > Thanks for explaining in detail. I misread some of the code. It is correct that we need to add +1 when we have found a parallelism which yields a greater value for `num_partitions / p` than the initial `num_partitions / new_parallelism` because we have found the tipping point where we achieve the most utilization in terms of partitions per task.
> 
> I think we should return `new_parallelism` if all adaptation logic fails because using a potentially very small configured lower parallelism could make things a lot worse due to resource constraints.

I want to explain the reason for using `parallelismLowerLimit`,  example:

`numPartitions=35 ，newParallelism=24， upperBound = 30, parallelismLowerLimit = 19`

Step1 cannot get a result, so it goes to step2, but step2 still cannot get a result because parallelismLowerLimit = 19 and the expected value of step2 is 18, so it will eventually return 19

mxm: On second thought, it's probably ok to use the lower limit. As you said, we would already be approaching the limit. Most users will never run into this because they haven't configured a minimum parallelism.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
mxm: Can we call them like this?

1. adjustableMaxParallelism => numKeyGroupsOrPartitions
2. adjustableUpperBound =>  upperBoundForAlignment  

Adjustable just doesn't tell someone who is unfamiliar with the code very much.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
```diff
@@ -389,15 +413,51 @@ protected static int scale(
...
+            consumer.accept(
+                    String.format(
+                            SCALE_LIMITED_MESSAGE_FORMAT,
+                            vertex,
+                            newParallelism,
+                            finalParallelism,
+                            String.format(
+                                    "numPartitions : %s，upperBound(maxParallelism or parallelismUpperLimit): %s, "
+                                            + "parallelismLowerLimit: %s.",
+                                    numPartitions, upperBound, parallelismLowerLimit)));
```
mxm, line 457: I would prefer if we worked on the EventHandler directly here instead of going through the consumer. The reason is that I would like to keep event generation code in one place.

mxm: There is a test implementation of the event handler which we can simply pass in for tests.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
```diff
@@ -389,15 +407,61 @@ protected static int scale(
...
 
-        // When the shuffle type of vertex inputs contains keyBy or vertex is a source, we try to
-        // adjust the parallelism such that it divides the maxParallelism without a remainder
-        // => data is evenly spread across subtasks
-        for (int p = newParallelism; p <= maxParallelism / 2 && p <= upperBound; p++) {
-            if (maxParallelism % p == 0) {
+        int numKeyGroupsOrPartitions = maxParallelism;
+        int upperBoundForAlignment;
+        if (numPartitions <= 0) {
+            upperBoundForAlignment = Math.min(maxParallelism / 2, upperBound);
```
1996fanrui, line 413: I'm curious why `maxParallelism / 2`?

Assuming the upstream edge of vertex is keyBy(hash):
- The maxParallelism  is 100
- newParallelism is 80

We will use 80 as the result, right? If so, it will meet same issue with source partition.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
```diff
@@ -389,15 +407,61 @@ protected static int scale(
...
+        } else {
+            upperBoundForAlignment = Math.min(numPartitions, upperBound);
```
1996fanrui, line 415: Should the source vertex respect the `maxParallelism`?

If maxParallelism is 100, source has 1000 partition, and `upperBound` is 200. The new parallelism may be greater than 100, right? 

IIUC, the flink job won't run when `parallelism > maxParallelism` even if without key group.

1996fanrui (changes requested):
Thanks @huyuanfeng2018 for the contribution, and @mxm  for the review!

I left some comments, please take a look in your free time, thanks~

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
```diff
@@ -389,15 +407,61 @@ protected static int scale(
             return newParallelism;
```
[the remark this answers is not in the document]
huyuanfeng2018: Not necessarily, Source Vertex may also have keyedstate, and it should be matched with MaxParallelism.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
```diff
@@ -345,15 +356,22 @@ private boolean detectIneffectiveScaleUp(
      * <p>Also, in order to ensure the data is evenly spread across subtasks, we try to adjust the
      * parallelism for source and keyed vertex such that it divides the maxParallelism without a
      * remainder.
+     *
+     * <p>This method also attempts to adjust the parallelism to ensure it aligns well with the
+     * number of partitions if a vertex has a known partition count.
```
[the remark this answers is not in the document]
huyuanfeng2018: agree

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
huyuanfeng2018: yes, but I think not all tasks after hash are in large flink keyedState. In this scenario, miss alignment has almost no impact on the task. I think this is the trade-off between performance and resources that was taken into consideration by the previous logic. 

But this becomes less acceptable for consuming kafka or pulsar

@mxm @1996fanrui Maybe we can discuss this logic further

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
huyuanfeng2018: upperBound is already the smaller value obtained by comparing parallelismUpperLimit and maxParallelism.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
```diff
@@ -389,15 +407,61 @@ protected static int scale(
...
+            var message =
+                    String.format(
+                            SCALE_LIMITED_MESSAGE_FORMAT,
+                            vertex,
+                            newParallelism,
+                            p,
+                            String.format(
+                                    "numPartitions : %s，upperBoundForAlignment(maxParallelism or parallelismUpperLimit): %s, "
+                                            + "parallelismLowerLimit: %s.",
+                                    numPartitions, upperBound, parallelismLowerLimit));
```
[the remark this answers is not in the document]
huyuanfeng2018: You are right.

huyuanfeng2018 (commented):
Thank  @1996fanrui for your review, I have made some changes to the code, PTAL.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
1996fanrui: As I understand, the key group is totally similar with source partition(kafka or plusar). They determine how many partitions or groups a Flink parallelism can consume.

The performance is unbalanced even if without large state. For example, the maxParallelism(number of keyGroups) is 100, and the actual parallelism is 70.

-  It means that 30 instances process 2 keyGroups each, and the remaining 40 instances process 2 keyGroups each.
- Assuming that the data of each keyGroup is balanced, the 30 instances processing 2 keyGroups will become the bottleneck of the job.

For this scenario, there is no difference when the parallelism is set to 50 and 99.

IIUC, this situation is exactly the source partition problem you want to solve, and it works exactly the same for keyGroup as well.

Please correct me if anything is wrong.

huyuanfeng2018: Well, you are right, the specific entrance is here [KeyGroupStreamPartitioner](https://github.com/apache/flink/blob/release-1.19/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/partitioner/KeyGroupStreamPartitioner.java#L63)In this case, We can unify our logic.

mxm: @1996fanrui The code path you are pointing to hasn't been changed by this PR. It has merely been refactored. In the scenario of maxParallelism=100 and newParallelism=80, the resulting parallelism would always be 80.

Are you asking to expand the source logic introduced here to hash keyed state? 

>For this scenario, there is no difference when the parallelism is set to 50 and 99.

That depends on the amount of state in the key groups and other factors like hot keys. 50 could be the same as 99, it could also be much worse. For sources, the problem is much more amplified because they usually have pretty evenly balanced partitions and there is extra overhead to fetch the partition data.

In this scenario though, we should not be going down to 50, we probably should be going up to 100. I think the `maxParallelism / 2` stems from the idea that the maximum parallelism won't be reached because it is set to a number `parallelism <= maxParallelism / 2` which would mean that it doesn't make sense to continue beyond maxParallelism/2 because there aren't more possible divisors. However, when `parallelism > maxParallelism / 2`, this logic is flawed because maxParallelism itself could be a possible divisor. We should really be going up to `maxParallelism` for the initial `parallelism > maxParallelism / 2`. We could just skip this (premature) optimization entirely.

I agree that we should replace the current key alignment logic with the generalized source logic introduced here. Something like this:

```java
final int numKeyGroupsOrPartitions;
final int upperBoundForAlignment;
if (numSourcePartitions <= 0) {
    numKeyGroupsOrPartitions = maxParallelism;
    upperBoundForAlignment = Math.min(
        // Optimize the case where newParallelism <= maxParallelism / 2
        newParallelism > maxParallelism / 2 ? maxParallelism : maxParallelism / 2, 
        upperBound
    );
} else {
    numKeyGroupsOrPartitions = numSourcePartitions;
    upperBoundForAlignment = Math.min(numSourcePartitions, upperBound);
}

// When the shuffle type of vertex inputs contains keyBy or vertex is a source,
// we try to adjust the parallelism such that it divides
// the adjustableMaxParallelism without a remainder => data is evenly spread across subtasks
for (int p = newParallelism; p <= upperBoundForAlignment; p++) {
    if (numKeyGroupsOrPartitions % p == 0) {
        return p;
    }
}

// When adjust the parallelism after rounding up cannot be evenly divided by source
// numSourcePartitions, Try to find the smallest parallelism that can satisfy the
// current
// consumption rate.
int p = newParallelism;
for (; p > 0; p--) {
    if (numKeyGroupsOrPartitions / p > numKeyGroupsOrPartitions / newParallelism) {
        if (numKeyGroupsOrPartitions % p != 0) {
            p++;
        }
        break;
    }
}

p = Math.max(p, parallelismLowerLimit);
return p;
```

mxm (commented):
Great discussion @huyuanfeng2018 @1996fanrui! I think this is a good opportunity to unify the logic further and to address a gap with the current key alignment logic for hash partitioning.

huyuanfeng2018:
> Great discussion @huyuanfeng2018 @1996fanrui! I think this is a good opportunity to unify the logic further and to address a gap with the current key alignment logic for hash partitioning.

Thanks @mxm for review,LGTM for  these suggestions , I fixed code.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
```diff
@@ -383,23 +403,68 @@ protected static int scale(
...
+                    Math.min(
+                            // Optimize the case where newParallelism <= maxParallelism / 2
+                            newParallelism > maxParallelism / 2
+                                    ? maxParallelism
+                                    : maxParallelism / 2,
+                            upperBound);
+        } else {
+            numKeyGroupsOrPartitions = numSourcePartitions;
+            upperBoundForAlignment = Math.min(numSourcePartitions, upperBound);
+        }
```
1996fanrui, line 426: >  Optimize the case where newParallelism <= maxParallelism / 2

Why need this this optimization? Reducing the count of `for loop`?

I'm curious why source partition doesn't use this optimization? If both of source and keygroup could use this optimization, does the following code work?

```suggestion
        var numKeyGroupsOrPartitions = numSourcePartitions <= 0 ? maxParallelism : numSourcePartitions;
        var upperBoundForAlignment = 
                    Math.min(
                            // Optimize the case where newParallelism <= maxParallelism / 2
                            newParallelism > numKeyGroupsOrPartitions / 2
                                    ? numKeyGroupsOrPartitions
                                    : numKeyGroupsOrPartitions / 2,
                            upperBound);
```

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
```diff
@@ -383,23 +403,68 @@ protected static int scale(
...
+
+        // When the shuffle type of vertex inputs contains keyBy or vertex is a source,
+        // we try to adjust the parallelism such that it divides
+        // the adjustableMaxParallelism without a remainder => data is evenly spread across subtasks
```
1996fanrui, line 430:
```suggestion
        // the numKeyGroupsOrPartitions without a remainder => data is evenly spread across subtasks
```

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
```diff
@@ -383,23 +403,68 @@ protected static int scale(
...
+        // current consumption rate.
+        int p = newParallelism;
+        for (; p > 0; p--) {
+            if (numKeyGroupsOrPartitions / p > numKeyGroupsOrPartitions / newParallelism) {
+                if (numKeyGroupsOrPartitions % p != 0) {
+                    p++;
+                }
+                break;
+            }
+        }
```
1996fanrui, line 448: i found our discussion cannot cover all cases during I review this part in detail.

For example: sourcePartition is 199, and new parallelism is 99. IIUC, the final parallelism is 67(every subtask consume 3 source partitions, except for the last subtask), right?

But 100 as the final parallelism makes sense to me(every subtask consume 2 source partitions, except for the last subtask).

huyuanfeng2018: > i found our discussion cannot cover all cases during I review this part in detail.
> 
> For example: sourcePartition is 199, and new parallelism is 99. IIUC, the final parallelism is 67(every subtask consume 3 source partitions, except for the last subtask), right?
> 
> But 100 as the final parallelism makes sense to me(every subtask consume 2 source partitions, except for the last subtask).

Well, I actually considered this situation at the very beginning, that is, whether to use the same logic of fetching downwards when fetching upwards, so as to maximize the utilization of resources.

But our ultimate goal is to achieve balanced consumption of tasks? This is the healthiest situation. Often max parallelism or partition will have quite a few divisors. You will rarely set partition num to 1, 3, 7, 199, etc., so when I develop, I first think about balance. If It is due to the limitation of upperbound that we will take a reasonable value downwards, and there will be an event happend.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
```diff
@@ -383,23 +403,68 @@ protected static int scale(
...
+        } else {
+            numKeyGroupsOrPartitions = numSourcePartitions;
+            upperBoundForAlignment = Math.min(numSourcePartitions, upperBound);
+        }
+
+        // When the shuffle type of vertex inputs contains keyBy or vertex is a source,
+        // we try to adjust the parallelism such that it divides
+        // the adjustableMaxParallelism without a remainder => data is evenly spread across subtasks
+        for (int p = newParallelism; p <= upperBoundForAlignment; p++) {
+            if (numKeyGroupsOrPartitions % p == 0) {
```
1996fanrui, line 432: About this  comment https://github.com/apache/flink-kubernetes-operator/pull/879#discussion_r1764419949, I'm thinking whether the following change is more reasonable?

Note: `numKeyGroupsOrPartitions / p` means how many source partitions or key groups every subtask consume.

```suggestion
            if (numKeyGroupsOrPartitions % p == 0 || numKeyGroupsOrPartitions / p < numKeyGroupsOrPartitions / newParallelism) {
```

For example: maxParallelism is 200, and new parallelism is 60. (Some subtasks consume 4 keyGroups, the rest of subtask consume 3 keyGroups)

- The final parallelism is 100 based on the main branch code due to we only return p when `maxParallelism % p == 0`.
- But I think 67 is more reasonable here. (One subtask consumes 2 key groups. The remaining 66 subtasks, each subtask consumes 3 key groups.)


@mxm @gyfora , WDYT? 

Also, it's a bit beyond the scope of this PR. I could file a separate PR if you think it makes sense. Of course, it's acceptable to be done at this PR.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
1996fanrui: Follow https://github.com/apache/flink-kubernetes-operator/pull/879#discussion_r1764471043 . I found the current logic isn't perfect even if `sourcePartitionNumber` is 200.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
1996fanrui: Hey @mxm , sorry for the late reply. Because there are too many comments, I missed this one.

> @1996fanrui The code path you are pointing to hasn't been changed by this PR. It has merely been refactored. In the scenario of maxParallelism=100 and newParallelism=80, the resulting parallelism would always be 80.
> 
> Are you asking to expand the source logic introduced here to hash keyed state?

I'm asking hash keyed state, I don't know why we recommend the result is 80 instead of 100 for hash keyed state case. But I think you have answered my question in this comment.

> > For this scenario, there is no difference when the parallelism is set to 50 and 99.
> 
> That depends on the amount of state in the key groups and other factors like hot keys. 50 could be the same as 99, it could also be much worse. For sources, the problem is much more amplified because they usually have pretty evenly balanced partitions and there is extra overhead to fetch the partition data.

Good point! Hot keys may happen in the flink job. In general, Source partition without data skew.
 
> However, when `parallelism > maxParallelism / 2`, this logic is flawed because maxParallelism itself could be a possible divisor. We should really be going up to `maxParallelism` for the initial `parallelism > maxParallelism / 2`. We could just skip this (premature) optimization entirely.

This is exactly my question, I think we should use `maxParallelism` as the final parallelism. (For the above example, 100 instead of 80).

> ```java
>     upperBoundForAlignment = Math.min(
>         // Optimize the case where newParallelism <= maxParallelism / 2
>         newParallelism > maxParallelism / 2 ? maxParallelism : maxParallelism / 2, 
>         upperBound
>     );
> ```

Great, this part solved my concern. thank you~

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
mxm: I think that makes sense, but it makes the scaling more aggressive and less balanced. If we want to be more conservative, maybe 100 is ok in this scenario, where there is actually a divisor without a remainder. When there isn't, I think what you propose is way better than just using the initially provided parallelism.

In summary, I'm proposing to do a two-step process, similarly as for the partitions, where we first try to find a parallelism that divides the key groups without a remainder, and if that fails we do what you propose.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
mxm: Good observation, I think there are two opposing ideas in this discussion:

1. Reducing the parallelism as much as possible when we can't evenly balance partitions / key groups
2. Increasing the parallelism such that we can have the most evenly balanced partitions / key groups, even if we cannot do it perfectly (https://github.com/apache/flink-kubernetes-operator/pull/879#discussion_r1764471043)

Both are good ideas. For the sake of stability, I think we probably want to default to doing (2). I could imagine adding an option for (1) but I'm not sure it should be the default mode.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
huyuanfeng2018: Thanks for the review, I agree with this, We can introduce an additional parameter to enable a more aggressive strategy: https://github.com/apache/flink-kubernetes-operator/pull/879#discussion_r1764419949. But by default, a divisor of the number of partitions is still used to ensure balanced consumption.
WDYT？ @mxm @1996fanrui

huyuanfeng2018: > I think that makes sense, but it makes the scaling more aggressive and less balanced. If we want to be more conservative, maybe 100 is ok in this scenario, where there is actually a divisor without a remainder. When there isn't, I think what you propose is way better than just using the initially provided parallelism.
> 
> In summary, I'm proposing to do a two-step process, similarly as for the partitions, where we first try to find a parallelism that divides the key groups without a remainder, and if that fails we do what you propose.


Thanks for the review, I agree with this, We can introduce an additional parameter to enable a more aggressive strategy: https://github.com/apache/flink-kubernetes-operator/pull/879#discussion_r1764419949. But by default, a divisor of the number of partitions is still used to ensure balanced consumption.
WDYT？ @mxm @1996fanrui

1996fanrui: Sounds make sense to me, let us choose a conservative solution in this PR first. 

We could improve it in the future if it's needed.

1996fanrui: Sounds make sense to me, we could introduce an `additional parameter to enable a more aggressive strategy`.

I wanna check with @huyuanfeng2018 and @mxm :  The strategy will work for both source partition and key group, right? As I understand, we could unify the strategy for these 2 cases.

huyuanfeng2018: > I wanna check with @huyuanfeng2018 and @mxm : The strategy will work for both source partition and key group, right? As I understand, we could unify the strategy for these 2 cases.

I think we can unify strategies

mxm: Sounds good!

huyuanfeng2018:
@mxm @1996fanrui  Has been updated according to the latest discussion results. You can take a look again when you have time.

1996fanrui (commented):
Hi @huyuanfeng2018 , I saw [this comment](https://github.com/apache/flink-kubernetes-operator/pull/879/files#r1764471043) is not addressed in your latest change. 

For example, introducing an additional option and try to rounding up based on ` if (numKeyGroupsOrPartitions / p < numKeyGroupsOrPartitions / newParallelism)`. Do you mean introducing it in a separate PR?

Also, the CI is failed, it seems not related to your change. Would you mind rebasing the main branch?

huyuanfeng2018:
> For example, introducing an additional option and try to rounding up based on ` if (numKeyGroupsOrPartitions / p < numKeyGroupsOrPartitions / newParallelism)`. Do you mean introducing it in a separate PR?

Yes, i think a separate PR is required

> Also, the CI is failed, it seems not related to your change. Would you mind rebasing the main branch?

done.

flink-autoscaler/src/test/java/org/apache/flink/autoscaler/JobVertexScalerTest.java
```diff
@@ -775,11 +931,119 @@ public void testSendingIneffectiveScalingEvents(Collection<ShipStrategy> inputSh
         assertThat(event.getReason()).isEqualTo(INEFFECTIVE_SCALING);
     }
 
+    @Test
+    public void testNumPartitionsAdjustment() {
+        final int minParallelism = 1;
+        final int maxParallelism = Integer.MAX_VALUE;
```
1996fanrui, line 937: They are `parallelismLowerLimit` and `parallelismUpperLimit`.

It's better to distinguish `parallelismUpperLimit` and `maxParallelism` here.

1996fanrui (commented):
Thanks @huyuanfeng2018  for the quick update!

Overall LGTM, I left 2 minor comments, please take a look in your free time, thanks~

huyuanfeng2018:
> Thanks @huyuanfeng2018 for the quick update!
> 
> Overall LGTM, I left 2 minor comments, please take a look in your free time, thanks~

Thanks for your suggestion, fixed !

mxm (approved):
LGTM

1996fanrui (approved):
Thanks @huyuanfeng2018 for the hard work and @mxm for the detailed discussion!

huyuanfeng2018:
@mxm @1996fanrui 
Thank you both for your patient review

PR #881, opened 2024-09-20
FLINK-36332 Add option to select the Fabric8 httpclient implemention

SamBarker:

Enable the building of the operator using alternative Fabric8 http clients.

## Brief change log

Enable the building of the operator using alternative Fabric8 http clients.

## Verifying this change

The build should still work even when run with `-Dfabric8.httpclinent.impl=vertx` or `jdk` or `jetty` for that matter 😁 

  - Dependencies (does it add or upgrade a dependency): yes. 

## Documentation

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,7 +67,7 @@ jobs:
           cd flink-autoscaler-plugin-jdbc
           mvn -B verify -Dit.skip=false
           cd ..
-  e2e_smoketest:
+  e2e_smoke_test:
     name: HTTP Client smoke test
     runs-on: ubuntu-latest
     strategy:
@@ -131,6 +131,7 @@ jobs:
           source e2e-tests/utils.sh
           stop_minikube
   e2e_ci:
+    needs: e2e_smoke_test
     runs-on: ubuntu-latest
     strategy:
       matrix:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,6 +67,69 @@ jobs:
           cd flink-autoscaler-plugin-jdbc
           mvn -B verify -Dit.skip=false
           cd ..
+  e2e_smoketest:
+    name: HTTP Client smoke test
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        http-client: [ "okhttp", "jdk", "jetty", "vertx" ]
+        version: ["v1_20"]
+        mode: ["native"]
+        namespace: ["default"]
+        java-version: ["21"]
+        test:
+          - test_application_operations.sh
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up JDK ${{ matrix.java-version }}
+        uses: actions/setup-java@v4
+        with:
+          java-version: ${{ matrix.java-version }}
+          distribution: 'temurin'
+          cache: 'maven'
+      - name: Start minikube
+        run: |
+          source e2e-tests/utils.sh
+          start_minikube
+      - name: Install cert-manager
+        run: |
+          kubectl get pods -A
+          kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v1.8.2/cert-manager.yaml
+          kubectl -n cert-manager wait --all=true --for=condition=Available --timeout=300s deploy
+      - name: Build image
+        run: |
+          export SHELL=/bin/bash
+          export DOCKER_BUILDKIT=1
+          eval $(minikube -p minikube docker-env)
+          HTTP_CLIENT=${{ matrix.http-client }}
+          docker build --progress=plain --no-cache -f ./Dockerfile -t flink-kubernetes-operator:ci-latest --progress plain --build-arg HTTP_CLIENT="${HTTP_CLIENT:-okhttp}" .
+          docker images
+      - name: Start the operator
+        run: |
+          if [[ "${{ matrix.test }}" == "test_flink_operator_ha.sh" ]]; then
+            sed -i "s/# kubernetes.operator.leader-election.enabled: false/kubernetes.operator.leader-election.enabled: true/" helm/flink-kubernetes-operator/conf/flink-conf.yaml
+            sed -i "s/# kubernetes.operator.leader-election.lease-name: flink-operator-lease/kubernetes.operator.leader-election.lease-name: flink-operator-lease/" helm/flink-kubernetes-operator/conf/flink-conf.yaml
+            sed -i "s/replicas: 1/replicas: 2/" helm/flink-kubernetes-operator/values.yaml
+          fi
+          helm --debug install flink-kubernetes-operator -n ${{ matrix.namespace }} helm/flink-kubernetes-operator --set image.repository=flink-kubernetes-operator --set image.tag=ci-latest ${{ matrix.extraArgs }}
+          kubectl wait --for=condition=Available --timeout=120s -n ${{ matrix.namespace }} deploy/flink-kubernetes-operator
+          kubectl get pods -n ${{ matrix.namespace }}
+      - name: Run Flink e2e tests
+        run: |
+          sed -i "s/image: flink:.*/image: ${{ matrix.image }}/" e2e-tests/data/*.yaml
+          sed -i "s/flinkVersion: .*/flinkVersion: ${{ matrix.version }}/" e2e-tests/data/*.yaml
+          sed -i "s/mode: .*/mode: ${{ matrix.mode }}/" e2e-tests/data/*.yaml
+          git diff HEAD
+          echo "Running e2e-tests/$test"
+          bash e2e-tests/${{ matrix.test }} || exit 1
+          git reset --hard
+      - name: Stop the operator
+        run: |
+          helm uninstall -n ${{ matrix.namespace }} flink-kubernetes-operator
+      - name: Stop minikube
+        run: |
+          source e2e-tests/utils.sh
+          stop_minikube
   e2e_ci:
     runs-on: ubuntu-latest
     strategy:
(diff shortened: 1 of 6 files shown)


flink-kubernetes-operator/pom.xml
```diff
@@ -413,6 +425,22 @@ under the License.
...
+                    <name>fabric8.httpclinent.impl</name>
+                    <value>okhttp</value>
+                </property>
+            </activation>
+            <dependencies>
+                <dependency>
+                    <groupId>com.squareup.okhttp3</groupId>
+                    <artifactId>okhttp</artifactId>
+                    <version>${okhttp.version}</version>
+                </dependency>
```
SamBarker, line 441: By default `kubernetes-httpclient-okhttp` pulls in OkHttp 3.X which is reported to have problems with Kubernetes IPV6 stacks e.g. https://issues.apache.org/jira/browse/FLINK-31928

pom.xml
```diff
@@ -110,6 +110,7 @@ under the License.
         <surefire.module.config/>
 
         <derby.version>10.15.2.0</derby.version>
+        <jetbrains.annotations.version>24.1.0</jetbrains.annotations.version>
```
SamBarker, line 113: Required by `KubernetesAutoScalerStateStore` https://github.com/apache/flink-kubernetes-operator/blob/968a578515d6269bbd5637594a7a342d74c1cd5c/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/state/KubernetesAutoScalerStateStore.java#L205-L212 

though looking again that looks like a bug and it should use `@nonnull` instead.

gyfora:
Do you have any recommendation how we should test this and make sure it works? Maybe we can add a case to the e2e-s somehow?

csviri:
Eventually we should move to Vertx as default, which is the preferred one in fabric8.

SamBarker:
> Do you have any recommendation how we should test this and make sure it works? Maybe we can add a case to the e2e-s somehow?

Hmm thats a good question. It looks fairly easy to add it to maven build matrix (though at what cost?) and thus test on every build. That is however of debatable value as there is no change to the projects code and thus all it would really be doing is testing a third party dependency and is that very interesting at unit/integration test level? 

A more interesting angle as suggested is to include it in the e2e suite. Which again has a massive implication for the spread of tests but is probably more interesting as it validates it works against various different minikube setups. 

I've added both options in different commits so we can easily drop them if they aren't considered helpful.

pom.xml
SamBarker: @gyfora @csviri Do you have any preferences as to how to resolve this? Should the annotation be updated or add the extra dependency?

.github/workflows/ci.yml
```diff
@@ -78,6 +79,7 @@ jobs:
         namespace: ["default","flink"]
         mode: ["native", "standalone"]
         java-version: [ 11, 17, 21 ]
+        http-client: [ "okhttp", "vertx", "jetty",  "jdk" ]
```
gyfora, line 82: I suggest to only test the different HTTP clients in a single test case such as:
`v1_20, default, native, java21 and test_application_operations.sh `

we don't want to blow up our matrix by a factor of 4 :D

.github/workflows/ci.yml
```diff
@@ -47,7 +48,7 @@ jobs:
             ${{ runner.os }}-maven-
       - name: Build with Maven
         run: |
-          mvn -B clean install javadoc:javadoc -Pgenerate-docs
+          mvn -B clean install javadoc:javadoc -Pgenerate-docs -Dfabric8.httpclinent.impl=${{ matrix.httpclient }}
```
gyfora, line 51: I prefer the e2e approach, unit tests mock most http interactions anyways

pom.xml
SamBarker: I've removed `@NotNull` from `KubernetesAutoScalerStateStore` as `@NonNull` is the default.

.github/workflows/ci.yml
SamBarker: Reverted in [57b341b](https://github.com/apache/flink-kubernetes-operator/pull/881/commits/57b341bd841404c4d31d083adcc0d1385f22dffc)

.github/workflows/ci.yml
SamBarker: I think https://github.com/apache/flink-kubernetes-operator/pull/881/commits/7b74210e8d64acb881a268ac8f37271de155c060 does the job

.github/workflows/ci.yml
```diff
@@ -74,10 +74,11 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        version: ["v1_20","v1_19","v1_18","v1_17","v1_16"]
+        version: ["v1_20"]
         namespace: ["default","flink"]
-        mode: ["native", "standalone"]
-        java-version: [ 11, 17, 21 ]
```
gyfora, line 80: I think this will also get rid of our existing tests. Maybe what you need is an exclusion like in https://github.com/apache/flink-kubernetes-operator/blob/main/.github/workflows/ci.yml#L104

SamBarker: Yeah. I've been trying to run the workflow on my fork to validate it rather than waiting for approval to run on the main flink operator repo. But to try and keep things comprehensible I've tried to turn most of the suite off while I work out if it's building and running properly.

My hope is to drop the final two commits before this actually gets merged.

SamBarker: To be a bit more explicit [ca99f0f](https://github.com/apache/flink-kubernetes-operator/pull/881/commits/ca99f0fde68f5dd4a95a890396d9ee2d368e6cff) and [aa96389](https://github.com/apache/flink-kubernetes-operator/pull/881/commits/aa963890c6524d7c3ec1579dd943d536d16e0d72) will be dropped before this is mergeable, but I'm hoping the rest of the PR stays the same. 

Trying to find my 4 jobs in the other 2-4 hundred was a bit to painful

gyfora: makes sense!

SamBarker:
I think the e2e workflow is now correctly setup (and has passed on [my fork](https://github.com/SamBarker/flink-kubernetes-operator/actions/runs/11061322378/job/30735483464?pr=1)

SamBarker:
I should note I'm not convinced the workflow is doing what its expected to do I've documented my findings as [FLINK-36392](https://issues.apache.org/jira/browse/FLINK-36392) which I'm happy to work on next. 

I've also noticed a bunch of warnings about deprecated node version from the actions so have opened [FLINK-36393](https://issues.apache.org/jira/browse/FLINK-36393) which I'm currently testing a PR for on my fork.

SamBarker:
> I think the e2e workflow is now correctly setup (and has passed on [my fork](https://github.com/SamBarker/flink-kubernetes-operator/actions/runs/11061322378/job/30735483464?pr=1)

Nope its not. The includes with the http-client seem to be overriding tests and not adding an extra run as intended... 

I'm tempted to suggest splitting out `test_application_operations.sh` as a separate "smoke test" workflow that could run with each http-client with a fixed combination of the other matrix args. 

My other thought is to have a workflow per Java version as that controls the supported flink versions as well and possibly the test suite as well as some of the tests have version exclusions. 

A workflow per Java version feels easier for people to reason about too, but I don't have a lot of project context or history to work from so could be missing something...

gyfora:
@SamBarker I think it makes sense to split the test workflows. It has grown too much and it's getting a bit annoying as you can see :) 

Please feel free to propose any reasonable split, I think your understanding is probably as good as anyone's at this point.

.github/workflows/ci.yml
SamBarker: I've changed tack and created a smoke test job in https://github.com/apache/flink-kubernetes-operator/pull/881/commits/28535546dbf887d7bd85f7f13dce1211540c27dc.

SamBarker:
> Please feel free to propose any reasonable split, I think your understanding is probably as good as anyone's at this point.

In the scope of this PR i've just pulled out a smoke test job which validates the HTTP clients can do the basics. I don't see that as being a great split for future testing and or other combinations however I'd like to look at a wider restructure in a separate PR. 

I'd also like to call out https://github.com/apache/flink-kubernetes-operator/pull/881/commits/bb6c5772f3d0c8e4a279a0ee6cf1d04872937303 my primary motivation for adding it was to have a clearer picture of what was actually being built (in my fork) but I think its reasonable to merge. As I don't see any point in running the full job matrix if the basics aren't working (and thus saving actions minutes/ other resources).  The same argument could be extended to making the smoke test dependent on the `test_ci` job as well, but I haven't done that here as it might slow down feedback from the edge to edge tests for some un-related or flanky unit test failures.

.github/workflows/ci.yml
```diff
@@ -64,7 +64,71 @@ jobs:
...
+  e2e_smoke_test:
+    name: HTTP Client smoke test
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        http-client: [ "okhttp", "jdk", "jetty", "vertx" ]
+        version: ["v1_20"]
+        mode: ["native"]
+        namespace: ["default"]
+        java-version: ["21"]
```
gyfora, line 76: I think we should actually move the namespace , and java-version matrix also to the smoke_test and fix-it (remove-it) in the regular e2es.

The http-client, java-version, namespace matrix doesn't need all the different tests, it either works or it doesn't (famous last words).

gyfora: actually the namespace flink/default doesn't really need more than 1 test (regardless of other params...)

SamBarker: > I think we should actually move the namespace , and java-version matrix also to the smoke_test and fix-it (remove-it) in the regular e2es.
> 
> 

I agree it's a bit strange. What I was intending to do was make the smallest change here and revisit it in a follow up hence the matrix of single values.

> actually the namespace flink/default doesn't really need more than 1 test (regardless of other params...)

Thanks that’s the kind of insight I need for restructuring🧩.

gyfora: @SamBarker , how would you like to proceed with this? As long as we are adding the matrix changes I suggested in a followup , this is good from my side.

Should we squash it and merge it or would you want to add the restructuring here as a "second" commit

PR #883, opened 2024-09-26
[FLINK-36264] Fix can not setup other values for parameter `rest-service.exposed.type`

chenyuzhi459:

Fix can not setup other values for parameter `rest-service.exposed.type` when using ingress

## Brief change log

Only if using ingress and the `rest-service.exposed.type=null`, operator will setup as `ClusterIp` when 

## Verifying this change

## Documentation

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
@@ -249,17 +249,6 @@ public class FlinkConfigBuilder {
         return this;
     }
 
-    protected FlinkConfigBuilder applyIngressDomain() {
-        // Web UI
-        if (spec.getIngress() != null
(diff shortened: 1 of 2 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
```diff
@@ -243,7 +243,8 @@ protected FlinkConfigBuilder applyPodTemplate() throws IOException {
 
     protected FlinkConfigBuilder applyIngressDomain() {
         // Web UI
-        if (spec.getIngress() != null) {
+        if (spec.getIngress() != null
```
mateczagany, line 246: I think we can get rid of `applyIngressDomain()` altogether, this default configuration is already applied in method `applyFlinkConfiguration()`, where it's only set if the configuration is not yet set by the user.

chenyuzhi459: OK, I have removed `applyIngressDomain()`

PR #886, opened 2024-10-04
[FLINK-36425][refactor] Replace jobStatus.state string with Flink JobStatus enum

gyfora:

Use / enforce the Flink JobStatus enum directly in job status.state this how we already use it and it would simplify the code in a lot of places to avoid string juggling .

## Verifying this change

Tests have been updated to use/expect the enum and e2es should further cover existing behaviour.

The change have been manually verified with multiple envs and Flink versions.


gaborgsomogyi (commented):
The change looks trivial and good. What I can't justify is why `RestApiMetricsCollectorTest` failed. Is that flaky?

gaborgsomogyi (approved):

gyfora:
> The change looks trivial and good. What I can't justify is why `RestApiMetricsCollectorTest` failed. Is that flaky?

Yes that test is flaky and not related/dependent on these changes.

mxm (approved):

PR #887, opened 2024-10-04
[FLINK-36109][snapshot] Add labels to snapshot resources

mateczagany:

Adds labels to FlinkStateSnapshot resources managed by the operator to allow for filtering. Currently we have no way to easily filter snapshots resources (e.g. by job reference, snapshot type or state), since the API server only allows to filter by labels for custom resources. 

The only way currently is to download all resources from the API-server and filter the resources client-side. Filtering by labels allows the API-server to serve the snapshots in a more optimized way.

Once [KEP-4358](https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/4358-custom-resource-field-selectors/README.md) is available as GA, we can hopefully gradually switch to using spec and status fields as field selectors.

## Brief change log

- Before the end of each reconciliation phase of a snapshot, update its labels to reflect the latest spec/status.
- Change `UpdateControl` logic to also update the resource if the labels have changed.

## Verifying this change

- Unit testing
- Manual testing

  - The public API, i.e., is any changes to the `CustomResourceDescriptors`: yes, but only labels were added/changed

## Documentation

### DIFF
--- a/docs/content/docs/custom-resource/snapshots.md
+++ b/docs/content/docs/custom-resource/snapshots.md
@@ -127,7 +127,6 @@ This however requires the referenced Flink resource to be alive, as this operati
 
 This feature is not available for checkpoints.
 
-
 ## Triggering snapshots
 
 Upgrade savepoints are triggered automatically by the system during the upgrade process as we have seen in the previous sections.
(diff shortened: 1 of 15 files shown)


gyfora (approved):
Looks good! Please rebase on main

PR #891, opened 2024-10-09
[hotfix] Fix testfail due to data race with after migration to Fabric8 interceptor

aplyusnin:
Tests that perform requests are usualy fail now on CI due to race conditions. 

To fix it, I addeв busywait for metrics with timeout as it is done in RestApiMetricsCollectorTest#testJmMetricCollection().
Also I added an extra delay in the RestApiMetricsCollectorTest#testJmMetricCollection(), it helps to avoid CI failing on my workload runs. 

Alternative fix:
Add blocks inside tests until requests are performed, but this requreis a lot of modifications 

## Documentation

### DIFF
--- a/flink-autoscaler/src/test/java/org/apache/flink/autoscaler/RestApiMetricsCollectorTest.java
+++ b/flink-autoscaler/src/test/java/org/apache/flink/autoscaler/RestApiMetricsCollectorTest.java
@@ -160,6 +160,8 @@ public class RestApiMetricsCollectorTest {
                             (c, e) ->
                                     new StandaloneClientHAServices(
                                             miniCluster.getRestAddress().get().toString()));
+            // wait a little to ensure data is ready
+            Thread.sleep(500);
             do {
                 var collector = new RestApiMetricsCollector<>();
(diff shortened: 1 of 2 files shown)


SamBarker (commented):
I usually treat thread.sleep as a smell in tests.  Do you have a theory as to why its needed? 

I'll try and investigate but its nots a problem I observed when running the tests while doing the interceptor work.

aplyusnin:
```kubernetesClient.resource(deployment).delete()``` has to be a blocking call, but when I run the tests wtih some extra logging I saw, that 404 response code doesn't appear before assertions on CI (I don't actually know why it is, I didn't get that problem locally).  After some experiments, I came up with a conclusion, that such way fixes the problem.

I agree, that it is not good way for fixing the problem, but I don't know how to properly add some extra blocking for the main testing Thread to get rid of the problem.

SamBarker:
I've raised https://github.com/apache/flink-kubernetes-operator/pull/896 to actually wait for the metrics to become available.

gyfora:
I think with @SamBarker 's PR merged, we can close this one right?

PR #893, opened 2024-10-10
[FLINK-36460] Refactoring the CI matrix

SamBarker:

As observed in [FLINK-36332](https://github.com/apache/flink-kubernetes-operator/pull/881#issuecomment-2378765602) the CI workflow has become difficult to mange. This PR is attempting to split tests along sensible lines to make the matrix more manageable. 

To do so this PR splits the edge to edge workflow in two. A common workflow definition: [e2e.yaml](https://github.com/apache/flink-kubernetes-operator/pull/893/files#diff-a2de7550554c0198ac7f959c78a19ee0996921c98034411de47a7dd49b0c209b)  which is then reused from multiple different test matrices. 

I think this split allows better control over the variables without having to vary each dimension every time. By making the simplifying assumption that variations in the Java version are adequately tested by the maven builds and the smoke tests it effectively reduces the number of test jobs from ~225 to  about 100 (TBC)

## Brief change log

Simplify CI workflows.

## Verifying this change

## Documentation

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,15 +68,15 @@ jobs:
           mvn -B verify -Dit.skip=false
           cd ..
   e2e_smoke_test:
-    name: HTTP Client smoke test
+    name: Operator smoke test
     runs-on: ubuntu-latest
     strategy:
       matrix:
         http-client: [ "okhttp", "jdk", "jetty", "vertx" ]
         version: ["v1_20"]
         mode: ["native"]
-        namespace: ["default"]
-        java-version: ["21"]
+        namespace: ["default", "flink"]
+        java-version: ["11", "17", "21"]
         test:
           - test_application_operations.sh
     steps:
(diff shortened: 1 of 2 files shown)


SamBarker:
@gyfora alternative CI matrix as requested. Even if the matrix proposed in this PR doesn't feel "right" it sets things up so that the matrix is considerably easier to change.

SamBarker:
This shows it in action on my fork: Screenshot 2024-10-25 at 8 58 16 pm

gyfora (approved):
Looks great @SamBarker !

SamBarker:
I think I've finally got the excludes to align with main. *sigh* That was a lot more fuss than I wanted and doesn't end up as clean as I was hoping. However I think this represents a much cleaner way to setup test matrices. 

The tests are still very slow to run as each test has to build its own copies of the containers. I think thats resolvable without publishing temporary artefacts to a container registry. Some combination of [docker advice](https://docs.docker.com/build/ci/github-actions/share-image-jobs/) and GitHub [sharing data docs](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/storing-and-sharing-data-from-a-workflow) along with this [discussion](https://github.com/docker/build-push-action/issues/225). 

That feels like a separate PR.

@gyfora could you trigger another run of the PR please (its gone green on my fork)

mxm (approved):
Cool! This looks a lot cleaner.

PR #904, opened 2024-10-17
[FLINK-36527][autoscaler] Introduce a parameter to support autoscaler adopt a more radical strategy when source vertex or upstream shuffle is keyBy

huyuanfeng2018:
Introduce a parameter to support autoscaler adopt a more radical strategy when source vertex or upstream shuffle is keyBy

## Brief change log

  - *Add a new option: `scaling.radical.enabled`*
  - *Support use a more aggressive strategy（Resource utilization takes priority） to determine the degree of parallelism after Source or after keyby without first considering balanced consumption.*

## Verifying this change

in `org.apache.flink.autoscaler.JobVertexScalerTest.JobVertexScalerTest`
* `testParallelismComputationWithAdjustment` and `testNumPartitionsAdjustment` add logic to test

## Documentation

### DIFF
--- a/docs/layouts/shortcodes/generated/auto_scaler_configuration.html
+++ b/docs/layouts/shortcodes/generated/auto_scaler_configuration.html
@@ -189,6 +189,12 @@
             <td>Time interval to resend the identical event</td>
         </tr>
         <tr>
+            <td><h5>job.autoscaler.scaling.radical.enabled</h5></td>
+            <td style="word-wrap: break-word;">false</td>
+            <td>Boolean</td>
+            <td>If this option is enabled, The determination of parallelism will be more radical, which will maximize resource utilization, but may also cause data skew in some vertex.</td>
+        </tr>
+        <tr>
             <td><h5>job.autoscaler.stabilization.interval</h5></td>
             <td style="word-wrap: break-word;">5 min</td>
             <td>Duration</td>
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
@@ -416,21 +416,31 @@ public class JobVertexScaler<KEY, Context extends JobAutoScalerContext<KEY>> {
                         // Optimize the case where newParallelism <= maxParallelism / 2
                         newParallelism > numKeyGroupsOrPartitions / 2
                                 ? numKeyGroupsOrPartitions
-                                : numKeyGroupsOrPartitions / 2,
+                                : numKeyGroupsOrPartitions / 2 + numKeyGroupsOrPartitions % 2,
                         upperBound);
+
+        boolean scalingRadical =
+                context.getConfiguration().get(AutoScalerOptions.SCALING_RADICAL_ENABLED);
 
         // When the shuffle type of vertex inputs contains keyBy or vertex is a source,
         // we try to adjust the parallelism such that it divides
         // the numKeyGroupsOrPartitions without a remainder => data is evenly spread across subtasks
         for (int p = newParallelism; p <= upperBoundForAlignment; p++) {
-            if (numKeyGroupsOrPartitions % p == 0) {
+            if (numKeyGroupsOrPartitions % p == 0
+                    ||
+                    // When scaling radical is enabled, Try to find the smallest parallelism that
+                    // can satisfy the
+                    // current consumption rate.
+                    (scalingRadical
+                            && numKeyGroupsOrPartitions / p
+                                    < numKeyGroupsOrPartitions / newParallelism)) {
                 return p;
             }
         }
 
-        // When adjust the parallelism after rounding up cannot be evenly divided by
-        // numKeyGroupsOrPartitions, Try to find the smallest parallelism that can satisfy the
-        // current consumption rate.
+        // When adjust the parallelism after rounding up cannot be
+        // find the right degree of parallelism to meet requirements,
+        // Try to find the smallest parallelism that can satisfy the current consumption rate.
         int p = newParallelism;
         for (; p > 0; p--) {
             if (numKeyGroupsOrPartitions / p > numKeyGroupsOrPartitions / newParallelism) {
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
@@ -351,4 +351,13 @@ public class AutoScalerOptions {
                     .withFallbackKeys(oldOperatorConfigKey("quota.cpu"))
                     .withDescription(
                             "Quota of the CPU count. When scaling would go beyond this number the the scaling is not going to happen.");
+
+    public static final ConfigOption<Boolean> SCALING_RADICAL_ENABLED =
+            autoScalerConfig("scaling.radical.enabled")
+                    .booleanType()
+                    .defaultValue(false)
+                    .withFallbackKeys(oldOperatorConfigKey("scaling.radical.enabled"))
+                    .withDescription(
+                            "If this option is enabled, The determination of parallelism will be more radical, which"
+                                    + " will maximize resource utilization, but may also cause data skew in some vertex.");
 }
--- a/flink-autoscaler/src/test/java/org/apache/flink/autoscaler/JobVertexScalerTest.java
+++ b/flink-autoscaler/src/test/java/org/apache/flink/autoscaler/JobVertexScalerTest.java
@@ -379,6 +379,37 @@ public class JobVertexScalerTest {
                         maxParallelism,
                         eventCollector,
                         context));
+
+        assertEquals(
+                32,
+                JobVertexScaler.scale(
+                        vertex,
+                        10,
+                        inputShipStrategies,
+                        0,
+                        128,
+                        2.5,
+                        minParallelism,
+                        maxParallelism,
+                        eventCollector,
+                        context));
+
+        // scaling.radical.enabled  = true
+        Configuration conf = context.getConfiguration();
+        conf.set(AutoScalerOptions.SCALING_RADICAL_ENABLED, true);
+        assertEquals(
+                26,
+                JobVertexScaler.scale(
+                        vertex,
+                        10,
+                        inputShipStrategies,
+                        0,
+                        128,
+                        2.5,
+                        minParallelism,
+                        maxParallelism,
+                        eventCollector,
+                        context));
     }
 
     @ParameterizedTest
@@ -1000,6 +1031,37 @@ public class JobVertexScalerTest {
                         200,
                         128,
                         1.4,
+                        parallelismLowerLimit,
+                        parallelismUpperLimit,
+                        eventCollector,
+                        context));
+
+        assertEquals(
+                199,
+                JobVertexScaler.scale(
+                        vertex,
+                        24,
+                        List.of(),
+                        199,
+                        256,
+                        4,
+                        parallelismLowerLimit,
+                        parallelismUpperLimit,
+                        eventCollector,
+                        context));
+
+        // scaling.radical.enabled  = true
+        Configuration conf = context.getConfiguration();
+        conf.set(AutoScalerOptions.SCALING_RADICAL_ENABLED, true);
+        assertEquals(
+                100,
+                JobVertexScaler.scale(
+                        vertex,
+                        24,
+                        List.of(),
+                        199,
+                        256,
+                        4,
                         parallelismLowerLimit,
                         parallelismUpperLimit,
                         eventCollector,


docs/layouts/shortcodes/generated/auto_scaler_configuration.html
```diff
@@ -188,6 +188,12 @@
             <td>Duration</td>
             <td>Time interval to resend the identical event</td>
         </tr>
+        <tr>
+            <td><h5>job.autoscaler.scaling.radical.enabled</h5></td>
```
SamBarker, line 192: Coming at this cold, its not at all clear to me what radical means. While the description goes some way towards clarifying the intent it doesn't feel like a great term, additionally following through the JIRA links `radical` feels like a very off term for a default (assuming I'm following properly). I wonder if `job.autoscaler.scaling.maximizeUtilisation.enabled` would make things more explicit?

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
```diff
@@ -416,21 +416,31 @@ protected static <KEY, Context extends JobAutoScalerContext<KEY>> int scale(
...
         for (int p = newParallelism; p <= upperBoundForAlignment; p++) {
-            if (numKeyGroupsOrPartitions % p == 0) {
+            if (numKeyGroupsOrPartitions % p == 0
+                    ||
+                    // When scaling radical is enabled, Try to find the smallest parallelism that
+                    // can satisfy the
+                    // current consumption rate.
+                    (scalingRadical
+                            && numKeyGroupsOrPartitions / p
+                                    < numKeyGroupsOrPartitions / newParallelism)) {
```
SamBarker, line 436: I think extracting this as a method `canMaximiseUtilisation` the intent is to make the intent of the condition easier to understand when working through the code.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
```diff
@@ -416,21 +416,31 @@ protected static <KEY, Context extends JobAutoScalerContext<KEY>> int scale(
...
                 return p;
             }
         }
 
-        // When adjust the parallelism after rounding up cannot be evenly divided by
-        // numKeyGroupsOrPartitions, Try to find the smallest parallelism that can satisfy the
-        // current consumption rate.
+        // When adjust the parallelism after rounding up cannot be
+        // find the right degree of parallelism to meet requirements,
+        // Try to find the smallest parallelism that can satisfy the current consumption rate.
```
SamBarker, line 443: nits
```suggestion
        // When adjusting the parallelism after rounding up cannot
        // find the right degree of parallelism to meet requirements.
        // Try to find the smallest parallelism that can satisfy the current consumption rate.
```

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
```diff
@@ -351,4 +351,13 @@ private static ConfigOptions.OptionBuilder autoScalerConfig(String key) {
...
                             "Quota of the CPU count. When scaling would go beyond this number the the scaling is not going to happen.");
+
+    public static final ConfigOption<Boolean> SCALING_RADICAL_ENABLED =
+            autoScalerConfig("scaling.radical.enabled")
+                    .booleanType()
+                    .defaultValue(false)
+                    .withFallbackKeys(oldOperatorConfigKey("scaling.radical.enabled"))
+                    .withDescription(
+                            "If this option is enabled, The determination of parallelism will be more radical, which"
+                                    + " will maximize resource utilization, but may also cause data skew in some vertex.");
```
SamBarker, line 362: I think it might be helpful to give consumers/users some more context as to how/why it would potentially cause skew.

SamBarker (commented):
I'm lacking context for the change so my comments could be missing the mark but I think that means I represent a lot of people trying to consume the proposed config, if the naming confuses me its likely to confuse others.

huyuanfeng2018:
> I'm lacking context for the change so my comments could be missing the mark but I think that means I represent a lot of people trying to consume the proposed config, if the naming confuses me its likely to confuse others.

@SamBarker  Thank you very much for your review. `scaling.radical.enabled` is really not easy to understand. I think your concern is warranted. I will think about the naming of this option.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/NumKeyGroupsOrPartitionsParallelismAdjuster.java
```diff
@@ -0,0 +1,134 @@
...
+import static org.apache.flink.autoscaler.JobVertexScaler.SCALE_LIMITED_MESSAGE_FORMAT;
+import static org.apache.flink.autoscaler.JobVertexScaler.SCALING_LIMITED;
+import static org.apache.flink.autoscaler.config.AutoScalerOptions.SCALING_EVENT_INTERVAL;
+import static org.apache.flink.configuration.description.TextElement.text;
+
+/**
+ * Component responsible adjusts the parallelism of a vertex that knows the number of partitions or
+ * a vertex whose upstream shuffle is key by.
+ */
+public class NumKeyGroupsOrPartitionsParallelismAdjuster {
```
SamBarker, line 35: To me the number of key groups or partitions is an implementation details the role this class plays is simply:

```suggestion
public class ParallelismAdjuster {
```

One could make an argument for an interface with an implementation named `PartitionBoundedParallelismAdjuster` or `KeyGroupBoundedParallelismAdjuster` but I don't have any context to suggest that is required at this point. Having said that the fact we have an enum and thus two implementations already maybe it jus justified.

flink-autoscaler/src/test/java/org/apache/flink/autoscaler/JobVertexScalerTest.java
```diff
@@ -379,6 +379,38 @@ public void testParallelismComputationWithAdjustment(
                         maxParallelism,
                         eventCollector,
                         context));
+
+        assertEquals(
+                32,
```
SamBarker, line 384: It would be nice to tie this answer to something so it was clear why it was picked. 
e.g. 

```suggestion
                EXPECTED_KEY_GROUPS,
```

flink-autoscaler/src/test/java/org/apache/flink/autoscaler/JobVertexScalerTest.java
```diff
@@ -1004,6 +1036,38 @@ public void testNumPartitionsAdjustment() {
                         parallelismUpperLimit,
                         eventCollector,
                         context));
+
+        assertEquals(
+                199,
```
SamBarker, line 1041:
```suggestion
                PARTITION_COUNT,
```

flink-autoscaler/src/test/java/org/apache/flink/autoscaler/JobVertexScalerTest.java
```diff
@@ -1004,6 +1036,38 @@ public void testNumPartitionsAdjustment() {
...
+                        parallelismUpperLimit,
+                        eventCollector,
+                        context));
+
+        Configuration conf = context.getConfiguration();
+        conf.set(
+                AutoScalerOptions.SCALING_KEY_GROUP_PARTITIONS_ADJUST_MODE,
+                NumKeyGroupsOrPartitionsParallelismAdjuster.Mode.MAXIMIZE_UTILISATION);
+        assertEquals(
+                100,
```
SamBarker, line 1059:
```suggestion
                MAXIMUM_UTILISATION_PARALLELISM,
```

Its not the number that matters but what it means.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/NumKeyGroupsOrPartitionsParallelismAdjuster.java
huyuanfeng2018: Many thanks to @SamBarker for the review.

Why `NumKeyGroupsOrPartitionsParallelismAdjuster` is used here, because there has been a judgment before 
https://github.com/apache/flink-kubernetes-operator/blob/d2a68eb1eaa2624fce47107b2c371b5cda428830/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java#L406
So the responsibility of `NumKeyGroupsOrPartitionsParallelismAdjuster` is single(**for key groups or partitions**).

But what you said makes sense. Maybe we should change it to this to cope with new extensions that may exist in the future:
```suggestion
public class FinalParallelismAdjuster {

   public static <KEY, Context extends JobAutoScalerContext<KEY>> int adjust(
            JobVertexID vertex,
            Context context,
            AutoScalerEventHandler<KEY, Context> eventHandler,
            int maxParallelism,
            int numSourcePartitions,
            int newParallelism,
            int upperBound,
            int parallelismLowerLimit) {
        var adjustByMaxParallelismOrPartitions =
            numSourcePartitions > 0 || inputShipStrategies.contains(HASH);
        if (!adjustByMaxParallelismOrPartitions) {
            return newParallelism;
        }
        ...
            }

}
```

The corresponding enum is changed to

```
public class KeyGroupOrPartitionsAdjustMode {
...
}
```

WDYT?

SamBarker: I'm not sure what `Final` adds to the class name but otherwise it looks good to me.

huyuanfeng2018: thanks for  @SamBarker, fixed, PTAL

SamBarker (approved):
LGTM

Thanks for listening @huyuanfeng2018

huyuanfeng2018:
Hey @1996fanrui @mxm , Would you help reviewing this PR ?

1996fanrui (commented):
Thanks @huyuanfeng2018 for the ping, I will review it this week.

mxm (commented):
Thanks for the PR! This is an interesting idea. See the comments inline. I would prefer to do the refactoring separately, as it distracts from the actual changes at hand (which are relatively small). There is also some value in preserving the Git history of the changes.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ParallelismAdjuster.java
```diff
@@ -0,0 +1,148 @@
...
+    /** The mode of the key group or parallelism adjustment. */
+    public enum KeyGroupOrPartitionsAdjustMode implements DescribedEnum {
+        DEFAULT(
+                "This mode ensures that the parallelism adjustment attempts to evenly distribute data across subtasks"
+                        + ". It is particularly effective for source vertices that are aware of partition counts or vertices after "
+                        + "'keyBy' operation. The goal is to have the number of key groups or partitions be divisible by the set parallelism, ensuring even data distribution and reducing data skew."),
+
+        MAXIMIZE_UTILISATION(
+                "This model is to maximize resource utilization. In this mode, an attempt is made to set"
+                        + " the minimum degree of parallelism that meets the current consumption rate requirements. Unlike the default mode, it is not enforced that the number of key groups or partitions is divisible by the degree of parallelism."),
```
1996fanrui, line 134: Could the `degree of parallelism` be updated to `parallelism`?

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ParallelismAdjuster.java
```diff
@@ -0,0 +1,148 @@
...
+        p = Math.max(p, parallelismLowerLimit);
+        return p;
+    }
+
+    /** The mode of the key group or parallelism adjustment. */
+    public enum KeyGroupOrPartitionsAdjustMode implements DescribedEnum {
+        DEFAULT(
+                "This mode ensures that the parallelism adjustment attempts to evenly distribute data across subtasks"
+                        + ". It is particularly effective for source vertices that are aware of partition counts or vertices after "
+                        + "'keyBy' operation. The goal is to have the number of key groups or partitions be divisible by the set parallelism, ensuring even data distribution and reducing data skew."),
```
1996fanrui, line 130: Can `DEFAULT` be renamed to `ABSOLUTELY_EVENLY_DISTRIBUTION`?

It's more clear for users, and we may update the default value for `scaling.key-group.partitions.adjust.mode` option in the future. For example, `MAXIMIZE_UTILISATION` may be more reasonable in the future. And `KeyGroupOrPartitionsAdjustMode.DEFAULT` is a little wired if it's not the default enum.

1996fanrui (commented):
>  I would prefer to do the refactoring separately

I agree with @mxm , refactoring should at least be a separate commit.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ParallelismAdjuster.java
huyuanfeng2018: ok，great advice.

huyuanfeng2018:
> Thanks for the PR! This is an interesting idea. See the comments inline. I would prefer to do the refactoring separately, as it distracts from the actual changes at hand (which are relatively small). There is also some value in preserving the Git history of the changes.



> > I would prefer to do the refactoring separately
> 
> I agree with @mxm , refactoring should at least be a separate commit.

ok, I will revert the changes

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ParallelismAdjuster.java
```diff
@@ -0,0 +1,148 @@
...
+        // we try to adjust the parallelism such that it divides
+        // the numKeyGroupsOrPartitions without a remainder => data is evenly spread across subtasks
+        for (int p = newParallelism; p <= upperBoundForAlignment; p++) {
+            if (numKeyGroupsOrPartitions % p == 0
+                    ||
+                    // When Mode is MAXIMIZE_UTILISATION , Try to find the smallest parallelism
+                    // that can satisfy the current consumption rate.
+                    (mode == MAXIMIZE_UTILISATION
+                            && numKeyGroupsOrPartitions / p
+                                    < numKeyGroupsOrPartitions / newParallelism)) {
```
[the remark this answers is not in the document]
huyuanfeng2018: Sorry for replying so late.

> I understand the idea behind this but I wonder whether it has the desired effect. The autoscaling algorithm isn't aware of any state/partition imbalance, it assumes linear scale through state/partition balance can be achieved. A slight adjustment to the parallelism won't drastically improve the situation.

When the partition or keygroups is evenly distributed and there are relatively many divisors such as (60, 120, 720), we can indeed easily achieve linear scaling through state/partition balancing, which is the most ideal situation.

If the number of partitions is unreasonable, such as 35, hope adjust the parallelism to linearly increase the consumption rate, Although it is not possible to consume evenly, it is meaningful.

Imagine a situation: when running with 7 parallelism degrees, the busy value of the operator is large. No matter what the `scaleFactor` , it can be expected that the autoscaler will expand the parallelism degree to 35, which may cause the operator to be very idle, but we know that the linear assumption may not be completely linear when changing 7->35, and it may not be possible to obtain a 5 (35/7) times increase in processing speed (this is related to many factors), it may be in the next cycle Trigger scale down again, and then the cycle repeats

In addition, due to the limitations of the **`scale-down.max-factor`** parameter, we may never be able to reduce the degree of parallelism

If  use the MAXIMIZE_UTILISATION mode, can significantly improve this phenomenon. 

> It looks like this change could help in situations where the number of partition / key groups do not have many divisors, but its also kind of hard to reason about.

Any suggestions?

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ParallelismAdjuster.java
```diff
@@ -106,4 +107,43 @@ public static <KEY, Context extends JobAutoScalerContext<KEY>> int adjust(
...
+            if (numKeyGroupsOrPartitions / p > numKeyGroupsOrPartitions / newParallelism) {
+                if (numKeyGroupsOrPartitions % p != 0) {
+                    p++;
+                }
+                break;
+            }
+        }
+        p = Math.max(p, parallelismLowerLimit);
+        return p;
+    }
```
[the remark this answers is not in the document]
huyuanfeng2018: nice catch. I missed this

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ParallelismAdjuster.java
```diff
@@ -0,0 +1,109 @@
...
+            Collection<ShipStrategy> inputShipStrategies) {
+
+        var adjustByMaxParallelismOrPartitions =
+                numSourcePartitions > 0 || inputShipStrategies.contains(HASH);
+        if (!adjustByMaxParallelismOrPartitions) {
+            return newParallelism;
+        }
+
+        var numKeyGroupsOrPartitions =
+                numSourcePartitions <= 0 ? maxParallelism : numSourcePartitions;
```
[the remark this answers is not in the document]
huyuanfeng2018: IMHO, in our subsequent work, there may be other ways to adjust the logic of parallelism. It’s not just based on the numPartitions or keygroups, so I want to unify it with ParallelismAdjuster. WDYT?

mxm (commented):
Looks good to me overall, but I would like to ask you to drop the refactoring commit. The reason is that this code won't be reused via an extra class. It isn't more easily testable. It also doesn't make the code easier to read, and it removes the valuable Git history from the JobVertexScaler file. I hope that makes sense.

huyuanfeng2018:
> Looks good to me overall, but I would like to ask you to drop the refactoring commit. The reason is that this code won't be reused via an extra class. It isn't more easily testable. It also doesn't make the code easier to read, and it removes the valuable Git history from the JobVertexScaler file. I hope that makes sense.

Thanks for the review. Regarding this part of the refactoring, I originally thought that there might be some other logic to modify the parallelism in the future, just like fine-tuning based on numpartitionOrKeygroup here, so I refactored it, But I think your concerns are valid (this would change the git history). 

So, i have no problem with canceling this part of the refactoring.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ParallelismAdjuster.java
```diff
@@ -100,4 +109,28 @@ private static int calculateMinimumParallelism(
         p = Math.max(p, parallelismLowerLimit);
         return p;
     }
+
+    /** The mode of the key group or parallelism adjustment. */
+    public enum KeyGroupOrPartitionsAdjustMode implements DescribedEnum {
+        ABSOLUTELY_EVENLY_DISTRIBUTION(
```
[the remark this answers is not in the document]
huyuanfeng2018: > The name is confusing because this doesn't guarantee absolutely even distribution.
> 
> How about `EVENLY_SPREAD`?

Looks like `EVENLY_SPREAD` is more reasonable, I'm fine with that
@1996fanrui What do you think about this ?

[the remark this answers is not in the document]
1996fanrui: Sounds good to me.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
```diff
@@ -411,26 +416,30 @@ protected static <KEY, Context extends JobAutoScalerContext<KEY>> int scale(
...
-                        // Optimize the case where newParallelism <= maxParallelism / 2
-                        newParallelism > numKeyGroupsOrPartitions / 2
-                                ? numKeyGroupsOrPartitions
-                                : numKeyGroupsOrPartitions / 2,
-                        upperBound);
+
+        KeyGroupOrPartitionsAdjustMode mode =
+                context.getConfiguration().get(SCALING_KEY_GROUP_PARTITIONS_ADJUST_MODE);
+
+        var upperBoundForAlignment = Math.min(numKeyGroupsOrPartitions, upperBound);
```
1996fanrui, line 423: Why the `upperBoundForAlignment` logic is updated? Would you mind sharing one case?

mxm: The optimization to use half of the keyspace when the parallelism is less/equal to the keyspace doesn't work for the new parallelism adjustment mode. It was anyways just a shortcut, to avoid checking all divisors until the maximum. No harm removing it.

mxm, line 423: Can we keep `var upperBoundForAlignment = ...` on line 414, as previously?

huyuanfeng2018: > Why the `upperBoundForAlignment` logic is updated? Would you mind sharing one case?

When I ran the test case, i found  this logic is wrong when keygroups or partitionnum are not even numbers.

huyuanfeng2018: > Can we keep `var upperBoundForAlignment = ...` on line 414, as previously?

fixed

1996fanrui: It's seems it's wrong logic when numKeyGroupsOrPartitions is odd number. 

Thanks for the clarification! This change makes sense to me.

1996fanrui (approved):
Thanks @huyuanfeng2018 for the update!

LGTM

huyuanfeng2018:
Thanks for the review @mxm @1996fanrui @SamBarker

PR #906, opened 2024-10-21
[FLINK-36332] Missed webhook okhttp reference.

SamBarker:

There is still a stray reference to OkHTTP via the webhook.

## Brief change log

*(for example:)*
  - Fully exclude OkHttp
  
## Verifying this change

## Documentation

I'd also like to back port this to release-1.10 if possible.

### DIFF
--- a/flink-kubernetes-webhook/pom.xml
+++ b/flink-kubernetes-webhook/pom.xml
@@ -24,7 +24,7 @@ under the License.
         <groupId>org.apache.flink</groupId>
         <artifactId>flink-kubernetes-operator-parent</artifactId>
         <version>1.11-SNAPSHOT</version>
-        <relativePath>..</relativePath>
+        <relativePath>../pom.xml</relativePath>
     </parent>
 
     <artifactId>flink-kubernetes-webhook</artifactId>
@@ -78,22 +78,6 @@ under the License.
             <scope>test</scope>
         </dependency>
 
-        <!-- okhttp -->
-        <!--
-            Regarding the okhttp explicit version
-            see https://github.com/fabric8io/kubernetes-client/issues/4290
-            and https://issues.apache.org/jira/browse/FLINK-28637
-            -->
-        <dependency>
-            <groupId>com.squareup.okhttp3</groupId>
-            <artifactId>okhttp</artifactId>
-            <version>${okhttp.version}</version>
-        </dependency>
-        <dependency>
-            <groupId>com.squareup.okhttp3</groupId>
-            <artifactId>logging-interceptor</artifactId>
-            <version>${okhttp.version}</version>
-        </dependency>
         <dependency>
             <groupId>com.squareup.okhttp3</groupId>
             <artifactId>mockwebserver</artifactId>
--- a/flink-kubernetes-webhook/pom.xml
+++ b/flink-kubernetes-webhook/pom.xml
@@ -63,12 +63,6 @@ under the License.
             <artifactId>kubernetes-server-mock</artifactId>
             <version>${fabric8.version}</version>
             <scope>test</scope>
-            <exclusions>
-                <exclusion>
-                    <groupId>com.squareup.okhttp3</groupId>
-                    <artifactId>mockwebserver</artifactId>
-                </exclusion>
-            </exclusions>
         </dependency>
 
         <dependency>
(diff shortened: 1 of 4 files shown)


flink-kubernetes-webhook/pom.xml
```diff
@@ -24,7 +24,7 @@ under the License.
         <groupId>org.apache.flink</groupId>
         <artifactId>flink-kubernetes-operator-parent</artifactId>
         <version>1.11-SNAPSHOT</version>
-        <relativePath>..</relativePath>
```
tinaselenge, line 27: is this necessary? should it be consistent with the other modules?

tinaselenge (approved):
LGTM :)

k-wall:
I think there is a misunderstand about the way that Maven profiles are activated.  Maven profile can be activated by [system properties](https://maven.apache.org/guides/introduction/introduction-to-profiles.html#property), not maven properties.  The two concepts are separate.  So if the intent is that build activates `okhttp` by default, that's not happening.

You can show this against main

```
mvn   help:all-profiles  | grep http
Profile Id: depend-on-okhttp4 (Active: false , Source: pom)
```

```
mvn   help:all-profiles -Dfabric8.httpclient.impl=okhttp | grep http
  Profile Id: depend-on-okhttp4 (Active: true , Source: pom)
```

flink-kubernetes-webhook/pom.xml
SamBarker: IntelliJ was complaining that `..` was invalid and as I was moving the profile up to the parent it felt important to fix the warning. I haven't checked if the intellij error is correct or if it should be applied else where.

SamBarker:
Thanks @k-wall 

> I think `depend-on-okhttp4` should be activated by default.  If a user wants a alternative client, they should set the `fabric8.httpclient.impl` property and deactivate the profile.

Grr that is annoying distinction between properties that I missed. I'd been testing by toggling the -D flag when building with maven. 

I'll have a look at the `mockwebserver` inconsistency as that was something I'd seen initially and thought I'd fixed.

SamBarker:
I've added eede213 to address the issues highlighted by @k-wall 

```shell
mvn help:effective-pom -pl :flink-kubernetes-operator -Doutput=/tmp/default-pom.xml
mvn help:effective-pom -pl :flink-kubernetes-operator -P\!depend-on-okhttp4 -Dfabric8.httpclient.impl=jdk -Doutput=/tmp/jdk-pom.xml
diff /tmp/*-pom.xml
4c4
< <!-- Generated by Maven Help Plugin on 2024-10-23T13:21:57+13:00            -->
---
> <!-- Generated by Maven Help Plugin on 2024-10-23T13:23:54+13:00            -->
551c551
<       <artifactId>kubernetes-httpclient-okhttp</artifactId>
---
>       <artifactId>kubernetes-httpclient-jdk</artifactId>
693,698d692
<     </dependency>
<     <dependency>
<       <groupId>com.squareup.okhttp3</groupId>
<       <artifactId>okhttp</artifactId>
<       <version>4.12.0</version>
<       <scope>compile</scope>

```

k-wall:
Thanks @SamBarker.  However when running ` mvn dependency:tree  -Dfabric8.httpclient.impl=jdk -P'!depend-on-okhttp4'`.  I still see okhttp on module test classpaths at versions 3.12.12 and 4.12.0.

k-wall:
Just testing out an idea locally. I hope to have a PR soon

flink-kubernetes-webhook/pom.xml
SamBarker: So checking https://maven.apache.org/ref/3.9.9/maven-model/maven.html#parent intellij is correct `..` is not a valid reference, but based on stackoverflow and some other sources it might have been at one point or at least commonly thought to be.

gyfora:
What's the status of this? Could you guys @SamBarker @k-wall sort out the issues that you have found?
Does this mean that the changes we released in 1.10.0 don't really do anything? (we can then consider adding this to a patch release if there ever will be one)

SamBarker:
@gyfora Should I open a pr to apply this change to the `release-1.10` branch? Or is there another process for potential backports?

gyfora:
@SamBarker if you would like to backport it for a potential bug fix release, then please open a separate PR for `release-1.10` please :)

PR #910, opened 2024-11-06
[FLINK-36646] Test different versions of the JDK in the Flink image

SamBarker:

* adding tests to demonstrate FLINK-36646

## Brief change log

* adding tests to demonstrate FLINK-36646

## Verifying this change

This change is already covered by existing tests, such as `test_application_operations.sh` should pass with each of the supported JDK versions.

## Documentation

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,19 +89,18 @@ jobs:
     strategy:
       matrix:
         http-client: [ "okhttp" ]
-        java-version: [ "11", "17", "21" ]
+        java-version: [ "11", "17"]
         flink-version:
           - "v1_20"
-          - "v1_18"
           - "v1_19"
-          - "v1_17"
-          - "v1_16"
+          - "v1_18"
     uses: ./.github/workflows/e2e.yaml
     with:
       java-version: ${{ matrix.java-version }}
       flink-version: ${{ matrix.flink-version }}
       http-client: ${{ matrix.http-client }}
       test: test_application_operations.sh
+      append-java-version: true
   e2e_namespace_tests:
     name: Alternative namespace tests
     needs: e2e_smoke_test
@@ -109,8 +108,8 @@ jobs:
       matrix:
         flink-version:
           - "v1_20"
-          - "v1_18"
           - "v1_19"
+          - "v1_18"
           - "v1_17"
           - "v1_16"
         mode:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,6 +83,24 @@ jobs:
       flink-version: "v1_20"
       http-client: ${{ matrix.http-client }}
       test: test_application_operations.sh
+  java_rutimes:
+    name: Java runtimes smoke test
+    strategy:
+      matrix:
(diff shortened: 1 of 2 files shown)


SamBarker:
cc: @tomncooper & @robobario

.github/workflows/ci.yml
```diff
@@ -83,15 +83,33 @@ jobs:
       flink-version: "v1_20"
       http-client: ${{ matrix.http-client }}
       test: test_application_operations.sh
+  java_rutimes:
```
robobario, line 86:
```suggestion
  java_runtimes:
```

.github/workflows/ci.yml
```diff
@@ -83,15 +83,33 @@ jobs:
...
+      matrix:
+        http-client: [ "okhttp" ]
+        java-version: [ "11", "17"]
+        flink-version:
+          - "v1_20"
+          - "v1_19"
+          - "v1_18"
+    uses: ./.github/workflows/e2e.yaml
+    with:
+      java-version: ${{ matrix.java-version }}
```
robobario, line 99: I'm wondering if it's a bit funky that we are using `java-version` to control both the JDK/JRE used to build/run the operator and the runtime of the flink image. Aren't they different dimensions of the matrix?

This would show that a JDK17 operator works with a JDK17 flink image, but not cover the current default operator runtime with a JDK17 flink image.

.github/workflows/e2e.yaml
```diff
@@ -27,6 +27,9 @@ on:
       create-namespace:
         type: boolean
         default: false
+      append-java-version:
```
robobario, line 30: maybe this should be an optional `flink-java-version` and the matrix could contain the versions we want to test

.github/workflows/ci.yml
SamBarker: Yeah, I had debated making a `flink-tag` parameter but didn't as we would then have to manually encode the matrix. 

>This would show that a JDK17 operator works with a JDK17 flink image, but not cover the current default operator runtime with a JDK17 flink image.

I'm not convinced that the operator JDK really matters so don't think its particularly important to have a JDK11 operator deploying a JDK 17 Flink. I can go that direction if others see it as valuable.

robobario: yeah you're right we aren't likely to see some incompatibility arising from different JRE combinations. Maybe it would be better to hold the operator JDK constant and vary only the flink runtime. Just to make it clearer which thing is varied.

tomncooper: Yeah I agree with @SamBarker. The issue here is actually not the interplay of JDK versions but rather the config supplied by the operator's logic not being compatible (or more accuratly _enabling_) Java 17 support. So that's what we need to test.

tomncooper (approved):
LGTM, just need to update the changelog in the PR description.

SamBarker:
E2E test failed as expected: https://github.com/apache/flink-kubernetes-operator/actions/runs/11696768142/job/32793460638?pr=910#step:9:4256

gyfora:
Closing this as it was merged in another PR

PR #869, opened 2024-08-22
[FLINK-36129][autoscaler] Autoscaler is compatible with Flink 1.20

1996fanrui:
## What is the purpose of the change

AggregatedMetric added the skew field in Flink 1.20, it caused the AggregatedMetric of old version cannot parse the AggregatedMetric of 1.20.

Following is the root exception:

 ```
Caused by: org.apache.flink.shaded.jackson2.com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: Unrecognized field "skew" (class org.apache.flink.runtime.rest.messages.job.metrics.AggregatedMetric), not marked as ignorable (5 known properties: "min", "id", "max", "avg", "sum"])
 at [Source: UNKNOWN; byte offset: #UNKNOWN] (through reference chain: java.util.ArrayList[0]->org.apache.flink.runtime.rest.messages.job.metrics.AggregatedMetric["skew"])
```

## Brief change log

Copy AggregatedMetric to this repo, and removing it after upgrading maven dependency to 1.20. 

## Verifying this change

Added the CI to run with 1.20.

## Does this pull request potentially affect one of the following parts:

  - Dependencies (does it add or upgrade a dependency): no
  - The public API, i.e., is any changes to the `CustomResourceDescriptors`: no
  - Core observer or reconciler logic that is regularly executed: no

## Documentation

  - Does this pull request introduce a new feature? no

PR #912, opened 2024-11-09
[hot-fix][docs] Reflect v1.20 support

fqaiser94:

Update docs to reflect that v1.20 is supported. 

Currently [this page](https://nightlies.apache.org/flink/flink-kubernetes-operator-docs-main/docs/concepts/overview/) shows that the operator only supports up to Flink v1.19. 

## Verifying this change

CMIIW but my understanding is that v1.20 support was added as part of https://issues.apache.org/jira/browse/FLINK-36129 and in fact v1.20 is supported since the flink-kubernetes-operator 1.10 release. I'm assuming this was just overlooked in https://github.com/apache/flink-kubernetes-operator/pull/869

### DIFF
--- a/docs/content/docs/concepts/overview.md
+++ b/docs/content/docs/concepts/overview.md
@@ -36,7 +36,7 @@ Flink Kubernetes Operator aims to capture the responsibilities of a human operat
   - Stateful and stateless application upgrades
   - Triggering and managing savepoints
   - Handling errors, rolling-back broken upgrades
-- Multiple Flink version support: v1.16, v1.17, v1.18, v1.19
+- Multiple Flink version support: v1.16, v1.17, v1.18, v1.19, v1.20
 - [Deployment Modes]({{< ref "docs/custom-resource/overview#application-deployments" >}}):
   - Application cluster
   - Session cluster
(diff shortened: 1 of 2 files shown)


docs/content/docs/concepts/overview.md
```diff
@@ -36,7 +36,7 @@ Flink Kubernetes Operator aims to capture the responsibilities of a human operat
   - Stateful and stateless application upgrades
   - Triggering and managing savepoints
   - Handling errors, rolling-back broken upgrades
-- Multiple Flink version support: v1.16, v1.17, v1.18, v1.19
+- Multiple Flink version support: v1.16, v1.17, v1.18, v1.19, v1.20
```
fqaiser94, line 39: To avoid confusion, I would also suggest backporting this commit to the 1.10 release branch (if that's how the 1.10 docs can be updated).

fqaiser94:
cc @gyfora @1996fanrui (since you were involved in https://github.com/apache/flink-kubernetes-operator/pull/869)

docs/content/docs/concepts/overview.md
1996fanrui: Sounds make sense, I have cherry picked it to 1.10.

PR #913, opened 2024-11-11
FLINK-36545: [OLM] fix uid/gid mixup when generating the OLM bundle.

k-wall:
The docker container was being run with the uid/gid arguments switched about.

As pointed out by https://issues.apache.org/jira/browse/FLINK-36545, when generating the OLM bundle, the container is run with uid/gid values switched around.

## Brief change log

- *Correct uid/gid values used when creating the OLM bundle.*
 

## Verifying this change

## Documentation


tomncooper (approved):

SamBarker (approved):
LGTM 

but as you mention on the JIRA its not clear to me why we specify the user at all

PR #921, opened 2024-12-04
[FLINK-36836][Autoscaler] Supports config the upper and lower limits of target utilization

huyuanfeng2018:

detail: FLINK-36836 

## Brief change log

  - *Introduce two new parameters `job.autoscaler.utilization.min` and `job.autoscaler.utilization.max` to determine the upper and lower limits of utilization*

## Verifying this change

  - Change the previous test cases related to `job.autoscaler.target.utilization.boundary` to `job.autoscaler.utilization.min` and `job.autoscaler.utilization.max` 
  - add test `ScalingExecutorTest#testUtilizationBoundariesAndUtilizationMinMaxCompatibility` Verify compatibility of parameters before and after

  - Dependencies (does it add or upgrade a dependency): (yes / *no*)
  - The public API, i.e., is any changes to the `CustomResourceDescriptors`: (yes / *no*)

## Documentation

### DIFF
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricEvaluator.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricEvaluator.java
@@ -287,6 +287,7 @@ public class ScalingMetricEvaluator {
             Duration restartTime) {
 
         double targetUtilization = conf.get(UTILIZATION_TARGET);
+        double utilizationBoundary = conf.get(TARGET_UTILIZATION_BOUNDARY);
 
         double upperUtilization;
         double lowerUtilization;
@@ -297,16 +298,12 @@ public class ScalingMetricEvaluator {
             upperUtilization = 1.0;
             lowerUtilization = 0.0;
         } else {
-            if (conf.getOptional(UTILIZATION_MAX).isPresent()
-                    || conf.getOptional(UTILIZATION_MIN).isPresent()
-                    || conf.getOptional(TARGET_UTILIZATION_BOUNDARY).isEmpty()) {
-                upperUtilization = conf.get(UTILIZATION_MAX);
-                lowerUtilization = conf.get(UTILIZATION_MIN);
-            } else {
-                Double boundary = conf.get(TARGET_UTILIZATION_BOUNDARY);
-                upperUtilization = targetUtilization + boundary;
-                lowerUtilization = targetUtilization - boundary;
-            }
+            upperUtilization =
+                    conf.getOptional(UTILIZATION_MAX)
+                            .orElse(targetUtilization + utilizationBoundary);
+            lowerUtilization =
+                    conf.getOptional(UTILIZATION_MIN)
+                            .orElse(targetUtilization - utilizationBoundary);
         }
 
         double scaleUpThreshold =
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricEvaluator.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricEvaluator.java
@@ -44,8 +44,10 @@ import java.util.Optional;
 import java.util.SortedMap;
 
 import static org.apache.flink.autoscaler.config.AutoScalerOptions.BACKLOG_PROCESSING_LAG_THRESHOLD;
-import static org.apache.flink.autoscaler.config.AutoScalerOptions.TARGET_UTILIZATION;
 import static org.apache.flink.autoscaler.config.AutoScalerOptions.TARGET_UTILIZATION_BOUNDARY;
+import static org.apache.flink.autoscaler.config.AutoScalerOptions.UTILIZATION_MAX;
+import static org.apache.flink.autoscaler.config.AutoScalerOptions.UTILIZATION_MIN;
+import static org.apache.flink.autoscaler.config.AutoScalerOptions.UTILIZATION_TARGET;
 import static org.apache.flink.autoscaler.metrics.ScalingMetric.CATCH_UP_DATA_RATE;
 import static org.apache.flink.autoscaler.metrics.ScalingMetric.GC_PRESSURE;
 import static org.apache.flink.autoscaler.metrics.ScalingMetric.HEAP_MAX_USAGE_RATIO;
@@ -284,8 +286,7 @@ public class ScalingMetricEvaluator {
             boolean processingBacklog,
             Duration restartTime) {
 
-        double utilizationBoundary = conf.getDouble(TARGET_UTILIZATION_BOUNDARY);
-        double targetUtilization = conf.get(TARGET_UTILIZATION);
+        double targetUtilization = conf.get(UTILIZATION_TARGET);
 
         double upperUtilization;
         double lowerUtilization;
@@ -296,8 +297,16 @@ public class ScalingMetricEvaluator {
             upperUtilization = 1.0;
             lowerUtilization = 0.0;
         } else {
-            upperUtilization = targetUtilization + utilizationBoundary;
-            lowerUtilization = targetUtilization - utilizationBoundary;
+            if (conf.getOptional(UTILIZATION_MAX).isPresent()
+                    || conf.getOptional(UTILIZATION_MIN).isPresent()
+                    || conf.getOptional(TARGET_UTILIZATION_BOUNDARY).isEmpty()) {
+                upperUtilization = conf.get(UTILIZATION_MAX);
+                lowerUtilization = conf.get(UTILIZATION_MIN);
+            } else {
+                Double boundary = conf.get(TARGET_UTILIZATION_BOUNDARY);
+                upperUtilization = targetUtilization + boundary;
+                lowerUtilization = targetUtilization - boundary;
+            }
         }
 
         double scaleUpThreshold =
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
@@ -89,13 +89,17 @@ public class AutoScalerOptions {
                                     + "seconds suffix, daily expression's formation is startTime-endTime, such as 9:30:30-10:50:20, when exclude from 9:30:30-10:50:20 in Monday and Thursday "
                                     + "we can express it as 9:30:30-10:50:20 && * * * ? * 2,5");
 
-    public static final ConfigOption<Double> TARGET_UTILIZATION =
-            autoScalerConfig("target.utilization")
+    public static final ConfigOption<Double> UTILIZATION_TARGET =
+            autoScalerConfig("utilization.target")
                     .doubleType()
                     .defaultValue(0.7)
-                    .withFallbackKeys(oldOperatorConfigKey("target.utilization"))
+                    .withDeprecatedKeys(autoScalerConfigKey("target.utilization"))
+                    .withFallbackKeys(
+                            oldOperatorConfigKey("utilization.target"),
+                            oldOperatorConfigKey("target.utilization"))
                     .withDescription("Target vertex utilization");
 
+    @Deprecated
     public static final ConfigOption<Double> TARGET_UTILIZATION_BOUNDARY =
             autoScalerConfig("target.utilization.boundary")
                     .doubleType()
@@ -103,6 +107,20 @@ public class AutoScalerOptions {
                     .withFallbackKeys(oldOperatorConfigKey("target.utilization.boundary"))
                     .withDescription(
                             "Target vertex utilization boundary. Scaling won't be performed if the processing capacity is within [target_rate / (target_utilization - boundary), (target_rate / (target_utilization + boundary)]");
+
+    public static final ConfigOption<Double> UTILIZATION_MAX =
+            autoScalerConfig("utilization.max")
+                    .doubleType()
+                    .defaultValue(1.)
+                    .withFallbackKeys(oldOperatorConfigKey("utilization.max"))
+                    .withDescription("Max vertex utilization");
+
+    public static final ConfigOption<Double> UTILIZATION_MIN =
+            autoScalerConfig("utilization.min")
+                    .doubleType()
+                    .defaultValue(0.4)
+                    .withFallbackKeys(oldOperatorConfigKey("utilization.min"))
+                    .withDescription("Min vertex utilization");
 
     public static final ConfigOption<Duration> SCALE_DOWN_INTERVAL =
             autoScalerConfig("scale-down.interval")
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
@@ -111,14 +111,14 @@ public class AutoScalerOptions {
     public static final ConfigOption<Double> UTILIZATION_MAX =
             autoScalerConfig("utilization.max")
                     .doubleType()
-                    .defaultValue(1.)
+                    .noDefaultValue()
                     .withFallbackKeys(oldOperatorConfigKey("utilization.max"))
                     .withDescription("Max vertex utilization");
 
     public static final ConfigOption<Double> UTILIZATION_MIN =
             autoScalerConfig("utilization.min")
                     .doubleType()
-                    .defaultValue(0.4)
+                    .noDefaultValue()
                     .withFallbackKeys(oldOperatorConfigKey("utilization.min"))
                     .withDescription("Min vertex utilization");
 
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultValidator.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultValidator.java
@@ -65,6 +65,10 @@ import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import static org.apache.flink.autoscaler.config.AutoScalerOptions.UTILIZATION_MAX;
+import static org.apache.flink.autoscaler.config.AutoScalerOptions.UTILIZATION_MIN;
+import static org.apache.flink.autoscaler.config.AutoScalerOptions.UTILIZATION_TARGET;
+
 /** Default validator implementation for {@link FlinkDeployment}. */
 public class DefaultValidator implements FlinkResourceValidator {
 
@@ -605,9 +609,19 @@ public class DefaultValidator implements FlinkResourceValidator {
         return firstPresent(
                 validateNumber(flinkConfiguration, AutoScalerOptions.MAX_SCALE_DOWN_FACTOR, 0.0d),
                 validateNumber(flinkConfiguration, AutoScalerOptions.MAX_SCALE_UP_FACTOR, 0.0d),
-                validateNumber(flinkConfiguration, AutoScalerOptions.TARGET_UTILIZATION, 0.0d),
+                validateNumber(flinkConfiguration, UTILIZATION_TARGET, 0.0d, 1.0d),
                 validateNumber(
                         flinkConfiguration, AutoScalerOptions.TARGET_UTILIZATION_BOUNDARY, 0.0d),
+                validateNumber(
+                        flinkConfiguration,
+                        UTILIZATION_MAX,
+                        flinkConfiguration.get(UTILIZATION_TARGET),
+                        1.0d),
+                validateNumber(
+                        flinkConfiguration,
+                        UTILIZATION_MIN,
+                        0.0d,
+                        flinkConfiguration.get(UTILIZATION_TARGET)),
                 CalendarUtils.validateExcludedPeriods(flinkConfiguration));
     }
 
--- a/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/validation/DefaultValidatorTest.java
+++ b/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/validation/DefaultValidatorTest.java
@@ -802,22 +802,34 @@ public class DefaultValidatorTest {
         var result =
                 testAutoScalerConfiguration(
                         flinkConf ->
-                                flinkConf.put(AutoScalerOptions.TARGET_UTILIZATION.key(), "-0.6"));
+                                flinkConf.put(AutoScalerOptions.UTILIZATION_TARGET.key(), "-0.6"));
         assertErrorContains(
-                result, getFormattedErrorMessage(AutoScalerOptions.TARGET_UTILIZATION, 0.0d));
+                result, getFormattedErrorMessage(AutoScalerOptions.UTILIZATION_TARGET, 0.0d, 1.0d));
     }
 
     @Test
     public void testAutoScalerDeploymentWithInvalidNegativeUtilizationBoundary() {
-        var result =
+        var resultMaxUtilization =
                 testAutoScalerConfiguration(
                         flinkConf ->
-                                flinkConf.put(
-                                        AutoScalerOptions.TARGET_UTILIZATION_BOUNDARY.key(),
-                                        "-0.6"));
+                                flinkConf.put(AutoScalerOptions.UTILIZATION_MAX.key(), "-0.6"));
         assertErrorContains(
-                result,
-                getFormattedErrorMessage(AutoScalerOptions.TARGET_UTILIZATION_BOUNDARY, 0.0d));
+                resultMaxUtilization,
+                getFormattedErrorMessage(
+                        AutoScalerOptions.UTILIZATION_MAX,
+                        AutoScalerOptions.UTILIZATION_TARGET.defaultValue(),
+                        1.0));
+
+        var resultMinUtilization =
+                testAutoScalerConfiguration(
+                        flinkConf ->
+                                flinkConf.put(AutoScalerOptions.UTILIZATION_MIN.key(), "-0.6"));
+        assertErrorContains(
+                resultMinUtilization,
+                getFormattedErrorMessage(
+                        AutoScalerOptions.UTILIZATION_MIN,
+                        0.0d,
+                        AutoScalerOptions.UTILIZATION_TARGET.defaultValue()));
     }
 
     @Test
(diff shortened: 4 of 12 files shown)


flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
```diff
@@ -89,21 +89,35 @@ private static ConfigOptions.OptionBuilder autoScalerConfig(String key) {
                                     + "seconds suffix, daily expression's formation is startTime-endTime, such as 9:30:30-10:50:20, when exclude from 9:30:30-10:50:20 in Monday and Thursday "
                                     + "we can express it as 9:30:30-10:50:20 && * * * ? * 2,5");
 
-    public static final ConfigOption<Double> TARGET_UTILIZATION =
-            autoScalerConfig("target.utilization")
+    public static final ConfigOption<Double> UTILIZATION_TARGET =
+            autoScalerConfig("utilization.target")
                     .doubleType()
                     .defaultValue(0.7)
                     .withFallbackKeys(oldOperatorConfigKey("target.utilization"))
```
1996fanrui, line 96: The new option key is ` job.autoscaler.utilization.target`, but it should be compatible with the following options:

- job.autoscaler.target.utilization (Deprecated key)
- kubernetes.operator.job.autoscaler.target.utilization (fallback key)
- kubernetes.operator.job.autoscaler.utilization.target (fallback key)

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
```diff
@@ -89,21 +89,35 @@ private static ConfigOptions.OptionBuilder autoScalerConfig(String key) {
...
 
-    public static final ConfigOption<Double> TARGET_UTILIZATION_BOUNDARY =
-            autoScalerConfig("target.utilization.boundary")
+    public static final ConfigOption<Double> UTILIZATION_TARGET_BOUNDARY =
+            autoScalerConfig("utilization.target.boundary")
                     .doubleType()
                     .defaultValue(0.3)
                     .withFallbackKeys(oldOperatorConfigKey("target.utilization.boundary"))
                     .withDescription(
                             "Target vertex utilization boundary. Scaling won't be performed if the processing capacity is within [target_rate / (target_utilization - boundary), (target_rate / (target_utilization + boundary)]");
```
1996fanrui, line 105: As we discussed in FLINK-36836, we only need the following 3 options:

```
job.autoscaler.utilization.target
job.autoscaler.utilization.min
job.autoscaler.utilization.max
```

So `target.utilization.boundary` should be deprecated in this PR. 

But we need to ensure the backward compatibility. For example, if `job.autoscaler.utilization.min` or `job.autoscaler.utilization.max` are not set, and `job.autoscaler.target.utilization.boundary` is set, we could use the `job.autoscaler.target.utilization.boundary`.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
```diff
@@ -89,21 +89,35 @@ private static ConfigOptions.OptionBuilder autoScalerConfig(String key) {
...
 
+    public static final ConfigOption<Double> TARGET_UTILIZATION_MAX =
+            autoScalerConfig("utilization.target.max")
```
1996fanrui, line 108:
```suggestion
            autoScalerConfig("utilization.max")
```

`job.autoscaler.utilization.max` is the option key we discussed in FLINK-36836.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
```diff
@@ -89,21 +89,35 @@ private static ConfigOptions.OptionBuilder autoScalerConfig(String key) {
...
+                    .doubleType()
+                    .defaultValue(-1d)
```
1996fanrui, line 110: IIUC, if we don't update the default value, it should be 1.0d.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
```diff
@@ -89,21 +89,35 @@ private static ConfigOptions.OptionBuilder autoScalerConfig(String key) {
...
+                    .withFallbackKeys(oldOperatorConfigKey("utilization.target.max"))
+                    .withDescription("Max vertex utilization");
+
+    public static final ConfigOption<Double> TARGET_UTILIZATION_MIN =
+            autoScalerConfig("utilization.target.min")
+                    .doubleType()
+                    .defaultValue(-1d)
```
1996fanrui, line 117: IIUC, if we don't update the default value, it should be 0.4d.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
```diff
@@ -89,21 +89,35 @@ private static ConfigOptions.OptionBuilder autoScalerConfig(String key) {
...
 
+    public static final ConfigOption<Double> TARGET_UTILIZATION_MAX =
+            autoScalerConfig("utilization.target.max")
+                    .doubleType()
+                    .defaultValue(-1d)
+                    .withFallbackKeys(oldOperatorConfigKey("utilization.target.max"))
+                    .withDescription("Max vertex utilization");
+
+    public static final ConfigOption<Double> TARGET_UTILIZATION_MIN =
+            autoScalerConfig("utilization.target.min")
```
1996fanrui, line 115:
```suggestion
            autoScalerConfig("utilization.min")
```

1996fanrui (commented):
Thanks @huyuanfeng2018 for the PR.

Although it's just a draft PR, I checked the changes of AutoScalerOptions, something is unexpected from me. So I would give some feedbacks in advance.

huyuanfeng2018:
> Thanks @huyuanfeng2018 for the PR.
> 
> Although it's just a draft PR, I checked the changes of AutoScalerOptions, something is unexpected from me. So I would give some feedbacks in advance.

Thanks for the quick suggestion!

I have made the suggested modifications and added the test case, PTAL.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricEvaluator.java
```diff
@@ -296,8 +297,14 @@ protected static void computeProcessingRateThresholds(
...
+                    conf.getOptional(TARGET_UTILIZATION_BOUNDARY)
+                            .map(boundary -> targetUtilization + boundary)
+                            .orElseGet(() -> conf.get(UTILIZATION_MAX));
+            lowerUtilization =
+                    conf.getOptional(TARGET_UTILIZATION_BOUNDARY)
+                            .map(boundary -> targetUtilization - boundary)
+                            .orElseGet(() -> conf.get(UTILIZATION_MIN));
         }
 
         double scaleUpThreshold =
```
1996fanrui, line 310: In `AutoScalerUtils.getTargetProcessingCapacity`, we adjust `targetUtilization` between 0.0 and 1.0, it's reasonable.

After introducing `UTILIZATION_MAX` and `UTILIZATION_MIN`, it's better to ensure:
-  1.0 >= UTILIZATION_MAX >= UTILIZATION_TARGET
- UTILIZATION_TARGET >= UTILIZATION_MIN >= 0.0

And it's needed to add some tests to check them.

Also, could we unify the logic in this method instead of `AutoScalerUtils.getTargetProcessingCapacity`?


Of course, another solution is when `UTILIZATION_MAX < UTILIZATION_TARGET` or `UTILIZATION_TARGET < UTILIZATION_MIN`, we could throw exception.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/validation/DefaultValidatorTest.java
```diff
@@ -972,8 +990,9 @@ private Map<String, String> getDefaultTestAutoScalerFlinkConfigurationMap() {
         conf.put(AutoScalerOptions.MAX_SCALE_UP_FACTOR.key(), "100000.0");
         conf.put(AutoScalerOptions.MAX_SCALE_DOWN_FACTOR.key(), "0.6");
         conf.put(AutoScalerOptions.SCALING_EFFECTIVENESS_DETECTION_ENABLED.key(), "0.1");
-        conf.put(AutoScalerOptions.TARGET_UTILIZATION.key(), "0.7");
-        conf.put(AutoScalerOptions.TARGET_UTILIZATION_BOUNDARY.key(), "0.4");
+        conf.put(AutoScalerOptions.UTILIZATION_TARGET.key(), "0.7");
+        conf.put(AutoScalerOptions.UTILIZATION_MAX.key(), "1.");
+        conf.put(AutoScalerOptions.UTILIZATION_MIN.key(), "0.1");
```
1996fanrui, line 995: Why not 0.3? As I understand, the original logic is 0.7 - 0.4.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricEvaluator.java
```diff
@@ -296,8 +297,14 @@ protected static void computeProcessingRateThresholds(
...
-            upperUtilization = targetUtilization + utilizationBoundary;
-            lowerUtilization = targetUtilization - utilizationBoundary;
+            upperUtilization =
+                    conf.getOptional(TARGET_UTILIZATION_BOUNDARY)
+                            .map(boundary -> targetUtilization + boundary)
+                            .orElseGet(() -> conf.get(UTILIZATION_MAX));
+            lowerUtilization =
+                    conf.getOptional(TARGET_UTILIZATION_BOUNDARY)
+                            .map(boundary -> targetUtilization - boundary)
+                            .orElseGet(() -> conf.get(UTILIZATION_MIN));
```
1996fanrui, line 307: IIUC, the compatibility logic is not as expected. `TARGET_UTILIZATION_BOUNDARY` should be only used when  `UTILIZATION_MAX` or `UTILIZATION_MIN` are not set by users.

If `UTILIZATION_MAX` or `UTILIZATION_MIN` is set by users, it means users are using the latest options. So we should ignore `TARGET_UTILIZATION_BOUNDARY` even if it's set as well. WDYT?

Also, it should be tested as well.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricEvaluator.java
huyuanfeng2018: Thanks for review.

The last commit was not fully checked and there were some errors in some places, which have been fixed.


> In `AutoScalerUtils.getTargetProcessingCapacity`, we adjust `targetUtilization` between 0.0 and 1.0, it's reasonable.
> 
> After introducing `UTILIZATION_MAX` and `UTILIZATION_MIN`, it's better to ensure:
> 
> * 1.0 >= UTILIZATION_MAX >= UTILIZATION_TARGET
> * UTILIZATION_TARGET >= UTILIZATION_MIN >= 0.0
> 
> And it's needed to add some tests to check them.
> 

Added `DefaultValidator#validateNumberOrder` to verify that multiple parameters need to follow the logic of size order, and added test cases. Also`validateNumber` can also limit the maximum value

> Also, could we unify the logic in this method instead of `AutoScalerUtils.getTargetProcessingCapacity`?

`AutoScalerUtils.getTargetProcessingCapacity` I think it is reasonable to keep them in this pr, Because it is used in many places

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricEvaluator.java
huyuanfeng2018: fixed.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricEvaluator.java
1996fanrui: `DefaultValidator` is a class of `flink-kubernetes-operator` module. It doesn't work for Autoscaler Standalone or other scenarios. Could we validate autoscaler related options inside of `flink-autoscaler` module? and call it in `JobAutoScalerImpl#scale`?

1996fanrui: > But how does Autoscaler standlone handle exceptions where these parameters are unreasonable.
by event report?
> 
> I'm not sure if it needs to be resolved via another PR?

It's acceptable that Autoscaler standlone check options in a separate PR.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultValidator.java
```diff
@@ -605,9 +605,16 @@ public static Optional<String> validateAutoScalerFlinkConfiguration(
         return firstPresent(
                 validateNumber(flinkConfiguration, AutoScalerOptions.MAX_SCALE_DOWN_FACTOR, 0.0d),
                 validateNumber(flinkConfiguration, AutoScalerOptions.MAX_SCALE_UP_FACTOR, 0.0d),
-                validateNumber(flinkConfiguration, AutoScalerOptions.TARGET_UTILIZATION, 0.0d),
+                validateNumber(flinkConfiguration, AutoScalerOptions.UTILIZATION_TARGET, 0.0d),
                 validateNumber(
                         flinkConfiguration, AutoScalerOptions.TARGET_UTILIZATION_BOUNDARY, 0.0d),
+                validateNumber(flinkConfiguration, AutoScalerOptions.UTILIZATION_MAX, 0.0d, 1.0d),
+                validateNumber(flinkConfiguration, AutoScalerOptions.UTILIZATION_MIN, 0.0d, 1.0d),
```
1996fanrui, line 612: It seems we don't need a new `validateNumberOrder`, and following change is enough. 

```suggestion
                validateNumber(flinkConfiguration, AutoScalerOptions.UTILIZATION_MAX, flinkConfiguration.get(AutoScalerOptions.UTILIZATION_TARGET), 1.0d),
                validateNumber(flinkConfiguration, AutoScalerOptions.UTILIZATION_MIN, 0.0d, flinkConfiguration.get(AutoScalerOptions.UTILIZATION_TARGET)),
```

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/validation/DefaultValidatorTest.java
```diff
@@ -940,13 +958,78 @@ public void testNonEnabledAutoScalerSessionJob() {
...
-                                    AutoScalerOptions.TARGET_UTILIZATION_BOUNDARY.key(), "-1.6");
+                            flinkConf.put(AutoScalerOptions.UTILIZATION_TARGET.key(), "-1.6");
+                            flinkConf.put(AutoScalerOptions.UTILIZATION_MAX.key(), "-1.6");
+                            flinkConf.put(AutoScalerOptions.UTILIZATION_MIN.key(), "-1.6");
                         });
         assertErrorNotContains(result);
     }
 
+    @Test
+    public void testAutoScalerConfigurationNumberOrder() {
```
1996fanrui, line 969: If the last comment makes sense, this test needs to be refactored.

1996fanrui (commented):
Thanks @huyuanfeng2018  for the update!

Overall LGTM,  I only left 2 minor comments, please take a look in your free time, thanks~

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricEvaluator.java
```diff
@@ -296,8 +297,16 @@ protected static void computeProcessingRateThresholds(
             upperUtilization = 1.0;
             lowerUtilization = 0.0;
         } else {
-            upperUtilization = targetUtilization + utilizationBoundary;
-            lowerUtilization = targetUtilization - utilizationBoundary;
+            if (conf.getOptional(UTILIZATION_MAX).isPresent()
+                    || conf.getOptional(UTILIZATION_MIN).isPresent()
+                    || conf.getOptional(TARGET_UTILIZATION_BOUNDARY).isEmpty()) {
+                upperUtilization = conf.get(UTILIZATION_MAX);
+                lowerUtilization = conf.get(UTILIZATION_MIN);
```
mxm, line 304: I'm wondering whether the logic should be the following? 

```suggestion
            if (conf.getOptional(UTILIZATION_MAX).isPresent()
                    && conf.getOptional(UTILIZATION_MIN).isPresent()) {
                upperUtilization = conf.get(UTILIZATION_MAX);
                lowerUtilization = conf.get(UTILIZATION_MIN);
```

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
```diff
@@ -89,13 +89,17 @@ private static ConfigOptions.OptionBuilder autoScalerConfig(String key) {
                                     + "seconds suffix, daily expression's formation is startTime-endTime, such as 9:30:30-10:50:20, when exclude from 9:30:30-10:50:20 in Monday and Thursday "
                                     + "we can express it as 9:30:30-10:50:20 && * * * ? * 2,5");
 
-    public static final ConfigOption<Double> TARGET_UTILIZATION =
-            autoScalerConfig("target.utilization")
+    public static final ConfigOption<Double> UTILIZATION_TARGET =
+            autoScalerConfig("utilization.target")
```
mxm, line 93: Do we have to change this well-established key? This might confuse users, even though we have a fallback.

mxm: I understand its cleaner, but perhaps we can simply add the new keys and leave this unchanged?

mxm (commented):
Thanks for the PR!

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricEvaluator.java
huyuanfeng2018: I think as long as one of UTILIZATION_MAX and UTILIZATION_MIN is set, it means that the user is already using the new config.

At the same time, if the user does not set TARGET_UTILIZATION_BOUNDARY, we still use the new config

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
huyuanfeng2018: > I understand its cleaner, but perhaps we can simply add the new keys and leave this unchanged?
> Do we have to change this well-established key? This might confuse users, even though we have a fallback.

I think methed `withFallbackKeys` good enough
WDYT?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultValidator.java
huyuanfeng2018: makes sense

huyuanfeng2018:
Thanks @1996fanrui @mxm review.

Changes and replies have been made based on comments，PTAL.

 thanks~

1996fanrui (approved):

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
gyfora: I think in the long term it's preferable to have consistent well named keys. I think the fallback keys should be enough. I would expect that users will gravitate towards the new min/max options and at that point this will make a lot of sense for everyone.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricEvaluator.java
gyfora: I think we should not hard-code the defaults in the MIN_MAX options but it should be dynamically computed based on the target.

For example if the users sets target to 0.4 and min to 0.1, then it doesn't make sense to have the default max on 1.. Previously the boundary was 0.4 by default I think we should preserve this

huyuanfeng2018: I don't know if there is something wrong with my understanding:

Should we keep the boundary and then the  max and min values ​​are derived by `boundary + target` and `target - boundary` by default? Unless the user explicitly sets min max?

If that's the case, I think it's reasonable. 

But in this case, should we change the min max parameter like to `utilization.lower.boundary` and `utilization.upper.boundary` ? 
Then `max = upper.boundary+ target` , `min = target- upper.boundary`

gyfora: As @mxm suggested I think the logic should be enabled if either MIN or MAX is defined. If only one of them is defined then we can compute the default for the other one from the target utilization +- boundary.

I think we should not have default values for the min/max options

mxm: >As @mxm suggested I think the logic should be enabled if either MIN or MAX is defined

@gyfora Did you mean, if BOTH min and max are defined, the new logic should be enabled? If only one of them is defined, then we compute the other one using the configured target utilization and the boundary. I think that makes sense 👍

mxm: +1 for not having static defaults for the new options because that would be confusing to users.

huyuanfeng2018: Thanks @mxm  @gyfora  for review.

> compute the default for the other one from the target utilization +- boundary.

+1 for  this .

I've modified and added compatibility test cases.

PTAL~

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
```diff
@@ -89,21 +89,41 @@ private static ConfigOptions.OptionBuilder autoScalerConfig(String key) {
...
+                    .noDefaultValue()
+                    .withFallbackKeys(oldOperatorConfigKey("utilization.max"))
+                    .withDescription("Max vertex utilization");
+
+    public static final ConfigOption<Double> UTILIZATION_MIN =
+            autoScalerConfig("utilization.min")
+                    .doubleType()
+                    .noDefaultValue()
+                    .withFallbackKeys(oldOperatorConfigKey("utilization.min"))
+                    .withDescription("Min vertex utilization");
```
1996fanrui, line 125: These 2 options have no default value, I'm not sure whether it's expected.

If the `target.utilization.boundary` is removed in the future, how do we handle the default value?

Usually, when a new option is introduced, a default value is also designed. When deprecated options are removed in subsequent versions, we only need to remove unnecessary compatibility code.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
```diff
@@ -89,21 +89,41 @@ private static ConfigOptions.OptionBuilder autoScalerConfig(String key) {
...
+                    .withDeprecatedKeys(autoScalerConfigKey("target.utilization"))
+                    .withFallbackKeys(
+                            oldOperatorConfigKey("utilization.target"),
+                            oldOperatorConfigKey("target.utilization"))
                     .withDescription("Target vertex utilization");
 
-    public static final ConfigOption<Double> TARGET_UTILIZATION_BOUNDARY =
-            autoScalerConfig("target.utilization.boundary")
+    public static final ConfigOption<Double> UTILIZATION_TARGET_BOUNDARY =
+            autoScalerConfig("utilization.target.boundary")
```
1996fanrui, line 103: This option is deprecated, do we need to add a `@Deprecated` annotation here?

Also, IIUC, we don't need to rename the option key name since it's deprecated. It means new users don't need it and we don't maintain it in the future. (Compatibility needs to be considered only in the short term.)

1996fanrui (changes requested):
Hi @huyuanfeng2018 , thanks for the update!

I left 2 comments for the last commit, and they are related to https://github.com/apache/flink-kubernetes-operator/pull/921#discussion_r1879605290 (Compatibility issues), please @gyfora and @mxm  help double confirm, thanks a lot.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
huyuanfeng2018: Well, maybe retaining the `target.utilization.boundary` parameter is also a feasible solution,like latest code changes, I don't know if my understanding is consistent with @gyfora and @mxm.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
huyuanfeng2018: If `target.utilization.boundary` will not be deprecated in the future，Is it reasonable?

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
1996fanrui: As I understand, we will only keep the following three options in the end, and make it compatible with `job.autoscaler.target.utilization.boundary` before it is deprecated.

```
job.autoscaler.utilization.target
job.autoscaler.utilization.min
job.autoscaler.utilization.max
```

Do you mean we keep all 4 options in the end?

gyfora: I think we should remove the boundary at the end (deprecate now), but still min/max should not have fixed defaults as the default would be derived from the current target.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
gyfora: we should not rename this option also, the name you set is incorrect anyways. Keys should not be prefixes of each other (not yaml compliant)

huyuanfeng2018:
Thanks @1996fanrui  @gyfora  for the reply

Well, let me summarize our conclusion, 
* we need to remove the boundary parameter in the future (deprecate now), So there is no need to rename it now
* default min/max need to be derived from target

Then the logic for determining the default value of min/max is ：

now（boundary exist） : `max = upper.boundary+ target , min = target- upper.boundary`

future（boundary remove） : `max = 0.3 + target , min = target- 0.3`,  `0.3` is a static value

I don’t know if there is anything wrong with my understanding.

gyfora:
> Thanks @1996fanrui @gyfora for the reply
> 
> Well, let me summarize our conclusion,
> 
> * we need to remove the boundary parameter in the future (deprecate now), So there is no need to rename it now
> * default min/max need to be derived from target
> 
> Then the logic for determining the default value of min/max is ：
> 
> now（boundary exist） : `max = upper.boundary+ target , min = target- upper.boundary`
> 
> future（boundary remove） : `max = 0.3 + target , min = target- 0.3`, `0.3` is a static value
> 
> I don’t know if there is anything wrong with my understanding.

That sounds fully correct :)

1996fanrui (approved):
Thanks for the update, LGTM assuming the CI is green.

huyuanfeng2018:
> Thanks for the update, LGTM assuming the CI is green.

Thanks for your quick reply

gyfora (approved):

PR #834, opened 2024-06-03
[FLINK-33803] Remove deprecated `lastReconciledSpec` meta generation in favour of `observedGeneration`

angelachenn:
## What is the purpose of the change
Followup to https://issues.apache.org/jira/browse/FLINK-33803 (https://github.com/apache/flink-kubernetes-operator/pull/755)

> The `observedGeneration` field in the `FlinkDeployment` CRD status is used to persist the generation that was last acted on by the operator. At the end of a successful reconciliation, the operator sets the `observedGeneration` field to the current `generation` number (from the CRD metadata).

This PR removes internal usages of the `lastReconciledSpec` metadata (and generation) that were kept in the initial PR to avoid downstream breakage. `observedGeneration` replaces this semantically. Removing the deprecated fields is to end up with a cleaner codebase.

Note: the removed fields were nested within the `lastReconciledSpec` `String`, thus there is no documentation to be updated (please advise if I missed something).
 
## Brief change log
- There is no more `ObjectMeta` `metadata` attribute (containing `generation`) within the `ReconciliationMetadata` class
  - that field will not exist to the operator within `FlinkDeploymentStatus` w.r.t. reconciliation
- removed the `metadata` field (from the CRD metadata) during deserialization before passing the content into a `ReconciliationMetadata` instance such that its constructor does not mutate the resources
  - if not, Jackson throws a `JsonMappingException`: `Unrecognized field 'metadata' (class org.apache.flink.kubernetes.operator.api.reconciler.ReconciliationMetadata), not marked as ignorable`, causing pods to fail.

## Verifying this change

This change added tests and can be verified as follows:

  - Modified existing tests that checked for `specWithMeta.getMeta().getMetadata().getGeneration()` to assert against `status.getObservedGeneration()` instead (tests pass)
  - Manually verified the change by running a cluster with  a `FlinkDeployment` custom resource that depends on the operator (with 2 JobManagers & 2 TaskManagers), and checking its CRD, verifying that fields are mapped accordingly.
<details> <summary> Logs </summary>

Before:
```yaml
k get flinkdeployment flink --context=<context> -n <ns>  -o yaml | yq .status.reconciliationStatus.lastReconciledSpec | jq .resource_metadata
{
  "apiVersion": "flink.apache.org/v1beta1",
  "metadata": {
    "generation": 5
  },
  "firstDeployment": false
}
```
```yaml
k get flinkdeployment flink --context=<context> -n <ns>  -o yaml | yq .status.observedGeneration
5
```
After:
[...]

PR #923, opened 2024-12-11
[FLINK-36890] Update observedGeneration on NOOP reconciliation

mxm:
The observedGeneration field wasn't updated anymore when the spec update results in a NOOP reconciliation.

This is a regression of a10fb4501011be4e88fb6164156e3b9e8527950e / https://github.com/apache/flink-kubernetes-operator/pull/834


flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconcilerTest.java
```diff
@@ -1235,10 +1235,6 @@ public void testUpgradeReconciledGeneration() throws Exception {
         // Submit no-op upgrade
         deployment.getSpec().getFlinkConfiguration().put("kubernetes.operator.test", "value");
```
gyfora, line 1236: This is not a no-op change if we change the spec?

gyfora: ah yes, you are right sorry

gyfora (approved):

PR #926, opened 2025-01-06
[FLINK-36932][metrics] Added resource-level metrics for different states/statuses

pgrefviau:

This PR adds new metrics that help track the current value of different states/statuses at the resource level. In some cases, metrics already exists for some of these statuses/states, but those metrics represent namespace or system-wide counts, as opposed to per-resource gauges that indicate whether or not a deployment/session job is in a particular state.

In other cases, some statuses/states that weren't yet tracked through a dedicated metric (ex: job status) now have a resource-level gauge and namespace-level counter.

## Brief change log

Summary of the changes for each state/status:
- `JobManagerDeploymentStatus`: state gauge added at resource-level (FlinkDeployment only)
- `JobStatus`: status gauge added at resource-level (FlinkDeployment only), status counter at namespace-level
- `ResourceLifecycleState`: state gauge added at resource-level

## Verifying this change

- Updated test cases for resource lifecycle metrics and Flink deployment metrics to account for new resource-level metrics
  - Also added utility methods to test classes to reduce duplicated test logic
- Changes were deployed and tested using our own fork/instance of the operator

## Documentation

N.B. While these changes might not represent a full-on "feature", I'm planning to update the documentation that generates [this](https://nightlies.apache.org/flink/flink-kubernetes-operator-docs-main/docs/operations/metrics-logging/) page. However, I've held off doing this as part of this initial commit in order to settle the naming and implementation. Once this is done, I can update the documentation accordingly.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/metrics/FlinkDeploymentMetrics.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/metrics/FlinkDeploymentMetrics.java
@@ -17,6 +17,7 @@
 
 package org.apache.flink.kubernetes.operator.metrics;
 
+import org.apache.flink.api.common.JobStatus;
 import org.apache.flink.configuration.Configuration;
 import org.apache.flink.kubernetes.operator.api.FlinkDeployment;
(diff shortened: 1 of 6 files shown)


gyfora:
> @gyfora any updates here?

Not sure what you mean @blaskovicz . I have asked some questions and got no answers, I don't have any updates :D

PR #929, opened 2025-01-09
[FLINK-36696] [flink-autoscaler-plugin-jdbc] Switch sql connection usages to datasource

sharath1709:

This pull request makes changes to replace the usage of SQL connection with DataSource to improve the multithreaded performance of the JDBC plugin.

## Brief change log

  - *Replace all usages of SQL connection with DataSource *

## Verifying this change

This change is already covered by existing unit tests. The test cases have been modified appropriately to use DataSource rather than Connection everywhere

## Documentation

### DIFF
--- a/flink-autoscaler-plugin-jdbc/src/test/java/org/apache/flink/autoscaler/jdbc/event/AbstractJdbcAutoscalerEventHandlerITCase.java
+++ b/flink-autoscaler-plugin-jdbc/src/test/java/org/apache/flink/autoscaler/jdbc/event/AbstractJdbcAutoscalerEventHandlerITCase.java
@@ -42,6 +42,7 @@ import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
 
 import javax.annotation.Nonnull;
+import javax.sql.DataSource;
 
 import java.sql.Connection;
 import java.sql.PreparedStatement;
@@ -73,20 +74,22 @@ abstract class AbstractJdbcAutoscalerEventHandlerITCase implements DatabaseTest 
             generateScalingSummaries(currentParallelism, newParallelism, metricAvg, metricCurrent);
     private final Clock defaultClock = Clock.fixed(createTime, ZoneId.systemDefault());
 
+    private DataSource dataSource;
     private CountableJdbcEventInteractor jdbcEventInteractor;
     private JdbcAutoScalerEventHandler<JobID, JobAutoScalerContext<JobID>> eventHandler;
     private JobAutoScalerContext<JobID> ctx;
 
     @BeforeEach
     void beforeEach() throws Exception {
-        jdbcEventInteractor = new CountableJdbcEventInteractor(getConnection());
+        dataSource = getDataSource();
+        jdbcEventInteractor = new CountableJdbcEventInteractor(dataSource);
         jdbcEventInteractor.setClock(defaultClock);
         eventHandler = new JdbcAutoScalerEventHandler<>(jdbcEventInteractor, Duration.ZERO);
         ctx = createDefaultJobAutoScalerContext();
     }
 
     @AfterEach
-    void tearDown() {
+    void tearDown() throws Exception {
         eventHandler.close();
     }
 
@@ -459,7 +462,7 @@ abstract class AbstractJdbcAutoscalerEventHandlerITCase implements DatabaseTest 
                         .max(Comparable::compareTo)
                         .orElseThrow();
 
-        try (Connection connection = getConnection();
+        try (Connection connection = dataSource.getConnection();
                 PreparedStatement ps =
                         connection.prepareStatement(
                                 "update t_flink_autoscaler_event_handler set id = ? where id = ?")) {
@@ -505,6 +508,7 @@ abstract class AbstractJdbcAutoscalerEventHandlerITCase implements DatabaseTest 
             Duration eventHandlerTtl,
             int unexpiredRecordsNum)
             throws Exception {
+
         eventHandler = new JdbcAutoScalerEventHandler<>(jdbcEventInteractor, eventHandlerTtl);
 
         // Init the expired records.
@@ -525,7 +529,7 @@ abstract class AbstractJdbcAutoscalerEventHandlerITCase implements DatabaseTest 
 
         eventHandler.cleanExpiredEvents();
 
-        try (Connection connection = getConnection();
+        try (Connection connection = dataSource.getConnection();
                 PreparedStatement ps =
                         connection.prepareStatement(
                                 "select count(1) from t_flink_autoscaler_event_handler");
@@ -542,7 +546,7 @@ abstract class AbstractJdbcAutoscalerEventHandlerITCase implements DatabaseTest 
         if (minId == null) {
             return;
         }
-        try (Connection connection = getConnection();
+        try (Connection connection = dataSource.getConnection();
                 PreparedStatement ps =
                         connection.prepareStatement(
                                 "delete from t_flink_autoscaler_event_handler where id = ?")) {
--- a/flink-autoscaler-plugin-jdbc/src/test/java/org/apache/flink/autoscaler/jdbc/event/AbstractJdbcEventInteractorITCase.java
+++ b/flink-autoscaler-plugin-jdbc/src/test/java/org/apache/flink/autoscaler/jdbc/event/AbstractJdbcEventInteractorITCase.java
@@ -46,8 +46,7 @@ abstract class AbstractJdbcEventInteractorITCase implements DatabaseTest {
 
         // The datetime precision is seconds in MySQL by default.
         var createTime = Instant.now().truncatedTo(ChronoUnit.SECONDS);
-        try (var conn = getConnection()) {
-            var jdbcEventInteractor = new JdbcEventInteractor(conn);
+        try (var jdbcEventInteractor = new JdbcEventInteractor(getDataSource())) {
             jdbcEventInteractor.setClock(Clock.fixed(createTime, ZoneId.systemDefault()));
 
             jdbcEventInteractor.createEvent(
--- a/flink-autoscaler-plugin-jdbc/src/test/java/org/apache/flink/autoscaler/jdbc/state/AbstractJdbcStateInteractorITCase.java
+++ b/flink-autoscaler-plugin-jdbc/src/test/java/org/apache/flink/autoscaler/jdbc/state/AbstractJdbcStateInteractorITCase.java
@@ -42,8 +42,7 @@ abstract class AbstractJdbcStateInteractorITCase implements DatabaseTest {
         var value1 = "value1";
         var value2 = "value2";
         var value3 = "value3";
-        try (var conn = getConnection()) {
-            var jdbcStateInteractor = new JdbcStateInteractor(conn);
+        try (var jdbcStateInteractor = new JdbcStateInteractor(getDataSource())) {
             assertThat(jdbcStateInteractor.queryData(jobKey)).isEmpty();
 
             // Test for creating data.
--- a/flink-autoscaler-plugin-jdbc/src/test/java/org/apache/flink/autoscaler/jdbc/state/AbstractJdbcStateStoreITCase.java
+++ b/flink-autoscaler-plugin-jdbc/src/test/java/org/apache/flink/autoscaler/jdbc/state/AbstractJdbcStateStoreITCase.java
@@ -28,7 +28,8 @@ import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
-import java.sql.Connection;
+import javax.sql.DataSource;
+
 import java.sql.SQLException;
 import java.util.List;
 import java.util.Map;
@@ -45,21 +46,21 @@ import static org.assertj.core.api.Assertions.assertThatThrownBy;
 abstract class AbstractJdbcStateStoreITCase implements DatabaseTest {
 
     private static final String DEFAULT_JOB_KEY = "jobKey";
-    private Connection conn;
+    private DataSource dataSource;
     private CountableJdbcStateInteractor jdbcStateInteractor;
     private JdbcStateStore jdbcStateStore;
 
     @BeforeEach
     void beforeEach() throws Exception {
-        this.conn = getConnection();
-        this.jdbcStateInteractor = new CountableJdbcStateInteractor(conn);
+        this.dataSource = getDataSource();
+        this.jdbcStateInteractor = new CountableJdbcStateInteractor(dataSource);
         this.jdbcStateStore = new JdbcStateStore(jdbcStateInteractor);
     }
 
     @AfterEach
-    void afterEach() throws SQLException {
-        if (conn != null) {
-            conn.close();
+    void afterEach() throws Exception {
+        if (dataSource instanceof AutoCloseable) {
+            ((AutoCloseable) dataSource).close();
         }
     }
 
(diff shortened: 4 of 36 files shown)


flink-autoscaler-plugin-jdbc/src/test/java/org/apache/flink/autoscaler/jdbc/state/AbstractJdbcStateStoreITCase.java
```diff
@@ -45,24 +45,17 @@
...
+        this.jdbcStateInteractor = new CountableJdbcStateInteractor(dataSource);
         this.jdbcStateStore = new JdbcStateStore(jdbcStateInteractor);
     }
 
-    @AfterEach
-    void afterEach() throws SQLException {
-        if (conn != null) {
-            conn.close();
-        }
-    }
```
1996fanrui, line 64: Why don't close dataSource here?

As I understand, each test will create a new dataSource. The old dataSource will leak some connections if we don't close dataSource here.

flink-autoscaler-plugin-jdbc/src/test/java/org/apache/flink/autoscaler/jdbc/state/AbstractJdbcStateInteractorITCase.java
```diff
@@ -42,33 +42,32 @@ void testAllOperations() throws Exception {
         var value1 = "value1";
         var value2 = "value2";
         var value3 = "value3";
-        try (var conn = getConnection()) {
-            var jdbcStateInteractor = new JdbcStateInteractor(conn);
-            assertThat(jdbcStateInteractor.queryData(jobKey)).isEmpty();
+        var dataSource = getDataSource();
```
1996fanrui, line 45: It's similar with the last comment, we didn't close the dataSource.

1996fanrui (commented):
Thanks @sharath1709 for the contribution!

Overall LGTM! And I still have some comments:

1. The first question is related to connection leak for testing. This PR will create lots of new dataSources, but the close isn't called. IIUC, it will have connection leak.
2. The CI is failed, please take a look in your free time.
3. We encountered a bug related to this fix recently. The database master is crashed, the slave is promoted to be the master. The jdbc state store and event handler won't work due to they are using the fixed connection. 
    - As I understand,  DataSource could solve this problem as well. 
    - When old connection is crashed, DataSource will create a new physical connection to connect the new database master server.
    - Would you mind adding a test to check it? In the ITCase, we only have one databse server, I'm not sure could it be reproduced when the database server is restarted.

sharath1709:
Thanks a lot @1996fanrui for the quick review. Please find my replies below

1. Typically, DataSource connections will be relinquished automatically after a time out and DataSource isn't Closeable. However, some DataSource implementations like HikariDataSource implement the Closeable interface to close the ConnectionPool. We can add a check to see if the DataSource is Closeable and then close it. Additionally, I've attempted to make the classes that hold the datasource objects Autocloseable as well so that they could be created in a try-with-resources block
2. Thanks, I've tried to fix most of the issues but might need a couple more iterations.
3. Great point, I will try to add a test case for this scenario soon

flink-autoscaler-plugin-jdbc/src/test/java/org/apache/flink/autoscaler/jdbc/testutils/databases/derby/DerbyExtension.java
```diff
@@ -34,8 +37,10 @@ public class DerbyExtension implements BeforeAllCallback, AfterAllCallback, Afte
             List.of("t_flink_autoscaler_state_store", "t_flink_autoscaler_event_handler");
     private static final String JDBC_URL = "jdbc:derby:memory:test";
 
-    public Connection getConnection() throws Exception {
-        return DriverManager.getConnection(JDBC_URL);
+    public DataSource getDataSource() {
```
1996fanrui, line 40: Similar with the last comment.

flink-autoscaler-standalone/src/test/java/org/apache/flink/autoscaler/standalone/AutoscalerStateStoreFactoryTest.java
```diff
@@ -67,14 +66,12 @@ void testCreateJdbcStateStore() throws Exception {
...
         conf.set(STATE_STORE_TYPE, JDBC);
         conf.set(JDBC_URL, String.format("%s;create=true", jdbcUrl));
-        HikariJDBCUtil.getConnection(conf).close();
 
         var stateStore = AutoscalerStateStoreFactory.create(conf);
         assertThat(stateStore).isInstanceOf(JdbcAutoScalerStateStore.class);
 
         try {
             conf.set(JDBC_URL, String.format("%s;shutdown=true", jdbcUrl));
-            HikariJDBCUtil.getConnection(conf).close();
```
1996fanrui, line 77: IIUC, `conf.set(JDBC_URL, String.format("%s;shutdown=true", jdbcUrl))` means shutdown the derby server. Don't we need to call it here?

flink-autoscaler-standalone/src/test/java/org/apache/flink/autoscaler/standalone/AutoscalerEventHandlerFactoryTest.java
```diff
@@ -68,14 +67,12 @@ void testCreateJdbcEventHandler() throws Exception {
...
         conf.set(EVENT_HANDLER_TYPE, JDBC);
         conf.set(JDBC_URL, String.format("%s;create=true", jdbcUrl));
-        HikariJDBCUtil.getConnection(conf).close();
 
         var eventHandler = AutoscalerEventHandlerFactory.create(conf);
         assertThat(eventHandler).isInstanceOf(JdbcAutoScalerEventHandler.class);
 
         try {
             conf.set(JDBC_URL, String.format("%s;shutdown=true", jdbcUrl));
-            HikariJDBCUtil.getConnection(conf).close();
```
1996fanrui, line 78: Same with the last comment.

flink-autoscaler-plugin-jdbc/src/test/java/org/apache/flink/autoscaler/jdbc/event/AbstractJdbcAutoscalerEventHandlerITCase.java
```diff
@@ -459,7 +462,8 @@ void testDeleteCounterWhenIdNotConsecutive() throws Exception {
                         .max(Comparable::compareTo)
                         .orElseThrow();
 
-        try (Connection connection = getConnection();
+        var dataSource = getDataSource();
```
1996fanrui, line 465: This temporary data source isn't closed. 

Or could we reuse the dataSource of `beforeEach` instead of creating a new one?

flink-autoscaler-plugin-jdbc/src/test/java/org/apache/flink/autoscaler/jdbc/event/AbstractJdbcEventInteractorITCase.java
```diff
@@ -46,63 +46,63 @@ void testAllOperations() throws Exception {
 
         // The datetime precision is seconds in MySQL by default.
         var createTime = Instant.now().truncatedTo(ChronoUnit.SECONDS);
-        try (var conn = getConnection()) {
-            var jdbcEventInteractor = new JdbcEventInteractor(conn);
```
1996fanrui, line 50:
```
        try(var jdbcEventInteractor = new JdbcEventInteractor(getDataSource())) {
            
        }
```

It maybe a simpler choice, we don't need to check `dataSource` and close it in the end.

flink-autoscaler-plugin-jdbc/src/test/java/org/apache/flink/autoscaler/jdbc/state/AbstractJdbcStateInteractorITCase.java
```diff
@@ -42,32 +42,35 @@ void testAllOperations() throws Exception {
         var value1 = "value1";
         var value2 = "value2";
         var value3 = "value3";
-        try (var conn = getConnection()) {
-            var jdbcStateInteractor = new JdbcStateInteractor(conn);
```
1996fanrui, line 46:
```
        try(var jdbcStateInteractor = new JdbcStateInteractor(getDataSource())) {
            
        }
```

It maybe a simpler choice, we don't need to check `dataSource` and close it in the end.

flink-autoscaler-plugin-jdbc/src/test/java/org/apache/flink/autoscaler/jdbc/state/JobStateViewTest.java
```diff
@@ -36,25 +35,18 @@
...
+        this.dataSource = getDataSource();
+        this.jdbcStateInteractor = new CountableJdbcStateInteractor(dataSource);
         this.jobStateView = new JobStateView(jdbcStateInteractor, DEFAULT_JOB_KEY);
         jdbcStateInteractor.assertCountableJdbcInteractor(1, 0, 0, 0);
     }
 
-    @AfterEach
-    void afterEach() throws SQLException {
-        if (conn != null) {
-            conn.close();
```
1996fanrui, line 54: Why don't close `jdbcStateInteractor` here?

1996fanrui (commented):
Hi @sharath1709 , thanks for the update!

I left some comments, and the CI is still failed (It cannot be compiled for now. ). 

You could try to execute ` mvn  clean install -DskipTests` in your local before next push, thanks~

flink-autoscaler-plugin-jdbc/src/test/java/org/apache/flink/autoscaler/jdbc/event/AbstractJdbcEventInteractorITCase.java
sharath1709: Mb. Agree.

flink-autoscaler-standalone/src/test/java/org/apache/flink/autoscaler/standalone/AutoscalerStateStoreFactoryTest.java
sharath1709: I see, I wasn't familiar with this and assumed it only closes the connection. We can continue to keep the same logic.

flink-autoscaler-plugin-jdbc/src/test/java/org/apache/flink/autoscaler/jdbc/state/AbstractJdbcStateInteractorITCase.java
sharath1709: updated

flink-autoscaler-plugin-jdbc/src/test/java/org/apache/flink/autoscaler/jdbc/state/JobStateViewTest.java
sharath1709: Thanks, updated.

flink-autoscaler-standalone/src/test/java/org/apache/flink/autoscaler/standalone/AutoscalerEventHandlerFactoryTest.java
```diff
@@ -68,14 +68,14 @@ void testCreateJdbcEventHandler() throws Exception {
...
+        HikariJDBCUtil.getDataSource(conf).close();
 
         var eventHandler = AutoscalerEventHandlerFactory.create(conf);
         assertThat(eventHandler).isInstanceOf(JdbcAutoScalerEventHandler.class);
+        conf.set(JDBC_URL, String.format("%s;shutdown=true", jdbcUrl));
 
-        try {
-            conf.set(JDBC_URL, String.format("%s;shutdown=true", jdbcUrl));
-            HikariJDBCUtil.getConnection(conf).close();
+        try (var datasource = HikariJDBCUtil.getDataSource(conf)) {
```
1996fanrui, line 77: It's similar with the last comment.

1996fanrui (commented):
Thanks @sharath1709 for the update, overall LGTM. I left 2 minor comments, please take a look in your free time, thanks~

> 3. We encountered a bug related to this fix recently. The database master is crashed, the slave is promoted to be the master. The jdbc state store and event handler won't work due to they are using the fixed connection.
> 
>     - As I understand, DataSource could solve this problem as well.
>     - When old connection is crashed, DataSource will create a new physical connection > to connect the new database master server.
>     - Would you mind adding a test to check it? In the ITCase, we only have one databse server, I'm not sure could it be reproduced when the database server is restarted.

I have checked it manually on my Mac, it works well with this PR. Also, as we discussed before, it's better to add a test to check it.

sharath1709:
Thanks @1996fanrui for patience with the review. I addressed the remaining comments and added a test case for db restart. I also verified that the test case fails without the change in this PR.

flink-autoscaler-standalone/src/test/java/org/apache/flink/autoscaler/standalone/AutoscalerEventHandlerFactoryTest.java
sharath1709: good catch, fixed

1996fanrui:
> I addressed the remaining comments and added a test case for db restart to `AbstractJdbcStateInteractorITCase`. The same could be added to `AbstractJdbcEventInteractorITCase` if needed but it's a bit redundant.

Sounds make sense to me, testing restart in AbstractJdbcStateInteractorITCase is enough, but the test is failed. 
https://github.com/apache/flink-kubernetes-operator/actions/runs/12939940745/job/36097810802?pr=929#step:8:1025

Also, I may need to do another round of overall review after the test passes.

sharath1709:
@1996fanrui It appears that MySQL/Postgresql DB restart test cases are failing in CI with SQLTransientException. Let me update the test to ignore transient exception as typically subsequent connection should be successful. We may also consider adding retries but I feel it's overkill. lmk your thoughts

1996fanrui:
> @1996fanrui It appears that MySQL/Postgresql DB restart test cases are failing in CI with SQLTransientException. Let me update the test to ignore transient exception as typically subsequent connection should be successful. Before this PR, the test case produces SQLNonTransientConnectionException. We may also consider adding retries but I feel it's overkill. lmk your thoughts

After I debug it on my Mac, I don't think catch `SQLTransientException` is reasonable.  Please check the following figure, the `bbbbb` is not executed. It means we cannot ensure the connection works after restarting. So I think retry mechanism is needed, we need to retry queryData from data base until the result is expected.

Also, I added the retry mechanism on my Mac, but it still doesn't work, I found after restart, the port of MySQL Container is changed. It means the original data source won't be used.

sharath1709:
That sounds great, let me remove the test. I'm not the expert on this situation either and ChatGPT originally suggested starting and stopping the container to simulate a restart. I will try to explore it in my free time and add the test case later

1996fanrui:
One test fails since ` too many clients already`, I'm not sure is there any client or connection leak?

```
Error:  Tests run: 17, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 33.745 s <<< FAILURE! - in org.apache.flink.autoscaler.jdbc.event.PostgreSQLJdbcAutoscalerEventHandlerITCase
Error:  org.apache.flink.autoscaler.jdbc.event.PostgreSQLJdbcAutoscalerEventHandlerITCase.testEventIntervalWithoutMessageKey  Time elapsed: 1.024 s  <<< ERROR!
com.zaxxer.hikari.pool.HikariPool$PoolInitializationException: Failed to initialize pool: FATAL: sorry, too many clients already
Caused by: org.postgresql.util.PSQLException: FATAL: sorry, too many clients already
```


https://github.com/apache/flink-kubernetes-operator/actions/runs/12958019250/job/36164630541?pr=929#step:8:2293

flink-autoscaler-plugin-jdbc/src/test/java/org/apache/flink/autoscaler/jdbc/testutils/databases/postgres/PostgreSQLExtension.java
```diff
@@ -66,7 +70,8 @@ public void afterAll(ExtensionContext extensionContext) {
 
     @Override
     public void afterEach(ExtensionContext extensionContext) throws Exception {
-        try (var conn = getConnection();
+        try (var dataSource = getDataSource();
+                var conn = getDataSource().getConnection();
```
1996fanrui, line 74:
```suggestion
                var conn = dataSource.getConnection();
```

@sharath1709 I found a bug here, it should use data source instead of ` getDataSource()` here.

The second ` getDataSource()` is not closed. That's why connection is leaked.

https://github.com/apache/flink-kubernetes-operator/pull/929#issuecomment-2614195894

1996fanrui (approved):
Thanks @sharath1709 for the patience!

LGTM. I will rebase master later, and merge it if the CI is green.

PR #930, opened 2025-01-09
[FLINK-36645] [flink-autoscaler] Gracefully handle null execution pla…

sharath1709:
…n from autoscaler

This pull request makes change to gracefully handle null execution plan returned from the Flink Rest API on autoscaler. This scenario can happen when a rescale operation is already in progress.

## Brief change log

  - *Gracefully handle null execution plan from autoscaler*

## Verifying this change

This change added a unit test case to verify the ScalingMetricCollector throws NotReadyException when the execution plan is empty

## Documentation


flink-autoscaler/src/main/java/org/apache/flink/autoscaler/topology/JobTopology.java
```diff
@@ -150,6 +151,11 @@ public static JobTopology fromJsonPlan(
         ObjectNode plan = objectMapper.readValue(jsonPlan, ObjectNode.class);
         ArrayNode nodes = (ArrayNode) plan.get("nodes");
 
+        if (nodes == null || nodes.isEmpty()) {
+            throw new NotReadyException(
+                    new RuntimeException("No nodes found in the plan, job is not ready yet"));
```
gyfora, line 156: Can we maybe add a new constructor to the `NotReadyException` so we don't need to wrap an unnecessary RuntimeException?

sharath1709: Thanks for the quick review! Updated.

PR #936, opened 2025-01-20
[FLINK-37126] Add Validator for Autoscaler

huyuanfeng2018:

Add Validator for Autoscler, Facilitates StandaloneAutoscaler to detect configuration anomalies in advance

## Brief change log

  - Migrate the verification part of Autoscaler in `DefaultAutoscalerValidator` to the Autoscaler module
  - Verify the Autoscaler config separately in StandaloneAutoscaler. If there are illegal parameters, report them through events.

## Verifying this change

  - add `StandaloneAutoscalerValidatorTest` Used to verify parameter verification under standloneAutoscaler
  - The old test cases cover the test for Autoscaler parameter verification under the **operator**

## Documentation

  - If yes, how is the feature documented? (not documented)


huyuanfeng2018:
> Any update on this?

Thanks for @gyfora ping, it's been too long I forgot about this PR, I'll keep pushing forward

huyuanfeng2018:
@gyfora Could you help review this PR when you have time? Thank you very much ~

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/validation/DefaultAutoscalerValidator.java
```diff
@@ -0,0 +1,92 @@
...
+
+import java.util.Optional;
+
+import static org.apache.flink.autoscaler.config.AutoScalerOptions.OBSERVED_SCALABILITY_COEFFICIENT_MIN;
+import static org.apache.flink.autoscaler.config.AutoScalerOptions.UTILIZATION_MAX;
+import static org.apache.flink.autoscaler.config.AutoScalerOptions.UTILIZATION_MIN;
+import static org.apache.flink.autoscaler.config.AutoScalerOptions.UTILIZATION_TARGET;
+
+/** Default implementation of {@link AutoscalerValidator}. */
+public class DefaultAutoscalerValidator implements AutoscalerValidator {
```
gyfora: Instead of making this an interface with a default implementation, can we simply make this a regular class and just use it directly? `AutoscalerValidator` that contains the implementation. If later we want to make it pluggable then we can extract the interface but now this is just boilerplate

gyfora (changes requested):
The changes look good, but I would remove the extra interface added

PR #938, opened 2025-01-30
[FLINK-37236] Flink 2.0 preview support

gyfora:

This PR adds support for the 2.0 preview release of Flink. As a necessary prerequisite it also includes a bumping of the operator Flink dependency to 1.20.

The major incompatibility that was addressed here is related to the YAML config handling. Flink 2.0 only supports the new standardized yaml runtime config while 1.20 and older versions supported both the old and the new.

The operator previously only used the legacy flat yaml config. Now a new logic has been put in place for Flink 2.0 and above to automatically convert and use the new yaml configs.

## Brief change log

  - *Bump Flink dependency to 1.20*
  - *Update examples to 1.20*
  - *Update NOTICE Files*
  - *Update e2e version matrix to include 2.0*
  - *Introduce custom FlinkConfMountDecorator to support standard yaml configs*
  
## Verifying this change

- Existing e2es, 
- [TODO] unit tests for FlinkConfMountDecorator

## Documentation


mxm (approved):

flink-kubernetes-operator-api/pom.xml
```diff
@@ -55,6 +55,18 @@ under the License.
             </exclusions>
         </dependency>
 
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-core-api</artifactId>
+            <version>${flink.version}</version>
+            <exclusions>
```
afedulov, line 62: Maybe a comment why such exclusion is needed?

afedulov (approved):

flink-kubernetes-operator-api/pom.xml
gyfora: This hasn't changed compared to what was before here, we already had the exclusions in the api package as we do not require the transitive dependencies for that. The operator core still pulls the transitive deps as well

PR #940, opened 2025-02-04
FLINK-37179 - remove priority: 0 from CRDs

luismacosta:
Hello @gyfora

Following https://issues.apache.org/jira/browse/FLINK-37179, remove **priority: 0** from CRDs

### DIFF
--- a/helm/flink-kubernetes-operator/crds/flinkdeployments.flink.apache.org-v1.yml
+++ b/helm/flink-kubernetes-operator/crds/flinkdeployments.flink.apache.org-v1.yml
@@ -17,13 +17,11 @@ spec:
     - description: Last observed state of the job.
       jsonPath: .status.jobStatus.state
       name: Job Status
-      priority: 0
       type: string
     - description: "Lifecycle state of the Flink resource (including being rolled\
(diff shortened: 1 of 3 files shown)


gyfora (changes requested):
I don't think that this approach is good. These are generated classes, next time we run the mvn build, those fields will be readded unless we fix it somewhere else

luismacosta:
There is an issue raised with argo-cd https://github.com/argoproj/argo-cd/issues/3889
I will close this PR
Thanks

wwakas:
> Hello @wwakas
> 
> My workaround was to add _ignoreDifferences_ in the AppSet
> 
> ```
>   ignoreDifferences:
>     - group: apiextensions.k8s.io
>       kind: CustomResourceDefinition
>       jsonPointers:
>         - "/spec/versions/0/additionalPrinterColumns/0/priority"
>         - "/spec/versions/0/additionalPrinterColumns[]/priority"
>       jqPathExpressions:
>         - ".spec.versions[].additionalPrinterColumns[].priority"
> ```

Thanks a lot, seems its working!

PR #941, opened 2025-02-04
[FLINK-37253] Add state size in application status and deployment metrics

mxm:
This adds state size, i.e. the size of the last completed checkpoint, to the
deployment status. It also exposes the state size as a deployment metric.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
@@ -73,6 +73,8 @@ import org.apache.flink.runtime.rest.messages.checkpoints.CheckpointStatusHeader
 import org.apache.flink.runtime.rest.messages.checkpoints.CheckpointStatusMessageParameters;
 import org.apache.flink.runtime.rest.messages.checkpoints.CheckpointTriggerHeaders;
 import org.apache.flink.runtime.rest.messages.checkpoints.CheckpointTriggerRequestBody;
+import org.apache.flink.runtime.rest.messages.checkpoints.CheckpointingStatistics;
+import org.apache.flink.runtime.rest.messages.checkpoints.CheckpointingStatisticsHeaders;
 import org.apache.flink.runtime.rest.messages.job.metrics.JobMetricsHeaders;
 import org.apache.flink.runtime.rest.messages.job.savepoints.SavepointDisposalRequest;
 import org.apache.flink.runtime.rest.messages.job.savepoints.SavepointDisposalTriggerHeaders;
@@ -161,6 +163,7 @@ public abstract class AbstractFlinkService implements FlinkService {
     private static final String EMPTY_JAR_FILENAME = "empty.jar";
     public static final String FIELD_NAME_TOTAL_CPU = "total-cpu";
     public static final String FIELD_NAME_TOTAL_MEMORY = "total-memory";
+    public static final String FIELD_NAME_STATE_SIZE = "state-size";
 
     protected final KubernetesClient kubernetesClient;
     protected final ExecutorService executorService;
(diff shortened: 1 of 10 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
```diff
@@ -757,16 +778,35 @@ public Map<String, String> getClusterInfo(Configuration conf) throws Exception {
...
+                                        EmptyRequestBody.getInstance())
+                                .get();
+                CheckpointStatistics.CompletedCheckpointStatistics completedCheckpointStatistics =
+                        checkpointingStatistics
+                                .getLatestCheckpoints()
+                                .getCompletedCheckpointStatistics();
+                if (completedCheckpointStatistics != null) {
+                    clusterInfo.put(
+                            FIELD_NAME_STATE_SIZE,
+                            String.valueOf(completedCheckpointStatistics.getCheckpointedSize()));
```
gyfora, line 806: Will this be any useful for jobs with incremental checkpointing enabled (which is I guess what most people would use with RocksDB )

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkUtils.java
```diff
@@ -403,6 +403,28 @@ public static Long calculateClusterMemoryUsage(Configuration conf, int taskManag
         return tmTotalMemory + jmTotalMemory;
     }
 
+    public static Long calculateClusterStateSize(Configuration conf, int taskManagerReplicas) {
```
gyfora, line 406: Shouldn't this be called totalClusterMemorySize instead of state?

gyfora: Also I don't think this is used anywhere... :)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
mxm: Good question. It is a bit tricky to find this out, even after studying the code. Good news is that there should be a metric that is somewhat approximate to the actual full state / checkpoint size. It was introduced via https://issues.apache.org/jira/browse/FLINK-25557.

Turns out, I was using the wrong metric because of a misleading JavaDoc in CheckpointStatistics. I've updated the PR. In the Rest API, there is `stateSize` and `checkpointedSize`. The former is the full state size, while the latter is the checkpointed size which various depending on whether checkpointing is incremental or not. 

StateSize is based on https://github.com/apache/flink/blob/d0d44d6196c02552179ec23c799d9769e128a8ae/flink-runtime/src/main/java/org/apache/flink/runtime/state/StateObject.java#L66. I'm not 100% sure whether it always yields the exact state size but it should be close.

CheckpointedSize is based on https://github.com/apache/flink/blob/d0d44d6196c02552179ec23c799d9769e128a8ae/flink-runtime/src/main/java/org/apache/flink/runtime/state/CompositeStateHandle.java#L62.

I wonder whether we should expose both?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkUtils.java
mxm: State size or checkpoint size isn't directly related to the cluster memory size. For the heap memory backend, we would expect the state size to be lower than the overall memory. For RocksDB, it could even exceed the cluster memory.

gyfora: Ok but I still don't get 3 things:
 - Where is this used?
 - Why do we need this bad approximation if state size metrics are available from Flink?
 - This is basically just total memory, why do we call it state size?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
gyfora: I think we should only expose one. Checkpointed size is not very useful in this context but instead we should aim to somehow get the actual state size.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkUtils.java
mxm: I didn't see your comment in https://github.com/apache/flink-kubernetes-operator/pull/941#discussion_r1943008300, it was somehow hidden when I replied.

Sorry, this code was unused code. I have removed it.

gyfora (approved):
If the manually tested for correctness for the supported Flink versions and e2es pass then good to go

mxm:
Tried it out on a local k8s cluster with various Flink versions:

image
image
image
image

PR #948, opened 2025-02-21
[FLINK-37370] [Observer] Finished batch jobs throw ReconciliationException and never reach FINISHED in the CR

luca-p-castelli:

This pull request fixes a bug where in the observation phase, the operator tries to observe savepoint information for batch jobs and fails since checkpointing is not enabled for batch jobs.

More information in: https://issues.apache.org/jira/browse/FLINK-37370

## Brief change log

1. Modifies JobDetailsInfo and downstream tests to include job-type
2. Adds a method to FlinkService + AbstractFlinkService to fetch JobDetailsInfo for the given job
3. Adds a method to SnapshotObserver to check if a job is a batch job. This method is used to skip observing checkpoints/savepoints for batch jobs.

### Some questions

1. There is a local version of [JobDetailsInfo](https://github.com/apache/flink-kubernetes-operator/blob/main/flink-autoscaler/src/main/java/org/apache/flink/runtime/rest/messages/job/JobDetailsInfo.java) in the operator code. Based on the existing comment, saying it can be removed when the client upgrades to 1.20, I tried to remove the local version, but ran into issues with the non-null requirement for slotSharingGroupId and jobType - particularly with the flink15 and flink16 compatibility tests. I've kept the local version and added job-type. Open to suggestions if you think there is a better way to handle this.
2. When running a batch job there are still some exceptions being logged as warnings that originate in [populateStateSize](https://github.com/apache/flink-kubernetes-operator/blob/main/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java#L783) - also related to checkpointing not being enabled for batch jobs. I didn't change anything there since those warnings/exceptions don't crash anything. Open to suggestions if you think we shouldn't leave this as is.
3. [JobAutoScalerImplTest.testMetricReporting](https://github.com/apache/flink-kubernetes-operator/blob/main/flink-autoscaler/src/test/java/org/apache/flink/autoscaler/JobAutoScalerImplTest.java#L87) seemed flaky when running tests locally. Has this been observed before?

## Verifying this change

- All existing tests pass (modified a couple tests to include job-type)
- Tested locally with the manifest attached to the issue to make sure the job successfully reaches FINISHED JOB STATUS

If you think my approach is good, as a next step, happy to add some tests for the new logic I've introduced. Where would you like to see tests?

## Documentation

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/SnapshotObserver.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/SnapshotObserver.java
@@ -37,9 +37,7 @@ import org.apache.flink.kubernetes.operator.utils.ConfigOptionUtils;
 import org.apache.flink.kubernetes.operator.utils.EventRecorder;
 import org.apache.flink.kubernetes.operator.utils.FlinkStateSnapshotUtils;
 import org.apache.flink.kubernetes.operator.utils.SnapshotUtils;
-import org.apache.flink.runtime.rest.util.RestClientException;
 import org.apache.flink.util.CollectionUtil;
-import org.apache.flink.util.ExceptionUtils;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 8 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/SnapshotObserver.java
```diff
@@ -83,6 +84,11 @@ public void observeSavepointStatus(FlinkResourceContext<CR> ctx) {
         var jobStatus = resource.getStatus().getJobStatus();
         var jobId = jobStatus.getJobId();
 
+        if (isBatchJob(ctx, jobId)) {
+            LOG.debug("Skipping checkpoint observation for BATCH job");
+            return;
```
gyfora, line 89: It would be good to avoid this extra rest api call all the time and also if in the future batch jobs can also take a checkpoint (which is where Flink is going) this is going to have to be rewritten.

Could we instead catch the checkpoint error in the `observeLatestCheckpoint` method and do this check then?

Also please add a test for this new logic in the observer

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/SnapshotObserver.java
```diff
@@ -441,21 +443,33 @@ private long getMaxCountForSnapshotType(
...
+                                if (ReconciliationUtils.isJobCancelled(status)) {
+                                    // For cancelled jobs the observed savepoint is always definite,
+                                    // so if empty we know the job doesn't have any
+                                    // checkpoints/savepoints
+                                    jobStatus.setUpgradeSavepointPath(null);
+                                }
+                            });
+        } catch (Exception e) {
+            if (ExceptionUtils.findThrowable(e, RestClientException.class)
+                    .map(ex -> ex.getMessage().contains("Checkpointing has not been enabled"))
```
gyfora, line 464: Is message always not null here?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/SnapshotObserver.java
```diff
@@ -441,21 +443,33 @@ private long getMaxCountForSnapshotType(
...
+                    .orElse(false)) {
+                LOG.warn(
+                        "Checkpointing not enabled for job {}, skipping checkpoint observation",
+                        jobId,
+                        e);
```
gyfora, line 469: I wonder if the try/catch logic should be part of `getLastCheckpoint` of the flink service. That would mean that anywhere else we call this in the future we get a consistently good behaviour for batch jobs

gyfora (changes requested):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/SnapshotObserver.java
luca-p-castelli: Good point. I'll add logic to check for null.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/SnapshotObserver.java
luca-p-castelli: Yeah, I thought about that. I'm open to either. We would just have `getLastCheckpoint` catch the exception and return `optional.empty()`

gyfora: I think that would make sense 👍

gyfora (approved):

PR #949, opened 2025-02-24
[FLINK-37372] Fix infinite loop bug in savepoint error handling

gyfora:

A bug in the ExceptionUtils#findThrowableSerializedAware can cause it to enter an infinite loop if it cannot deserialize the SerializedThrowable error.

This affects the Kubernetes operator and can cause it to hang indefinitely.

## Brief change log

 - Eliminate bug in SerializedThrowable parsing logic

## Verifying this change

Unit test added

## Documentation

no


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/ExceptionUtils.java
```diff
@@ -0,0 +1,60 @@
...
+
+package org.apache.flink.kubernetes.operator.utils;
+
+import org.apache.flink.types.DeserializationException;
+import org.apache.flink.util.SerializedThrowable;
+
+import java.util.Optional;
+
+/** Exception utils. * */
+public class ExceptionUtils {
```
afedulov, line 26: I wonder if it is worth porting this class instead of simply adding a wrapper that handles the `deserialized == t` bit on the return value of the existing util method.

gyfora: I would like to avoid jumping through extra complexity here. I will submit a similar patch to Flink core itself then we can simply delete this utility method in the future. I think this is pretty easy and straightforward.

Wdyt?

afedulov: sounds good to me if thats the plan.

afedulov (approved):

csviri (approved):

PR #950, opened 2025-02-27
[FLINK-33525] Move ImpulseSource to new Source API

Poorvankbhatia:

This pull request migrates the ImpulseSource (used in LoadSimulationPipeline) to use Flink's DataGeneratorSource, replacing the older SourceFunction.

## Brief change log

- Replaced the old ImpulseSource (which implemented SourceFunction<Long>) with a new implementation using Flink's DataGeneratorSource.
- Updated LoadSimulationPipeline to use env.fromSource(...) instead of env.addSource(...).
- Ensured that the new implementation preserves the same functionality as the old one.

## Verifying this change

This change is already covered by existing tests, such as:

- Existing load simulation tests that verify the source is producing expected records.
- The Flink job runs with the updated ImpulseSource without functional changes.
- The change was manually verified by running a Flink job with the new ImpulseSource.

## Documentation


afedulov (commented):
@Poorvankbhatia thanks for the PR. Did you evaluate if this can be achieved based off of the existing `DataGeneratorSource`/`GeneratorFunction`? Generally, our approach should be to try to capture usage patterns that might also be required for generating synthetic data by Flink users and make them available by providing utilities in the `generator` package.

Poorvankbhatia:
> @Poorvankbhatia thanks for the PR. Did you evaluate if this can be achieved based off of the existing `DataGeneratorSource`/`GeneratorFunction`? Generally, our approach should be to try to capture usage patterns that might also be required for generating synthetic data by Flink users and make them available by providing utilities in the `generator` package.

Yes, absolutely! However, using DataGeneratorSource requires adding a dependency on flink-connector-datagen in autoscaling's pom.xml. I wasn't sure if that was acceptable, so I replicated the original source instead.

afedulov:
> However, using DataGeneratorSource requires adding a dependency on flink-connector-datagen in autoscaling's pom.xml.

I think it is fine - datagen's only major transitive dependency is on `flink-core` and which is already present in the operator. 
@gyfora what do you think?

gyfora:
Since this is only an example we should focus on simplicity and add the extra dep to use the built in generator functionality.

Poorvankbhatia:
> > However, using DataGeneratorSource requires adding a dependency on flink-connector-datagen in autoscaling's pom.xml.
> 
> I think it is fine - datagen's only major transitive dependency is on `flink-core` which is already present in the operator. @gyfora what do you think?

Updated the code. Please have a look. Thanks.

examples/autoscaling/src/main/java/autoscaling/LoadSimulationPipeline.java
```diff
@@ -74,8 +78,38 @@ public static void main(String[] args) throws Exception {
         for (String branch : maxLoadPerTask.split("\n")) {
             String[] taskLoads = branch.split(";");
 
+            /*
+             * Creates an unbounded stream that continuously emits the constant value 42L.
+             * Flink's DataGeneratorSource with RateLimiterStrategy is used to control the emission rate.
+             *
+             * Rate Calculation:
+             * - samplingIntervalMs / 10 gives maxSleepTimeMs, which represents the interval between emissions.
```
afedulov, line 86: I believe this section needs to be adjusted. Since we do not control the maxSleepTimeMs directly, but it is rather controlled internally by the Guava's token bucket algorithm, it should not be explicitly mentioned here. It is also not strictly guaranteed to be max - at the startup there can be a short burst. The value 10 seems to just be hardcoded value meaning that we want at least 10 impulses per sampling interval. Basically it should rather explain: check how many sampling intervals are there within a second, make sure that 10 impulses are generated for each sampling interval (IMPULSES_PER_SAMPLING_INTERVAL).

Poorvankbhatia: Updated the comment. Let me know if this makes sense.

examples/autoscaling/src/main/java/autoscaling/LoadSimulationPipeline.java
```diff
@@ -82,20 +82,32 @@ public static void main(String[] args) throws Exception {
...
+             *      - `samplingIntervalsPerSecond = 1000 / 1000 = 1`
+             *      - `impulsesPerSecond = 10 * 1 = 10 records per second`
+             * - If `samplingIntervalMs = 500 ms`:
+             *      - `samplingIntervalsPerSecond = 1000 / 500 = 2`
+             *      - `impulsesPerSecond = 10 * 2 = 20 records per second`
+             * - If `samplingIntervalMs = 2000 ms`:
+             *      - `samplingIntervalsPerSecond = 1000 / 2000 = 0.5`
+             *      - `impulsesPerSecond = 10 * 0.5 = 5 records per second`
+             *
+             * This approach ensures that the number of records emitted dynamically scales
```
afedulov, line 108: This sounds very AI-generated :) I the intention is clear without the last paragraph. 
Also, one example would suffice.

examples/autoscaling/src/main/java/autoscaling/LoadSimulationPipeline.java
```diff
@@ -74,8 +78,50 @@ public static void main(String[] args) throws Exception {
...
+             */
             DataStream<Long> stream =
-                    env.addSource(new ImpulseSource(samplingIntervalMs)).name("ImpulseSource");
+                    env.fromSource(
+                            new DataGeneratorSource<>(
+                                    (GeneratorFunction<Long, Long>)
+                                            (index) -> 42L, // Emits constant value 42
+                                    Long.MAX_VALUE, // Unbounded stream
+                                    RateLimiterStrategy.perSecond(
+                                            (double) 1000
```
afedulov, line 119: More importantly, actually adjust the calculation to align with the description
a/b/c -> (a/b)*c
c: 10 -> IMPULSES_PER_SAMPLING_INTERVAL

Poorvankbhatia: Corrected this.

examples/autoscaling/src/main/java/autoscaling/LoadSimulationPipeline.java
```diff
@@ -74,8 +78,37 @@ public static void main(String[] args) throws Exception {
...
+             *   exactly 10 impulses should be generated.
+             *
+             * To calculate the total number of records emitted per second:
+             * 1. Determine how many sampling intervals fit within one second:
+             *      samplingIntervalsPerSecond = 1000 / samplingIntervalMs;
+             * 2. Multiply this by the number of impulses per interval to get the total rate:
+             *      impulsesPerSecond = IMPULSES_PER_SAMPLING_INTERVAL * samplingIntervalsPerSecond;
+             *
+             * Example:
+             * - If `samplingIntervalMs = 500 ms` and `IMPULSES_PER_SAMPLING_INTERVAL = 10`:
```
afedulov, line 98: nit: either introduce this constant and use it in the calculations or use plain text in the comment.

examples/autoscaling/src/main/java/autoscaling/LoadSimulationPipeline.java
```diff
@@ -74,8 +78,37 @@ public static void main(String[] args) throws Exception {
...
             String[] taskLoads = branch.split(";");
 
+            /*
+             * Creates an unbounded stream that continuously emits the constant value 42L.
+             * Flink's DataGeneratorSource with RateLimiterStrategy is used to control the emission rate.
+             *
+             * Emission Rate Logic:
+             * - The goal is to generate a fixed number of impulses per sampling interval.
+             * - `samplingIntervalMs` defines the duration of one sampling interval in milliseconds.
+             * - We define `IMPULSES_PER_SAMPLING_INTERVAL = 10`, meaning that for every sampling interval,
```
afedulov, line 88: nit: either introduce this constant and use it in the calculations or use plain text in the comment.

Poorvankbhatia: Added the constant.

afedulov (approved):
Looks good, thanks @Poorvankbhatia

PR #952, opened 2025-03-04
Draft PR for plugin based approach for evaluation hook for scaling metric evaluation

pchoudhury22:
This PR is a prototype implementation for review and feedback.


pchoudhury22:
There is another evaluated metric, EXPECTED_PROCESSING_RATE, which gets computed at JobVertexScaler.computeScaleTargetParallelism .
Since we calculate this first and use it for calculating new parallelism and then put it, did not add the evaluator hook to it. Let me know if we would want to have hook for EXPECTED_PROCESSING_RATE aswell? in that case would it make sense to move calculation of new parallelism after this metric is put to evaluated metric?

PR #944, opened 2025-02-18
[FLINK-37320] [Observer] FINISHED finite streaming jobs incorrectly being set to RECONCILING

luca-p-castelli:
## What is the purpose of the change

This pull request fixes a bug where the observer tries changing the JOB STATUS of a FINISHED finite/bounded streaming job back to RECONCILING if the JM is MISSING.

## Brief change log
Wraps [AbstractFlinkDeploymentObserver.observeJmDeployment](https://github.com/apache/flink-kubernetes-operator/blob/main/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/AbstractFlinkDeploymentObserver.java#L155) in an if-statement that only executes if the job is NOT in a terminal state.

## Verifying this change

- Tested locally
- Unit test to guard against accidental future regression

## Does this pull request potentially affect one of the following parts:
  - Dependencies (does it add or upgrade a dependency): No
  - The public API, i.e., is any changes to the `CustomResourceDescriptors`: No
  - Core observer or reconciler logic that is regularly executed: Yes

## Documentation
  - Does this pull request introduce a new feature? No
  - If yes, how is the feature documented? not applicable

PR #955, opened 2025-03-12
[FLINK-37370] [Observer] Fix exception caught when handling checkpointing not enabled for batch jobs and add batch e2e test

luca-p-castelli:

This pull request is a follow-up to https://github.com/apache/flink-kubernetes-operator/pull/948. It corrects the exception that is caught from `org.apache.flink.runtime.rest.util.RestClientException` to `java.util.concurrent.ExecutionException`.

To make batch testing more robust, I have also introduced a batch e2e test with this PR.

## Brief change log

* Updates the Exception and corresponding test
* Introduces a batch e2e test for the operator that uses the bundled Flink WordCount example

## Verifying this change

* Existing tests
* Newly added batch e2e test
* Tested locally with Minikube and the batch manifest in the JIRA ticket

## Documentation

### DIFF
--- a/e2e-tests/data/flinkdep-batch-cr.yaml
+++ b/e2e-tests/data/flinkdep-batch-cr.yaml
@@ -31,10 +31,6 @@ spec:
       nginx.ingress.kubernetes.io/rewrite-target: "/$2"
   flinkConfiguration:
     taskmanager.numberOfTaskSlots: "2"
-    high-availability.type: kubernetes
-    high-availability.storageDir: file:///opt/flink/volume/flink-ha
-    state.checkpoints.dir: file:///opt/flink/volume/flink-cp
-    state.savepoints.dir: file:///opt/flink/volume/flink-sp
     kubernetes.operator.snapshot.resource.enabled: "false"
   serviceAccount: flink
   podTemplate:
@@ -59,14 +55,9 @@ spec:
           volumeMounts:
             - mountPath: /opt/flink/usrlib
               name: flink-artifact
-            - mountPath: /opt/flink/volume
-              name: flink-volume
       volumes:
         - name: flink-artifact
           emptyDir: { }
-        - name: flink-volume
-          persistentVolumeClaim:
-            claimName: flink-example-wordcount-batch
   jobManager:
     resource:
       memory: "1024m"
@@ -82,19 +73,6 @@ spec:
     parallelism: 2
     upgradeMode: stateless
   mode: native
-
----
-apiVersion: v1
-kind: PersistentVolumeClaim
-metadata:
-  name: flink-example-wordcount-batch
-spec:
-  accessModes:
-    - ReadWriteOnce
-  volumeMode: Filesystem
-  resources:
-    requests:
-      storage: 1Gi
 
 ---
 apiVersion: networking.k8s.io/v1
--- a/e2e-tests/data/flinkdep-batch-cr.yaml
+++ b/e2e-tests/data/flinkdep-batch-cr.yaml
@@ -1,0 +1,109 @@
+################################################################################
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+# limitations under the License.
+################################################################################
+
+apiVersion: flink.apache.org/v1beta1
+kind: FlinkDeployment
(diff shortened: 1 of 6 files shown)


gyfora:
Thanks for the fix, I think we should include the e2e with this PR to make sure it's solid now :)

luca-p-castelli:
> Thanks for the fix, I think we should include the e2e with this PR to make sure it's solid now :)

I've updated the PR with the e2e. Let me know what you think.

e2e-tests/data/flinkdep-batch-cr.yaml
```diff
@@ -0,0 +1,109 @@
...
+    template: "/{{namespace}}/{{name}}(/|$)(.*)"
+    className: "nginx"
+    annotations:
+      nginx.ingress.kubernetes.io/rewrite-target: "/$2"
+  flinkConfiguration:
+    taskmanager.numberOfTaskSlots: "2"
+    high-availability.type: kubernetes
+    high-availability.storageDir: file:///opt/flink/volume/flink-ha
+    state.checkpoints.dir: file:///opt/flink/volume/flink-cp
+    state.savepoints.dir: file:///opt/flink/volume/flink-sp
```
gyfora, line 37: Do we need HA/checkpoint configs? We could remove this and also simplify the pod template

e2e-tests/test_batch_job.sh
```diff
@@ -0,0 +1,60 @@
...
+wait_for_jobmanager_running $CLUSTER_ID $TIMEOUT
+
+# Wait for the job to reach the FINISHED state (first check).
+wait_for_status $APPLICATION_IDENTIFIER '.status.jobStatus.state' FINISHED $TIMEOUT || exit 1
+
+echo "Job reached FINISHED state.  Sleeping for $SLEEP_DURATION seconds..."
+sleep "$SLEEP_DURATION"
+
+# Verify the job is *still* in the FINISHED state (second check).
+wait_for_status $APPLICATION_IDENTIFIER '.status.jobStatus.state' FINISHED $TIMEOUT || exit 1
```
gyfora, line 50: instead of sleeping and checking again, would it make sense to maybe send in a spec change and make sure the operator re-triggers the batch job correctly? I think that would rule out the savepoint observe errors etc.

gyfora (commented):
Looks good, added some minor comments , could you verify that without the fix the test fails?

e2e-tests/data/flinkdep-batch-cr.yaml
```diff
@@ -0,0 +1,109 @@
...
+  taskManager:
+    resource:
+      memory: "1Gi"
+      cpu: 0.5
+  job:
+    jarURI: local:///opt/flink/usrlib/myjob.jar
+    entryClass: org.apache.flink.streaming.examples.wordcount.WordCount
+    args: ["--execution-mode", "BATCH"]
+    parallelism: 2
+    upgradeMode: stateless
```
gyfora, line 83: Should we set this to last-state?
I feel that with stateless we won't be able to verify that the savepoint observe logic is successful

e2e-tests/test_batch_job.sh
luca-p-castelli: The idea behind the sleeping and checking again came from the fact that in https://github.com/apache/flink-kubernetes-operator/pull/944, I fixed a bug where a bounded streaming job would reach finished and then would be set back to reconciling. 

I do like the idea of adding a spec change and verifying the batch job reruns and finishes though.

e2e-tests/data/flinkdep-batch-cr.yaml
luca-p-castelli: We could. Although the operator logic still reaches `getLastCheckpoint` even in `stateless` mode. Does last-state even make sense for batch jobs given they don't support checkpointing?

luca-p-castelli:
Thanks for the review @gyfora. I've made some updates based on your comments. 

I have also:

1. Made sure the e2e test fails without the fix (i.e. fails with what's currently in main)
image

2. I had to change the flink-examples JAR to a newer one since `--execution-mode` wasn't available in `"https://repo1.maven.org/maven2/org/apache/flink/flink-examples-streaming_2.12/1.14.4/flink-examples-streaming_2.12-1.14.4.jar"`

e2e-tests/data/flinkdep-batch-cr.yaml
gyfora: I think you are right, we could leave this in stateless. To avoid sleeping and make sure that is reconciled and stays in finished we can simply add a no-op spec change (such as change some operator config that doesn't result in a new job depoyment) or changing of the upgrade mode

luca-p-castelli:
Full CI test matrix just finished running for our internal fork. 

The batch tests for Flink versions 1.18 and under failed with the below error. I think it would be ok to just the run the batch tests for versions 1.19+. Thoughts?

```
Error: m2025-03-14 13:09:31,565 o.a.f.k.o.o.JobStatusObserver  [ERROR][default/flink-example-wordcount-batch] Job 063b25ead9e781bfaad8a07e2b587f38 failed with error: java.lang.NoSuchMethodError: 'org.apache.flink.streaming.api.datastream.DataStreamSource org.apache.flink.streaming.api.environment.StreamExecutionEnvironment.fromData(java.lang.Object[])
```

e2e-tests/data/flinkdep-batch-cr.yaml
luca-p-castelli: @gyfora Makes sense. Updated.

PR #956, opened 2025-03-13
[FLINK-36876] Add proxy for restClient

chenyuzhi459:

Add proxy for restClient, which supports to use shared `EventLoopGroup`

## Brief change log

- *Introduce a new class `RestClientProxy` as the proxy of restClient*
- *Introduce a new configuration: `kubernetes.operator.flink.client.io.threads` to specify the total thread of sharded eventLoopGroup*
- *RestClientProxy supports to use sharded eventLoopGroup*

## Verifying this change

## Documentation

### DIFF
--- a/docs/layouts/shortcodes/generated/kubernetes_operator_config_configuration.html
+++ b/docs/layouts/shortcodes/generated/kubernetes_operator_config_configuration.html
@@ -154,7 +154,7 @@
         </tr>
         <tr>
             <td><h5>kubernetes.operator.flink.client.io.threads</h5></td>
-            <td style="word-wrap: break-word;">60</td>
+            <td style="word-wrap: break-word;">10</td>
             <td>Integer</td>
             <td>The maximum number of io threads used by the flink rest client.</td>
         </tr>
--- a/docs/layouts/shortcodes/generated/kubernetes_operator_config_configuration.html
+++ b/docs/layouts/shortcodes/generated/kubernetes_operator_config_configuration.html
@@ -153,6 +153,12 @@
             <td>The timeout for the observer to wait the flink rest client to return.</td>
         </tr>
         <tr>
+            <td><h5>kubernetes.operator.flink.client.io.threads</h5></td>
+            <td style="word-wrap: break-word;">60</td>
+            <td>Integer</td>
+            <td>The maximum number of io threads used by the flink rest client.</td>
+        </tr>
+        <tr>
             <td><h5>kubernetes.operator.health.canary.resource.timeout</h5></td>
             <td style="word-wrap: break-word;">1 min</td>
             <td>Duration</td>
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/RestClientProxy.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/RestClientProxy.java
@@ -1,0 +1,111 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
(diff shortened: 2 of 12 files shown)


gyfora (commented):
Makes sense on a high level, reflection is not great but it's okay if we have a better solution in newer Flink versions.

Which Flink version will contain the Fix for this?

chenyuzhi459:
It depends on when will the pr https://github.com/apache/flink/pull/25788 could be merged. 

If necessary, I can submit the pr fix in both 1.20 and 2.x versions of Flink.

@gyfora

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/RestClientProxy.java
```diff
@@ -0,0 +1,110 @@
...
+            bootstrapField.setAccessible(true);
+            terminationFutureField.setAccessible(true);
+            groupField.setAccessible(true);
+
+            this.terminationFuture = (CompletableFuture<Void>) terminationFutureField.get(this);
+            this.bootstrap = (Bootstrap) bootstrapField.get(this);
+            this.groupField = groupField;
+
+            // close previous group
+            bootstrap.config().group().shutdown();
```
gyfora, line 74: How will this actually solve the original problem? The rest client still creates an individual group (that we close here) won't that cause the same issue anyways?

gyfora (commented):
Added a minor clarifying question, also please rebase to retrigger the CI

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/RestClientProxy.java
chenyuzhi459: Because PoolThreadCache is a request-thread-related cache, it is created and initialized when a request is made using the NioEventLoopGroup; 

here it is created and closed immediately, with no requests in between, so there is no PoolThreadCache leakage issue.

gyfora (approved):

chenyuzhi459:
Could you help take a look at the latest CI failure issue, which doesn't seem to be related to this code commit?

gyfora:
Seems to be consistently broken, did you manage to build/test locally?

gaborgsomogyi:
This feature looks promising. Can you plz answer because we're intended to do a release soon?

chenyuzhi459:
I have reproduced the `Too many open files` issue of CI at local compile env,  I will resolved it as soon as possible

chenyuzhi459:
I think the default value of `flink.client.io.threads=60`  new parameter  is too big in the test environment, because multiple FlinkResourceContextFactory and flinkClientEventLoopGroup will be created simultaneously in parallel tests, which leads to too many files; but there is only one FlinkResourceContextFactory/flinkClientEventLoopGroup in production environment; 

Is there any way to differentiate between the test and the production environment for this param, casue I am not very familiar with the test of operator, I hope it can be helped

chenyuzhi459:
Because `flink.client.io.threads=60` specifies the number of EventExecutors that handle event in NioEventLoopGroup, when creating NioEventLoopGroup, a corresponding number of EventExecutors will be created in the construction method of the parent class MultithreadEventExecutorGroup.

In MultithreadEventExecutorGroup class, the implementation of EventExecutor is NioEventLoop.  When NioEventLoop is created, temporary files will be created through openSelector method.

image

docs/layouts/shortcodes/generated/kubernetes_operator_config_configuration.html
```diff
@@ -152,6 +152,12 @@
             <td>Duration</td>
             <td>The timeout for the observer to wait the flink rest client to return.</td>
         </tr>
+        <tr>
+            <td><h5>kubernetes.operator.flink.client.io.threads</h5></td>
+            <td style="word-wrap: break-word;">10</td>
+            <td>Integer</td>
+            <td>The maximum number of io threads used by the flink rest client.</td>
```
davidradl, line 159: it would be worth detailing what the default is. Can we say there is no maximum number of io thread?

davidradl: nit: io -> I/O

gyfora:
closing stale work / PR

PR #957, opened 2025-03-20
[FLINK-33634] Add Conditions to Flink CRD's Status field

lajith2006:

This PR address the [FLIP](https://cwiki.apache.org/confluence/display/FLINK/FLIP-493%3A+Add+K8S+conditions+to+CRD%27s+Status) to add conditions to flink deployment status related to issue https://issues.apache.org/jira/browse/FLINK-33634

## Brief change log

  - Added list of conditions to `FlinkDeploymentStatus`
  - A new class `ConditionUtils` introduced as utility class to build condition. 

## Verifying this change

  - Deployed and validated  session job and application custer deployment. 
  - Validated the status transitions. 

## Documentation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,5 @@ buildNumber.properties
 
 # VSCode settings
 .vscode/
+.metals
+.bloop
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentController.java
@@ -166,8 +166,10 @@ public class FlinkDeploymentController
             throw new ReconciliationException(e);
         }
 
-        flinkApp.getStatus()
-                .setConditions(ConditionUtils.createConditionFromStatus(flinkApp.getStatus()));
+        var conditions = ConditionUtils.createConditionFromStatus(flinkApp.getStatus());
+        LOG.debug("Setting conditions: {}", conditions);
+        flinkApp.getStatus().setConditions(conditions);
+        LOG.debug("Conditions after set: {}", flinkApp.getStatus().getConditions());
         LOG.debug("End of reconciliation");
         statusRecorder.patchAndCacheStatus(flinkApp, ctx.getKubernetesClient());
         return ReconciliationUtils.toUpdateControl(


davidradl (changes requested):
remove comment

davidradl:
@lajith2006 this implementation does not match the Flip words and has not been discussed in the slack thread, please can you update the discussion thread to ensure that you get consensus, and confirm that the Flip vote still is valid after this change.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/FlinkDeploymentStatus.java
```diff
@@ -55,4 +60,175 @@ public class FlinkDeploymentStatus extends CommonStatus<FlinkDeploymentSpec> {
...
+                case MISSING:
+                case ERROR:
+                case DEPLOYING:
+                    phase = "Pending";
+                    break;
+            }
+        } else if (getJobStatus() != null && getJobStatus().getState() != null) {
+            switch (getJobStatus().getState()) {
+                case RECONCILING:
+                    phase = "Pending";
```
davidradl, line 190: why is this one not RECONCILING like the pattern the others follow others? I suggest a comment, also a constant is better then a an inline literal.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/FlinkDeploymentStatus.java
```diff
@@ -55,4 +60,175 @@ public class FlinkDeploymentStatus extends CommonStatus<FlinkDeploymentSpec> {
...
+    public List<Condition> getConditions() {
+        if (reconciliationStatus != null
+                && reconciliationStatus.deserializeLastReconciledSpec() != null
+                && reconciliationStatus.deserializeLastReconciledSpec().getJob() == null) {
+            switch (jobManagerDeploymentStatus) {
+                case READY:
+                    updateCondition(
+                            conditions,
+                            ConditionUtils.runningTrue(
+                                    "JobManager is running and ready to receive REST API call",
```
davidradl, line 78: nit: call -> calls

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/FlinkDeploymentStatus.java
```diff
@@ -55,4 +60,175 @@ public class FlinkDeploymentStatus extends CommonStatus<FlinkDeploymentSpec> {
...
+    /** Condition of the CR . */
+    private List<Condition> conditions = new ArrayList<>();
+
+    private String phase;
+
+    public List<Condition> getConditions() {
+        if (reconciliationStatus != null
+                && reconciliationStatus.deserializeLastReconciledSpec() != null
+                && reconciliationStatus.deserializeLastReconciledSpec().getJob() == null) {
+            switch (jobManagerDeploymentStatus) {
```
davidradl, line 73: can we have a datastructure, keyed off the jobManagerDeploymentStatus, that we interrogate to get the inserts for the updateConditions.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/FlinkDeploymentStatus.java
```diff
@@ -55,4 +60,175 @@ public class FlinkDeploymentStatus extends CommonStatus<FlinkDeploymentSpec> {
...
+                            conditions,
+                            ConditionUtils.runningFalse(
+                                    JobStatus.FAILING.name(), "Job has failed"));
+                    break;
+                case RESTARTING:
+                    updateCondition(
+                            conditions,
+                            ConditionUtils.runningFalse(
+                                    JobStatus.RESTARTING.name(),
+                                    "The job is currently undergoing a restarting"));
```
davidradl, line 139: nit:  "The job is currently undergoing a restarting" ->  "The job is currently restarting"

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/FlinkDeploymentStatus.java
```diff
@@ -55,4 +60,175 @@ public class FlinkDeploymentStatus extends CommonStatus<FlinkDeploymentSpec> {
...
     /** Information about the TaskManagers for the scale subresource. */
     private TaskManagerInfo taskManager;
+
+    /** Condition of the CR . */
+    private List<Condition> conditions = new ArrayList<>();
+
+    private String phase;
+
+    public List<Condition> getConditions() {
+        if (reconciliationStatus != null
```
davidradl, line 70: I am curious which parts of the code refer to application mode and which are session cluster. It would be good to have some comments to detail this and maybe use the mode name in method name or variable names to make this more intuitive to read.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/FlinkDeploymentStatus.java
```diff
@@ -55,4 +60,175 @@ public class FlinkDeploymentStatus extends CommonStatus<FlinkDeploymentSpec> {
...
+                            ConditionUtils.runningFalse(
+                                    "JobManager deployment not found",
+                                    "JobManager deployment not found"));
+                    break;
+                case DEPLOYING:
+                    updateCondition(
+                            conditions,
+                            ConditionUtils.runningFalse(
+                                    "JobManager process is starting up",
+                                    "JobManager process is starting up"));
```
davidradl, line 93: I am curious why the message and reaon is the same . I would expect them to be different. In this case I would expect the reason to be "A New JVM deployment exists and is being created" - i.e. the word on the arrow in the UML.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentControllerTest.java
```diff
@@ -448,6 +465,15 @@ public void verifyUpgradeFromSavepointLegacyMode(FlinkVersion flinkVersion) thro
         assertEquals(
                 "savepoint_1", appCluster.getStatus().getJobStatus().getUpgradeSavepointPath());
 
+        // Validate status conditions
```
davidradl, line 468: can we use a java parameterized test (or more than one as appropriate)  to cover all the permutations of the tests.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/utils/ConditionUtils.java
```diff
@@ -0,0 +1,51 @@
...
+package org.apache.flink.kubernetes.operator.api.utils;
+
+import io.fabric8.kubernetes.api.model.Condition;
+import io.fabric8.kubernetes.api.model.ConditionBuilder;
+
+import java.text.SimpleDateFormat;
+import java.util.Date;
+
+/**
+ * Creates a condition object with the specified parameters.
```
davidradl, line 27: nit: with the specified parameters -> whether this is running , the message and the reason.
then for return just say  @return A condition

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/utils/ConditionUtils.java
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * @return A condition object with the type, message, status, reason and timestamp.
+ */
+public class ConditionUtils {
+    public static Condition runningTrue(final String message, final String reason) {
+        return crCondition("Running", "True", message, reason);
```
davidradl, line 33: no need to pass "Running" as a parameter, we can hard code that in the method,

nit: rename crCondition -> crRunningCondition

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/utils/ConditionUtils.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    public static Condition runningFalse(final String message, final String reason) {
+        return crCondition("Running", "False", message, reason);
```
davidradl, line 37: are there constants / enums for "False" and "True" we can use?

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/utils/ConditionUtils.java
lajith2006: I would like to keep method name as  `crCondition` rather `crRunningCondition` , just because , `crCondition' will keep it as more generic , so that if in case,  if we need to address any other condition with `type`  other than `Running` , we just need to pass parameter to build respective condition.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/utils/ConditionUtils.java
lajith2006: I haven't see any existing enum/constants for  "False" / "True , instead looks like used as string itself in wherever required in existing code.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/FlinkDeploymentStatus.java
lajith2006: As per description [here] (https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties)  which says  `Reason is intended to be a one-word, CamelCase representation of the category of cause of the current status, and Message is intended to be a human-readable phrase or sentence`, I would use the existing enum https://github.com/apache/flink-kubernetes-operator/blob/main/flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/JobManagerDeploymentStatus.java, to use as reason and respective explanation as message

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/FlinkDeploymentStatus.java
lajith2006: I will add comments in the respective code.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/FlinkDeploymentStatus.java
lajith2006: Right, as mentioned [here](https://issues.apache.org/jira/browse/FLINK-33634?focusedCommentId=17933676&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17933676) , `phase` which is intended to keep the current status of FlinkDeployment especially useful in Openshift environment, as Openshift UI can  render the value from `status.phase` and populate the current status of deployment.  

As when FlinkDeployment applied in Application Mode,  in the initial phase, as the JM  brining up , job state would be in `RECONCILING` state, so status.phase is kept as "Pending".

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/FlinkDeploymentStatus.java
lajith2006: Wondering will that add any value?. Any way to build the Map , we have to call `ConditionUtils` to build the Condition, so rather than have a Prebuild Map , we can directly call `ConditionUtils` to build them right. Your thoughts?.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/utils/ConditionUtils.java
davidradl: I suggest doing that refactor when we need to, in the spirit of keeping the code lean.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/FlinkDeploymentStatus.java
davidradl: I see [https://github.com/apache/flink-kubernetes-operator/pull/961/files](https://github.com/apache/flink-kubernetes-operator/pull/961/files), gives examples of camelcase single words reasons. We should follow that style for the cases that PR does not cover.

lajith2006: Done a refactoring for the conditions.

csviri:
Would it be useful also to set also observedGeneration on the conditions, I believe it is in general a good practice.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/FlinkDeploymentStatus.java
```diff
@@ -55,4 +60,188 @@ public class FlinkDeploymentStatus extends CommonStatus<FlinkDeploymentSpec> {
...
+            conditions.add(condition);
+            return;
+        }
+        // If new condition is same as last condition, ignore
+        Condition existingCondition = conditions.get(conditions.size() - 1);
+        if (existingCondition.getType().equals(condition.getType())
+                && existingCondition.getMessage().equals(condition.getMessage())) {
+            return;
+        }
+        conditions.add(condition);
```
csviri, line 245: It's maybe just me, just don't get how this list is truncated eventually, in other words, how is it prevented from growing indefinitely?

lajith2006: Thanks for review @csviri . Right now it's  not preventing from growing indefinitely. As conditions reflects the different transitions state of Job/Deployment , thinking what could be the use case where  can grow indefinitely?.

gyfora: I think there should only ever be a single condition of type `Running` in the list that we return. Since we only have a single condition type right now, then the list should only have a single element. The latestTransition timestamp needs to represent when running changed from true->false or false->true. We can however keep updating the message if we want.

lajith2006: Thanks @gyfora for review , If I read correctly , you meant that we need to have only condition in the list at right now as we have only one type `Running` instead of multiple conditions in the list as currently this PR having.  And the  `lastTransitionTime`  in the condition must represent when the `Running` type changed its status from true>false or false > true. Is that correct?.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/FlinkDeploymentStatus.java
```diff
@@ -55,4 +60,188 @@ public class FlinkDeploymentStatus extends CommonStatus<FlinkDeploymentSpec> {
...
+                case ERROR:
+                    updateCondition(
+                            conditions,
+                            ConditionUtils.crCondition(
+                                    ConditionUtils.SESSION_MODE_CONDITION.get(
+                                            JobManagerDeploymentStatus.ERROR.name())));
+            }
+        } else if (getJobStatus() != null && getJobStatus().getState() != null) {
+            // Populate conditions for ApplicationMode deployment
+            switch (getJobStatus().getState()) {
```
davidradl, line 112: can this switch be replaced with
 updateCondition(
                            conditions,
                            ConditionUtils.crCondition(
                                    ConditionUtils.APPLICATION_MODE_CONDITION.get(
                                            getJobStatus().getState().name())));

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/utils/ConditionUtils.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    public static final Map<String, Condition> SESSION_MODE_CONDITION =
+            Map.of(
+                    JobManagerDeploymentStatus.READY.name(),
+                    new ConditionBuilder()
+                            .withType("Running")
+                            .withStatus("True")
+                            .withMessage("Ready")
+                            .withReason("JobManager is running and ready to receive REST API calls")
```
davidradl, line 46: isn't the reason - how it got to this state, not a description of the state itself?

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/FlinkDeploymentStatus.java
davidradl: I think you have pre built the map now - with map.of - looks good - thanks.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/FlinkDeploymentStatus.java
gyfora: Yes @lajith2006 , there should be a single condition of type Running and lastTransitionTime when that changed

lajith2006: Thanks for clarification @gyfora 

So just to be align with what you have mentioned  , for an example , when we do an  application cluster deployment, as Job cycles goes through CREATED>RUNNING

initially conditions will have as below 

```
status
  conditions:
    - type: Running
      status: "False"
      reason: Job is created
      message: "Job is newly created, no task has started to run"
      lastTransitionTime: 2025-04-030T06:17:08Z  
```


And then when the Job  starts running at `2025-04-030T06:19:01Z` , condition will have as below 

```
status
  conditions:
    - type: Running
      status: "True"
      reason: Job is running
      message: "Job is running"
      lastTransitionTime: 2025-04-030T06:19:01Z 
```

and then later at `2025-04-030T07:00:01Z`. when job gets finished , condition will have as below. 

```
status
  conditions:
    - type: Running
      status: "False"
      reason: Job's tasks have successfully finished
      message: Job's tasks have successfully finished
      lastTransitionTime: 2025-04-030T07:00:01Z 
```

which means ideally we will not have an history in the condition when the Job goes through different job lifecycle , instead only single condition and lastTransitionTime will reflect when the status was changed.

gyfora: Yes, I think this makes sense for a `Running` condition

gyfora:
@lajith2006 , I think we should try to somehow try to wrap this PR up. We have been going back and forth on this for months, a lot of reviews, different people involved (many unaddressed comments from @davidradl). This is a significant time investment from all involved.

Overall I feel that you need to iterate over the changes once more and try to eliminate the parts that were added with a much more complex design in mind, and just simplify the methods, eliminate the large switch statements etc.

I would love to include this in the upcoming release but for that we need to merge it by the end of next week.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/utils/ConditionUtils.java
lajith2006: @davidradl this has been updated and reason as is made as one-word, CamelCase representation of current status.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/utils/ConditionUtils.java
lajith2006: @davidradl There were some refactoring done on this class.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/FlinkDeploymentStatus.java
```diff
@@ -55,4 +60,188 @@ public class FlinkDeploymentStatus extends CommonStatus<FlinkDeploymentSpec> {
...
+
+    private String phase;
+
+    public List<Condition> getConditions() {
+        if (reconciliationStatus != null
+                && reconciliationStatus.deserializeLastReconciledSpec() != null
+                && reconciliationStatus.deserializeLastReconciledSpec().getJob() == null) {
+            // Populate conditions for SessionMode deployment
+            switch (jobManagerDeploymentStatus) {
+                case READY:
```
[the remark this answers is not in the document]
lajith2006: @davidradl this has been revisited and made it simple by removing switch.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/FlinkDeploymentStatus.java
lajith2006: Yes, @gyfora and @davidradl this has been revisited and addressed by removing switch and made it simple.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/FlinkDeploymentStatus.java
```diff
@@ -55,4 +63,56 @@ public class FlinkDeploymentStatus extends CommonStatus<FlinkDeploymentSpec> {
...
+                                                            .equals(newCondition.getReason())
+                                                    && c.getMessage()
+                                                            .equals(newCondition.getMessage()))
+                            .findFirst();
+            // Until there is a condition change which reflects the latest state, no need to add
+            // condition to list.
+            if (existingCondition.isPresent()) {
+                return;
+            }
+            // Remove existing Condition with type running and then add a new condition that
```
davidradl, line 108: nit: the comment is not quite right I think. 
I suggest this comment be 
  // Remove existing Condition with type running if not the same as the new condition

then `add a new condition that reflects the current state.` should be after the if

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/FlinkDeploymentStatus.java
```diff
@@ -55,4 +63,56 @@ public class FlinkDeploymentStatus extends CommonStatus<FlinkDeploymentSpec> {
...
+            // reflects the current state.
+            conditions.removeIf(
+                    c ->
+                            c.getType().equals(CONDITION_TYPE_RUNNING)
+                                    && !c.getMessage().equals(newCondition.getMessage())
```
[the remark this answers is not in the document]
lajith2006: >So we should just remove the existing condition without any conditions.
I think, if in later point of time, if we have some other conditions other than Running,  just removing the existing condition without of any check will cause an issue. 

>As-is the condition will not be removed if existing condition message or reason matches the new condition; which is not correct.
That's right,  but for Running type , we don't have any such case were reason or message can be matched with existing condition.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/FlinkDeploymentStatus.java
```diff
@@ -55,4 +63,56 @@ public class FlinkDeploymentStatus extends CommonStatus<FlinkDeploymentSpec> {
...
+                                ConditionUtils.SESSION_MODE_CONDITION.get(
+                                        jobManagerDeploymentStatus.name())));
+            } else if (jobStatus != null) {
+                // Populate conditions for ApplicationMode deployment
+                updateCondition(
+                        conditions,
+                        ConditionUtils.crCondition(
+                                ConditionUtils.APPLICATION_MODE_CONDITION.get(jobStatus.name())));
+            }
+        }
```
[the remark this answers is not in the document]
lajith2006: Thanks @gyfora . I believe , If we have them in ConditionUtils block, we will not be able to utilise  the `jobManagerDeploymentStatus` ConditionUtils. Please Correct me if I am wrong.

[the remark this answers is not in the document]
gyfora: This could be implemented as a static method that operates on the status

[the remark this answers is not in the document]
davidradl: > This could be implemented as a static method that operates on the status

@gyfora Yes - we were thinking changing to something like 
```
if (getJobStatus() != null) {
            JobStatus jobStatus = getJobStatus().getState();
            Condition conditionToAdd = null; 
            if (jobStatus == null) {
                // Populate conditions for SessionMode deployment
              conditionToAdd =
                                ConditionUtils.SESSION_MODE_CONDITION.get(
                                        jobManagerDeploymentStatus.name());
            } else {
                // Populate conditions for ApplicationMode deployment
                conditionToAdd =
                                ConditionUtils.APPLICATION_MODE_CONDITION.get(jobStatus.name());
            }
          conditions = ConditionsUtils.createListOfConditions(conditionToAdd);
}
```

Can you see a slicker way of doing this?

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/FlinkDeploymentStatus.java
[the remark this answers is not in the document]
gyfora: lastTransitionTime time should refer to the time when the status changes (true -> false or false->true) at least this is how it is used everywhere I know and everywhere I find online. Changed message alone should not update the timestamp as it would break the semantics and makes it impossible to tell when the job actually stopped running for example.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/FlinkDeploymentStatus.java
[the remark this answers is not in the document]
gyfora: I just wanted to factor out this condition logic from the FlinkDeploymentStatus class, it doesn't seem to belong there. I don't have an issue with he logic itself

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/utils/ConditionUtils.java
```diff
@@ -0,0 +1,146 @@
...
+                            .build(),
+                    JobManagerDeploymentStatus.ERROR.name(),
+                    new ConditionBuilder()
+                            .withType(CONDITION_TYPE_RUNNING)
+                            .withStatus("False")
+                            .withReason("Error")
+                            .withMessage("JobManager deployment failed")
+                            .build());
+
+    public static final Map<String, Condition> APPLICATION_MODE_CONDITION =
```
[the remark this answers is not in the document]
lajith2006: Thanks, I have a them in method. As reason is expected to be in camelcase, I have added a method to make it as camecase from status.name().

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/FlinkDeploymentStatus.java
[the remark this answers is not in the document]
lajith2006: Thanks @gyfora , modified in such a way that  `lastTransitionTimeStamp` gets updated when status changes from true> false or vice versa.

lajith2006: It's been refactored..

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/utils/ConditionUtils.java
```diff
@@ -0,0 +1,131 @@
...
+                            .build(),
+                    JobManagerDeploymentStatus.ERROR.name(),
+                    new ConditionBuilder()
+                            .withType(CONDITION_TYPE_RUNNING)
+                            .withStatus(JobManagerDeploymentConditionStatus.ERROR.getStatus())
+                            .withReason(JobManagerDeploymentConditionStatus.ERROR.getReason())
+                            .withMessage(JobManagerDeploymentConditionStatus.ERROR.getMessage())
+                            .build());
+
+    public static Condition getCondition(FlinkDeploymentStatus flinkDeploymentStatus) {
```
davidradl, line 82: I this the parameters that are required should be passed not the complete FlinkDeploymentStatus.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/JobManagerDeploymentConditionStatus.java
```diff
@@ -0,0 +1,35 @@
+package org.apache.flink.kubernetes.operator.api.status;
+
+/** Condition Status of the Flink JobManager Kubernetes deployment. */
+public enum JobManagerDeploymentConditionStatus {
+    READY("True", "JobManagerReady", "JobManager is running and ready to receive REST API calls"),
+    MISSING("False", "JobManagerDeploymentMissing", "JobManager deployment not found"),
+    DEPLOYING("False", "JobManagerIsDeploying", "JobManager process is starting up"),
```
davidradl, line 7: is deploying a starting up the same thing?

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/JobManagerDeploymentConditionStatus.java
```diff
@@ -0,0 +1,35 @@
...
+    DEPLOYED_NOT_READY(
+            "False",
+            "DeployedNotReady",
+            "JobManager is running but not ready yet to receive REST API calls"),
```
davidradl, line 11: nit ready yet -> yet ready

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/utils/ConditionUtils.java
```diff
@@ -0,0 +1,131 @@
...
+    private static Condition getApplicationModeCondition(JobStatus jobStatus) {
+        return new ConditionBuilder()
+                .withType(CONDITION_TYPE_RUNNING)
+                .withStatus(jobStatus == RUNNING ? "True" : "False")
+                .withReason(toCameCase(jobStatus.name()))
+                .withMessage("Job state " + jobStatus.name())
+                .build();
+    }
+
+    private static String toCameCase(String reason) {
```
davidradl, line 121: I think you mens toCamelCase

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/utils/ConditionUtils.java
lajith2006: I think from [here](https://github.com/apache/flink-kubernetes-operator/pull/957#discussion_r2089493297) @gyfora meant by operate on status on that util method .

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/utils/ConditionUtils.java
```diff
@@ -0,0 +1,131 @@
...
+        reason = reason.toLowerCase();
+        return reason.substring(0, 1).toUpperCase() + reason.substring(1);
```
davidradl, line 123: Camel case is not just upper casing the first letter. We may need to upper case in the string as well. I suggest mapping the lower case to the appropriate camel cased reason.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/utils/ConditionUtils.java
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+        return conditionToAdd;
+    }
+
+    public static void updateLastTransitionTime(List<Condition> conditions, Condition condition) {
+        if (condition == null) {
+            return;
+        }
+        if (isLastTransactionTimeStampUpdateRequired(conditions, condition)) {
```
davidradl, line 104: would it not be simpler to have an existing condition that is null or a value, then we do not need to get the first element of the list twice.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/utils/ConditionUtils.java
lajith2006: Thank you, my bad.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/utils/ConditionUtils.java
```diff
@@ -0,0 +1,131 @@
...
+            condition.setLastTransitionTime(conditions.get(0).getLastTransitionTime());
+        }
+    }
+
+    private static Condition getApplicationModeCondition(JobStatus jobStatus) {
+        return new ConditionBuilder()
+                .withType(CONDITION_TYPE_RUNNING)
+                .withStatus(jobStatus == RUNNING ? "True" : "False")
+                .withReason(toCameCase(jobStatus.name()))
+                .withMessage("Job state " + jobStatus.name())
```
davidradl, line 117: should the message say Job status - as that is what we are reading?

lajith2006: Yes,

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/utils/ConditionUtils.java
```diff
@@ -0,0 +1,131 @@
...
+import java.util.Date;
+import java.util.List;
+import java.util.Map;
+
+import static org.apache.flink.api.common.JobStatus.RUNNING;
+
+/** Creates a condition object with the type, status, message and reason. */
+public class ConditionUtils {
+    public static final String CONDITION_TYPE_RUNNING = "Running";
+    private static final Map<String, Condition> SESSION_MODE_CONDITION =
```
davidradl, line 38: @gyfora are you happy with this map in the utils class - it might be cleaner to be in its own appropriately named class, same for the application mode map

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/utils/ConditionUtils.java
```diff
@@ -0,0 +1,131 @@
...
+            Map.of(
+                    JobManagerDeploymentStatus.READY.name(),
+                    new ConditionBuilder()
```
davidradl, line 41: nit: you could refactor to a method and pass JobManagerDeploymentConditionStatus.READY, as the rest of the condition builder is the same apart from the status . 
So 
```
JobManagerDeploymentStatus.READY.name(),
createRunningConditionBuilder(JobManagerDeploymentConditionStatus.READY),
 JobManagerDeploymentStatus.MISSING.name(),
createRunningConditionBuilder(JobManagerDeploymentConditionStatus.MISSING),
...

File: flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/JobManagerDeploymentConditionStatus.java (hunk 20 above)
Comment (davidradl) [reply flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/JobManagerDeploymentConditionStatus.java:7]: OK that makes sense. Maybe link to that class in a comment to say where you got the descriptions from.

File: flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/utils/ConditionUtils.java (hunk 19 above)
Comment (davidradl) [reply flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/utils/ConditionUtils.java:82]: > I think from [here](https://github.com/apache/flink-kubernetes-operator/pull/957#discussion_r2089493297) @gyfora meant by operate on status on that util method .

OK I see

File: flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/utils/ConditionUtils.java [hunk 28]
```diff
@@ -0,0 +1,131 @@
...
+
+        return conditionToAdd;
+    }
+
+    public static void updateLastTransitionTime(List<Condition> conditions, Condition condition) {
+        if (condition == null) {
+            return;
+        }
+        if (isLastTransactionTimeStampUpdateRequired(conditions, condition)) {
+            condition.setLastTransitionTime(
```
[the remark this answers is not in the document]
lajith2006: Above is from `flink-autoscaler ` , but `flink-kubernetes-operator-api` , doesn't have dependency to `flink-autoscaler `, so we can't utilise it.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/utils/ConditionUtils.java
lajith2006: It has been moved to method and reason and messages are getting from enum.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/utils/ConditionUtils.java
lajith2006: updated,

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/utils/ConditionUtils.java
lajith2006: Right, but not as of now as per [here](https://github.com/apache/flink/blob/master/flink-core/src/main/java/org/apache/flink/api/common/JobStatus.java). If we go with mapping, we end up with new method, do we required them?.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/utils/ConditionUtils.java
```diff
@@ -0,0 +1,97 @@
...
+
+            JobStatus jobStatus = status.getState();
+
+            conditionToAdd =
+                    jobStatus == null
+                            ? getSessionModeCondition(
+                                    flinkDeploymentStatus.getJobManagerDeploymentStatus())
+                            : getApplicationModeCondition(jobStatus);
+        }
+
```
davidradl, line 52: can we not update the last transition time here?

davidradl:
> Would it be useful also to set also observedGeneration on the conditions, I believe it is in general a good practice.

It seems that this would be useful if the condition.observedGeneration does not matches the metadata.generation value.  
@gyfora WDYT? We could add this is a following on pr if you think this is a reasonable addition.

lajith2006:
> Would it be useful also to set also observedGeneration on the conditions, I believe it is in general a good practice.

by looking into some docs, It seems like condition offers a meaningful status info if the cond.observedGeneration is same as  metadata.generation, provided condition is calculated based on the last configuration of the object. So in our case here , as condition represents the different status of job, not sure if we have cond.observedGeneration, that  will matches with metadata.generation. WDYT?

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/FlinkDeploymentStatus.java
```diff
@@ -55,4 +63,146 @@ public class FlinkDeploymentStatus extends CommonStatus<FlinkDeploymentSpec> {
...
     /** Information about the TaskManagers for the scale subresource. */
     private TaskManagerInfo taskManager;
+
+    /** Condition of the CR . */
+    private List<Condition> conditions = new ArrayList<>();
+
+    public List<Condition> getConditions() {
+        if (reconciliationStatus != null
+                && reconciliationStatus.deserializeLastReconciledSpec() != null
+                && reconciliationStatus.deserializeLastReconciledSpec().getJob() == null) {
```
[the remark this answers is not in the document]
lajith2006: If we derive the mode from Jobstatus.state, it looks like , there is a chance of misleading information in the condition status. 

For example, it is observed that, when an existing session cluster is updated with some information , for example with checkpoint interval changed to new value, we can see that state is updated with value as FINISHED as below 

status:
  clusterInfo:
  jobManagerDeploymentStatus: READY
  jobStatus:
    checkpointInfo:
      lastPeriodicCheckpointTimestamp: 0
    savepointInfo:
      lastPeriodicSavepointTimestamp: 0
      savepointHistory: []
    state: FINISHED

So, if we derive the running state from JobStatus, by looking at the state value , we end up condition for above as Running false. But if we look at the  jobManagerDeploymentStatus: READY, we are supposed to say that condition as Running true as it is session cluster, and it's  says that session is ready to receive the request.

PR #958, opened 2025-03-25
[FLINK-37430] Operator hides the actual error on deployment issues

rodmeneses:

The operator often hides the actual issue for deployment errors and do not correctly report the error events.

## Brief change log

We are reporting the first 3 possible exceptions in the exception hierarchy in the error events.
## Verifying this change
- Added new unit tests

## Documentation

### DIFF
--- a/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/utils/ExceptionUtilsTest.java
+++ b/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/utils/ExceptionUtilsTest.java
@@ -1,0 +1,130 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.kubernetes.operator.utils;
+
+import org.apache.flink.util.SerializedThrowable;
+
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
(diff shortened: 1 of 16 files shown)


flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/utils/ExceptionUtilsTest.java
```diff
@@ -0,0 +1,130 @@
...
+package org.apache.flink.kubernetes.operator.utils;
+
+import org.apache.flink.util.SerializedThrowable;
+
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/** Test for {@link ExceptionUtils}. */
+public class ExceptionUtilsTest {
```
gyfora, line 27: So many test cases feel a bit over the top, could we narrow it down to 2-3 cases that cover the functionality?

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/utils/ExceptionUtilsTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+    @Test
+    void testGetExceptionMessage_moreThanThreeWithSerializedAndNotWithDifferentOrdering() {
+        var ex4 = new RuntimeException("Cause 4");
+        var ex3 = new RuntimeException("Cause 3", ex4);
+        var ex2 = new RuntimeException("Cause 2", new SerializedThrowable(ex3));
+        var ex = new RuntimeException("Cause 1", new SerializedThrowable(ex2));
+        assertThat(ExceptionUtils.getExceptionMessage(ex))
+                .isEqualTo("Cause 1 -> Cause 2 -> Cause 3");
+    }
```
gyfora, line 129: You could add a test for the non serializable case such as:

```
    @Test
    void testSerializedThrowableError() {
        var serializedException = new SerializedThrowable(new NonSerializableException());
        assertThat(ExceptionUtils.getExceptionMessage(serializedException))
                .isEqualTo("Unknown Error (SerializedThrowable)");
    }

    private static class NonSerializableException extends Exception {
        private void writeObject(java.io.ObjectOutputStream stream) throws IOException {
            throw new IOException();
        }
    }
```

gyfora (changes requested):

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/utils/ExceptionUtilsTest.java
rodmeneses: good idea!

gyfora (approved):

gyfora:
The e2es for Flink 2.0 started to fail after the official release, so we have new docker images. Merging this to fix the maven build then going to work on fixing the 2.0 compatibility.

PR #960, opened 2025-03-26
[FLINK-37530] Record upgrade savepoint correctly in savepointInfo as well

gyfora:

During the change to the snapshot resource logic we removed the logic that adds the upgrade savepoints to the old savepoint info status field and history. This causes inconsistencies for those who are still using that feature.

## Brief change log

Simply record upgrade savepoint in deprecated savepoint info to keep consistent 

## Verifying this change
Unit tests


morhidi (approved):

csviri (approved):
LGTM

PR #968, opened 2025-04-02
[FLINK-31406] Do not delete jobgraph on scale only last-state upgrades

pchoudhury22:
### What is the purpose of the change

Retain job graph on Scale only last state upgrades.

### Brief change log

Add mechanism to detect upgrade for SCALE only diff using lastStableSpec. retain job graph if Scale only last state upgrade. 

### Verifying this change

New unit tests added to cover this

### Does this pull request potentially affect one of the following parts:

Dependencies (does it add or upgrade a dependency): no
The public API, i.e., is any changes to the CustomResourceDescriptors: no
Core observer or reconciler logic that is regularly executed: yes

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractJobReconciler.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractJobReconciler.java
@@ -41,6 +41,7 @@ import org.apache.flink.kubernetes.operator.exception.UpgradeFailureException;
 import org.apache.flink.kubernetes.operator.reconciler.ReconciliationUtils;
 import org.apache.flink.kubernetes.operator.reconciler.SnapshotTriggerTimestampStore;
 import org.apache.flink.kubernetes.operator.reconciler.SnapshotType;
+import org.apache.flink.kubernetes.operator.reconciler.diff.ReflectiveDiffBuilder;
 import org.apache.flink.kubernetes.operator.service.CheckpointHistoryWrapper;
 import org.apache.flink.kubernetes.operator.service.SuspendMode;
 import org.apache.flink.kubernetes.operator.utils.EventRecorder;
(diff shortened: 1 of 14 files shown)


pchoudhury22:
Hi @gyfora , Please help review the PR! Thanks!

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractJobReconciler.java
```diff
@@ -575,7 +602,7 @@ protected void resubmitJob(FlinkResourceContext<CR> ctx, boolean requireHaMetada
                 && lastSavepointKnown) {
             specToRecover.getJob().setUpgradeMode(UpgradeMode.SAVEPOINT);
         }
-        restoreJob(ctx, specToRecover, ctx.getObserveConfig(), requireHaMetadata);
+        restoreJob(ctx, specToRecover, ctx.getObserveConfig(), requireHaMetadata, false);
```
gyfora, line 605: We could probably leverage the new flag here as well and not always set it to false during resubmit. Maybe when a lost deployment/unhealthy job is restarted it should be true as well. But we can follow up on a separate ticket

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractJobReconciler.java
```diff
@@ -202,13 +203,38 @@ protected boolean reconcileSpecChange(
...
+        if (lastStableSpec == null) {
+            return false;
+        }
+        lastReconciledSpec.getJob().setState(lastStableSpec.getJob().getState());
+        var specDiff =
+                new ReflectiveDiffBuilder<>(
+                                ctx.getDeploymentMode(), lastReconciledSpec, lastStableSpec)
+                        .build();
+        return specDiff.getType() == DiffType.SCALE;
+    }
```
gyfora, line 236: I think the logic here will not work. We cannot use the lastStableSpec to decide what triggered the last reconciliation as reconciliation (scaling/upgrade/etc) is never determined based on the lastStableSpec but always based on the lastReconciledSpec vs current spec. 

Currently the status may not have enough information to determine if this was a scale only change or not, we might have to record some extra metadata somewhere, we can sync offline.

pchoudhury22: I absolutely agree! honestly while working on this, I was able to come up with two options, 
1. use lastStableSpec to compare with lastReconciledSpec. which definitely goes agains the current design. But due the 2step upgrade, the other option was, as you mentioned we have to record some extra metadata somewhere. I was considering ReconciliationState for the same. Maybe to have like SCALING state. But again that does not sound very intuitive aswell. as Upgrading already does encompass Scaling. so definitely some guidance would be very helpful! Thanks!

gyfora:
Based on offline discussions, due to the introduced complexity and the very limited benefit we decided to drop this feature and close this jira for now.

pchoudhury22:
Thanks so much for all the notes!!

PR #969, opened 2025-04-08
[FLINK-37515] Basic support for Blue/Green deployments

schongloo:

This pull request adds basic support for Blue/Green deployments as outlined by FLIP-503 (https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=337677648).

## Brief change log
  - A new `FlinkBlueGreenDeploymentController` controller, with the capability of managing the lifecycle of these deployments.
  - A new corresponding CRD is introduced
  - The new `FlinkBlueGreenDeploymentController` will manage this CRD and hide from the user the details of the actual Blue/Green (Active/StandBy) jobs.
  - Delegate the lifecycle of the actual Jobs to the existing `FlinkDeployment` controller.

## Verifying this change

  - Test to verify a single basic deployment
  - Test to verify a proper transition Blue -> Green
  - Test to verify correct behavior when encountering an error before a transition
  - Test to verify correct behavior when encountering an error during a transition

## Documentation

  - If yes, how is the feature documented? docs / JavaDocs. UML diagrams for all scenarios added under `docs/content/docs`

### Overview Diagram
image


flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/bluegreen/DeploymentType.java
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.flink.kubernetes.operator.api.bluegreen;
+
+import org.apache.flink.kubernetes.operator.api.FlinkDeployment;
+
+/** Enumeration of the two possible Flink Blue/Green deployment types. */
+public enum DeploymentType {
+    BLUE,
```
davidradl, line 24: nit: could we have a description in the enum value please. It will be more self documenting then.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/FlinkBlueGreenDeploymentState.java
```diff
@@ -0,0 +1,26 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.kubernetes.operator.api.status;
+
+/** Enumeration of the possible states of the blue/green transition. */
+public enum FlinkBlueGreenDeploymentState {
+    ACTIVE_BLUE,
```
davidradl, line 22: could we have a descriptions for these enums as well please

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentController.java
```diff
@@ -0,0 +1,609 @@
...
+                    deploymentStatus,
+                    currentDeploymentType,
+                    josdkContext,
+                    currentDeployment,
+                    nextState);
+        } else {
+            // This phase requires rescheduling the reconciliation because the pod initialization
+            // could get stuck
+            //  (e.g. waiting for resources)
+            // TODO: figure out the course of action for error/failure cases
```
davidradl, line 198: It seems risky to put in a capability without error failure cases being dealt with, is there a way to put this under a feature flag , until it is mature enough and handles errors and failures?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentController.java
```diff
@@ -0,0 +1,609 @@
...
+                                        badEventPatterns.stream()
+                                                .anyMatch(
+                                                        p ->
+                                                                e.getReason()
+                                                                        .toUpperCase()
+                                                                        .contains(p)))
+                        .collect(Collectors.toList());
+
+        if (!badEvents.isEmpty()) {
+            LOG.warn("Bad events detected: " + badEvents);
```
davidradl, line 377: nit: does this log entry format reasonably?

davidradl:
I am concerned about the number of TODOs. Things like dealing with errors and timing conditions should be in place in the initial drop of code - unless we can mark this capability as beta or the like. WDYT @gyfora ?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentController.java
schongloo: Thanks for pointing this out, this comment is outdated and I'll clean it up. The edge cases are addressed with the current set of tests.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentController.java
schongloo: Yes, the events are logged like so:
```[Event(action=null, apiVersion=v1... message=MountVolume.SetUp failed for volume "checkpoint-volume" : hostPath type check failed: /tmp/checkpoints is not a directory, metadata=ObjectMeta(annotations={}, creationTimestamp=...```

**Comment by schongloo** [comment]
@davidradl thanks for your review/comments. I've cleaned up all the outdated/invalid TODOs which pointed me 1 missing unit test, added. At this point the few remaining TODOs do not interfere with the main functionality and are not critical/open for discussion.

**Comment by hansh0801** [comment]
when this feature gonna be released?

**Comment by schongloo** [comment]
Hi @hansh0801, trying to release it ASAP but I also want to make sure the logic and contracts are as robust as possible to minimize changes later. Looks pretty stable in my current testing and I've reopened it for review. Thanks!

**Comment by hansh0801** [comment]
cool ! and I'm also waiting for blue green phase 2 :)
https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=337677650

**Comment by schongloo** [comment]
@hansh0801 indeed... Phase 2 will take a little bit longer due to its complexity but already working on it.

File: flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/bluegreen/DeploymentType.java [hunk 5]
```diff
@@ -0,0 +1,38 @@
...
+public enum DeploymentType {
+    /** Identifier for the first or "Blue" deployment type. */
+    BLUE,
+
+    /** Identifier for the second or "Green" deployment type. */
+    GREEN;
+
+    public static DeploymentType fromDeployment(FlinkDeployment flinkDeployment) {
+        String typeAnnotation =
+                flinkDeployment.getMetadata().getLabels().get(DeploymentType.class.getSimpleName());
```
gyfora, line 35: Instead of using the label `DeploymentType` we should probably use a String constant that is less likely to clash with user annotations, such as:
`flink/blue-green-deployment-type`

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/bluegreen/TransitionMode.java
```diff
@@ -0,0 +1,33 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.kubernetes.operator.api.bluegreen;
+
+/** Possible transition modes supported by the `FlinkBlueGreenDeploymentController`. */
+public enum TransitionMode {
```
gyfora, line 21: I would remove this from this PR and only introduce it when it's implemented as part of the next flip

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/spec/FlinkDeploymentTemplateSpec.java
```diff
@@ -0,0 +1,60 @@
...
+    private ObjectMeta metadata;
+
+    @JsonProperty("deploymentDeletionDelayMs")
+    private int deploymentDeletionDelayMs;
+
+    @JsonProperty("abortGracePeriodMs")
+    private int abortGracePeriodMs;
+
+    @JsonProperty("reconciliationReschedulingIntervalMs")
+    private int reconciliationReschedulingIntervalMs;
```
gyfora, line 50: My gut feeling is that we will have more configurations in the past so maybe we should simply introduce a 


`Map<String, String> configuration` object and use Flink ConfigOptions for the timeout/interval settings. This is what we do basically everywhere else. And restrict the spec to key top level configs.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/FlinkBlueGreenDeploymentStatus.java
```diff
@@ -0,0 +1,58 @@
...
+    private String lastReconciledSpec;
+
+    /** Timestamp of last reconciliation. */
+    private Long lastReconciledTimestamp;
+
+    /** Computed from abortGracePeriodMs, timestamp after which the deployment should be aborted. */
+    private long abortTimestamp;
+
+    /** Timestamp when the deployment became READY/STABLE. Used to determine when to delete it. */
+    private long deploymentReadyTimestamp;
```
gyfora, line 54: All timestamps should be String timestamps (Instant.toString)

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/FlinkBlueGreenDeploymentStatus.java
```diff
@@ -0,0 +1,58 @@
...
+
+    /** Information about the TaskManagers for the scale subresource. */
+    private TaskManagerInfo taskManager;
```
gyfora, line 57: Do we want to support this? I feel that the scale logic using TM count is not really used feature for regular FlinkDeployments either

gyfora: We could remove it for now and only add it if there is a need

gyfora (changes requested):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentController.java
```diff
@@ -0,0 +1,703 @@
...
+            String wrapperKey,
+            Class<T> valueType)
+            throws JsonProcessingException {
+        String serializedSpec = SpecUtils.serializeObject(spec, wrapperKey);
+        String replacedSerializedSpec = serializedSpec.replace(deploymentName, childDeploymentName);
+        return SpecUtils.deserializeObject(replacedSerializedSpec, wrapperKey, valueType);
+    }
+
+    public static void logAndThrow(String message) {
+        LOG.error(message);
```
gyfora, line 700: Do we need to log this here? Maybe better to simply throw it and let error handling logic log it if needed

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentController.java
```diff
@@ -0,0 +1,703 @@
...
+                List.of(
+                        new OwnerReference(
+                                bgDeployment.getApiVersion(),
+                                true,
+                                false,
+                                bgDeployment.getKind(),
+                                bgMeta.getName(),
+                                bgMeta.getUid())));
+        return objectMeta;
+    }
```
gyfora, line 685: Should be moved to a static utility class

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentController.java
```diff
@@ -0,0 +1,703 @@
...
+            T spec,
+            String deploymentName,
+            String childDeploymentName,
+            String wrapperKey,
+            Class<T> valueType)
+            throws JsonProcessingException {
+        String serializedSpec = SpecUtils.serializeObject(spec, wrapperKey);
+        String replacedSerializedSpec = serializedSpec.replace(deploymentName, childDeploymentName);
+        return SpecUtils.deserializeObject(replacedSerializedSpec, wrapperKey, valueType);
+    }
```
gyfora, line 697: Should be moved to a utility class

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentController.java
```diff
@@ -0,0 +1,703 @@
...
+            FlinkBlueGreenDeploymentStatus deploymentStatus) {
+        deploymentStatus.setLastReconciledSpec(
+                SpecUtils.serializeObject(bgDeployment.getSpec(), "spec"));
+        deploymentStatus.setLastReconciledTimestamp(System.currentTimeMillis());
+    }
+
+    public void logPotentialWarnings(
+            FlinkDeployment flinkDeployment,
+            Context<FlinkBlueGreenDeployment> josdkContext,
+            long lastReconciliationTimestamp) {
```
gyfora, line 436: We are already logging these errors as part of the FlinkDeployment logic, this is duplicated code/logic here that should be removed

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentController.java
```diff
@@ -0,0 +1,703 @@
...
+
+        return patchStatusUpdateControl(
+                        bgDeployment, deploymentStatus, nextState, JobStatus.RECONCILING)
+                .rescheduleAfter(getReconciliationReschedInterval(bgDeployment));
+    }
+
+    private boolean isDeploymentReady(
+            FlinkDeployment deployment,
+            Context<FlinkBlueGreenDeployment> josdkContext,
+            FlinkBlueGreenDeploymentStatus deploymentStatus) {
```
gyfora, line 535: We should not check for pods here, we can eliminate all this logic I believe. This is too low level on this controller and if we feel that some info is missing from the FlinkDeployment status then it should be added there

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentController.java
```diff
@@ -0,0 +1,703 @@
...
+        var namespace = deployment.getMetadata().getNamespace();
+        var deploymentName = deployment.getMetadata().getName();
+
+        return josdkContext
+                .getClient()
+                .pods()
+                .inNamespace(namespace)
+                .withLabel("app", deploymentName)
+                .resources();
+    }
```
gyfora, line 573: probably not needed

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentController.java
```diff
@@ -0,0 +1,703 @@
...
+
+    private boolean hasSpecChanged(
+            FlinkBlueGreenDeploymentSpec newSpec, FlinkBlueGreenDeploymentStatus deploymentStatus) {
+
+        String lastReconciledSpec = deploymentStatus.getLastReconciledSpec();
+        String newSpecSerialized = SpecUtils.serializeObject(newSpec, "spec");
+
+        // TODO: in FLIP-504 check here the TransitionMode has not been changed
+
+        return !lastReconciledSpec.equals(newSpecSerialized);
```
gyfora, line 583: We may want to use the DiffBuilder/ existing utilities to get the type of the spec change here and make sure we don't trigger deployments on non-upgrade changes. 

Things like operator config changes etc should not go through the Blue/Green flow I believe but simply applied on the currently active deployment.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentController.java
```diff
@@ -0,0 +1,703 @@
...
+                        .getClient()
+                        .resources(FlinkDeployment.class)
+                        .inNamespace(currentDeployment.getMetadata().getNamespace())
+                        .withName(deploymentName)
+                        .delete();
+
+        boolean deleted =
+                deletedStatus.size() == 1
+                        && deletedStatus.get(0).getKind().equals("FlinkDeployment");
+
```
gyfora, line 663: do we want to wait here until the deletion is actually finished (the object is not there anymore?) that would require some additional logic here

gyfora (changes requested):

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/FlinkBlueGreenDeploymentState.java
davidradl: I was thinking that the enum would have the value and the description - this could help if we need to access the description programatically, e.g. in a UI.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/FlinkBlueGreenDeploymentState.java
```diff
@@ -0,0 +1,40 @@
...
+
+    /** Identifies the system is running normally with a "Green" deployment type. */
+    ACTIVE_GREEN,
+
+    /** Identifies the system is transitioning from "Green" to "Blue". */
+    TRANSITIONING_TO_BLUE,
+
+    /** Identifies the system is transitioning from "Blue" to "Green". */
+    TRANSITIONING_TO_GREEN,
```
davidradl, line 39: what state are we in during shutdown?

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/utils/SpecUtils.java
```diff
@@ -99,6 +99,28 @@ public static String writeSpecWithMeta(
         }
     }
 
+    public static String serializeObject(Object object, String wrapperKey) {
```
davidradl, line 102: I see that `serializeObject` seems to be only called for specs and I see the previous method is called `writeSpecWithMeta`. I suggest being consistent with the method names, maybe `writeSpecAsJSON`.
The key seems to always be called with "spec", could we hard code it here - if this is the case.

helm/flink-kubernetes-operator/templates/rbac.yaml
```diff
@@ -101,6 +103,7 @@ rules:
   - apiGroups:
       - flink.apache.org
     resources:
+      - flinkbluegreendeployments/status
       - flinkdeployments/status
```
davidradl, line 107: this maybe a basic question. Are flinkbluegreendeployments a type of flinkdeployments. Would we expect to see them as flinkdeployments?

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/utils/SpecUtils.java
schongloo: Yes I like that name suggestion, thanks. And indeed I'm using it on `FlinkBlueGreenDeploymentSpec` objects only but, unlike other specs, it doesn't extend `AbstractFlinkSpec`.

helm/flink-kubernetes-operator/templates/rbac.yaml
schongloo: That's the original idea but we shifted to a `has-is` relationship rather than a `is-a` one.

gyfora: To clarify this answer @davidradl . For every FlinkBlueGreenDeployment, there will also be 0-2 FlinkDeployment which are the currently running deployments.

maheepm-lyft:
Hi @schongloo @davidradl, I hope you're both doing well. I've been following this PR as blue-green deployments would solve a critical use case for our production needs. I noticed there have been no updates to this PR for a few weeks. Was wondering if there is anything blocking the release of this and if there is something myself or the OSS community could help with (testing, fixing any changes, etc). Thanks for all the work so far! 🙏

gyfora:
@schongloo can you please re-open this PR against the `main` branch?

schongloo:
> @schongloo can you please re-open this PR against the `main` branch?

Done @gyfora !

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/utils/SpecUtils.java
```diff
@@ -99,6 +99,28 @@ public static String writeSpecWithMeta(
...
+        ObjectNode wrapper = objectMapper.createObjectNode();
+        wrapper.set(wrapperKey, objectMapper.valueToTree(checkNotNull(object)));
+
+        try {
+            return objectMapper.writeValueAsString(wrapper);
+        } catch (JsonProcessingException e) {
+            throw new RuntimeException(
+                    "Could not serialize " + wrapperKey + ", this indicates a bug...", e);
```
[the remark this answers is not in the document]
schongloo: Yes it could be done, I just kept the same existing pattern as the rest of the functions in this class.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/spec/FlinkBlueGreenDeploymentConfigOptions.java
```diff
@@ -0,0 +1,54 @@
...
+
+/** Configuration options to be used by the Flink Blue/Green Deployments. */
+public class FlinkBlueGreenDeploymentConfigOptions {
+
+    public static final String BLUE_GREEN_CONF_PREFIX = "bluegreen.";
+
+    public static final int MIN_ABORT_GRACE_PERIOD_MS = 120000; // 2 mins
+
+    public static ConfigOptions.OptionBuilder operatorConfig(String key) {
+        return ConfigOptions.key(BLUE_GREEN_CONF_PREFIX + key);
```
gyfora, line 33: The prefix should be `KubernetesOperatorConfigOptions.K8S_OP_CONF_PREFIX + BLUE_GREEN_CONF_PREFIX`

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/spec/FlinkBlueGreenDeploymentConfigOptions.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    public static final ConfigOption<Integer> ABORT_GRACE_PERIOD_MS =
+            operatorConfig("abortGracePeriodMs")
```
gyfora: Config options should not be camel cased but yaml-key style. Look at KubernetesOperatorConfigOptions
`abort.grace-period`

Also this config should be `Duration` type

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/spec/FlinkBlueGreenDeploymentConfigOptions.java
```diff
@@ -0,0 +1,54 @@
...
+                    .intType()
+                    .defaultValue(0)
+                    .withDescription(
+                            "The max time to wait for a deployment to become ready before aborting it, in milliseconds. Cannot be smaller than 2 minutes.");
+
+    public static final ConfigOption<Integer> RECONCILIATION_RESCHEDULING_INTERVAL_MS =
+            operatorConfig("reconciliationReschedulingIntervalMs")
```
gyfora: Should be duration type and `reconciliation.reschedule.interval` or `reconciliation.reschedule-interval`

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/spec/FlinkBlueGreenDeploymentConfigOptions.java
```diff
@@ -0,0 +1,54 @@
...
+                    .intType()
+                    .defaultValue(15000) // 15 seconds
+                    .withDescription(
+                            "Configurable delay in milliseconds to use when the operator reschedules a reconciliation.");
+
+    public static final ConfigOption<Integer> DEPLOYMENT_DELETION_DELAY_MS =
+            operatorConfig("deploymentDeletionDelayMs")
```
gyfora: Should be duration and `deployment-deletion.delay`

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentController.java
```diff
@@ -0,0 +1,595 @@
...
+        long deploymentReadyTimestamp =
+                instantStrToMillis(deploymentStatus.getDeploymentReadyTimestamp());
+
+        if (deploymentReadyTimestamp == 0) {
+            LOG.info(
+                    "Deployment marked ready on "
+                            + System.currentTimeMillis()
+                            + ", rescheduling reconciliation in "
+                            + deploymentDeletionDelayMs
+                            + " ms.");
```
gyfora: should be `LOG.info(".... {}.... {}", a, b)`

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentController.java
```diff
@@ -0,0 +1,595 @@
...
+        }
+
+        var deletionTs = deploymentReadyTimestamp + deploymentDeletionDelayMs;
+
+        if (deletionTs < System.currentTimeMillis()) {
+            return deleteAndFinalize(
+                    bgDeployment, deploymentStatus, josdkContext, currentDeployment, nextState);
+        } else {
+            long delay = deletionTs - System.currentTimeMillis();
+            LOG.info("Rescheduling reconciliation (to delete) in " + delay + " ms.");
```
gyfora: should be `LOG.info(".... {}.... {}", a, b)`

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentController.java
```diff
@@ -0,0 +1,595 @@
...
+                            + previousState);
+
+            // If the current running FlinkDeployment is not in RUNNING/STABLE,
+            // we flag this Blue/Green as FAILING
+            return patchStatusUpdateControl(
+                    bgDeployment, deploymentStatus, null, JobStatus.FAILING);
+        } else {
+            // RETRY
+            var delay = abortTimestamp - System.currentTimeMillis();
+            LOG.info(
```
gyfora: should be `LOG.info(".... {}.... {}", a, b)`

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentController.java
```diff
@@ -0,0 +1,595 @@
...
+                throw new IllegalStateException("No blue/green FlinkDeployments found!");
+            }
+
+            deploymentStatus.setBlueGreenState(previousState);
+
+            LOG.warn(
+                    "Aborting deployment '"
+                            + deploymentName
+                            + "', rolling B/G deployment back to "
+                            + previousState);
```
gyfora: should be `LOG.warn(".... {}.... {}", a, b)`

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentController.java
```diff
@@ -0,0 +1,595 @@
...
+            FlinkBlueGreenDeploymentStatus deploymentStatus,
+            Context<FlinkBlueGreenDeployment> josdkContext,
+            FlinkDeployment currentDeployment,
+            FlinkBlueGreenDeploymentState nextState) {
+
+        if (currentDeployment != null) {
+            deleteDeployment(currentDeployment, josdkContext);
+            return UpdateControl.<FlinkBlueGreenDeployment>noUpdate().rescheduleAfter(500);
+        } else {
+            LOG.info(
```
gyfora: should be `LOG.info(".... {}.... {}", a, b)`

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentController.java
```diff
@@ -0,0 +1,595 @@
...
+                && JobStatus.RUNNING == deployment.getStatus().getJobStatus().getState();
+    }
+
+    private boolean hasSpecChanged(
+            FlinkBlueGreenDeploymentSpec newSpec, FlinkBlueGreenDeploymentStatus deploymentStatus) {
+
+        String lastReconciledSpec = deploymentStatus.getLastReconciledSpec();
+        String newSpecSerialized = SpecUtils.writeSpecAsJSON(newSpec, "spec");
+
+        return !lastReconciledSpec.equals(newSpecSerialized);
```
gyfora: I think this is too simplistic, you need to look at how the reconciler does it with:
```
var specDiff =
                new ReflectiveDiffBuilder<>(
                                ctx.getDeploymentMode(), lastReconciledSpec, currentDeploySpec)
```

A lot of spec changes should not trigger upgrades but should simply be applied on the currently running deployment.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentController.java
```diff
@@ -0,0 +1,595 @@
...
+                            josdkContext);
+                case TRANSITIONING_TO_GREEN:
+                    return monitorTransition(
+                            bgDeployment,
+                            FlinkBlueGreenDeployments.fromSecondaryResources(josdkContext),
+                            deploymentStatus,
+                            DeploymentType.BLUE,
+                            josdkContext);
+                default:
+                    return UpdateControl.noUpdate();
```
gyfora: Feels like this logic and the implementation should be moved to a new class as this is way to big.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentController.java
```diff
@@ -0,0 +1,595 @@
...
+            DeploymentType deploymentType,
+            Savepoint lastCheckpoint,
+            Context<FlinkBlueGreenDeployment> josdkContext,
+            boolean isFirstDeployment) {
+        ObjectMeta bgMeta = bgDeployment.getMetadata();
+
+        // Deployment
+        FlinkDeployment flinkDeployment = new FlinkDeployment();
+        flinkDeployment.setApiVersion("flink.apache.org/v1beta1");
+        flinkDeployment.setKind("FlinkDeployment");
```
gyfora: is this necessary?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentController.java
gyfora: Only UPGRADE / SCALE changes should trigger a transition. Other changes (such as operator/autoscaler config changes should simply be applied on the currently running FlinkDeployment (as they won't trigger an upgrade))

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/spec/FlinkBlueGreenDeploymentConfigOptions.java
schongloo: I cannot add a direct reference to `KubernetesOperatorConfigOptions.` due to circular dependency, but I'll add the "kubernetes.operator." prefix accordingly

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentController.java
```diff
@@ -0,0 +1,574 @@
...
+import static org.apache.flink.kubernetes.operator.api.spec.FlinkBlueGreenDeploymentConfigOptions.RECONCILIATION_RESCHEDULING_INTERVAL;
+import static org.apache.flink.kubernetes.operator.controller.FlinkBlueGreenDeploymentUtils.getConfigOption;
+import static org.apache.flink.kubernetes.operator.controller.FlinkBlueGreenDeploymentUtils.instantStrToMillis;
+import static org.apache.flink.kubernetes.operator.controller.FlinkBlueGreenDeploymentUtils.millisToInstantStr;
+
+/** Controller that runs the main reconcile loop for Flink Blue/Green deployments. */
+@ControllerConfiguration
+public class FlinkBlueGreenDeploymentController implements Reconciler<FlinkBlueGreenDeployment> {
+
+    private static final Logger LOG = LoggerFactory.getLogger(FlinkDeploymentController.class);
```
gyfora, line 69: Incorrect logger factory

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentController.java
```diff
@@ -0,0 +1,574 @@
...
+        }
+
+        var deletionTs = deploymentReadyTimestamp + deploymentDeletionDelayMs;
+
+        if (deletionTs < System.currentTimeMillis()) {
+            return deleteAndFinalize(
+                    bgDeployment, deploymentStatus, josdkContext, currentDeployment, nextState);
+        } else {
+            long delay = deletionTs - System.currentTimeMillis();
+            LOG.info("Rescheduling reconciliation (to delete) in {} ms.", delay);
```
gyfora: probably better to log in seconds

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentController.java
```diff
@@ -0,0 +1,574 @@
...
+            FlinkBlueGreenDeploymentState nextState) {
+        long deploymentDeletionDelayMs =
+                Math.max(getConfigOption(bgDeployment, DEPLOYMENT_DELETION_DELAY).toMillis(), 0);
+
+        long deploymentReadyTimestamp =
+                instantStrToMillis(deploymentStatus.getDeploymentReadyTimestamp());
+
+        if (deploymentReadyTimestamp == 0) {
+            LOG.info(
+                    "Deployment marked ready on {}, rescheduling reconciliation in {} ms.",
```
gyfora: Instead if `Deployment` we should write `FlinkDeployment` in all logs to avoid confusion. Also we should show seconds instead of milliseconds.

The timestamp is already in the log (when it was logged), not necessary to add

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentController.java
```diff
@@ -0,0 +1,574 @@
...
+                    previousState);
+
+            // If the current running FlinkDeployment is not in RUNNING/STABLE,
+            // we flag this Blue/Green as FAILING
+            return patchStatusUpdateControl(
+                    bgDeployment, deploymentStatus, null, JobStatus.FAILING);
+        } else {
+            // RETRY
+            var delay = abortTimestamp - System.currentTimeMillis();
+            LOG.info("Deployment '{}' not ready yet, retrying in {} ms", deploymentName, delay);
```
gyfora: *FlinkDeployment, log reschedule in seconds.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentController.java
```diff
@@ -0,0 +1,574 @@
...
+                        .delete();
+
+        boolean deleted =
+                deletedStatus.size() == 1
+                        && deletedStatus.get(0).getKind().equals("FlinkDeployment");
+
+        if (!deleted) {
+            LOG.info("Deployment '" + deploymentName + "' not deleted, will retry");
+        } else {
+            LOG.info("Deployment '" + deploymentName + "' deleted!");
```
gyfora: *FlinkDeployment

gyfora (changes requested):
Added some logging related comments.

We should also handle the `stateless` case a bit more gracefully. A simple stateless example fails with errors like:

```
flink-kubernetes-operator-7f95587856-cz8j7 2025-07-18 06:24:24,708 i.j.o.p.e.EventProcessor       [WARN ][default/basic-example] Uncaught error during event processing ExecutionScope{ resource id: ResourceID{name='basic-example', namespace='default'}, version: 11407039} - but another reconciliation will be attempted because a superseding event has been received or another retry attempt is pending.
flink-kubernetes-operator-7f95587856-cz8j7 org.apache.flink.kubernetes.operator.exception.UpgradeFailureException: Latest checkpoint not externally addressable, manual recovery required.
flink-kubernetes-operator-7f95587856-cz8j7     at org.apache.flink.kubernetes.operator.service.AbstractFlinkService.getLastCheckpoint(AbstractFlinkService.java:569)
flink-kubernetes-operator-7f95587856-cz8j7     at org.apache.flink.kubernetes.operator.controller.FlinkBlueGreenDeploymentController.configureSavepoint(FlinkBlueGreenDeploymentController.java:415)
flink-kubernetes-operator-7f95587856-cz8j7     at org.apache.flink.kubernetes.operator.controller.FlinkBlueGreenDeploymentController.checkAndInitiateDeployment(FlinkBlueGreenDeploymentController.java:378)
flink-kubernetes-operator-7f95587856-cz8j7     at org.apache.flink.kubernetes.operator.controller.FlinkBlueGreenDeploymentController.reconcile(FlinkBlueGreenDeploymentController.java:117)
flink-kubernetes-operator-7f95587856-cz8j7     at org.apache.flink.kubernetes.operator.controller.FlinkBlueGreenDeploymentController.reconcile(FlinkBlueGreenDeploymentController.java:65)
flink-kubernetes-operator-7f95587856-cz8j7     at io.javaoperatorsdk.operator.processing.Controller$1.execute(Controller.java:161)
flink-kubernetes-operator-7f95587856-cz8j7     at io.javaoperatorsdk.operator.processing.Controller$1.execute(Controller.java:117)
flink-kubernetes-operator-7f95587856-cz8j7     at org.apache.flink.kubernetes.operator.metrics.OperatorJosdkMetrics.timeControllerExecution(OperatorJosdkMetrics.java:82)
flink-kubernetes-operator-7f95587856-cz8j7     at io.javaoperatorsdk.operator.processing.Controller.reconcile(Controller.java:116)
flink-kubernetes-operator-7f95587856-cz8j7     at io.javaoperatorsdk.operator.processing.event.ReconciliationDispatcher.reconcileExecution(ReconciliationDispatcher.java:153)
flink-kubernetes-operator-7f95587856-cz8j7     at io.javaoperatorsdk.operator.processing.event.ReconciliationDispatcher.handleReconcile(ReconciliationDispatcher.java:130)
flink-kubernetes-operator-7f95587856-cz8j7     at io.javaoperatorsdk.operator.processing.event.ReconciliationDispatcher.handleDispatch(ReconciliationDispatcher.java:97)
flink-kubernetes-operator-7f95587856-cz8j7     at io.javaoperatorsdk.operator.processing.event.ReconciliationDispatcher.handleExecution(ReconciliationDispatcher.java:68)
flink-kubernetes-operator-7f95587856-cz8j7     at io.javaoperatorsdk.operator.processing.event.EventProcessor$ReconcilerExecutor.run(EventProcessor.java:475)
flink-kubernetes-operator-7f95587856-cz8j7     at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
flink-kubernetes-operator-7f95587856-cz8j7     at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
flink-kubernetes-operator-7f95587856-cz8j7     at java.base/java.lang.Thread.run(Unknown Source)
```

during upgrades.

In general if an upgrade would go through successfully for a regular FlinkDeployment it should also go through for a BlueGreenDeployment.

schongloo:
> Added some logging related comments.
> 
> We should also handle the `stateless` case a bit more gracefully. A simple stateless example fails with errors like:
> 
> ```
> ... org.apache.flink.kubernetes.operator.exception.UpgradeFailureException: Latest checkpoint not externally addressable, manual recovery required.
> ```
> 
> during upgrades.
> 
> In general if an upgrade would go through successfully for a regular FlinkDeployment it should also go through for a BlueGreenDeployment.

While not necessarily related to the `stateless` case. In this situation it's an error I throw when checkpoints are not externally accessible (which is mandatory), the confusing part I throw it during the upgrade only. I should also validate this type of restriction(s) before the initial deployment.

I've addressed all the simple refactoring/logging comments, remaining to be done asap:

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentController.java
schongloo: No, remains of early tests, removed.

gyfora:
> > Added some logging related comments.
> > We should also handle the `stateless` case a bit more gracefully. A simple stateless example fails with errors like:
> > ```
> > ... org.apache.flink.kubernetes.operator.exception.UpgradeFailureException: Latest checkpoint not externally addressable, manual recovery required.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > during upgrades.
> > In general if an upgrade would go through successfully for a regular FlinkDeployment it should also go through for a BlueGreenDeployment.
> 
> While not necessarily related to the `stateless` case. In this situation it's an error I throw when checkpoints are not externally accessible (which is mandatory), the confusing part I throw it during the upgrade only. I should also validate this type of restriction(s) before the initial deployment.
> 
> I've addressed all the simple refactoring/logging comments, remaining to be worked on asap: diff builder and e2e test

@schongloo , `stateless` Blue/Green upgrades should not use any state (even if available) and simply start the new job from an empty state. It should definitely not throw any error if checkpointing/external checkpoints are disabled

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentController.java
```diff
@@ -0,0 +1,128 @@
...
+                            ctxFactory);
+
+            LOG.debug(
+                    "Processing state: {} for deployment: {}",
+                    currentState,
+                    context.getDeploymentName());
+
+            BlueGreenStateHandler handler = handlerRegistry.getHandler(currentState);
+            return handler.handle(context);
+//            return stateMachine.processState(context);
```
gyfora: remove commented code

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/bluegreen/BlueGreenDeploymentService.java
```diff
@@ -0,0 +1,401 @@
...
+        final FlinkBlueGreenDeploymentState nextState;
+
+        DeploymentTransition(
+                DeploymentType nextDeploymentType, FlinkBlueGreenDeploymentState nextState) {
+            this.nextDeploymentType = nextDeploymentType;
+            this.nextState = nextState;
+        }
+    }
+
+    private static class TransitionState {
```
gyfora, line 654: Could use Lombok `@Value` annotation to remove boilerplate

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/bluegreen/BlueGreenStateHandlerRegistry.java
```diff
@@ -0,0 +1,82 @@
...
+     *
+     * @param state the Blue/Green deployment state
+     * @return the corresponding state handler
+     * @throws IllegalStateException if no handler is found for the state
+     */
+    public BlueGreenStateHandler getHandler(FlinkBlueGreenDeploymentState state) {
+        BlueGreenStateHandler handler = handlers.get(state);
+        if (handler == null) {
+            throw new IllegalStateException("No handler found for state: " + state);
+        }
```
gyfora: This feels like unnecessary logic as a handler for every state is required otherwise we would hit errors left and right in unit tests.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/bluegreen/handlers/AbstractBlueGreenStateHandler.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @Override
+    public FlinkBlueGreenDeploymentState getSupportedState() {
+        return supportedState;
+    }
+
+    protected Logger getLogger() {
+        return LOG;
+    }
```
gyfora: if we need access just make the logger field protected

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/bluegreen/BlueGreenDiffType.java
```diff
@@ -0,0 +1,38 @@
...
+    IGNORE,
+
+    /** Only top-level properties (metadata, configuration) have differences. */
+    PATCH_TOP_LEVEL,
+
+    /** Only the nested FlinkDeploymentSpec has differences. */
+    PATCH_CHILD,
+
+    /** Both top-level and nested specifications have differences. */
+    PATCH_BOTH,
```
gyfora: What's the functional difference between IGNORE the different PATCH diffs and TRANSITION? Feels like we should have only 3:

NO_CHANGE
OPERATOR_CONFIG_ONLY -> Non-upgrade/scale deployment changes
UPGRADE

.github/workflows/ci.yml
```diff
@@ -172,6 +172,8 @@ jobs:
           - test_flink_operator_ha.sh
           - test_snapshot.sh
           - test_batch_job.sh
+          - test_bluegreen_laststate.sh
+          - test_bluegreen_stateless.sh
```
gyfora: I think we will probably want to only test the blue/green stuff for a single/latest Flink Version for now as the logic is not related to Flink versions in any way I believe.

docs/content/docs/custom-resource/reference.md
```diff
@@ -57,6 +57,23 @@ This serves as a full reference for FlinkDeployment and FlinkSessionJob custom r
 | Parameter | Type | Docs |
 | ----------| ---- | ---- |
 
+### FlinkBlueGreenDeploymentConfigOptions
```
davidradl, line 76: I think we should define the Blue Green deployment concept some where in the docs, ideally with diagrams.

docs/content/docs/custom-resource/reference.md
```diff
@@ -78,6 +95,17 @@ This serves as a full reference for FlinkDeployment and FlinkSessionJob custom r
 | logConfiguration | java.util.Map<java.lang.String,java.lang.String> | Log configuration overrides for the Flink deployment. Format logConfigFileName -> configContent. |
 | mode | org.apache.flink.kubernetes.operator.api.spec.KubernetesDeploymentMode | Deployment mode of the Flink cluster, native or standalone. |
 
+### FlinkDeploymentTemplateSpec
```
davidradl, line 115: nit: I am curious why this class does not have blue green in its name - as the description says it is only for blue green. 

I suggest either changing the class name to include blue green, or make the description more generic, giving an example of blue green.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/bluegreen/DeploymentType.java
```diff
@@ -0,0 +1,39 @@
...
+
+package org.apache.flink.kubernetes.operator.api.bluegreen;
+
+import org.apache.flink.kubernetes.operator.api.FlinkDeployment;
+
+/**
+ * Enumeration of the two possible Flink Blue/Green deployment types. Only one of each type will be
+ * present at all times for a particular job.
+ */
+public enum DeploymentType {
```
davidradl: I suggest the enum has blue green in it's name

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/spec/FlinkBlueGreenDeploymentConfigOptions.java
```diff
@@ -0,0 +1,73 @@
...
+    public static final String K8S_OP_CONF_PREFIX = "kubernetes.operator.";
+
+    public static final String BLUE_GREEN_CONF_PREFIX = K8S_OP_CONF_PREFIX + "bluegreen.";
+
+    public static ConfigOptions.OptionBuilder operatorConfig(String key) {
+        return ConfigOptions.key(BLUE_GREEN_CONF_PREFIX + key);
+    }
+
+    /**
+     * NOTE: The string durations need to be in format "{length value}{time unit label}", e.g.
```
davidradl: nit: length value -> time unit value

docs/content/docs/custom-resource/reference.md
gyfora: I think the current name makes sense, because this is actually a template for a FlinkDeployment. 
In the same way the PodTemplateSpec doesn't have any reference to the actual CR-s that use it such as Deployment/ReplicaSet

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/spec/FlinkBlueGreenDeploymentConfigOptions.java
```diff
@@ -0,0 +1,73 @@
...
+     *
+     * <p>Supported time unit labels are:
+     *
+     * <ul>
+     *   <li>DAYS： "d", "day"
+     *   <li>HOURS： "h", "hour"
+     *   <li>MINUTES： "m", "min", "minute"
+     *   <li>SECONDS： "s", "sec", "second"
+     *   <li>MILLISECONDS： "ms", "milli", "millisecond"
+     *   <li>MICROSECONDS： "µs", "micro", "microsecond"
```
davidradl: I am concerned about `µs` as it is a non standard character. [Flink docs](https://nightlies.apache.org/flink/flink-docs-release-1.20/docs/deployment/config/) use `us`. I suggest we use that here to be consistent.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/spec/FlinkBlueGreenDeploymentConfigOptions.java
```diff
@@ -0,0 +1,73 @@
...
+     *   <li>NANOSECONDS： "ns", "nano", "nanosecond"
+     * </ul>
+     */
+    public static final ConfigOption<Duration> ABORT_GRACE_PERIOD =
+            operatorConfig("abort.grace-period")
+                    .durationType()
+                    .defaultValue(Duration.ofMinutes(10))
+                    .withDescription(
+                            "The max time to wait in milliseconds for a deployment to become ready before aborting it. Cannot be smaller than 10 minutes.");
```
davidradl: nit: it is a duration type not milliseconds

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/spec/FlinkBlueGreenDeploymentConfigOptions.java
```diff
@@ -0,0 +1,73 @@
...
+
+    public static final ConfigOption<Duration> RECONCILIATION_RESCHEDULING_INTERVAL =
+            operatorConfig("reconciliation.reschedule-interval")
+                    .durationType()
+                    .defaultValue(Duration.ofSeconds(15))
+                    .withDescription(
+                            "Configurable delay in milliseconds to use when the operator reschedules a reconciliation.");
```
davidradl: I suggest including some advice as to what to set this to in different circumstances. Same for deployment-deletion.delay

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/FlinkBlueGreenDeploymentState.java
```diff
@@ -0,0 +1,46 @@
...
+
+    /** Identifies the system is savepointing "Blue" before it transitions to "Green". */
+    SAVEPOINTING_BLUE,
```
davidradl, line 42: nit: do we need an error state if a transition fails?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentController.java
```diff
@@ -0,0 +1,142 @@
...
+import static org.apache.flink.kubernetes.operator.api.status.FlinkBlueGreenDeploymentState.INITIALIZING_BLUE;
+
+/**
+ * Controller that runs the main reconcile loop for Flink Blue/Green deployments.
+ *
+ * <p>State Machine Flow
+ *
+ * <p>Deployment States
+ *
+ * <p>1. INITIALIZING_BLUE - First-time deployment setup 2. ACTIVE_BLUE - Blue environment serving
```
davidradl, line 54: nit maybe use a html list for this list

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentController.java
```diff
@@ -0,0 +1,142 @@
...
+ * traffic, monitoring for updates 3. TRANSITIONING_TO_GREEN - Deploying Green environment while
+ * Blue serves traffic 4. ACTIVE_GREEN - Green environment serving traffic, monitoring for updates
+ * 5. TRANSITIONING_TO_BLUE - Deploying Blue environment while Green serves traffic
+ *
+ * <p>Orchestration Process
+ *
+ * <p>FlinkBlueGreenDeploymentController.reconcile() ↓ 1. Create BlueGreenContext with current
+ * deployment state ↓ 2. Query StateHandlerRegistry for appropriate handler ↓ 3. Delegate to
```
davidradl: what are the `↓` characters?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentController.java
```diff
@@ -0,0 +1,142 @@
...
+            var context =
+                    new BlueGreenContext(
+                            bgDeployment,
+                            new FlinkBlueGreenDeploymentStatus(),
+                            josdkContext,
+                            null,
+                            ctxFactory);
+            return BlueGreenDeploymentService.patchStatusUpdateControl(
+                            context, INITIALIZING_BLUE, null)
+                    .rescheduleAfter(100);
```
davidradl: should this 100 be using the configured  RECONCILIATION_RESCHEDULING_INTERVAL?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeployments.java
```diff
@@ -0,0 +1,80 @@
...
+        }
+        return counter;
+    }
+
+    static FlinkBlueGreenDeployments fromSecondaryResources(
+            Context<FlinkBlueGreenDeployment> context) {
+        Set<FlinkDeployment> secondaryResources =
+                context.getSecondaryResources(FlinkDeployment.class);
+
+        if (secondaryResources.isEmpty() || secondaryResources.size() > 2) {
```
davidradl, line 53: looking at this logic:
0 secondary deployments is valid. 
2 secondary deployments are valid as long as one is blue and one is green
1 secondary deployment is allowed - when is this valid?

schongloo:
@davidradl thank you for all your feedback. The past few weeks I was focused on mostly the high level refactoring, I'll address your comments ASAP.

docs/content/docs/custom-resource/reference.md
schongloo: I've (temporarily) added a diagram to the PR description of the actual state machine, we can find it a final location later. 

Also the pages for both [FLIP-503](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=337677648) and [FLIP-504](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=337677650) have higher level (simplified) concept diagrams. 

Does this help?

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/FlinkBlueGreenDeploymentState.java
schongloo: The (old) deployments are deleted once the new ones are RUNNING and STABLE, this is the last step during the TRANSITION_TO_* state.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/FlinkBlueGreenDeploymentState.java
schongloo: The rationale has been to let the system "rollback", e.g. return to the previous healthy state (ACTIVE_BLUE/ACTIVE_GREEN) as far as state management is concerned, this eases the continuation/retries. However, depending on the situation, the FlinkBlueGreenDeployment resource will be marked as FAILED and the applicable offending FlinkDeployment will be left in their failed state for forensic analysis. Corresponding messages are logged as well.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentController.java
schongloo: I'll change it to "0". I wanted to reschedule a reconciliation "as soon as possible" and I initially just thought of 100.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeployments.java
schongloo: The parent is always the `FlinkBlueGreenDeployment` CR and normally there's always 1 secondary `FlinkDeployment` (either Blue or Green) running, 2 secondary deployments indicate a transition is occurring, 0 or more than 2 secondary FlinkDeployments would be an anomaly.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/spec/FlinkBlueGreenDeploymentConfigOptions.java
```diff
@@ -0,0 +1,73 @@
...
+                            "The max time to wait for a deployment to become ready before aborting it.");
+
+    public static final ConfigOption<Duration> RECONCILIATION_RESCHEDULING_INTERVAL =
+            operatorConfig("reconciliation.reschedule-interval")
+                    .durationType()
+                    .defaultValue(Duration.ofSeconds(15))
+                    .withDescription(
+                            "Configurable delay to use when the operator reschedules a reconciliation.");
+
+    public static final ConfigOption<Duration> DEPLOYMENT_DELETION_DELAY =
```
vikasap, line 67: These configs are not being used. Is there an intent to use them later ?

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/spec/FlinkDeploymentTemplateSpec.java
```diff
@@ -0,0 +1,49 @@
...
+@NoArgsConstructor
+@Data
+@SuperBuilder
+public class FlinkDeploymentTemplateSpec {
+
+    @JsonProperty("metadata")
+    private ObjectMeta metadata;
+
+    @JsonProperty("configuration")
+    private Map<String, String> configuration;
```
gyfora: I think this configuration field should be in the top level FlinkBlueGreen deployment spec and not in the template. The template should only contain FlinkDeployment related stuff and this configuration is what controls the management of the bg deployment itself.

Once this is changed we have to update the test data as well to reflect the schema change

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/bluegreen/BlueGreenUtils.java
```diff
@@ -0,0 +1,387 @@
...
+     * @return adjusted spec with name references updated
+     */
+    public static <T> T adjustNameReferences(
+            T spec,
+            String deploymentName,
+            String childDeploymentName,
+            String wrapperKey,
+            Class<T> valueType) {
+        String serializedSpec = SpecUtils.writeSpecAsJSON(spec, wrapperKey);
+        String replacedSerializedSpec = serializedSpec.replace(deploymentName, childDeploymentName);
```
vikasap, line 79: This will do replacement of the image tag as well. We would probably need to exclude that.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/spec/FlinkBlueGreenDeploymentConfigOptions.java
schongloo: Not sure I follow. These values are used throughout the `BlueGreenDeploymentService`.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/bluegreen/BlueGreenDeploymentService.java
```diff
@@ -0,0 +1,659 @@
...
+                throw new RuntimeException(
+                        String.format(
+                                "Could not fetch savepoint with triggerId: %s. Error: %s",
+                                triggerId, savepointFetchResult.getError()));
+            }
+
+            return getSavepointObject(ctx, savepointFetchResult.getLocation());
+        }
+
+        // Currently not using last checkpoint recovery for LAST_STATE upgrade mode
```
vikasap, line 310: Any particular reason this is disabled explicitly ?

schongloo: We realized using checkpoints to preserve state across Blue/Green transitions can be a brittle strategy since it would depend entirely on the users configuration (whether they're even enabled or not, externally retained, for how long, etc.). So we opted to go for Savepoints instead, at least initially.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/bluegreen/BlueGreenDeploymentService.java
```diff
@@ -0,0 +1,659 @@
...
+            }
+        }
+
+        return UpdateControl.noUpdate();
+    }
+
+    private UpdateControl<FlinkBlueGreenDeployment> patchFlinkDeployment(
+            BlueGreenContext context, BlueGreenDeploymentType blueGreenDeploymentTypeToPatch) {
+
+        String childDeploymentName =
```
maheepm-lyft, line 174: Hi @schongloo , sorry to leave a comment after the PR is merged. I have been doing some testing with BlueGreen deployments, and I think this is replacing the name in several places where it should not be. For example, if I am using an `initialSavepointPath` with savepoint path in S3, it is replacing the app name in the path with app name + "-blue" or "-green", and the app is unable to find the corresponding savepoint. E.g. `app_name` becomes `app_name-blue`. If the savepoint path already has -blue in it (e.g. `app_name-blue`, then it is updating it to `app_name-blue-blue`. I am seeing similar issues related to a few other things like configmap names, although I haven't followed that code path all the way.

EDIT: added a comment [elsewhere](https://github.com/apache/flink-kubernetes-operator/pull/969#discussion_r2430190274) after looking into the bug a little more.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/bluegreen/BlueGreenUtils.java
maheepm-lyft: +1 to what vikas said. This will change the name in places that are problematic. In my testing, I am facing problems so far in expected configmap name, savepoint path, and metrics reporters.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/bluegreen/BlueGreenDeploymentService.java
schongloo: @maheepm-lyft just wanted to make sure this is not a loose end. The proposed fix for this problem is tracked by [this PR](https://github.com/apache/flink-kubernetes-operator/pull/1039), hopefully you're already aware.

maheepm-lyft: Hi, thanks for following up! Yes I'm aware of the PR, the committer is in my org and we collaborated on that change. Thanks again for pushing out this blue green feature.

PR #973, opened 2025-04-28
Update publish-snapshot workflow actions

kornys:

Update version of used github actions in publish-snapshot.yaml

## Brief change log
- *Update cache, setup java and checkout to version 4*


gyfora:
What is the reason for this change?

kornys:
GitHub is retiring the old cache service on 1 Feb 2025. -> https://stackoverflow.com/questions/77897660/github-actions-node-js-16-actions-are-deprecated-warning

Older majors of both actions run on Node 16, which is now deprecated. Upgrading to v4 moves the workflow to Node 20

PR #975, opened 2025-04-29
[FLINK-37748][Kubernetes Operator]Remove JDK only JVM arguments from Flink operator Helm chart

ekazakas:

There is an issue with Flink Kubernetes operator where JVM reports unrecognized options. These options are:
```
--add-exports=jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED
--add-exports=jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED
--add-exports=jdk.compiler/com.sun.tools.javac.parser=ALL-UNNAMED
--add-exports=jdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED
--add-exports=jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED
```

The above are only valid in JDK environments. When running in JRE environments, where most of the Flink applications will be running, the VM reports the following warnings:

```
WARNING: Unknown module: jdk.compiler specified to --add-exports
WARNING: Unknown module: jdk.compiler specified to --add-exports
WARNING: Unknown module: jdk.compiler specified to --add-exports
<...>
```

## Brief change log

- Removed JDK only JVM options from operator Helm chart

## Verifying this change

## Documentation


kostas-vinted (approved):

gyfora:
These were added to improve java 17 compatibility for Flink apps, will this affect that? If so having the warnings doesn't seem like a big deal

ekazakas:
> These were added to improve java 17 compatibility for Flink apps, will this affect that? If so having the warnings doesn't seem like a big deal

It should not affect it, especially if the apps are running in JRE environment. When running in JRE, these options are not taken into account at all, they simply do not exist in those environments

gyfora:
> > These were added to improve java 17 compatibility for Flink apps, will this affect that? If so having the warnings doesn't seem like a big deal
> 
> It should not affect it, especially if the apps are running in JRE environment. When running in JRE, these options are not taken into account at all, they simply do not exist in those environments.
> 
> P.S. The operator [image](https://github.com/apache/flink-kubernetes-operator/blob/main/Dockerfile) itself is running in JRE environment

This is about the Flink jobs, and the question is whether these params make Java 17 compatibility better in JDK env. If they do we should keep that and accept the warning as that doesn't hurt

ekazakas:
It doesn't make the compatibility better, since Flink jobs do not require compiler modules. In case of very specific applications that for some reason would need to access the javac compiler and it's modules, users could specify that in `env.java.opts.all`.  But it's up to Flink maintainers to decide whether they would agree with this change. It would avoid the unnecessarry log pollution for 99% of Flink jobs

gyfora (approved):

helm/flink-kubernetes-operator/conf/flink-conf.yaml
```diff
@@ -24,8 +24,8 @@ parallelism.default: 1
 # These should be kept in-sync with the flink dist env.java.opts.all defaults (for the given flink version) in: flink-dist/src/main/resources/config.yaml
 # Flink 1.18 uses env.java.opts.all, if a user supplies their own version of these opts in their FlinkDeployment the options below will be overridden.
 # env.java.default-opts.all is used for 1.19 onwards so users can supply their own opts.all in their Job deployments and have these appended.
```
davidradl, line 26: I see in the comment above that the below content should be kept in sync with `flink-dist/src/main/resources/config.yaml.`

I suggest that we amend the comment - as it will no longer be true.

ekazakas: I have removed this comment

ferenc-csaky:
@gyfora we good to merge this?

gyfora:
Thanks for the reminder @ferenc-csaky , merging

PR #977, opened 2025-05-01
[FLINK-37662][helm] Support init and side car containers

tweise:
Customization points for init and side car containers to the helm chart, which are common in non-trivial deployments.

### DIFF
--- a/helm/flink-kubernetes-operator/templates/flink-operator.yaml
+++ b/helm/flink-kubernetes-operator/templates/flink-operator.yaml
@@ -75,6 +75,10 @@ spec:
       {{- if .Values.operatorPod.topologySpreadConstraints }}
       topologySpreadConstraints: {{ toYaml .Values.operatorPod.topologySpreadConstraints | nindent 8 }}
       {{- end }}
+      {{- if .Values.operatorPod.initContainers }}
+      initContainers:
+      {{- toYaml .Values.operatorPod.initContainers | nindent 8 }}
+      {{- end }}
       containers:
         - name: {{ .Chart.Name }}
           image: {{ include "flink-operator.imagePath" . }}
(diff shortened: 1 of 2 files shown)


helm/flink-kubernetes-operator/values.yaml
```diff
@@ -190,6 +195,7 @@ fullnameOverride: ""
 jvmArgs:
   webhook: ""
   operator: ""
+  operatorLog: "-Dlog4j.configurationFile=/opt/flink/conf/log4j-operator.properties"
```
gyfora, line 198: Could the log args be simply part of jvmArgs.operator? (Although that may be a breaking change)

tweise: entry point simply combines them for both operator and webhook: `$LOG_CONFIG $JVM_ARGS`

however moving the log default to jvmArgs.operator and jvmArgs.webhook would introduce duplication and not be compatible with existing usage of those settings (log settings would be lost for everyone that uses jvmArgs currently.

helm/flink-kubernetes-operator/templates/flink-operator.yaml
```diff
@@ -115,7 +119,7 @@ spec:
             - name: FLINK_PLUGINS_DIR
               value: /opt/flink/plugins
             - name: LOG_CONFIG
-              value: -Dlog4j.configurationFile=/opt/flink/conf/log4j-operator.properties
+              value: {{ .Values.jvmArgs.operatorLog }}
```
morhidi, line 122: shouldn't this be `.Values.jvmArgs.operatorLogConfig` or `.Values.jvmArgs.logConfig`

mxm (approved):

helm/flink-kubernetes-operator/templates/flink-operator.yaml
tweise: @morhidi renamed to `logConfig`, assuming it should be shared.

morhidi (approved):

PR #978, opened 2025-05-06
[FLINK-37730][Job Manager] Expose JM exception as K8s exceptions

vsantwana:

This pull requests adds a new feature to periodically check for job exceptions using the FLINK REST API for getting the exceptions and raise them as kubernetes events. This feature will be helpful for monitoring systems that want to do a post processing on the job exceptions.

This is ONLY introduced for Application mode and NOT Session mode.

## Brief change log

*(for example:)*
 - Periodic pulling of job exceptions (only done when the job manager is NOT in a terminal state)
 - New `SYSTEM_ADVANCED` config for configuring the max number of exceptions reported and max lenght for stacktrace (defaults are 5 and 10 respectively)
  - Most of the information is in the k8s event
  - Introduced new utility method to trigger the event with annotations 
  - Introduced new test method to help testing the configuration

## Verifying this change

*(or)*

*(or)*

*(example:)*
  - *Added integration tests for end-to-end deployment with large payloads (100MB)*

Apart from the unit tests in this PR, this was tested using two simulations:
sql-test: Good running job, exception simulated by manually killing the TM
sql-test-failing: Job that has exception in the open method, repeatedly fails.

Both the exceptions were produced one at a time, and both simultaneously.

Screenshot 2025-05-22 at 9 34 55 PM

Screenshot 2025-05-22 at 9 32 02 PM

  - Core observer or reconciler logic that is regularly executed: Yes, this changes the observer.

## Documentation

  - Does this pull request introduce a new feature? Yes 
  - If yes, how is the feature documented?  I have to figure out the documentation part


rmetzger (commented):
I like this draft a lot. 

@gyfora can you take a quick look as well, or would you prefer this to be fully ready before you take a look?

gyfora:
> I like this draft a lot.
> 
> @gyfora can you take a quick look as well, or would you prefer this to be fully ready before you take a

> I like this draft a lot.
> 
> @gyfora can you take a quick look as well, or would you prefer this to be fully ready before you take a look?

Thank you!
I will try to take a look in the next 1-2 days :)

morhidi:
> > I like this draft a lot.
> > @gyfora can you take a quick look as well, or would you prefer this to be fully ready before you take a
> 
> > I like this draft a lot.
> > @gyfora can you take a quick look as well, or would you prefer this to be fully ready before you take a look?
> 
> Thank you! I will try to take a look in the next 1-2 days :)



> > I like this draft a lot.
> > @gyfora can you take a quick look as well, or would you prefer this to be fully ready before you take a
> 
> > I like this draft a lot.
> > @gyfora can you take a quick look as well, or would you prefer this to be fully ready before you take a look?
> 
> Thank you! I will try to take a look in the next 1-2 days :)



> > I like this draft a lot.
> > @gyfora can you take a quick look as well, or would you prefer this to be fully ready before you take a
> 
> > I like this draft a lot.
> > @gyfora can you take a quick look as well, or would you prefer this to be fully ready before you take a look?
> 
> Thank you! I will try to take a look in the next 1-2 days :)

It'd be great to catch and turn every job exception into a k8s event, not just for terminal job failures. It'd simplify collecting the historical diagnostic data before an actual crash occurs.

vsantwana:
> It'd be great to catch and turn every job exception into a k8s event, not just for terminal job failures. It'd simplify collecting historical diagnostic data before an actual crash occurs.

@morhidi Sorry I do not understand this. I am not checking for only terminal job failures. I am checking for all the failures, when the job is not in one of the terminal states.
Lmk if I have misunderstood your comment.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
```diff
@@ -299,9 +303,92 @@ public boolean reconcileOtherChanges(FlinkResourceContext<FlinkDeployment> ctx)
...
+        try {
+            var jobId = JobID.fromHexString(deployment.getStatus().getJobStatus().getJobId());
+            var history = ctx.getFlinkService().getJobExceptions(deployment, jobId, observeConfig);
+            if (history == null || history.getExceptionHistory() == null) {
+                return;
+            }
+            var exceptionHistory = history.getExceptionHistory();
+            var exceptions = exceptionHistory.getEntries();
+            if (exceptions.isEmpty()) {
+                LOG.info(String.format("No exceptions found in job exception history for jobId '%s'.", jobId));
```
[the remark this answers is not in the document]
vsantwana: Makes sense.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
```diff
@@ -299,9 +303,92 @@ public boolean reconcileOtherChanges(FlinkResourceContext<FlinkDeployment> ctx)
             return true;
         }
 
+        // check for JobManager exceptions if the REST API server is still up.
+        if (!ReconciliationUtils.isJobInTerminalState(deployment.getStatus())) {
+            observeJobManagerExceptions(ctx, deployment, observeConfig);
```
gyfora, line 308: Feels like the exception observation logic should be part of the observer and not the reconciler. The method names and everything points to that. No action is taken here, the job is not modified etc just observed.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/EventRecorder.java
```diff
@@ -315,6 +316,7 @@ public enum Reason {
         UnsupportedFlinkVersion,
         SnapshotError,
         SnapshotAbandoned,
+        JobManagerException,
```
gyfora, line 319: Should this be simply `JobException`?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/EventRecorder.java
```diff
@@ -280,7 +280,8 @@ public KubernetesClient getKubernetesClient() {
     /** The type of the events. */
     public enum Type {
         Normal,
-        Warning
+        Warning,
+        Error
```
gyfora, line 284: We could keep using the `Warning` type here. I wouldn't consider these events more severe than other operator errors that are recorded as warnings especially that the job often recovers from them

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
```diff
@@ -299,9 +303,92 @@ public boolean reconcileOtherChanges(FlinkResourceContext<FlinkDeployment> ctx)
...
+        if (taskName != null) {
+            combined.append("Task: ").append(taskName).append("\n");
+        }
+        if (endpoint != null) {
+            combined.append("Endpoint: ").append(endpoint).append("\n");
+        }
+        if (tmId != null) {
+            combined.append("TaskManager ID: ").append(tmId).append("\n");
+        }
+
```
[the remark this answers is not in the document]
gyfora: I think we should try to simplify the reported error here, I think adding labels, full stack trace etc is way out of scope for kubernetes events.

As a matter of comparison for errors that happen within the operator or where we have an Exception object we report it in a simplified format such as:
```
Cause 1 -> Cause 2 -> Cause 3
```
For this we use `ExceptionUtils#getExceptionMessage`.

I understand that here the format is a bit different but we should try to make this slimmer and more digestible :)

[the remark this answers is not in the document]
rmetzger: There might be scenarios where people want the full stack trace, in particular if other tools are listening for exception events on K8s.
How about adding labels, and introducing a config parameter to control whether we want the full stack trace or only the exception message in the event message?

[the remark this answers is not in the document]
gyfora: If you think the stack trace is useful in general and you would like to have it we can start with that. And introduce configs later if we feel that it's too verbose.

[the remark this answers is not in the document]
gyfora: But at least we should make it less verbose. For example remove from the message: `JobManager Exception at 2025-05-06 20:08:04:` these are included in the event itself.

Labels we could event put as labels on the event (this may be a wild idea but would make searching pretty convenient)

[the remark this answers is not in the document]
rmetzger: > Labels we could even put as labels on the event

Yes, that's what I meant by: "Instead of putting metadata like task, time etc. into the message itself, is it possible to put this maybe into metadata.annotations?" in my initial comment.
I wouldn't use labels, I would use annotations, because annotations are meant to be used for system provided information, whereas labels are meant for users to add additional names.

I'm obviously fine with just adding the exception to the message w/o the config option.

morhidi:
> > It'd be great to catch and turn every job exception into a k8s event, not just for terminal job failures. It'd simplify collecting historical diagnostic data before an actual crash occurs.
> 
> @morhidi Sorry I do not understand this. I am not checking for only terminal job failures. I am checking for all the failures, when the job is not in one of the terminal states. Lmk if I have misunderstood your comment.

nm I miss-read it at first glance

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
```diff
@@ -299,9 +303,92 @@ public boolean reconcileOtherChanges(FlinkResourceContext<FlinkDeployment> ctx)
...
+            if (exceptions.isEmpty()) {
+                LOG.info(String.format("No exceptions found in job exception history for jobId '%s'.", jobId));
+                return;
+            }
+            if (exceptionHistory.isTruncated()) {
+                LOG.warn(String.format("Job exception history is truncated for jobId '%s'. "
+                        + "Some exceptions are not shown.", jobId));
+            }
+            for (var exception : exceptions) {
+                emitJobManagerExceptionEvent(ctx, deployment, exception);
```
gyfora, line 335: We should be careful here as the currently logic has a very high potential for introducing large load on the kubernetes api server.

We need to add 2 safeguards here:
1. Only trigger events for exceptions that happened since the last reconciliation. We can filter based on timestamp, cache the last sent event or some other mechanism but we cannot rely on the EventRecorder here as it will destroy the k8s api server

2. In general have a limit on number of exception events triggered per reconciliation and simply drop the rest. Jobs can fail/generate exceptions very quickly. So I would not record more than let's say 5-10 errors per reconciliation this way (we can make this configurable but let's have a conservative starting limit here to avoid some bad surprises)

vsantwana: I had thought about the first safeguard but that would have involved changing the `FlinkDeploymentStatus` to store something like the last exception time, which I wasn't sure of.

gyfora: A simpler alternative implementation that would not involve any status changes would be simply to have an in-memory cache of the last emitted exception timestamp on a per resource level.  I would go down this route instead

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
vsantwana: Moved to ApplicationObserver.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/status/FlinkDeploymentStatus.java
```diff
@@ -55,4 +55,7 @@ public class FlinkDeploymentStatus extends CommonStatus<FlinkDeploymentSpec> {
 
     /** Information about the TaskManagers for the scale subresource. */
     private TaskManagerInfo taskManager;
+
+    /** Information about when the last exception was recorded for this deployment. */
+    private String lastRecordedExceptionTimestamp;
```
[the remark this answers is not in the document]
vsantwana: Yes, sorry I missed that last message, changing this

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/ApplicationObserver.java
```diff
@@ -17,22 +17,50 @@
...
+        final long lastTimestamp;
+
+        ExceptionCacheEntry(String jobId, long lastTimestamp) {
+            this.jobId = jobId;
+            this.lastTimestamp = lastTimestamp;
+        }
+    }
+
+    @VisibleForTesting
+    final Map<String, ExceptionCacheEntry> lastRecordedExceptionCache = new ConcurrentHashMap<>();
```
gyfora, line 58: The cache key should be ResourceID (you can call `ResourceID.fromResource(resource)`) which will be the namespace + name. Job Id is not globally unique, and users can override it arbitrarily. It also doesn't really matter as we have only a single job active per resource

helm/flink-kubernetes-operator/crds/flinkdeployments.flink.apache.org-v1.yml
```diff
@@ -1,4 +1,4 @@
-# Generated by Fabric8 CRDGenerator, manual edits might get overwritten!
+helm/flink-kubernetes-operator/crds/flinkdeployments.flink.apache.org-v1.yml# Generated by Fabric8 CRDGenerator, manual edits might get overwritten!
```
gyfora, line 1: What caused this change?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/ApplicationObserver.java
```diff
@@ -17,22 +17,50 @@
...
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 import static org.apache.flink.kubernetes.operator.config.KubernetesOperatorConfigOptions.OPERATOR_CLUSTER_HEALTH_CHECK_ENABLED;
 
 /** The observer of {@link org.apache.flink.kubernetes.operator.config.Mode#APPLICATION} cluster. */
 public class ApplicationObserver extends AbstractFlinkDeploymentObserver {
 
+    /** The cache entry for the last recorded exception timestamp. */
+    public static final class ExceptionCacheEntry {
```
gyfora, line 47: Also one other consideration for the cache is that we need a mechanism to remove the resource from it once it is deleted. So the controller cleanup method should remove from the cache.

Due to this reason, I think we should not have the cache within the Observer module but a nicer place for it would be to put it into the `FlinkResourceContextFactory` and the we could expose this extra cached information through the `FlinkResourceContext`. If we make the info mutable then the observer can update it through the context.

I know this sounds a bit over engineered right now but we actually have some other similar logic scattered around the codebase that we could then standardize on this approach :)

gyfora (changes requested):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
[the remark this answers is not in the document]
vsantwana: @gyfora @rmetzger I have truncated the stacktrace, and moved everything else to annotation. Can you please take a look? If this looks okay, I will work on formalizing the PR.

Thanks

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/ApplicationObserver.java
vsantwana: I have made resource key 
```
String resourceKey =
                    deployment.getMetadata().getNamespace()
                            + "/"
                            + deployment.getMetadata().getName()
```

Wasn't aware of the ResourceID class. Will update this

helm/flink-kubernetes-operator/crds/flinkdeployments.flink.apache.org-v1.yml
vsantwana: I don't know. Will revert manually

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/ApplicationObserver.java
vsantwana: The JobID is in ExceptionCacheEntry to make sure that we are tracking the exception of the correct JobId, if for the same ResourceID, the jobId has changed, then it is treated like we are seeing the exception for the first time and only the number of exception limit is applied

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/ApplicationObserver.java
```diff
@@ -56,6 +84,149 @@ protected void observeFlinkCluster(FlinkResourceContext<FlinkDeployment> ctx) {
         }
     }
 
+    @Override
+    protected void observeJobManagerExceptions(FlinkResourceContext<FlinkDeployment> ctx) {
```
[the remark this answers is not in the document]
vsantwana: I have limited understanding of what this change would mean for Session Clusters and hence did not want to disrupt anything more than needed. If the changes look good for Session Clusters as well, I will move it to JobStatusObserver.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/ApplicationObserver.java
gyfora: Ah sorry that makes sense, I didn't check it carefully enough and thought that the jobid was used as the key too

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/ApplicationObserver.java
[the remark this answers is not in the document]
gyfora: I think this logic is valid for both Applications and SessionJobs (and not for Session Clusters) the place where job related observation happens is the JobStatusObserver

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
vsantwana: Moved to JobStatusObserver

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/EventUtils.java
```diff
@@ -104,6 +107,33 @@ public static Event findExistingEvent(
...
+            setAnnotations(event, annotations);
+            setEventTime(event, Instant.ofEpochMilli(eventTime));
+            createOrReplaceEvent(client, event).ifPresent(eventListener);
+            return true;
+        }
+    }
+
     public static boolean createIfNotExists(
             KubernetesClient client,
             HasMetadata target,
```
rmetzger, line 139: Why can't we call `createWithAnnotationsIfNotExists()` from this method to avoid code duplication?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkResourceContextFactory.java
```diff
@@ -98,15 +115,21 @@ public FlinkStateSnapshotContext getFlinkStateSnapshotContext(
             var flinkDep = (FlinkDeployment) resource;
             return (FlinkResourceContext<CR>)
                     new FlinkDeploymentContext(
-                            flinkDep, josdkContext, resMg, configManager, this::getFlinkService);
+                            flinkDep,
+                            josdkContext,
+                            resMg,
+                            configManager,
+                            this::getFlinkService,
+                            lastRecordedExceptionCache);
```
gyfora, line 123: I recommend passing something like `lastRecordedExceptionCache.computeIfAbsent(resourceId, () -> new CacheEntry())`

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/EventUtils.java
vsantwana: I ha thought about it but I did not do it because of the event time. In our case we had decided to set the exception time as event time, but I am not aware of how should it happen for other k8s events, so I kept them separate with the cost of duplicated code.
cc @gyfora

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
```diff
@@ -69,17 +75,23 @@ public boolean observe(FlinkResourceContext<R> ctx) {
...
+                var newJobStatus = newJobStatusOpt.get();
+                updateJobStatus(ctx, newJobStatus);
                 ReconciliationUtils.checkAndUpdateStableSpec(resource.getStatus());
+                // now check if the job is in a terminal state. This might not be need as we have
+                // already
+                // verified that the REST API server is available
+                // now check if the job is NOT in a terminal state
+                if (!newJobStatus.getJobState().isGloballyTerminalState()) {
+                    observeJobManagerExceptions(ctx);
+                }
```
gyfora, line 94: Maybe a better logic would be to check that old job status is not in a terminal state. 
With the current logic if the job fails with a fatal error (goes into terminal FAILED state) the exception(s) wouldn't be reported if I understand correctly.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
```diff
@@ -95,6 +107,143 @@ public boolean observe(FlinkResourceContext<R> ctx) {
...
+        eventRecorder.triggerEventOnceWithAnnotations(
+                exceptionTime.toEpochMilli(),
+                ctx.getResource(),
+                EventRecorder.Type.Warning,
+                EventRecorder.Reason.JobException,
+                eventMessage.toString().trim(),
+                EventRecorder.Component.JobManagerDeployment,
+                "jobmanager-exception-" + keyMessage.hashCode(),
+                ctx.getKubernetesClient(),
+                annotations);
```
gyfora, line 244: It probably makes sense to trim the values of the annotations map here or in the `eventRecorder.triggerEventOnceWithAnnotations` as there is a 63 character limit and some character limitations on the annotations in Kubernetes. Maybe AI can help you create a filter/formatter :)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
```diff
@@ -95,6 +107,143 @@ public boolean observe(FlinkResourceContext<R> ctx) {
...
+        try {
+            var jobId = JobID.fromHexString(jobStatus.getJobId());
+            // TODO: Ideally the best way to restrict the number of events is to use the query param
+            // `maxExceptions`
+            //  but the JobExceptionsMessageParameters does not expose the parameters and nor does
+            // it have setters.
+            var history =
+                    ctx.getFlinkService()
+                            .getJobExceptions(
+                                    resource, jobId, ctx.getDeployConfig(resource.getSpec()));
```
gyfora, line 129: We should simply use `ctx.getObserveConfig()` here that's what we use for accessing the currently running cluster (to observe it)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
```diff
@@ -845,6 +849,37 @@ public RestClusterClient<String> getClusterClient(Configuration conf) throws Exc
                 (c, e) -> new StandaloneClientHAServices(restServerAddress));
     }
 
+    @Override
+    public JobExceptionsInfoWithHistory getJobExceptions(
+            AbstractFlinkResource resource, JobID jobId, Configuration deployConfig) {
```
gyfora, line 854: Should rename the config to observeConfig as that what's should be passed I noted in another comment

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkResourceContextFactory.java
```diff
@@ -52,6 +54,23 @@ public class FlinkResourceContextFactory {
 
     private static final Logger LOG = LoggerFactory.getLogger(FlinkResourceContextFactory.class);
 
+    /** The cache entry for the last recorded exception timestamp for a JobID. */
+    @Getter
+    @Setter
```
gyfora, line 59: Could be simply `@Data`

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserverTest.java
```diff
@@ -147,6 +157,128 @@ void testFailed() throws Exception {
...
+                kubernetesClient
+                        .v1()
+                        .events()
+                        .inNamespace(deployment.getMetadata().getNamespace())
+                        .list()
+                        .getItems();
+        assertEquals(1, events.size());
+        assertTrue(events.get(0).getMessage().contains("NewException"));
+    }
+
```
gyfora, line 281: Would probably good to add a test for a job that failed terminally, to make sure the events are recorded (see my other comment)

gyfora (changes requested):
I think this is looking pretty good now, I added a few minor comments still

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
vsantwana: For my understanding 
What is the difference between the two configs?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
vsantwana: But if the current status (i.e the newStatus) is in terminal state, the API server won't be available right?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
vsantwana: Add a preliminary check on annotations

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
gyfora: On a high level the observedConfig corresponds to the running job while the deployConfig corresponds to the config of the current spec. When the user makes a spec change these 2 can differ, so maybe they change some port settings etc, so its important to use the observeConfig when interacting with the job in the observer

gyfora (approved):
Looks good! As a future followup we could think about reducing the number of REST API calls we make to fetch exceptions. 

At the moment this is done on every step but based on the job details that we get in previous steps in the observers we may be able to deduct that the job did not fail since the last time we checked so exceptions do not need to be queried.

If you could open a follow up ticket for that I think that would be nice :)

gyfora:
You need to regenerate the docs: `mvn clean install -DskipTests -Pgenerate-doc`

gyfora:
I hit the following error while running locally: 
```
When I tested this locally I hit the following issue:

```
2025-05-21 15:42:22,998 o.a.f.k.o.o.JobStatusObserver  [WARN ][default/basic-example] Failed to fetch JobManager exception info.
io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: POST at: https://10.96.0.1:443/api/v1/namespaces/default/events. Message: Event in version "v1" cannot be handled as a Event: parsing time "2025-05-21T15:41:41.602Z[UTC]" as "2006-01-02T15:04:05.000000Z07:00": cannot parse ".602Z[UTC]" as ".000000". Received status: Status(apiVersion=v1, code=400, details=null, kind=Status, message=Event in version "v1" cannot be handled as a Event: parsing time "2025-05-21T15:41:41.602Z[UTC]" as "2006-01-02T15:04:05.000000Z07:00": cannot parse ".602Z[UTC]" as ".000000", metadata=ListMeta(_continue=null, remainingItemCount=null, resourceVersion=null, selfLink=null, additionalProperties={}), reason=BadRequest, status=Failure, additionalProperties={}).

```

So something seems to be off with the time handling

File: flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java [hunk 20]
```diff
@@ -95,6 +105,141 @@ public boolean observe(FlinkResourceContext<R> ctx) {
...
+                Instant exceptionTime = Instant.ofEpochMilli(exception.getTimestamp());
+                if (lastRecorded != null && exceptionTime.isBefore(lastRecorded)) {
+                    continue;
+                }
+
+                emitJobManagerExceptionEvent(ctx, exception, exceptionTime, maxStackTraceLines);
+                if (++count >= maxEvents) {
+                    break;
+                }
+            }
```
gyfora, line 171: I wonder if we should take the last `maxEvents` exceptions from the list and start with those first so that new errors are always reported, but that may cause us to not report some later

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
```diff
@@ -95,6 +108,148 @@ public boolean observe(FlinkResourceContext<R> ctx) {
...
+                if (lastRecorded != null && exceptionTime.isBefore(lastRecorded)) {
+                    continue;
+                }
+                emitJobManagerExceptionEvent(ctx, exception, exceptionTime, maxStackTraceLines);
+                if (++count >= maxEvents) {
+                    break;
+                }
+            }
+            ctx.getExceptionCacheEntry().setJobId(currentJobId);
+            ctx.getExceptionCacheEntry().setLastTimestamp(now.toEpochMilli());
```
gyfora, line 184: Shouldn't this be the max exception timestamp? It could happen that there are job exceptions between getting it in the rest api and emitting them here and those would be missed if we set a higher timestamp based on `now`

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
```diff
@@ -95,6 +108,148 @@ public boolean observe(FlinkResourceContext<R> ctx) {
...
+                    Comparator.comparingLong(
+                                    JobExceptionsInfoWithHistory.RootExceptionInfo::getTimestamp)
+                            .reversed());
+
+            int count = 0;
+            for (var exception : sortedExceptions) {
+                Instant exceptionTime = Instant.ofEpochMilli(exception.getTimestamp());
+                // Skip already recorded exceptions
+                if (lastRecorded != null && exceptionTime.isBefore(lastRecorded)) {
+                    continue;
```
gyfora, line 176: We can also break here as the events are sorted

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
```diff
@@ -95,6 +108,148 @@ public boolean observe(FlinkResourceContext<R> ctx) {
...
+        String exceptionName = exception.getExceptionName();
+        if (exceptionName == null || exceptionName.isBlank()) {
+            return;
+        }
+
+        Map<String, String> annotations = new HashMap<>();
+        annotations.put(
+                "event-time-readable",
+                DateTimeUtils.readable(exceptionTime, ZoneId.systemDefault()));
+        annotations.put("event-timestamp-millis", String.valueOf(exceptionTime.toEpochMilli()));
```
gyfora, line 205: I think this should be a single field called: `exception-timestamp` and should follow the Kubernetes timestamp formatting standard (Instant.toString should work I guess)

gyfora (changes requested):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
```diff
@@ -95,6 +108,155 @@ public boolean observe(FlinkResourceContext<R> ctx) {
...
+            for (var exception : sortedExceptions) {
+                Instant exceptionTime = Instant.ofEpochMilli(exception.getTimestamp());
+                // Skip already recorded exceptions
+                if (lastRecorded != null && !exceptionTime.isAfter(lastRecorded)) {
+                    break;
+                }
+                emitJobManagerExceptionEvent(ctx, exception, exceptionTime, maxStackTraceLines);
+                if (latestSeen == null || exceptionTime.isAfter(latestSeen)) {
+                    latestSeen = exceptionTime;
+                }
```
gyfora, line 181: logically speaking it's enough to check that `latestSeen == null` as the events are sorted

gyfora (approved):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
```diff
@@ -95,6 +108,155 @@ public boolean observe(FlinkResourceContext<R> ctx) {
...
+            exception
+                    .getFailureLabels()
+                    .forEach((k, v) -> annotations.put("failure-label-" + k, v));
+        }
+
+        StringBuilder eventMessage = new StringBuilder(exceptionName);
+        String stacktrace = exception.getStacktrace();
+        if (stacktrace != null && !stacktrace.isBlank()) {
+            String[] lines = stacktrace.split("\n");
+            eventMessage.append("\n\nStacktrace (truncated):\n");
```
gyfora, line 234: Can we maybe remove this line completely? it seems to just increase the messages and adds an empty line:

image

Also seems like the first line with the exception name and the stack trace is basically duplicated

vsantwana: Done

PR #981, opened 2025-05-19
[ENGT-2178] Publish in prod ECR

OctavioBR:
This enables https://github.com/grepr/grepr-server/blob/main/flink/kubernetes-operator/Dockerfile to pull it as base image both from dev and prod ECR repos.


OctavioBR:
Sorry!

PR #983, opened 2025-05-26
[FLINK-37730] Improve exception recording ts initialization + 2.0 compatibility

gyfora:

Initialize last triggered event timestamp correctly from kubernetes events + fix 2.0 compatibility

## Verifying this change

Manually verified (Flink 1.18-2.0) +  Unit tests


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
```diff
@@ -132,8 +134,50 @@ protected void observeJobManagerExceptions(FlinkResourceContext<R> ctx) {
...
+                        // If the last job exception was a long time ago, then there is no point in
+                        // checking in k8s.
+                        lastExceptionTs = maxJobExceptionTs;
+                    } else {
+                        // If there were recent exceptions, we check the triggered events from kube
+                        // to make sure we don't double trigger
+                        lastExceptionTs =
+                                EventUtils.findLastJobExceptionTsFromK8s(
+                                                ctx.getKubernetesClient(), resource)
+                                        .orElse(Instant.now().minus(MAX_K8S_EVENT_AGE));
```
vsantwana, line 167:
```suggestion
                                        .orElse(Instant.now().minus(k8sExpirationTs));
```

vsantwana:
Thanks @gyfora for the PR!
Left two very minor comments

vsantwana (approved):
LGTM!
Thanks @gyfora

rmetzger (approved):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
```diff
@@ -132,8 +134,50 @@ protected void observeJobManagerExceptions(FlinkResourceContext<R> ctx) {
...
+                    var maxJobExceptionTs =
+                            exceptions.stream()
+                                    .map(e -> Instant.ofEpochMilli(e.getTimestamp()))
+                                    .max(Comparator.naturalOrder())
+                                    .orElseThrow();
+
+                    if (maxJobExceptionTs.isBefore(k8sExpirationTs)) {
+                        // If the last job exception was a long time ago, then there is no point in
+                        // checking in k8s.
+                        lastExceptionTs = maxJobExceptionTs;
```
mxm: Any reason for this optimization? It complicates the code by adding another setting. It also requires the user to tune just another setting. There is no harm in calling out to the k8s api regularly to fetch events.

gyfora: there is no config for this (nothing to tune) and the optimization can be very important when the operator starts up because then the cache is empty and it would fetch events for every single job. In most cases this filter completely eliminates that so this greatly reduces the startup api server load

mxm: Fair point. The value is hardcoded. We would only query for the jobs with exceptions, but still those could amount to quite some jobs.

PR #987, opened 2025-07-02
[FLINK-38033] Fix accidental upgrade snapshot dispose bug

gyfora:

Fix a bug where if the snapshot where if periodic snapshot cleanup and snapshot CR are both enabled, upgrade save points can be accidentally deleted.

## Verifying this change

new unit tests added

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractJobReconciler.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractJobReconciler.java
@@ -409,25 +409,30 @@ public abstract class AbstractJobReconciler<
                         conf.get(KubernetesOperatorConfigOptions.OPERATOR_SAVEPOINT_FORMAT_TYPE)
                                 .name());
 
-        FlinkStateSnapshotUtils.createUpgradeSnapshotResource(
-                conf,
-                ctx.getOperatorConfig(),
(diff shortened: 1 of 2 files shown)


salvalcantara:
Thanks @gyfora @gaborgsomogyi!

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractJobReconciler.java
```diff
@@ -409,25 +409,30 @@ protected void setUpgradeSavepointPath(
...
+            // and disposed immediately
+            var savepoint =
+                    new Savepoint(
+                            cancelTs.toEpochMilli(),
+                            savepointLocation,
+                            SnapshotTriggerType.UPGRADE,
+                            savepointFormatType,
+                            null);
+            jobStatus.getSavepointInfo().updateLastSavepoint(savepoint);
+        }
```
salvalcantara, line 435: @gyfora Maybe you could  use `isSnapshotResourceEnabled` here? Looks more explicit / less indirect to me. What if creation of the CR fails for example? Could that happen?

gyfora: I don't think that it could fail without an error and even if that were the case it would probably be better then to record this info still. The bug was caused by the information recorded in 2 places and this interfering with the savepoint cleanup logic

PR #994, opened 2025-07-10
[FLINK-28648][Kubernetes Operator] Allow session deletion to block on any running job

nishita-09:

*This pull request adds a configuration to allow session cluster cleanup blocking in case unmanaged jobs are present. That is the `FlinkDeployment` deletion is blocked in presence of `FlinkSessionJobs` as well as if jobs submitted through CLI are in `non_terminal` states.*

## Brief change log

*(for example:)*
 - Added `getUnmanagedJobs() `method to detect CLI-submitted jobs not managed by FlinkSessionJob resources
- Modified `cleanupInternal()` to check for unmanaged jobs when blocking is enabled
- Improved error messages and logging.

## Verifying this change

 - *Added ` testGetUnmanagedJobs` test for validating `getUnmanagedJobs`*
   -  This test would validate if the function correctly identifies jobs which are not controlled by `SessionJob` and are in `non-terminal` state.

 - *Manually verified the change by running a  cluster with 2 JobManagers and  submitted a SessionJob as well as CLI Jobs.*
    - *Config: `session.block-on-unmanaged-jobs: true`*
    
    - *Deleted `flinkDeployment` -> Generates Event `CleanupFailed` in `flinkDeployment` due to presence of sessionjobs*
    - *Deleted `flinkSessionJob` -> Generates Event `CleanupFailed` in `flinkDeployment` due to presence of unmanaged Jobs -> `flinkSessionJob` was deleted.*
    - *Cancelled CLI submitted job -> Generates Event `Cleanup` after `ReconcileInterval` -> CLI job was cancelled and then the `flinkDeployment` was cleaned up.*
   
  Screenshot 2025-07-10 at 10 04 51 PM
  Screenshot 2025-07-10 at 10 06 54 PM

  
Screenshot 2025-07-14 at 2 15 32 PM

  
  -  Config: `session.block-on-unmanaged-jobs: false`
  -  Manually verified the change by running a  cluster with 2 JobManagers and  submitted a SessionJob as well as CLI Jobs.
  -  Deleted `flinkDeployment` -> Generates Event `CleanupFailed` in `flinkDeployment` due to presence of sessionjobs
  -  Deleted `flinkSessionJob` -> Generates Event `Cleanup` inspite of running CLI jobs being present. -> `SessionJob` is deleted , followed by `flinkDeployment` .

 

## Documentation

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/SessionReconciler.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/SessionReconciler.java
@@ -197,22 +197,20 @@ public class SessionReconciler
                                 nonTerminalJobs.stream()
                                         .map(JobID::toHexString)
                                         .collect(Collectors.toList()));
-                if (eventRecorder.triggerEvent(
+                eventRecorder.triggerEvent(
                         deployment,
                         EventRecorder.Type.Warning,
                         EventRecorder.Reason.CleanupFailed,
                         EventRecorder.Component.Operator,
                         error,
-                        ctx.getKubernetesClient())) {
-                    LOG.warn(error);
-                }
+                        ctx.getKubernetesClient());
                 return DeleteControl.noFinalizerRemoval()
                         .rescheduleAfter(ctx.getOperatorConfig().getReconcileInterval().toMillis());
             }
         }
 
         LOG.info("Stopping session cluster");
-        var conf = ctx.getDeployConfig(ctx.getResource().getSpec());
+        var conf = ctx.getObserveConfig();
         ctx.getFlinkService()
                 .deleteClusterDeployment(
                         deployment.getMetadata(), deployment.getStatus(), conf, true);


nishita-09:
@gyfora The build was failing due to missing documentation about added configuration. I have added those in the new commit now.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/reconciler/deployment/SessionReconcilerTest.java
```diff
@@ -139,4 +149,164 @@ public void testSetOwnerReference() throws Exception {
...
+
+        Set<FlinkSessionJob> sessionJobs = new HashSet<>();
+        sessionJobs.add(managedSessionJob1);
+        sessionJobs.add(managedSessionJob2);
+
+        // Test with blocking enabled - should identify running unmanaged jobs correctly
+        var context = TestUtils.createContextWithReadyFlinkDeployment(kubernetesClient);
+        var resourceContext = getResourceContext(deployment, context);
+
+        // Use reflection to access the private getUnmanagedJobs method
```
gyfora: We should not use reflection for this, we can make the method protected with the @visiblefor testing annotation

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/SessionReconciler.java
```diff
@@ -120,6 +126,61 @@ private void recoverSession(FlinkResourceContext<FlinkDeployment> ctx) throws Ex
...
+                            .filter(jobId -> jobId != null)
+                            .map(JobID::fromHexString)
+                            .collect(Collectors.toSet());
+
+            // Get job IDs that are not managed by FlinkSessionJob resources and are currently
+            // running
+            //    FAILED(TerminalState.GLOBALLY)
+            //    CANCELED(TerminalState.GLOBALLY)
+            //    FINISHED(TerminalState.GLOBALLY)
+            // Above terminal states are not considered since they are no longer active jobs
```
gyfora: This comment feels a bit too verbose, enough to say only consider non terminal jobs (feels a bit like AI generated)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/SessionReconciler.java
```diff
@@ -120,6 +126,61 @@ private void recoverSession(FlinkResourceContext<FlinkDeployment> ctx) throws Ex
                 .setJobManagerDeploymentStatus(JobManagerDeploymentStatus.DEPLOYING);
     }
 
+    /**
+     * Detects unmanaged jobs running in the session cluster. Unmanaged jobs are jobs that exist in
+     * the Flink cluster but are not managed by FlinkSessionJob resources.
+     */
+    private Set<JobID> getUnmanagedJobs(
+            FlinkResourceContext<FlinkDeployment> ctx, Set<FlinkSessionJob> sessionJobs) {
```
gyfora: I don't think that we need to pass sessionJobs here. if the flag is enabled, any running job should simply block it. We can simplify this logic a lot

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/SessionReconciler.java
```diff
@@ -120,6 +126,61 @@ private void recoverSession(FlinkResourceContext<FlinkDeployment> ctx) throws Ex
...
+        LOG.info(
+                "Starting unmanaged job detection for session cluster: {}",
+                ctx.getResource().getMetadata().getName());
```
gyfora: this should be on debug level, also no need to include resource name/info in the log message. Its in the MDC already

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/SessionReconciler.java
gyfora: We can simply replace this method with something like getNonTerminalJobIds() or boolean anyNonTerminalJobs()

That would be enough for this feature.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/SessionReconciler.java
```diff
@@ -143,13 +204,41 @@ public DeleteControl cleanupInternal(FlinkResourceContext<FlinkDeployment> ctx)
...
+                                        .collect(Collectors.toList()));
+                LOG.warn(error);
+                if (eventRecorder.triggerEvent(
+                        deployment,
+                        EventRecorder.Type.Warning,
+                        EventRecorder.Reason.CleanupFailed,
+                        EventRecorder.Component.Operator,
+                        error,
+                        ctx.getKubernetesClient())) {
+                    LOG.warn(error);
```
gyfora: you are logging the error twice, you don't need to log it at all as the event triggering already logs it.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/SessionReconciler.java
nishita-09: I actually thought this would be necessary  to clarify the functionality of this. 😢  Sure will remove if its too verbose

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/SessionReconciler.java
nishita-09: Sure, will simplify this further. Thanks for the review

nishita-09: @gyfora I have pushed another commit to address the comments here. I have stuck with getNonTerminalJobIds() to ensure the Event contains the list of job IDs that are not terminated for better observability for the user.

nishita-09:
@gyfora I ran these tests on my local, they seem to be passing. I am not sure what is causing the issue here. Can you have a look at it if possible?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/SessionReconciler.java
```diff
@@ -143,13 +181,41 @@ public DeleteControl cleanupInternal(FlinkResourceContext<FlinkDeployment> ctx)
...
+                                        .collect(Collectors.toList()));
+                if (eventRecorder.triggerEvent(
+                        deployment,
+                        EventRecorder.Type.Warning,
+                        EventRecorder.Reason.CleanupFailed,
+                        EventRecorder.Component.Operator,
+                        error,
+                        ctx.getKubernetesClient())) {
+                    LOG.warn(error);
+                }
```
gyfora: we can remove the if branch and the logging. Event triggering already creates logs we don't need both I think

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/SessionReconciler.java
```diff
@@ -143,13 +181,41 @@ public DeleteControl cleanupInternal(FlinkResourceContext<FlinkDeployment> ctx)
...
+                return DeleteControl.noFinalizerRemoval()
+                        .rescheduleAfter(ctx.getOperatorConfig().getReconcileInterval().toMillis());
+            }
+        }
+
+        LOG.info("Stopping session cluster");
+        var conf = ctx.getDeployConfig(ctx.getResource().getSpec());
+        ctx.getFlinkService()
+                .deleteClusterDeployment(
+                        deployment.getMetadata(), deployment.getStatus(), conf, true);
```
gyfora: I think there is a bug in this existing logic, instead of getting the deployconfig, here we should just use `ctx.getObserveConfig()`

gyfora (commented):
Added 2 minor comments, otherwise looks good!

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/SessionReconciler.java
nishita-09: @gyfora  I see this for sessionjob event as well , should we remove [here](https://github.com/apache/flink-kubernetes-operator/pull/994/files#diff-3e6121d0b2bd1cea1993203ec9882d1bc336f34fe983d69e0f1a6f5f713c52b0R180) too.

nishita-09:
> Added 2 minor comments, otherwise looks good!

@gyfora Thank you for reviewing, I have made the changes. Please do trigger the workflows on your end. Also do let me know if we should also remove the `LOG.warn()` statement in sessionjob section too?

gyfora:
> > Added 2 minor comments, otherwise looks good!
> 
> @gyfora Thank you for reviewing, I have made the changes. Please do trigger the workflows on your end. Also do let me know if we should also remove the `LOG.warn()` statement in sessionjob section too?

We can leave it as is for now :)

gyfora (approved):

PR #1000, opened 2025-07-27
[FLINK-35746][Kubernetes-Operator] Add getJobConfiguration and getJobCheckpointConfiguration

nishita-09:

*This pull request adds a new capability to the operator that ensures the configuration used for observation and health checks is consistent with the actual resultant configuration set both programmatically (in the user's `main()` method) and via the deployment spec.

Today, the operator infers running application/cluster settings solely from the K8s deployment spec and operator defaults. While this is mostly fine, many configs can be programmatically changed by the user in their app's main method (e.g., `env.enableCheckpointing(3000, AT_LEAST_ONCE)`), which takes precedence at runtime but is invisible to the operator. This leads to incorrect health check windows, missed in-place scaling opportunities, and inconsistent behavior in checkpoint stall detection.

Additionally, JM-level configuration (like `jobmanager.scheduler`, `state.savepoints.dir`, `high-availability.*`) set via operator defaults can drift if defaults are changed while a job is still running — the operator's reconstructed config diverges from the actual JM runtime config.*

## Brief change log

  - Add `getJobManagerConfiguration`, `getJobConfiguration`, and `getJobCheckpointConfiguration` to `FlinkService` interface, implemented in `AbstractFlinkService` using Flink REST API endpoints (`/jobs/:jobid/jobmanager/config`, `/jobs/:jobid/config`, `/jobs/:jobid/checkpoints/config`).
  - Add `FlinkRuntimeConfigurationUtils` utility class with compile-time safe `CheckpointConfigMapping` enum to map checkpoint REST API JSON fields to Flink configuration keys.
  - Add Guava-backed runtime configuration cache in `FlinkConfigManager` with `putRuntimeConfig`, `getRuntimeConfig`, `invalidateRuntimeConfig`, and `applyCachedRuntimeConfig` methods. Cache is keyed by `{namespace, name, jobId}` and expires after a configurable timeout.
  - Fetch runtime configuration in `JobStatusObserver` when a job reaches RUNNING and cache is empty. The fetch uses a 3-layer strategy: JM config (base layer) -> job execution config (override) -> checkpoint config (programmatic overrides win). REST calls are made **once per job lifecycle** and cached.
  - Apply cached runtime configuration overrides in `FlinkResourceContext.getObserveConfig()` so all downstream consumers (health evaluator, reconciler, scaler) see the actual runtime values.
  - Invalidate the runtime config cache in `ApplicationReconciler` and `SessionJobReconciler` on cancel/upgrade, ensuring fresh config is fetched when a new job starts.

## Verifying this change

*(or)*

[WIP]
*(or)*

- **Unit tests** (2126 pass, 0 failures):
    - `AbstractFlinkServiceTest`: `testGetJobManagerConfiguration`, `testGetJobManagerConfigurationReturnsEmptyOnFailure` : verifies JM config REST endpoint parsing and graceful degradation.
    - `AbstractFlinkServiceTest`: `testAllCheckpointConfigInfoFieldNamesCoveredByMapping`, `testAllExecutionConfigInfoFieldNamesCoveredByMapping` :  reflection-based tests that fail at build time if Flink adds new `FIELD_NAME_*` constants not covered by our mapping.
    - `AbstractFlinkServiceTest`: `testMapJobConfigurationMapsAllExpectedFields`, `testMapJobConfigurationHandlesNullGracefully` :  verifies job config mapping correctness and null safety.
    - `FlinkConfigManagerTest`: runtime config cache put/get/invalidate/apply tests.
    - `JobStatusObserverTest`: verifies `fetchAndCacheRuntimeConfig` populates cache on first observe, skips on cache hit, handles partial/full REST failures gracefully, and invalidates on job upgrade.

- **Integration validation**
 - **1. Deployed the job with application mode with deployment (checkpoint disabled) but enabled via the user in job logic.**
  - *Here, are the logs from `ClusterHealthEvaluator` which confirm that the operator now recognised the user defined configs.*
  - Before the change:
Screenshot 2025-07-24 at 5 09 16 PM

- After the change:
Screenshot 2025-07-24 at 5 13 06 PM

   - **2. Initial cache population** -  53 entries from 3 REST endpoints cached in ~80ms:
```
    15:09:55,712 JobStatusObserver    [DEBUG][flink/basic-example] Fetching runtime configuration
    15:09:55,712 AbstractFlinkService  [DEBUG][flink/basic-example] Fetching JobManager configuration
    15:09:55,764 AbstractFlinkService  [DEBUG][flink/basic-example] Fetched 40 JobManager configuration entries
    15:09:55,764 AbstractFlinkService  [DEBUG][flink/basic-example] Fetching job configuration
    15:09:55,776 AbstractFlinkService  [DEBUG][flink/basic-example] Fetched 2 job configuration entries
    15:09:55,776 AbstractFlinkService  [DEBUG][flink/basic-example] Fetching checkpoint configuration
    15:09:55,791 FlinkConfigManager   [DEBUG][flink/basic-example] Cached runtime configuration with 53 entries
```
   - **3. Cache hit on subsequent observe cycles** - zero re-fetches
```
       15:10:10,946 JobStatusObserver  [DEBUG][flink/basic-example] Runtime configuration already cached
       15:10:26,125 JobStatusObserver  [DEBUG][flink/basic-example] Runtime configuration already cached
```
  - **4. Cache invalidation on job suspension:**
`       15:11:53,127 FlinkConfigManager  [DEBUG][flink/basic-example] Invalidated runtime configuration cache
`
   - **5. Cache re-population after job resume** (new job ID, fresh REST fetch):
```
    15:13:11,133 JobStatusObserver    [DEBUG][flink/basic-example] Fetching runtime configuration
    15:13:11,133 AbstractFlinkService  [DEBUG][flink/basic-example] Fetching JobManager configuration
    15:13:11,139 AbstractFlinkService  [DEBUG][flink/basic-example] Fetched 40 JobManager configuration entries
    15:13:11,156 FlinkConfigManager   [DEBUG][flink/basic-example] Cached runtime configuration with 53 entries
```
 - **6. Parallelism change** triggers invalidation + re-population:
```
    15:14:53,150 FlinkConfigManager  [DEBUG][flink/basic-example] Invalidated runtime configuration cache
    15:15:32,508 JobStatusObserver   [DEBUG][flink/basic-example] Fetching runtime configuration
```
  - **7. Programmatic config override detection** - deployed with `interval=10000, EXACTLY_ONCE` in spec, app code overrides to `3000ms, AT_LEAST_ONCE`:
```
    15:29:00,908 FlinkRuntimeConfigurationUtils [DEBUG][flink/programmatic-ckpt-test] Mapped 15 checkpoint
      configuration entries: {execution.checkpointing.mode=AT_LEAST_ONCE,
      execution.checkpointing.interval=3000ms, execution.checkpointing.timeout=30000ms,
      execution.checkpointing.min-pause=1000ms, ...}
    15:29:00,908 FlinkConfigManager [DEBUG][flink/programmatic-ckpt-test] Cached runtime configuration with 55 entries

    | Config | Deployment Spec | App Code Override (detected) |
    |--------|----------------|---------------------------|
    | `execution.checkpointing.interval` | 10000ms | **3000ms** |
    | `execution.checkpointing.mode` | EXACTLY_ONCE | **AT_LEAST_ONCE** |
    | `execution.checkpointing.timeout` | 600000ms | **30000ms** |
    | `execution.checkpointing.min-pause` | 0ms (default) | **1000ms** |
```
      **A/B comparison:** `ClusterHealthEvaluator` computes `window = max(interval * 2, timeout * 2)`:
       - Before (main branch, spec values): `max(10000*2, 600000*2)` = **~20 min**
       - After (feature branch, runtime values): `max(3000*2, 30000*2)` = **~1 min** (20x faster detection)
  - **8. Platform config drift** - `jobmanager.scheduler: adaptive` set in operator defaults (NOT in K8s spec), then removed from defaults while job keeps running:
```
    16:03:35,205 AbstractFlinkService [DEBUG][flink/scheduler-test] Fetching JobManager configuration
    16:03:35,230 AbstractFlinkService [DEBUG][flink/scheduler-test] Fetched 40 JobManager configuration entries
    16:03:35,262 FlinkConfigManager  [DEBUG][flink/scheduler-test] Cached runtime configuration with 54 entries
    16:03:35,263 FlinkConfigManager  [DEBUG][flink/scheduler-test] Applying 54 cached runtime configuration overrides
```
    | Source | `jobmanager.scheduler` |
    |--------|----------------------|
    | JM REST endpoint (actual runtime) | `adaptive` |
    | K8s spec `flinkConfiguration` | _(not set)_ |
    | Operator defaults (after drift) | _(not set)_ |

    | Branch | `observeConfig` scheduler value | `supportsInPlaceScaling` |
    |--------|-------------------------------|--------------------------|
    | **Main** (reconstructed) | `Default` (wrong) | `false` - unnecessary restart |
    | **Feature** (JM cache) | `Adaptive` (correct) | `true` - in-place scaling works |

**Test Setup:**
- FlinkDeployment with `jobmanager.scheduler: adaptive`, `inplace-scaling.enabled: true`
- Autoscaler enabled with `stabilization.interval: 30s`, `metrics.window: 60s`
- Job: `StateMachineExample.jar` at parallelism 2 (deliberately over-provisioned)
- The autoscaler detects low utilization and writes `pipeline.jobvertex-parallelism-overrides`

**Job ID (unchanged throughout):** `8d7ea2427c1eb8d7f3c491e4691bfd53`

**Result: PASS**

### Phase 1 — Cache active before scaling (steady-state)

```
13:14:05,974 FlinkConfigManager [DEBUG][default/inplace-scale-test] Applying 54 cached runtime configuration overrides
13:14:05,984 JobStatusObserver  [DEBUG][default/inplace-scale-test] Runtime configuration already cached
13:14:21,069 FlinkConfigManager [DEBUG][default/inplace-scale-test] Applying 54 cached runtime configuration overrides
13:14:21,075 JobStatusObserver  [DEBUG][default/inplace-scale-test] Runtime configuration already cached
```

Cache with 54 entries is being applied every ~15s observe cycle. No REST calls.

### Phase 2 — Autoscaler triggers in-place scaling

```
13:14:21,164 AuditUtils         [INFO ][default/inplace-scale-test] >>> Event[Job]       | Info    | SCALINGREPORT   | Scaling execution enabled, begin scaling vertices:{ Vertex ID 20ba6b65f97481d5570070de90e4e791 | Parallelism 2 -> 1 | Processing capacity 258514.72 -> 103405.89 | Target data rate 2003.49}{ Vertex ID bc764cd8ddf7a0cff126f51c16239658 | Parallelism 2 -> 1 | Processing capacity 44651.35 -> 17860.54 | Target data rate 2667.92}
13:14:21,177 AuditUtils         [INFO ][default/inplace-scale-test] >>> Event[Job]       | Info    | SPECCHANGED     | SCALE change(s) detected (Diff: FlinkDeploymentSpec[flinkConfiguration.pipeline.jobvertex-parallelism-overrides : null -> 20ba6b65f97481d5570070de90e4e791:1,bc764cd8ddf7a0cff126f51c16239658:1]), starting reconciliation.
13:14:21,276 AuditUtils         [INFO ][default/inplace-scale-test] >>> Event[Job]       | Info    | SCALING         | In-place scaling triggered
```

The autoscaler wrote `pipeline.jobvertex-parallelism-overrides`, the reconciler detected a
`DiffType.SCALE` change (not UPGRADE), and called the Rescale API.

### Phase 3 — Cache invalidation (new code in `AbstractFlinkResourceReconciler.scale()`)

```
13:14:21,276 FlinkConfigManager [DEBUG][default/inplace-scale-test] Invalidated runtime configuration cache
```

Same millisecond as the in-place scale event. The stale cache (with old parallelism) is removed.

### Phase 4 — Cache re-population from the rescaled JM

```
13:14:36,306 JobStatusObserver  [DEBUG][default/inplace-scale-test] Fetching runtime configuration
13:14:36,306 AbstractFlinkService [DEBUG][default/inplace-scale-test] Fetching runtime configuration
13:14:36,317 AbstractFlinkService [DEBUG][default/inplace-scale-test] Fetched merged runtime configuration with 54 entries
13:14:36,317 FlinkConfigManager [DEBUG][default/inplace-scale-test] Cached runtime configuration with 54 entries
```

Next observe cycle (~15s later): observer sees empty cache, calls `getRuntimeConfiguration()` which
queries all 3 REST endpoints. Fresh 54 entries cached from the now-rescaled JM.

### Phase 5 — Subsequent cycles use fresh cache

```
13:14:51,490 FlinkConfigManager [DEBUG][default/inplace-scale-test] Applying 54 cached runtime configuration overrides
13:14:51,506 JobStatusObserver  [DEBUG][default/inplace-scale-test] Runtime configuration already cached
```

Back to steady-state with the new cache. No more REST calls.

### Timeline Summary

| Time | Log Line | Phase |
|------|----------|-------|
| 13:14:21,069 | `FlinkConfigManager: Applying 54 cached runtime configuration overrides` | Cache active (stale) |
| 13:14:21,164 | `AuditUtils: SCALINGREPORT \| Parallelism 2 -> 1` | Autoscaler decision |
| 13:14:21,177 | `AuditUtils: SPECCHANGED \| SCALE change(s) detected` | DiffType.SCALE reconciliation |
| 13:14:21,276 | `AuditUtils: SCALING \| In-place scaling triggered` | Rescale API called |
| 13:14:21,276 | `FlinkConfigManager: Invalidated runtime configuration cache` | **Stale cache cleared** |
| 13:14:36,306 | `JobStatusObserver: Fetching runtime configuration` | Re-fetch triggered |
| 13:14:36,306 | `AbstractFlinkService: Fetching runtime configuration` | `getRuntimeConfiguration()` |
| 13:14:36,317 | `AbstractFlinkService: Fetched merged runtime configuration with 54 entries` | Fresh config from rescaled JM |
| 13:14:36,317 | `FlinkConfigManager: Cached runtime configuration with 54 entries` | **New cache stored** |
| 13:14:51,490 | `FlinkConfigManager: Applying 54 cached runtime configuration overrides` | Steady-state with fresh cache |

## Documentation


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigManager.java
```diff
@@ -352,8 +393,193 @@ public Configuration getObserveConfig(FlinkDeployment deployment) {
...
+            // Check if JobManager is ready (only for FlinkDeployment)
+            if (resource instanceof FlinkDeployment) {
+                FlinkDeployment deployment = (FlinkDeployment) resource;
+                if (deployment.getStatus().getJobManagerDeploymentStatus()
+                        != JobManagerDeploymentStatus.READY) {
+                    LOG.debug(
+                            "JobManager is not ready, skipping runtime config fetch for job {}",
+                            jobId);
+                    return;
+                }
```
gyfora: In general I think it's not the right approach to have this logic here. We should move this to the observers where job state is tracked already and we know exactly when the rest api can be accessed.

So instead of accessing the rest api here, expose methods for setting/invalidating/getting the runtime config in the ConfigManager. Then the observer could check if the configmanager has a cached config already for the current job id , and if not update it. Same way, the reconciler could invalidate runtime configs after an cancel/upgrade.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigManager.java
```diff
@@ -352,8 +393,193 @@ public Configuration getObserveConfig(FlinkDeployment deployment) {
...
+            JobStatus jobState = jobStatus.getState();
+
+            // Only fetch runtime config if job is in a running state
+            if (jobState == null || !jobState.equals(JobStatus.RUNNING)) {
+                LOG.debug(
+                        "Job {} is not in RUNNING state (current: {}), skipping runtime config fetch",
+                        jobId,
+                        jobState);
+                return;
+            }
```
gyfora: Please see my other comment relating the observer logic, but overall once the runtime config is cached for the jobid we should apply it, even if the job is failing/restarting/etc. The configmanager doesn't need this logic after that

gyfora (changes requested):
I have added some notes, overall I think there is too much logic in the configmanager. We should move the getting/accessing of the configs to the Observers, then the FlinkConfigManager should simply always apply when available.

This is going to be very critical logic, so the exception handling is critical, it's better to fall back to the previous behaviour (simply use the previous logic if the runtime config cannot be accessed) so that part is good (has to go to the observer too).

Once we restructured this, we should add detailed test coverage for the core logic and error handling :)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigManager.java
nishita-09: Sure @gyfora thanks for the review. Will update this logic to move it to observers.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigManager.java
```diff
@@ -140,6 +161,12 @@ public Configuration load(Key k) {
                 cacheTimeout.toMillis(),
                 TimeUnit.MILLISECONDS);
 
+        executorService.scheduleWithFixedDelay(
+                runtimeConfigCache::cleanUp,
+                cacheTimeout.toMillis(),
+                cacheTimeout.toMillis(),
+                TimeUnit.MILLISECONDS);
```
gyfora: Is this necessary? I think we probably don't need it for the other cache either as we are continuously using the cache itself

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigManager.java
```diff
@@ -351,6 +378,89 @@ public Configuration getObserveConfig(FlinkDeployment deployment) {
...
+            String namespace, String name, String jobId, Map<String, String> config) {
+        RuntimeConfigCacheKey cacheKey =
+                RuntimeConfigCacheKey.builder()
+                        .namespace(namespace)
+                        .name(name)
+                        .jobId(jobId)
+                        .build();
+        runtimeConfigCache.put(cacheKey, Collections.unmodifiableMap(config));
+        LOG.debug(
+                "Cached runtime configuration for job {} (resource {}/{})", jobId, namespace, name);
```
gyfora: no need to put namespace/name in the log as it's already part of the MDC

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigManager.java
```diff
@@ -351,6 +378,89 @@ public Configuration getObserveConfig(FlinkDeployment deployment) {
...
+     * @param namespace Resource namespace
+     * @param name Resource name
+     */
+    public void invalidateRuntimeConfig(String namespace, String name) {
+        runtimeConfigCache
+                .asMap()
+                .keySet()
+                .removeIf(
+                        key -> namespace.equals(key.getNamespace()) && name.equals(key.getName()));
+        LOG.debug("Invalidated runtime configuration cache for resource {}/{}", namespace, name);
```
gyfora: no need to put namespace/name in the log as it's already part of the MDC

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentContext.java
```diff
@@ -59,7 +59,9 @@ public Configuration getDeployConfig(AbstractFlinkSpec spec) {
 
     @Override
     protected Configuration createObserveConfig() {
-        return configManager.getObserveConfig(getResource());
+        Configuration conf = configManager.getObserveConfig(getResource());
```
gyfora: should getObserveConfig automatically apply the overrides in the configmanager?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
```diff
@@ -111,6 +112,85 @@ public boolean observe(FlinkResourceContext<R> ctx) {
...
+
+        var resource = ctx.getResource();
+        var jobStatus = resource.getStatus().getJobStatus();
+        String namespace = resource.getMetadata().getNamespace();
+        String name = resource.getMetadata().getName();
+        String jobId = jobStatus.getJobId();
+
+        if (jobId == null) {
+            return;
+        }
```
gyfora: why would jobId be null?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
```diff
@@ -111,6 +112,85 @@ public boolean observe(FlinkResourceContext<R> ctx) {
...
+        try {
+            Map<String, String> jobConfig = flinkService.getJobConfiguration(conf, jobIdObj);
+            if (jobConfig != null) {
+                runtimeConfig.putAll(jobConfig);
+                LOG.debug(
+                        "Fetched {} job configuration entries for job {}", jobConfig.size(), jobId);
+            }
+        } catch (Exception e) {
+            fetchFailed = true;
+            LOG.warn("Failed to fetch job configuration for job {}: {}", jobId, e.getMessage());
```
gyfora: can we pass the exception to the warn ?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
```diff
@@ -111,6 +112,85 @@ public boolean observe(FlinkResourceContext<R> ctx) {
...
+                        "Fetched {} checkpoint configuration entries for job {}",
+                        checkpointConfig.size(),
+                        jobId);
+            }
+        } catch (Exception e) {
+            fetchFailed = true;
+            LOG.warn(
+                    "Failed to fetch checkpoint configuration for job {}: {}",
+                    jobId,
+                    e.getMessage());
```
gyfora: pass exception instead of message

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
```diff
@@ -111,6 +112,85 @@ public boolean observe(FlinkResourceContext<R> ctx) {
...
+            LOG.debug("Checkpoint configuration fetch exception details", e);
+        }
+
+        if (runtimeConfig.isEmpty() && fetchFailed) {
+            LOG.warn(
+                    "All runtime config REST fetches failed for job {}, will retry next cycle",
```
gyfora: I think in general we can remove all job ids also from the logs, I don't think that adds any info

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
```diff
@@ -873,6 +884,308 @@ public JobExceptionsInfoWithHistory getJobExceptions(
...
+            return jobConfig;
+        }
+    }
+
+    /**
+     * Maps REST API JSON field names from {@link CheckpointConfigInfo} to Flink {@link
+     * org.apache.flink.configuration.ConfigOption} keys. Both sides use Flink's own constants for
+     * compile-time safety.
+     */
+    private enum CheckpointConfigMapping {
```
gyfora: We should move all the mapping logic for the Execution + Checkpoint config to a new utility class.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
```diff
@@ -873,6 +884,308 @@ public JobExceptionsInfoWithHistory getJobExceptions(
...
+                                        jobConfigHeaders,
+                                        parameters,
+                                        EmptyRequestBody.getInstance())
+                                .get(
+                                        operatorConfig.getFlinkClientTimeout().toSeconds(),
+                                        TimeUnit.SECONDS);
+                if (configurationInfo != null) {
+                    if (configurationInfo.getExecutionConfigInfo() != null) {
+                        jobConfig.put(
+                                "parallelism.default",
```
gyfora: We should use the config options

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
```diff
@@ -873,6 +884,308 @@ public JobExceptionsInfoWithHistory getJobExceptions(
...
+                                String.valueOf(
+                                        configurationInfo
+                                                .getExecutionConfigInfo()
+                                                .getParallelism()));
+                        jobConfig.put(
+                                "pipeline.object-reuse",
```
gyfora: We should use the config options

gyfora (changes requested):
I agree that this manual mapping of configs is very fragile , and it duplicates a lot of logic from core Flink. A different approach was taken in https://github.com/apache/flink-kubernetes-operator/pull/1080/changes it uses a different endpoint to get the JobManagerJobConfiguration.

Can we simply use this instead of all runtimeConfig/checkpointConfig ? I think when the user programmatically sets a config in the StreamExecEnv those actually end up in the  JobManagerJobConfiguration. cc @Dennis-Mircea

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentContext.java
nishita-09: Here, I have added it as a subsequent step that it specifically applies overrides after getting default configs. We can have this incorporated in the getObserveConfig as well

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
nishita-09: Yes @gyfora Was going to do it for this as well, like it was done for checkpoint configs. Thanks!!

nishita-09:
> https://github.com/apache/flink-kubernetes-operator/pull/1080/changes



> I agree that this manual mapping of configs is very fragile , and it duplicates a lot of logic from core Flink. A different approach was taken in https://github.com/apache/flink-kubernetes-operator/pull/1080/changes it uses a different endpoint to get the JobManagerJobConfiguration.
> 
> Can we simply use this instead of all runtimeConfig/checkpointConfig ? I think when the user programmatically sets a config in the StreamExecEnv those actually end up in the JobManagerJobConfiguration. cc @Dennis-Mircea

@gyfora For checkpoints configurations , the programmatically set config does not end up in JobManagerJobConfiguration. This still shows the default configs set in the deployment, while the checkpoint/config endpoint is the place where the configurations correctly reflect.

When user code calls `env.enableCheckpointing(30000)`, it writes `CheckpointingOptions.CHECKPOINTING_INTERVAL = 30s` into the Configuration object inside StreamExecutionEnvironment. This is the same Configuration object that gets serialized into the JobGraph and eventually becomes the CheckpointCoordinatorConfiguration used by the CheckpointCoordinator at runtime.

CheckpointConfigHandler - This handler serves `/jobs/:jobid/checkpoints/config`. It reads from the execution graph, which gets its values from the CheckpointCoordinatorConfiguration, which was built from the JobGraph (which includes the programmatic overrides)

JobManagerJobConfigurationHander - This handler serves` /jobs/:jobid/jobmanager/config`. It reads from a pre-baked Configuration object passed at construction time i.e the JM's initial configuration.
The Configuration object passed to this handler is the JM's original startup Configuration. The user's main() method hasn't run yet when this handler is constructed, and even after it runs, it only modifies the StreamExecutionEnvironment's internal Configuration and not the JM-level Configuration.

gyfora:
> > https://github.com/apache/flink-kubernetes-operator/pull/1080/changes
> 
> > I agree that this manual mapping of configs is very fragile , and it duplicates a lot of logic from core Flink. A different approach was taken in https://github.com/apache/flink-kubernetes-operator/pull/1080/changes it uses a different endpoint to get the JobManagerJobConfiguration.
> > Can we simply use this instead of all runtimeConfig/checkpointConfig ? I think when the user programmatically sets a config in the StreamExecEnv those actually end up in the JobManagerJobConfiguration. cc @Dennis-Mircea
> 
> @gyfora For checkpoints configurations , the programmatically set config does not end up in JobManagerJobConfiguration. This still shows the default configs set in the deployment, while the checkpoint/config endpoint is the place where the configurations correctly reflect.
> 
> When user code calls `env.enableCheckpointing(30000)`, it writes `CheckpointingOptions.CHECKPOINTING_INTERVAL = 30s` into the Configuration object inside StreamExecutionEnvironment. This is the same Configuration object that gets serialized into the JobGraph and eventually becomes the CheckpointCoordinatorConfiguration used by the CheckpointCoordinator at runtime.
> 
> CheckpointConfigHandler - This handler serves `/jobs/:jobid/checkpoints/config`. It reads from the execution graph, which gets its values from the CheckpointCoordinatorConfiguration, which was built from the JobGraph (which includes the programmatic overrides)
> 
> JobManagerJobConfigurationHander - This handler serves` /jobs/:jobid/jobmanager/config`. It reads from a pre-baked Configuration object passed at construction time i.e the JM's initial configuration. The Configuration object passed to this handler is the JM's original startup Configuration. The user's main() method hasn't run yet when this handler is constructed, and even after it runs, it only modifies the StreamExecutionEnvironment's internal Configuration and not the JM-level Configuration.

Okay, that sounds suboptimal for our use-case here. We can start with the manual config mapping for now but we should try to fix this in Flink and have a rest endpoint that simply returns the effective job config.

nishita-09:
> > > https://github.com/apache/flink-kubernetes-operator/pull/1080/changes
> > 
> > 
> > > I agree that this manual mapping of configs is very fragile , and it duplicates a lot of logic from core Flink. A different approach was taken in https://github.com/apache/flink-kubernetes-operator/pull/1080/changes it uses a different endpoint to get the JobManagerJobConfiguration.
> > > Can we simply use this instead of all runtimeConfig/checkpointConfig ? I think when the user programmatically sets a config in the StreamExecEnv those actually end up in the JobManagerJobConfiguration. cc @Dennis-Mircea
> > 
> > 
> > @gyfora For checkpoints configurations , the programmatically set config does not end up in JobManagerJobConfiguration. This still shows the default configs set in the deployment, while the checkpoint/config endpoint is the place where the configurations correctly reflect.
> > When user code calls `env.enableCheckpointing(30000)`, it writes `CheckpointingOptions.CHECKPOINTING_INTERVAL = 30s` into the Configuration object inside StreamExecutionEnvironment. This is the same Configuration object that gets serialized into the JobGraph and eventually becomes the CheckpointCoordinatorConfiguration used by the CheckpointCoordinator at runtime.
> > CheckpointConfigHandler - This handler serves `/jobs/:jobid/checkpoints/config`. It reads from the execution graph, which gets its values from the CheckpointCoordinatorConfiguration, which was built from the JobGraph (which includes the programmatic overrides)
> > JobManagerJobConfigurationHander - This handler serves` /jobs/:jobid/jobmanager/config`. It reads from a pre-baked Configuration object passed at construction time i.e the JM's initial configuration. The Configuration object passed to this handler is the JM's original startup Configuration. The user's main() method hasn't run yet when this handler is constructed, and even after it runs, it only modifies the StreamExecutionEnvironment's internal Configuration and not the JM-level Configuration.
> 
> Okay, that sounds suboptimal for our use-case here. We can start with the manual config mapping for now but we should try to fix this in Flink and have a rest endpoint that simply returns the effective job config.

Agreed @gyfora . Sure, I will clean this up a bit more and have it ready for review in some time. Thanks.

nishita-09:
> > > https://github.com/apache/flink-kubernetes-operator/pull/1080/changes
> > 
> > 
> > > I agree that this manual mapping of configs is very fragile , and it duplicates a lot of logic from core Flink. A different approach was taken in https://github.com/apache/flink-kubernetes-operator/pull/1080/changes it uses a different endpoint to get the JobManagerJobConfiguration.
> > > Can we simply use this instead of all runtimeConfig/checkpointConfig ? I think when the user programmatically sets a config in the StreamExecEnv those actually end up in the JobManagerJobConfiguration. cc @Dennis-Mircea
> > 
> > 
> > @gyfora For checkpoints configurations , the programmatically set config does not end up in JobManagerJobConfiguration. This still shows the default configs set in the deployment, while the checkpoint/config endpoint is the place where the configurations correctly reflect.
> > When user code calls `env.enableCheckpointing(30000)`, it writes `CheckpointingOptions.CHECKPOINTING_INTERVAL = 30s` into the Configuration object inside StreamExecutionEnvironment. This is the same Configuration object that gets serialized into the JobGraph and eventually becomes the CheckpointCoordinatorConfiguration used by the CheckpointCoordinator at runtime.
> > CheckpointConfigHandler - This handler serves `/jobs/:jobid/checkpoints/config`. It reads from the execution graph, which gets its values from the CheckpointCoordinatorConfiguration, which was built from the JobGraph (which includes the programmatic overrides)
> > JobManagerJobConfigurationHander - This handler serves` /jobs/:jobid/jobmanager/config`. It reads from a pre-baked Configuration object passed at construction time i.e the JM's initial configuration. The Configuration object passed to this handler is the JM's original startup Configuration. The user's main() method hasn't run yet when this handler is constructed, and even after it runs, it only modifies the StreamExecutionEnvironment's internal Configuration and not the JM-level Configuration.
> 
> Okay, that sounds suboptimal for our use-case here. We can start with the manual config mapping for now but we should try to fix this in Flink and have a rest endpoint that simply returns the effective job config.

@gyfora I have updated the logic, moved the core logic from config manager to jobstatusobserver with utility classes in a separate file. I have also added jobManagerConfig configurations to the runtime config. Removed the jobIds, namespaces from logs. Kindly review this and do let me know if you find any other concern. Thank you!

nishita-09:
@gyfora Please do review this- addressed all the comments and made the changes accordingly, thank you !!

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
```diff
@@ -873,6 +879,110 @@ public JobExceptionsInfoWithHistory getJobExceptions(
...
+
+                Map<String, String> jmConfig = new HashMap<>();
+                configurationInfo.forEach(entry -> jmConfig.put(entry.getKey(), entry.getValue()));
+                LOG.debug("Fetched {} JobManager configuration entries", jmConfig.size());
+                return jmConfig;
+
+            } catch (Exception e) {
+                LOG.warn("Failed to fetch JobManager configuration via REST API", e);
+                return new HashMap<>();
+            }
```
gyfora, line 909: This catch is probably not required here as we already catch outside of this method, this way we just hide errors unnecessarily

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
```diff
@@ -873,6 +879,110 @@ public JobExceptionsInfoWithHistory getJobExceptions(
...
+                                        operatorConfig.getFlinkClientTimeout().toSeconds(),
+                                        TimeUnit.SECONDS);
+
+                Map<String, String> jobConfig =
+                        FlinkRuntimeConfigurationUtils.mapJobConfiguration(configurationInfo);
+                LOG.debug("Fetched {} job configuration entries", jobConfig.size());
+                return jobConfig;
+
+            } catch (Exception e) {
+                LOG.warn("Failed to fetch job configuration via REST API", e);
```
gyfora, line 939: This catch is probably not required here as we already catch outside of this method, this way we just hide errors unnecessarily

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
```diff
@@ -873,6 +879,110 @@ public JobExceptionsInfoWithHistory getJobExceptions(
...
+                                        parameters,
+                                        EmptyRequestBody.getInstance())
+                                .get(
+                                        operatorConfig.getFlinkClientTimeout().toSeconds(),
+                                        TimeUnit.SECONDS);
+
+                return FlinkRuntimeConfigurationUtils.mapCheckpointConfiguration(
+                        checkpointConfigInfo);
+
+            } catch (Exception e) {
```
gyfora, line 968: This catch is probably not required here as we already catch outside of this method, this way we just hide errors unnecessarily

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
```diff
@@ -873,6 +879,110 @@ public JobExceptionsInfoWithHistory getJobExceptions(
...
+                LOG.warn("Failed to fetch checkpoint config, falling back to job config", e);
+
+                if (e.getCause() instanceof RestClientException restException
+                        && restException.getHttpResponseStatus() == HttpResponseStatus.NOT_FOUND) {
+                    throw new RuntimeException("Job not found: " + jobId, e);
```
gyfora, line 973: I don't think we need the complex error handling here

gyfora (commented):
I wonder that instead of the chained logic in the observer of calling the individual getConfig methods, we should have a single method to getRuntimeConfiguration in the Flink service that will call the 3 rest endpoints and merge the results. This we can later replace with a better endpoint easily without touching the observer.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
```diff
@@ -111,6 +112,67 @@ public boolean observe(FlinkResourceContext<R> ctx) {
...
+            LOG.warn("Failed to fetch checkpoint configuration", e);
+        }
+
+        if (runtimeConfig.isEmpty() && fetchFailed) {
+            LOG.warn("All runtime config REST fetches failed, will retry next cycle");
+            return;
+        }
+
+        configManager.putRuntimeConfig(namespace, name, jobId, runtimeConfig);
+        ctx.clearObserveConfig();
```
Dennis-Mircea: Why to clear the `observeConfig` here? By having this in place it re-runs the entire config generation pipeline (cache lookup, key building, `ObjectMapper` serialization, cloning, spec config application) just to get an identical base config cache when the `getObserveConfig()` will be called.

I'd say that it makes more sense to apply the running config support in here:

```suggestion
var currentObserveConfig = ctx.getObserveConfig();
if (currentObserveConfig != null) {
    runtimeConfig.forEach(currentObserveConfig::setString);
}
```

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/sessionjob/SessionJobReconciler.java
```diff
@@ -111,10 +111,14 @@ public void deploy(
     @Override
     protected boolean cancelJob(FlinkResourceContext<FlinkSessionJob> ctx, SuspendMode suspendMode)
             throws Exception {
+        var resource = ctx.getResource();
+        ctx.getConfigManager()
+                .invalidateRuntimeConfig(
```
Dennis-Mircea, line 116: I see no explicit reason to do this as the cache TTL will be hit anyway. The existing `LoadingCache` for `observeConfig` doesn't do explicit invalidation either (it relies on key-based natural misses + TTL), so the `runtimeConfigCache` doesn't need it for consistency.

In this way, if we want to call `invalidateRuntimeConfig` in all valid scenarios, we should also include `reconcileOtherChanges` as one of those cases.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
```diff
@@ -111,6 +112,67 @@ public boolean observe(FlinkResourceContext<R> ctx) {
...
+            FlinkResourceContext<R> ctx, JobStatusMessage clusterJobStatus) {
+        if (clusterJobStatus.getJobState() != JobStatus.RUNNING) {
+            return;
+        }
+
+        var resource = ctx.getResource();
+        var jobStatus = resource.getStatus().getJobStatus();
+        String namespace = resource.getMetadata().getNamespace();
+        String name = resource.getMetadata().getName();
+        String jobId = jobStatus.getJobId();
```
Dennis-Mircea: There are a lot of mixtures between concrete types and `var`. I'd suggest to stick just to `var`.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigManager.java
```diff
@@ -88,6 +90,17 @@ public class FlinkConfigManager {
...
 
+    private final Cache<RuntimeConfigCacheKey, Map<String, String>> runtimeConfigCache;
+
+    /** Cache key for runtime configuration overrides, scoped to a specific job instance. */
+    @Value
+    @Builder
+    private static class RuntimeConfigCacheKey {
+        String namespace;
+        String name;
+        String jobId;
```
Dennis-Mircea, line 101: Using `jobId` as part of the cache key assumes that runtime configuration is immutable for the lifetime of a job instance. This assumption may not hold for in-place scaling (where parallelism changes without a new jobId) or future dynamic configuration features such as [FLIP-530](https://cwiki.apache.org/confluence/display/FLINK/FLIP-514%3A+Custom+Evaluator+plugin+for+Flink+Autoscaler) and [FLIP-571](https://cwiki.apache.org/confluence/display/FLINK/FLIP-571%3A+Support+Dynamically+Updating+Checkpoint+Configuration+at+Runtime+via+REST+API).

For the dynamic configuration we can adapt in the future (maybe with a scheduled watcher in the same way as `scheduleConfigWatcher`), but for the in-place scaling this is for sure a limiting factor and requires specific cache invalidation.

The reason why the in-place scaling explicitly requires a cache invalidation is because it changes the running Flink job's configuration (parallelism, task slots), without changing the jobId. The current implementation as part of `FlinkResourceContext#getObserveConfig` is placing the `runtimeConfig` with higher precedence over `observeConfig`, meaning that it will outdate a new scaled `observeConfig`.

To resolve this limitation my suggestion is to add the following code as part at `AbstractFlinkResourceReconciler.java:367`:
```java
ctx.getConfigManager()
        .invalidateRuntimeConfig(
                resource.getMetadata().getNamespace(),
                resource.getMetadata().getName());
```

Dennis-Mircea (changes requested):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigManager.java
nishita-09: Thanks @Dennis-Mircea , this is definitely needed. Will update

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/sessionjob/SessionJobReconciler.java
nishita-09: @Dennis-Mircea I was thinking of some scenarios, while I do understand in the current set of things we would not need this explicitly since TTL would take care of it eventually. I feel this could be a real use case in scenarios where same jobId is used in recovery/cancellation to run the job with updated config -> ensure invalidation of the job whenever cancel/resubmit is called should be good guardrail.
I totally agree with the point that if we call `invalidateRuntimeConfig` in cancelJob, it needs to be included in other valid scenarios as well.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigManager.java
nishita-09: @Dennis-Mircea I have also added the test results for this in the above PR description. I hope that covers this usecase. Thanks!!

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
nishita-09: Added the above suggestion, thanks @Dennis-Mircea

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkResourceContext.java
```diff
@@ -97,7 +97,27 @@ public Configuration getObserveConfig() {
...
+        }
+        return observeConfig;
+    }
+
+    /**
+     * Clear the memoized observe config so it will be regenerated on next access. This must be
+     * called after the observer updates the runtime configuration cache so that subsequent
+     * consumers (other observers, reconciler) see the merged configuration.
+     */
+    public void clearObserveConfig() {
```
Dennis-Mircea, line 112: This method is currently unused.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
```diff
@@ -111,6 +112,46 @@ public boolean observe(FlinkResourceContext<R> ctx) {
...
 
+    /**
+     * Fetch runtime configuration from the Flink REST API and cache it in the config manager when
+     * the job is in RUNNING state and no cached config exists for the current job ID. After
+     * caching, the runtime values are applied directly to the already-memoized observe config so
+     * subsequent consumers see the merged values without a full config regeneration.
+     */
+    private void fetchAndCacheRuntimeConfig(
+            FlinkResourceContext<R> ctx, JobStatusMessage clusterJobStatus) {
+        if (clusterJobStatus.getJobState() != JobStatus.RUNNING) {
```
Dennis-Mircea, line 123: I think this condition restricts too much. I'd rather go with something like this:
```java
if (clusterJobStatus.getJobState().isGloballyTerminalState()) {
    return;
}
```
Or not even have at all this condition. @gyfora what do you say?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigManager.java
```diff
@@ -351,6 +372,81 @@ public Configuration getObserveConfig(FlinkDeployment deployment) {
...
+
+    /**
+     * Get cached runtime configuration overrides for a specific job instance.
+     *
+     * @param namespace Resource namespace
+     * @param name Resource name
+     * @param jobId Job ID string
+     * @return Cached runtime config if present
+     */
+    public Optional<Map<String, String>> getRuntimeConfig(
```
Dennis-Mircea, line 403: Currently, within a single cycle, the cache is hit in:
1. `getObserveConfig()` -> `applyCachedRuntimeConfig()` -> `getRuntimeConfig(ns, name, jobId)`
2. `fetchAndCacheRuntimeConfig()` -> `configManager.getRuntimeConfig(ns, name, jobId).isPresent()`

So there are 2 hits per cycle at this point (and maybe in the future will be even more) for this `runtimeConfig`. For that, I'd suggest to introduce a `runtimeConfig` variable as part of `FlinkResourceContext` that will follow the same pattern as `observeConfig`. Having a `runtimeConfig` variable per context will reduce the cache hits to 1.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/sessionjob/SessionJobReconciler.java
Dennis-Mircea: By calling `invalidateRuntimeConfig` for each concrete case, I’ll focus even more on adding tests to cover all of these scenarios, and by assuming that @gyfora agrees this is the most suitable approach.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
```diff
@@ -873,6 +879,111 @@ public JobExceptionsInfoWithHistory getJobExceptions(
...
+            throws Exception {
+        LOG.debug("Fetching runtime configuration");
+        Map<String, String> runtimeConfig = new HashMap<>();
+        boolean fetchFailed = false;
+
+        try {
+            runtimeConfig.putAll(getJobManagerConfiguration(conf, jobId));
+        } catch (Exception e) {
+            fetchFailed = true;
+            LOG.warn("Failed to fetch JobManager configuration", e);
```
Dennis-Mircea, line 893: I'll log these messages as ERROR not WARNING, since we encountered an exception. Same for all the below cases.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkRuntimeConfigurationUtils.java
```diff
@@ -0,0 +1,254 @@
...
+                    "FileSystemCheckpointStorage", "filesystem",
+                    "JobManagerCheckpointStorage", "jobmanager");
+
+    /** All config keys that {@link #mapCheckpointConfiguration} can produce. */
+    public static final Set<String> CHECKPOINT_CONFIG_KEYS =
+            Stream.concat(
+                            Stream.of(CheckpointConfigMapping.values()).map(m -> m.configKey),
+                            Stream.of(
+                                    CheckpointingOptions.EXTERNALIZED_CHECKPOINT_RETENTION.key(),
+                                    CheckpointingOptions.STATE_BACKEND.key(),
```
Dennis-Mircea, line 137: I see `CheckpointingOptions.STATE_BACKEND` is deprecated. Can we use `StateBackendOptions.STATE_BACKEND` instead?

Dennis-Mircea:
@nishita-09 could you take a look at the conflicts on this PR and resolve them?

nishita-09:
> @nishita-09 could you take a look at the conflicts on this PR and resolve them?

Sure @Dennis-Mircea resolving it.

nishita-09:
> @nishita-09 could you take a look at the conflicts on this PR and resolve them?

@Dennis-Mircea Could you review this now? Do let me know if you see any other scope of improvements here. Thanks

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
```diff
@@ -111,6 +112,38 @@ public boolean observe(FlinkResourceContext<R> ctx) {
...
+            var runtimeConfig =
+                    ctx.getFlinkService()
+                            .getRuntimeConfiguration(
+                                    ctx.getObserveConfig(),
+                                    JobID.fromHexString(jobStatus.getJobId()));
+            ctx.putRuntimeConfig(runtimeConfig);
+            var currentObserveConfig = ctx.getObserveConfig();
+            if (currentObserveConfig != null) {
+                runtimeConfig.forEach(currentObserveConfig::setString);
+            }
```
gyfora: this doesn't seem right here, seem like the putRuntimeConfig should also update the observeConfig or null it out simply.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
```diff
@@ -363,7 +363,12 @@ private boolean scale(FlinkResourceContext<CR> ctx, Configuration deployConfig)
         var scaled = ctx.getFlinkService().scale(ctx, deployConfig);
 
         if (scaled) {
-            ReconciliationUtils.updateStatusForDeployedSpec(ctx.getResource(), deployConfig, clock);
```
gyfora, line 366: Shouldn't the deploy method also invalidate the runtimeConfig?

gyfora (changes requested):
Noticed some minor inconsistencies, otherwise looks good :)

nishita-09:
> Noticed some minor inconsistencies, otherwise looks good :)

Thanks @gyfora  This makes sense, thanks for catching these inconsistencies. The deploy method should have the invalidation instead of relying on  upstream callers.

I have added comments explicitly to call that out.

Please do approve the workflows for the builds, the earlier build failed due to mvn spotless:apply , which I have now run to ensure the build passes.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractFlinkResourceReconciler.java
nishita-09: Yes @gyfora  I was relying on upstream callers, or previous actions like cancelJob or other terminating function to invalidate it but its good to have it in deploy as well to ensure we don't face any inconsistency.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
nishita-09: Thanks @gyfora  This makes sense.
`FlinkResourceContext.putRuntimeConfig` now nulls the memoized  observeConfig so the next getObserveConfig() rebuilds it and picks up the runtime overrides through the existing composition path in getObserveConfig(). This replaces the manual re-apply loop in `JobStatusObserver.fetchAndCacheRuntimeConfig`, which is removed

gyfora (approved):

PR #1013, opened 2025-08-18
[FLINK-37406] Add support for structured YAML config in FlinkDeployment CRD

csviri:

Allow to set `flinkConfiguration` as `yaml`, not just key/value pairs.

## Brief change log

- changes CRD to allow set `flinkConfiguration` as `yaml` not just key/value pairs
- changed type of `AbstractFlinkSpec.flinkConfiguration` it now accepts yaml.
- adjusted CRD checker to allow such changes

## Verifying this change

- adjusted e2e tests to the new format
- added test for CRD migration

## Documentation

  - If yes, how is the feature documented?  JavaDocs

### DIFF
--- a/flink-kubernetes-operator-api/pom.xml
+++ b/flink-kubernetes-operator-api/pom.xml
@@ -125,14 +125,7 @@ under the License.
             <version>${flink.version}</version>
             <scope>test</scope>
         </dependency>
-
-        <dependency>
-            <groupId>org.assertj</groupId>
-            <artifactId>assertj-core</artifactId>
-            <version>${assertj.version}</version>
-            <scope>test</scope>
-        </dependency>
-
+        
         <dependency>
             <groupId>org.apache.flink</groupId>
             <artifactId>flink-runtime</artifactId>


examples/kubernetes-client-examples/src/main/java/org/apache/flink/examples/Basic.java
```diff
@@ -49,7 +50,7 @@ public static void main(String[] args) {
         flinkDeploymentSpec.setImage("flink:1.19");
         Map<String, String> flinkConfiguration =
                 Map.ofEntries(entry("taskmanager.numberOfTaskSlots", "2"));
-        flinkDeploymentSpec.setFlinkConfiguration(flinkConfiguration);
+        flinkDeploymentSpec.setFlinkConfiguration(toJsonNode(flinkConfiguration));
```
gyfora: Can we provide a `setFlinkConfiguration(Map<String, String> map)` method like before to avoid the code changes everywhere? We can mark it deprecated but seems useful?

csviri: If remember correctly jackson had some issues with that, but will take a look how to improve this API

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/spec/ConfigObjectNode.java
```diff
@@ -0,0 +1,84 @@
...
+            remove(name);
+        }
+    }
+
+    public void putAllFrom(Map<String, String> value) {
+        value.forEach(this::put);
+    }
+
+    public void set(Map<String, String> value) {
+        removeAll();
```
gyfora, line 55: This looks a bit confusing, we probably shouldn't introduce a varargs variant of the same method that does something completely different :)

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/spec/ConfigObjectNode.java
```diff
@@ -0,0 +1,84 @@
...
+        super(nc);
+    }
+
+    public void removeAll(String... names) {
+        for (String name : names) {
+            remove(name);
+        }
+    }
+
+    public void putAllFrom(Map<String, String> value) {
```
gyfora, line 50: Should we call it simply putAll?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/KubernetesScalingRealizer.java
```diff
@@ -63,15 +62,16 @@ public void realizeConfigOverrides(
         }
         FlinkDeployment flinkDeployment = ((FlinkDeployment) context.getResource());
         // Apply config overrides
-        Map<String, String> flinkConf = flinkDeployment.getSpec().getFlinkConfiguration();
-        for (String keyToRemove : configChanges.getRemovals()) {
-            flinkConf.remove(keyToRemove);
-        }
-        flinkConf.putAll(configChanges.getOverrides());
+
+        flinkDeployment.getSpec().getFlinkConfiguration().remove(configChanges.getRemovals());
```
gyfora, line 65: Should this be removeAll? Seems like we are removing a single key here which will do nothing

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentControllerTest.java
```diff
@@ -197,6 +197,7 @@ public void verifyReconcileLoopForInitialSuspendedDeploymentWithSavepoint(
                 TestUtils.buildApplicationCluster(flinkVersion, JobState.SUSPENDED);
         appCluster.getSpec().getJob().setUpgradeMode(UpgradeMode.SAVEPOINT);
         appCluster.getSpec().getJob().setInitialSavepointPath("s0");
+
```
gyfora: There are a bunch of files that only have these newline changes, can you please remove those from the PR?

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/spec/ConfigObjectNode.java
```diff
@@ -0,0 +1,84 @@
...
+        value.forEach(this::put);
+    }
+
+    public void set(Map<String, String> value) {
+        removeAll();
+        putAllFrom(value);
+    }
+
+    public Map<String, String> asFlatMap() {
```
gyfora, line 59: I suggest adding a helper method:

```
public Configuration asConfiguration() {
 return Configuration.fromMap(asFlatMap());
}
```

This is used in a lot of places.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentControllerTest.java
csviri: yes, sure will clean it up, thx

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/spec/ConfigObjectNode.java
csviri: Unfortunatelly there is not a deprecated `putAll` in `ObjectNode` so that would be a bit confusing, but if you have suggestion for better name we can change it.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/KubernetesScalingRealizer.java
csviri: actually this uses this method from jackson
```java
public ObjectNode remove(Collection<String> propertyNames) {
        this._children.keySet().removeAll(propertyNames);
        return this;
    }
```
would rather rename our `removeAll` to `remove` so we follow this convention/style

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/spec/ConfigObjectNode.java
csviri: will do thx!

csviri: added

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/spec/ConfigObjectNode.java
csviri: changed it to `public void setAllFrom(Map<String, String> value)`

flink-kubernetes-operator-api/pom.xml
```diff
@@ -126,6 +126,13 @@ under the License.
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
```
gyfora: This dependency is now added twice if I see correctly

gyfora (approved):
Looks good aside from the single duplicated test dependency :)

flink-kubernetes-operator-api/pom.xml
csviri: good catch fixed

PR #1018, opened 2025-08-22
[FLINK-37193] Feature flag if operator should manage ingress

csviri:

This change adds a feature flag to set if the controller should manage the ingress. If set to false controller won't register the related InformerEventSource, and won't create ingress even if set in the spec.

## Brief change log
- adds feature flag as described above and related tests

## Verifying this change

- Added unit tests for the non-trivial logic

## Documentation

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/IngressUtils.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/IngressUtils.java
@@ -74,7 +74,9 @@ public class IngressUtils {
             FlinkDeploymentSpec spec,
             Configuration effectiveConfig,
             KubernetesClient client) {
-
+        if (!ctx.getOperatorConfig().isManageIngress()) {
+            return;
+        }
         var objectMeta = ctx.getResource().getMetadata();
         if (spec.getIngress() != null) {
             HasMetadata ingress = getIngress(objectMeta, spec, effectiveConfig, client);
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/FlinkOperator.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/FlinkOperator.java
@@ -199,7 +199,8 @@ public class FlinkOperator {
                         observerFactory,
                         statusRecorder,
                         eventRecorder,
-                        canaryResourceManager);
+                        canaryResourceManager,
+                        configManager);
         registeredControllers.add(operator.register(controller, this::overrideControllerConfigs));
     }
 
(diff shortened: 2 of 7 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/IngressUtils.java
```diff
@@ -74,7 +74,9 @@ public static void reconcileIngress(
             FlinkDeploymentSpec spec,
             Configuration effectiveConfig,
             KubernetesClient client) {
-
+        if (!ctx.getOperatorConfig().isManageIngress()) {
+            return;
```
gyfora: may be better to throw a reconciliation error if the user actually needs an ingress and it is turned off. Or at least log a warning and trigger an event. Maybe the exception is too drastic but the event is a minimum

csviri: makes sense, changed it, PTAL, thx

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/IngressUtils.java
```diff
@@ -68,13 +70,30 @@ public class IngressUtils {
...
 
     public static void reconcileIngress(
             FlinkResourceContext<?> ctx,
             FlinkDeploymentSpec spec,
             Configuration effectiveConfig,
-            KubernetesClient client) {
+            KubernetesClient client,
+            EventRecorder eventRecorder) {
+        if (!ctx.getOperatorConfig().isManageIngress()) {
+            if (spec.getIngress() != null) {
```
gyfora, line 84: would be cleaner to move this logic in the already existing `if (spec.getIngress() != null) {` branch, and return from there on error

csviri: TBH I like such checks separated before the actual logic, but if you insist I can change it.

gyfora: ok, lets keep it

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/IngressUtils.java
```diff
@@ -68,13 +70,30 @@ public class IngressUtils {
...
+                log.warn(INGRESS_MANAGEMENT_OFF_BUT_SPEC_SET);
```
gyfora: No need to both log and trigger the event as events are already logged. Logging can be removed

gyfora (approved):
aside from the minor comment related to redundant logging, looks good

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/IngressUtils.java
csviri: removed log, thx!

PR #1022, opened 2025-08-28
Express is values.yaml that config.yaml can be used

csviri:

Add an commented out section to helm `values.yaml` that showcases how to use `config.yaml` instead of `flink-conf.yaml`.

## Verifying this change

The underlying mechanism was already implemented before, was just hard to find.

## Documentation

### DIFF
--- a/helm/flink-kubernetes-operator/values.yaml
+++ b/helm/flink-kubernetes-operator/values.yaml
@@ -169,6 +169,16 @@ defaultConfiguration:
   # If set to false, loads just the overrides as in (2).
   # This option has not effect, if create is equal to false.
   append: true
+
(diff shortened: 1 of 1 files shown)


helm/flink-kubernetes-operator/values.yaml
```diff
@@ -147,7 +147,7 @@ operatorSecurityContext: {}
 webhookSecurityContext: {}
 
 webhook:
-  create: true
+  create: false
```
gyfora: Seems like an unintentional change

csviri: yes, sorry, reverted

PR #1023, opened 2025-08-28
[FLINK-38293] Update Checkstyle version and related rules in Kubernetes Operator

csviri:

Update checkstyle so it supports newer Java language features like text blocks, etc.

## Brief change log

Updated checkstyle maven plugin and related rules.

## Verifying this change

Run the build, it passes. Try to use a text block in a unit test for exmaple, now passes the build, unlike before.

## Documentation

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -281,7 +281,7 @@ under the License.
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-checkstyle-plugin</artifactId>
-                <version>3.4.0</version>
+                <version>3.6.0</version>
                 <dependencies>
                     <dependency>
                         <groupId>com.puppycrawl.tools</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -282,14 +282,6 @@ under the License.
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-checkstyle-plugin</artifactId>
                 <version>3.6.0</version>
-                <dependencies>
-                    <dependency>
-                        <groupId>com.puppycrawl.tools</groupId>
-                        <artifactId>checkstyle</artifactId>
-                        <!-- Note: match version with docs/flinkDev/ide_setup.md -->
-                        <version>10.17.0</version>
-                    </dependency>
-                </dependencies>
                 <executions>
                     <execution>
                         <id>validate</id>
--- a/pom.xml
+++ b/pom.xml
@@ -281,13 +281,13 @@ under the License.
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-checkstyle-plugin</artifactId>
-                <version>2.17</version>
+                <version>3.4.0</version>
                 <dependencies>
                     <dependency>
                         <groupId>com.puppycrawl.tools</groupId>
                         <artifactId>checkstyle</artifactId>
                         <!-- Note: match version with docs/flinkDev/ide_setup.md -->
-                        <version>8.14</version>
+                        <version>10.17.0</version>
                     </dependency>
                 </dependencies>
                 <executions>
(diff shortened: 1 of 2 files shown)


tools/maven/checkstyle.xml
```diff
@@ -276,23 +276,15 @@ This file is based on the checkstyle file of Apache Beam.
...
 			<property name="severity" value="error"/>
-			<property name="allowMissingJavadoc" value="true"/>
 			<property name="allowMissingParamTags" value="true"/>
 			<property name="allowMissingReturnTag" value="true"/>
-			<property name="allowMissingThrowsTags" value="true"/>
-			<property name="allowThrowsTagsForSubclasses" value="true"/>
-			<property name="allowUndeclaredRTE" value="true"/>
-			<!-- This check sometimes failed for with "Unable to get class information for @throws tag" for custom exceptions -->
-			<property name="suppressLoadErrors" value="true"/>
 		</module>
```
mxm, line 282: Why are we removing these?

tools/maven/checkstyle.xml
```diff
@@ -276,23 +276,15 @@ This file is based on the checkstyle file of Apache Beam.
...
 
 		<!-- Check that paragraph tags are used correctly in Javadoc. -->
 		<module name="JavadocParagraph"/>
 
 		<module name="JavadocType">
-			<property name="scope" value="protected"/>
```
mxm: Same here.

tools/maven/checkstyle.xml
csviri: Those should be deprecated

tools/maven/checkstyle.xml
csviri: It is deprecated

tools/maven/checkstyle.xml
mxm: Does the build still pass with these removed?

csviri: yes (tested it with texblocks locally without this change it fails), it is bit surprizing I agree :)

pom.xml
```diff
@@ -281,13 +281,13 @@ under the License.
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-checkstyle-plugin</artifactId>
-                <version>2.17</version>
+                <version>3.4.0</version>
```
kaygorodov: Any issues with upgrading to the latest i.e. to 3.6.0 https://maven.apache.org/plugins/maven-checkstyle-plugin/dependency-info.html

```
<plugin>
  <groupId>org.apache.maven.plugins</groupId>
  <artifactId>maven-checkstyle-plugin</artifactId>
  <version>3.6.0</version>
```
?

tools/maven/checkstyle.xml
kaygorodov: I think it was removed for JavadocMethod, but for JavadocType it is still there, please check:

https://checkstyle.sourceforge.io/checks/javadoc/javadoctype.html#scope

removing it would result the config defaulting to _private_

pom.xml
csviri: no, good catch, upgraded, thx!

tools/maven/checkstyle.xml
csviri: yep, reverted this. thx

pom.xml
```diff
@@ -281,13 +281,13 @@ under the License.
...
                 <artifactId>maven-checkstyle-plugin</artifactId>
-                <version>2.17</version>
+                <version>3.6.0</version>
                 <dependencies>
                     <dependency>
                         <groupId>com.puppycrawl.tools</groupId>
                         <artifactId>checkstyle</artifactId>
                         <!-- Note: match version with docs/flinkDev/ide_setup.md -->
-                        <version>8.14</version>
+                        <version>10.17.0</version>
```
kaygorodov: BTW, what about the above comment:

> 

I don't see such file anymore or mentioning of _com.puppycrawl.tools_ in the docs, it seems to be obsolete and probably should be deleted

mxm: Is this dependency required? Why such an old version? Latest version is 11.0.1: https://mvnrepository.com/artifact/com.puppycrawl.tools/checkstyle

csviri: Good point, I did not found anything about this in checkstyle docs either so removed it for now.

mxm: Did you actually remove it? I still see it.

csviri: pushed it now, sorry

PR #1025, opened 2025-09-15
Fixing typos in README.md files

schongloo:

* Fixed typos in a couple of README.md files, one grammar and another configuration related


morhidi (approved):
Thanks for the fix @schongloo !

PR #1035, opened 2025-09-27
[FLINK-38448] Replace `burnett01/rsync-deployments` with a custom `rsync-deploy` action in `docs.yaml`

ferenc-csaky:


gyfora (commented):
Can you please add a minimal context on why this change is required and where the new action.yml is coming from?

gyfora:
How about we simply copy the appropriate action from the burnett01/rsync-deployments codebase: https://github.com/Burnett01/rsync-deployments/tree/7.1.0

We could copy the logic as is, and add the MIT license in: `LICENSE.rsync-deployments`

ferenc-csaky:
> How about we simply copy the appropriate action from the burnett01/rsync-deployments codebase: https://github.com/Burnett01/rsync-deployments/tree/7.1.0
> 
> We could copy the logic as is, and add the MIT license in: `LICENSE.rsync-deployments`

That action runs in a Docker img. So we need to copy over pretty much the whole codebase (action.yml, Dockerfile, entrypoint.sh). It has the benefit of a fixed rsync version, I'm 99% sure it won't matter at all. I'm okay to do that though, just let me know if that would be preferred.

PR #1072, opened 2026-03-19
Move setLastReconciledSpec to after successful transition start

jennifer-xiong25:
This fixes a bug where the last reconciled spec on the flink b/g deployment shows an updated spec even when the update failed.

## Summary
- Move `setLastReconciledSpec()` to after `startTransition()` and `startSavepointRedeployTransition()` succeed, so it's not stamped if the transition fails
- Previously, `lastReconciledSpec` was set before the transition call — if the transition failed or was later aborted, `lastReconciledSpec` reflected the new spec while the active child still ran the old one. Subsequent deploys would see no diff and fall into PATCH_CHILD (in-place upgrade) instead of a proper B/G transition
- Add assertion in `verifySavepointFetchFailureRecovery` confirming `lastReconciledSpec` does not contain the failed spec change

## Test plan
- [x] `verifySavepointFetchFailureRecovery` passes with new assertion
- [x] Full B/G controller test suite passes (72 tests)

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/bluegreen/BlueGreenDeploymentService.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/bluegreen/BlueGreenDeploymentService.java
@@ -162,10 +162,18 @@ public class BlueGreenDeploymentService {
                                 .rescheduleAfter(getReconciliationReschedInterval(context));
                     }
 
-                    setLastReconciledSpec(context);
                     try {
-                        return startTransition(
-                                context, currentBlueGreenDeploymentType, currentFlinkDeployment);
+                        var result =
+                                startTransition(
+                                        context,
+                                        currentBlueGreenDeploymentType,
+                                        currentFlinkDeployment);
+                        // Only stamp lastReconciledSpec after the transition
+                        // succeeds. If stamped before and the transition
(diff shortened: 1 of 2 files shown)


gyfora:
@jennifer-xiong25 can you please open a jira ticket for this?

gyfora:
@jennifer-xiong25 @jhung-lyft would be nice if we could wrap this PR up and get the fix in

jhung-lyft:
Hi @gyfora, thanks for checking in! Just to clarify, I cherry-picked @jennifer-xiong25's fix internally, so she'll still be  driving this pr to completion here.

Also, we found and fixed another bug (stale status.upgradeSavepointPath causing JM to restore from an old savepoint on retry). We’d love to submit this upstream as a follow-up PR once our internal CLA process is cleared up. Thank you!

gyfora:
thanks @jhung-lyft , looking forward to it!

gyfora:
Are you guys still working on this? It would be nice to merge it this week for the upcoming release. Please create a JIRA ticket for it

gyfora:
I created a JIRA ticket: https://issues.apache.org/jira/browse/FLINK-40333

gyfora (changes requested):
Overall I am not so sure the fix is good, the startTransition logic actually deploys deployments etc including patching the status afterwards, at that point the lastReconciledSpec Must be set. so we have to set it before I believe. It doesn't mean it's stable but once a deployment action is made the spec is reconciled.

PR #1075, opened 2026-03-19
[hotfix][autoscaler] Cover ScalingRealizer overrides with tests

Dennis-Mircea:

`JobAutoScalerImpl#scale` invokes `applyParallelismOverrides` and `applyConfigOverrides` from its `finally` block on every reconciliation cycle, so that the autoscaler's persisted decisions are continuously re-projected onto `spec.flinkConfiguration`. This is intentional: the unconditional projection doubles as a self-heal hook against in-flight upgrades that may externally reset the spec (last-state recovery, blue/green promotion, manual `kubectl edit`, JOSDK informer cache lag).

This contract was previously not pinned by tests. Earlier review on PR #1075 surfaced the concern that skipping the override application while the job is not yet `RUNNING` would cause the reconciler to detect a spec change and trigger a spurious second upgrade, but no regression guard existed to catch a future change re-introducing that skip.

This pull request adds the missing test coverage. No production code changes.

## Brief change log

- *[flink-autoscaler]* `JobAutoScalerImplTest#testOverridesAreReappliedWhenJobNotRunning`, regression guard asserting that while `JobStatus == INITIALIZING`, both `applyParallelismOverrides` and `applyConfigOverrides` are still invoked from `scale()`'s `finally` block, so the realizer keeps re-projecting the autoscaler's persisted state onto `spec.flinkConfiguration`.
- *[flink-kubernetes-operator]* `KubernetesScalingRealizerTest`, additional tests around `realizeParallelismOverrides` and `realizeConfigOverrides` covering the apply path on a fresh spec, the re-apply path when the spec already carries the intended values, and the drift path when the spec was externally reset.

## Verifying this change
This change is test-only. Production behavior is unchanged.

- `mvn -pl flink-autoscaler -Dtest=JobAutoScalerImplTest test` -> `BUILD SUCCESS`.
- `mvn -pl flink-kubernetes-operator -Dtest=KubernetesScalingRealizerTest test` -> `BUILD SUCCESS`.

## Documentation

### DIFF
--- a/flink-autoscaler/src/test/java/org/apache/flink/autoscaler/JobAutoScalerImplTest.java
+++ b/flink-autoscaler/src/test/java/org/apache/flink/autoscaler/JobAutoScalerImplTest.java
@@ -18,6 +18,7 @@
 package org.apache.flink.autoscaler;
 
 import org.apache.flink.api.common.JobID;
+import org.apache.flink.api.common.JobStatus;
 import org.apache.flink.autoscaler.config.AutoScalerOptions;
 import org.apache.flink.autoscaler.event.TestingEventCollector;
 import org.apache.flink.autoscaler.exceptions.NotReadyException;
@@ -293,6 +294,12 @@ public class JobAutoScalerImplTest {
         autoscaler.scale(context);
         assertParallelismOverrides(Map.of(v1, "1", v2, "2"));
 
+        // Test job not running: overrides must still be re-projected onto the spec so that an
+        // in-flight upgrade does not regress the autoscaler's last decision (self-heal).
+        var notRunningContext = context.toBuilder().jobStatus(JobStatus.INITIALIZING).build();
+        autoscaler.scale(notRunningContext);
+        assertParallelismOverrides(Map.of(v1, "1", v2, "2"));
+
         // Make sure cleanup removes everything
         assertTrue(stateStore.hasDataFor(context));
         autoscaler.cleanup(context);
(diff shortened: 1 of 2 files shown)


flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
```diff
@@ -97,6 +98,7 @@ public void scale(Context ctx) throws Exception {
             if (ctx.getJobStatus() != JobStatus.RUNNING) {
                 LOG.debug("Autoscaler is waiting for stable, running state");
                 lastEvaluatedMetrics.remove(ctx.getJobKey());
+                waiting = true;
```
davidradl: this change seems reasonable - is it possible to unit test?

Dennis-Mircea: Sure! I updated the PR!

gyfora (approved):

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
```diff
@@ -107,12 +109,16 @@ public void scale(Context ctx) throws Exception {
...
-                applyConfigOverrides(ctx);
-            } catch (Exception e) {
-                LOG.error("Error applying overrides.", e);
-                onError(ctx, autoscalerMetrics, e);
+            // Skip applying overrides when the job is not yet in a stable running state,
+            // as they may have already been applied during the previous scaling cycle.
+            if (!waiting) {
+                try {
+                    applyParallelismOverrides(ctx);
+                    applyConfigOverrides(ctx);
```
gyfora: Actually this makes me think, if the overrides are not applied that means the reconciler will detect it as a change ?

Because it started an upgrade, last reconciled spec contains the overrides, the job is not running yet and the overrides are not there now -> wouldn't it try to reconcile it?  sounds incorrect.

gyfora (changes requested):
Taking another look , I am not so sure if this change is correct

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
Dennis-Mircea: You're right, the `if (!waiting)` skip breaks the self-heal that the unconditional `finally` invocation was implicitly providing: if anything resets `spec.flinkConfiguration` between cycles while the job is mid-upgrade, the next reconcile diffs against `lastReconciledSpec` and fires a spurious second upgrade.

I've reworked the PR to keep the realizer always invoked from `finally `and pushed the idempotency check down into `KubernetesScalingRealizer`, so steady-state cycles do no spec writes and emit no debug noise, while drift is still self-healed. I also open a JIRA as part of it: [FLINK-39564](https://issues.apache.org/jira/browse/FLINK-39564).

gyfora (changes requested):
Feels to me like an unnecessary change. The log just says that any override that exist will be applied , it was correct because autoscaler parallelism settings are applied as an override on the spec in every reconciliation round.

I would prefer to simply add a test to guard the current behaviour

Dennis-Mircea:
> Feels to me like an unnecessary change. The log just says that any override that exist will be applied , it was correct because autoscaler parallelism settings are applied as an override on the spec in every reconciliation round.
> 
> I would prefer to simply add a test to guard the current behaviour

Done! I also closed the previous open JIRA and I kept this as a hotfix only. It should all be fine now.

gyfora (approved):

Dennis-Mircea:
@gyfora The CI workflow is failing on some e2e tests that are not related with the current PR changes. I opened the #1107 PR for fixing the flaky `minikube ingress addon enable` command that is executed across the `e2e` tests.

PR #1078, opened 2026-03-23
[FLINK-39306][flink-autoscaler] Align the busy-time TRUE_PROCESSING_RATE numerator estimator with the busy-time aggregator

Dennis-Mircea:

The busy-time `TRUE_PROCESSING_RATE` is a ratio, `busyTimeTpr = inputRate / (busyTimeAvg / 1000)`, but its two halves are estimated over the metric window with different methods. The denominator `busyTimeAvg` follows the configured busy-time aggregator, while the numerator `inputRate` always uses `getRate(NUM_RECORDS_IN)`. Under the default `MAX` aggregator the denominator is a per-second gauge mean (`getAverage(LOAD) * 1000`) while the numerator is a cumulative endpoint rate, so the ratio mixes two different temporal estimators.

`getRate` (time-weighted average) and `getAverage` of a per-second gauge (unweighted sample mean) agree under uniform sampling but diverge under non-uniform sampling. Because `busyTimeTpr` is a ratio of two co-varying quantities, using one shared estimator for both lets the common sampling weighting cancel, while mixing them leaves the denominator's sampling artifact in the result.

This change makes the numerator follow the same estimator as the denominator: per-second gauge mean under `MAX`/`MIN`, cumulative `getRate` under `AVG`. It is an internal-consistency fix scoped to the busy-time TPR numerator. It is not an accuracy claim about the underlying capacity model.

## Brief change log

  - Added `computeTprInputRate(conf, vertex, metricsHistory)` in `ScalingMetricEvaluator`: returns `getAverage(NUM_RECORDS_IN_PER_SECOND)` under `MAX`/`MIN` (matching the per-second `busyTimeAvg`) and `getRate(NUM_RECORDS_IN)` under `AVG` (matching the cumulative `busyTimeAvg`), with a fallback to `getRate(NUM_RECORDS_IN)` when the per-second gauge is unavailable (older Flink, missing metric).
  - `computeTrueProcessingRate` now takes its numerator from `computeTprInputRate`. The demand path (`computeTargetDataRate`), backlog detection, and the edge data-rate / output-ratio propagation are unchanged and keep the exact `getRate`, since they are not part of this ratio.
  - Collected and stored the `numRecordsInPerSecond` task metric for non-source vertices (`FlinkMetric.NUM_RECORDS_IN_PER_SEC`, `ScalingMetric.NUM_RECORDS_IN_PER_SECOND`, `ScalingMetricCollector`, `ScalingMetrics`).
  - Fixed a flaky `JobAutoScalerImplTest` by injecting a deterministic `Clock`, so successive metric collections get distinct timestamps and the metric history does not collapse to a single sample.

## Verifying this change

  - `ScalingMetricEvaluatorTest#computeTprInputRateTest` asserts the gating: `MAX`/`MIN` use the per-second gauge mean, `AVG` uses the cumulative `getRate`, and the per-second-unavailable case falls back to `getRate`.
  - Existing `ScalingMetricEvaluatorTest`, `MetricsCollectionAndEvaluationTest`, and `BacklogBasedScalingTest` continue to pass. Behavior is unchanged under `AVG` and under uniform sampling.

  - Core observer or reconciler logic that is regularly executed: yes (autoscaler metric evaluation), but only the busy-time TPR numerator, and only under non-uniform sampling with the default/`MAX` aggregator

## Documentation

### DIFF
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricEvaluator.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricEvaluator.java
@@ -158,7 +158,11 @@ public class ScalingMetricEvaluator {
                 TRUE_PROCESSING_RATE,
                 EvaluatedScalingMetric.avg(
                         computeTrueProcessingRate(
-                                busyTimeAvg, inputRateAvg, metricsHistory, vertex, conf)));
+                                busyTimeAvg,
+                                computeTprInputRate(conf, vertex, metricsHistory),
+                                metricsHistory,
+                                vertex,
+                                conf)));
 
         evaluatedMetrics.put(LOAD, EvaluatedScalingMetric.avg(busyTimeAvg / 1000.));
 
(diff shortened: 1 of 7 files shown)


flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricEvaluator.java
```diff
@@ -494,6 +503,49 @@ public static double getAverage(
...
+            if (Double.isNaN(value)) {
+                continue;
+            }
+            if (!Double.isNaN(prevValue)) {
+                long tsDiff = Duration.between(prevTs, entry.getKey()).toMillis();
+                if (tsDiff > 0) {
+                    sumRates += 1000.0 * (value - prevValue) / tsDiff;
+                    n++;
+                }
+            }
```
gyfora: Could you please explain the logic here on a high level? The original intention with only using the cumulative values to compute the average rate is that you actual get correct average input rate across the metric window. You have num records in the beginning, num records at the end and you get the average. 

Can you please explain the averaging logic here that would lead to a better result using the point in time per/second values? Feels like that can actually fluctuate quite a lot and would not necessarily lead to the correct value.

gyfora (approved):

PR #1082, opened 2026-04-02
[FLINK-33548] Replace taskManager/jobmanager.resource with resources conforming to ResourceRequirements

vsantwana:

This pull request introduces a new resources field on JobManagerSpec and TaskManagerSpec that conforms to the standard Kubernetes `ResourceRequirements` schema (with requests and limits), replacing the non-standard resource field which only supported a flat cpu/memory/ephemeralStorage structure. The old resource field is deprecated but remains fully functional as a fallback.

## Brief change log

 - Added resources field (Kubernetes ResourceRequirements) to JobManagerSpec and TaskManagerSpec, deprecated the old resource field
  - Updated FlinkConfigBuilder to prioritize resources over resource in applyJobManagerSpec(), applyTaskManagerSpec(), and applyPodTemplate()
  - Added setResourceRequirements() method that maps requests.memory and requests.cpu to Flink config, and computes CPU_LIMIT_FACTOR and MEMORY_LIMIT_FACTOR from the
  limits/requests ratio
  - Added configureCpuFromString() to parse Kubernetes CPU quantity strings (e.g. "500m", "1.5")
  - Updated DefaultValidator to validate ResourceRequirements fields and accept memory defined via either resource.memory or resources.requests.memory
  - Updated CRD schema for flinkdeployments to include the new resources field
  - Fixed FlinkBlueGreenDeploymentControllerTest to use builder pattern instead of all-args constructor

## Verifying this change

  - Added testApplyJobManagerSpecWithResourceRequirements — verifies JM memory and CPU are set from ResourceRequirements
  - Added testApplyTaskManagerSpecWithCpuResourceRequirements — verifies milliCPU parsing ("500m" -> 0.5)
  - Added testResourceRequirementsCpuLimitFactor — verifies CPU limit factor is computed as limits.cpu / requests.cpu
  - Added testResourceRequirementsMemoryLimitFactor — verifies memory limit factor is computed as limits.memory / requests.memory
  - Added testResourceRequirementsTakesPrecedenceOverResource — verifies resources wins when both resource and resources are set
  - Added testApplyResourceRequirementsToPodTemplate — verifies ResourceRequirements are correctly merged into pod template containers
  - Added testDeprecatedResourceStillWorksWhenResourcesIsNull — verifies backward compatibility of the old resource field
  - Added testTaskManagerSpecWith2_GiResourceRequirementsSetting — verifies Kubernetes memory format "2Gi" is correctly parsed
  - Added DefaultValidator tests verifying ResourceRequirements with valid memory, valid memory+CPU, and resources-only (no deprecated resource) all pass validation

## Documentation

  - If yes, how is the feature documented? Javadocs


helm/flink-kubernetes-operator/crds/flinksessionjobs.flink.apache.org-v1.yml
```diff
@@ -1,253 +1,250 @@
...
+      type: "string"
+    name: "v1beta1"
     schema:
       openAPIV3Schema:
         properties:
           spec:
             properties:
               deploymentName:
-                type: string
+                type: "string"
```
vsantwana, line 32: I need to check why this change and if it is needed.

lqy1992:
Just wondering, is there any plan to release this? I'd really love to have it.

Dennis-Mircea:
> lqy1992

@vsantwana are you still working on it?

vsantwana:
Hey @Dennis-Mircea 
Yes, did not get time in between. Taking it up again

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/spec/TaskManagerSpec.java
```diff
@@ -40,8 +41,18 @@
 @Builder
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class TaskManagerSpec implements Diffable<TaskManagerSpec> {
-    /** Resource specification for the TaskManager pods. */
-    private Resource resource;
+    /**
+     * Resource specification for the TaskManager pods.
+     *
+     * @deprecated Use resources instead for proper Kubernetes ResourceRequirements schema.
```
Dennis-Mircea: This @deprecated comment is ambiguous.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/spec/JobManagerSpec.java
```diff
@@ -36,8 +37,19 @@
 @Builder
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class JobManagerSpec {
-    /** Resource specification for the JobManager pods. */
-    private Resource resource;
+    /**
+     * Resource specification for the JobManager pods.
+     *
+     * @deprecated Use resourceRequirements instead for proper Kubernetes ResourceRequirements
```
Dennis-Mircea: This @deprecated comment is ambiguous and not well formatted.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
```diff
@@ -196,6 +198,61 @@ protected FlinkConfigBuilder applyLogConfiguration() throws IOException {
...
+        if (mainContainerOpt.isPresent()) {
+            mainContainer = mainContainerOpt.get();
+        } else {
+            mainContainer = new Container();
+            mainContainer.setName(Constants.MAIN_CONTAINER_NAME);
+            podSpec.getContainers().add(mainContainer);
+        }
+
+        if (mainContainer.getResources() == null) {
+            mainContainer.setResources(new io.fabric8.kubernetes.api.model.ResourceRequirements());
```
Dennis-Mircea: Why to use here full package naming?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
```diff
@@ -196,6 +198,61 @@ protected FlinkConfigBuilder applyLogConfiguration() throws IOException {
         return this;
     }
 
+    @VisibleForTesting
+    protected static PodTemplateSpec applyResourceRequirementsToPodTemplate(
+            PodTemplateSpec podTemplate, ResourceRequirements requirements) {
+        if (requirements == null
+                || (requirements.getRequests().isEmpty() && requirements.getLimits().isEmpty())) {
```
Dennis-Mircea: Here you are not checking `requirements.getRequests() != null` in order to have symmetry with `setResourceRequirements` check.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
```diff
@@ -480,6 +639,49 @@ private void configureCpu(Resource resource, Configuration conf, boolean isJM) {
...
+                        : KubernetesConfigOptions.TASK_MANAGER_CPU;
+        conf.set(cpuConfigOption, cpu);
+
+        if (!spec.getFlinkVersion().isEqualOrNewer(FlinkVersion.v1_17)) {
+            String legacyKey = isJM ? "kubernetes.jobmanager.cpu" : "kubernetes.taskmanager.cpu";
+            conf.setDouble(legacyKey, cpu);
+        }
+    }
+
+    private static double parseCpuValue(String cpuValue) {
```
Dennis-Mircea: In the DefaultValidator you using already `Quantity.parse` for CPU. Why not to use it also here?

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilderTest.java
```diff
@@ -561,6 +565,162 @@ public void testTaskManagerSpecWith2_GiSetting() {
...
+                        .build();
+        assertEquals(
+                MemorySize.parse("4294967296 b"),
+                configuration.get(JobManagerOptions.TOTAL_PROCESS_MEMORY));
+        assertEquals(
+                Double.valueOf(2.0), configuration.get(KubernetesConfigOptions.JOB_MANAGER_CPU));
+    }
+
+    @Test
+    public void testApplyResourceRequirementsToPodTemplate() {
```
Dennis-Mircea, line 685: It will be good to have a UT testing mixed units (e.g. `1Gi` request vs `2048Mi` limit, and `500m` vs `1` for CPU).

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
```diff
@@ -269,9 +348,89 @@ protected FlinkConfigBuilder applyJobManagerSpec() {
...
+        }
+
+        // Handle CPU from requests
+        if (requests != null && requests.containsKey("cpu")) {
+            String cpuValue = requests.get("cpu").toString();
+            configureCpuFromString(cpuValue, effectiveConfig, isJM);
+        }
+
+        // Handle CPU limit factor: limits.cpu / requests.cpu
+        if (limits != null
```
Dennis-Mircea: With the current implementation if only `limits` is set without `requests` the implementation doesn't do anything. Kubernetes itself defaults requests = limits when requests are omitted.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
```diff
@@ -232,8 +297,17 @@ protected FlinkConfigBuilder applyPodTemplate() throws IOException {
             tmPodTemplate =
                     mergePodTemplates(
                             commonPodTemplate, spec.getTaskManager().getPodTemplate(), mergeByName);
-            tmPodTemplate =
-                    applyResourceToPodTemplate(tmPodTemplate, spec.getTaskManager().getResource());
+
+            // Prioritize new ResourceRequirements, then fall back to old Resource
```
Dennis-Mircea: The comment is wrong, the if-else branch is not equal with a "fall back" mechanism.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkConfigBuilder.java
```diff
@@ -207,8 +264,16 @@ protected FlinkConfigBuilder applyPodTemplate() throws IOException {
                     mergePodTemplates(
                             commonPodTemplate, spec.getJobManager().getPodTemplate(), mergeByName);
 
-            jmPodTemplate =
-                    applyResourceToPodTemplate(jmPodTemplate, spec.getJobManager().getResource());
+            // Prioritize new ResourceRequirements, then fall back to old Resource
```
Dennis-Mircea: The comment is wrong, the if-else branch is not equal with a "fall back" mechanism.

Dennis-Mircea (changes requested):
Thanks for picking this up, it's a genuinely useful feature. To be honest though: this PR reads as heavily AI-generated with fairly light human review. Several of the issues raised in the review are the kind a careful pass or real testing would have caught.

The "Verifying this change" section also only lists unit-test descriptions and reads as generated, there's no sign the change was exercised on a real cluster. For something that touches the resource model of every JM/TM pod, this needs focused manual verification on a local cluster (like minikube) with concrete examples, at minimum:

- `resources.requests` only (memory + cpu): confirm the JM/TM pods get the expected requests and the right Flink memory/CPU config.
- `requests` + `limits`: confirm the computed CPU/memory limit factors produce the intended pod limits.
- limits-only, and a spec with both `resource` and `resources`, to confirm precedence behaves as intended.
- the deprecated `resource` path still behaves unchanged.

Per the project's AI-tooling guidance (official link: https://www.apache.org/legal/generative-tooling.html), the author owns the correctness of every change regardless of tooling, so it would help to run through these locally and update the verification section with what was actually observed, not just the unit tests.

vsantwana:
Hey @Dennis-Mircea
I was still developing this. I picked this again today about few hours ago, but I agree that it is my fault of not marking it as not ready for review. Gyula had mentioned that he will review tomorrow and I was using that time to human review it and am actively making changes. Please excuse that you had to spend time on a PR that was not fully self reviewed.

e2e-tests/data/resource-requirements-cr.yaml
```diff
@@ -0,0 +1,73 @@
...
+metadata:
+  namespace: default
+  name: flink-example-resource-requirements
+spec:
+  image: flink:1.20
+  flinkVersion: v1_20
+  flinkConfiguration:
+    taskmanager.numberOfTaskSlots: "2"
+    kubernetes.operator.snapshot.resource.enabled: "false"
+  serviceAccount: flink
```
gyfora: Instead if introducing a new test, I suggest updating the existing tests to use the new syntax (this is going to be the way forward). We should keep a single e2e that uses the old syntax (not necessary to add a separate test case)

helm/flink-kubernetes-operator/crds/flinkbluegreendeployments.flink.apache.org-v1.yml
```diff
@@ -10947,4 +10999,6 @@ spec:
     served: true
     storage: true
     subresources:
+      scale:
+        specReplicasPath: ".spec.template.spec.taskManager.replicas"
```
gyfora: why this change?

gyfora: I think there is some logic to explicitly remove this

gyfora (changes requested):
Overall the logic looks good but I think we should update our examples and our existing test defaults (like the default flinkdeployment we create in tests) to align with the new non-deprecated api. This will also extend our test coverage

helm/flink-kubernetes-operator/crds/flinkbluegreendeployments.flink.apache.org-v1.yml
vsantwana: https://github.com/apache/flink-kubernetes-operator/pull/1143

gyfora: can you please simply include this in the PR? no need for another PR / commit

gyfora (changes requested):
Overall looks good, please upgrade the examples embedded in our documentation as well to use the new format, that would make sense going forward.

Also it would be nice if we could move the memory config / parsing related utility methods into a single unit testable utility class and use that both in the FlinkConfigBuilder and ScalingRealizer. It feels like there is way too many methods for handling the memory configs , maybe there is also some opportunities to simplify

vsantwana:
@gyfora thanks for the review.

**One open question for you**
The old `resource` and new `resources` paths are *not* fully symmetric in validation. After this change the new path is stricter (cpu validated + positivity + requests<=limits), whereas the deprecated `validateResources` never validated cpu at all. More importantly, **both** paths share one gap: when memory is defined via `resource.memory` *or* `resources.requests.memory`, the Flink process-spec memory check (`validateJmMemoryConfig`/`validateTmMemoryConfig`) is skipped, so a too-small memory value passes validation and only fails at deploy. This is pre-existing behavior (not introduced here), so I left it as-is. Just wanted to record it here, if we want to revisit the behaviour with this change?

gyfora (approved):

moh-incom:
Can you confirm that there still is an implicit `<manager>.<resource>.limit-factor` of `1` even if only a resource request is specified? I would like for my task manager pods to only have requests and not limits, but it does not seem like that is possible.

PR #1084, opened 2026-04-08
[FLINK-39967] Fix `backoffLimit=-1` to mean unlimited retries in FlinkStateSnapshot

vsantwana:

`FlinkStateSnapshotSpec.backoffLimit` defaults to `-1`, documented as meaning "unlimited" retries. However, the retry decision in `FlinkStateSnapshotController` did a plain numeric comparison `failures > backoffLimit`. With the default `-1`, after the first failure `failures` is `1`, so `1 > -1` is always `true` and the snapshot was immediately marked failed with no retry - the exact opposite of the documented behavior.

This PR fixes the comparison so that a negative `backoffLimit` (the `-1` sentinel) correctly means "retry indefinitely", while `0` means no retries and `N` means up to `N` retries.

## Brief change log

  - Treat `backoffLimit < 0` as "unlimited retries" in `FlinkStateSnapshotController`, so the documented `-1` default actually retries indefinitely instead of failing on the first error.
  - Added a test proving the bug and verifying the fix (snapshot with default `backoffLimit=-1` is retried after a failure rather than immediately failing).

## Verifying this change

  - `FlinkStateSnapshotControllerTest` covers a snapshot failure with the default `backoffLimit=-1` and asserts the snapshot is rescheduled for retry (with exponential backoff) instead of being marked failed with no retry.

  - Core observer or reconciler logic that is regularly executed: yes (snapshot reconciler retry handling)

## Documentation

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkStateSnapshotController.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkStateSnapshotController.java
@@ -128,7 +128,8 @@ public class FlinkStateSnapshotController
                 resource.getStatus().getError(),
                 ctx.getKubernetesClient());
 
-        if (resource.getStatus().getFailures() > resource.getSpec().getBackoffLimit()) {
+        var backoffLimit = resource.getSpec().getBackoffLimit();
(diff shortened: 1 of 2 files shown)


gyfora:
Can you please attach a JIRA to this PR?

gyfora (approved):

PR #1085, opened 2026-04-10
[FLINK-39555] FLIP-575: Scaling Executor Plugin SPI for Flink Autoscaler

Dennis-Mircea:

This pull request introduces the `ScalingExecutorPlugin` SPI as defined in [FLIP-575: Scaling Executor Plugin SPI for Flink Autoscaler](https://cwiki.apache.org/confluence/display/FLINK/FLIP-575%3A+Scaling+Executor+Plugin+SPI+for+Flink+Autoscaler). The SPI allows users to plug in custom logic that intercepts, modifies, or vetoes computed scaling decisions before they are applied by the autoscaler. It is complementary to [FLIP-514: Custom Evaluator plugin for Flink Autoscaler](https://cwiki.apache.org/confluence/display/FLINK/FLIP-514%3A+Custom+Evaluator+plugin+for+Flink+Autoscaler), which extends the metric evaluation layer; this contribution extends the scaling decision execution layer.

Unlike `FlinkAutoscalerEvaluator` (capped at one instance per job), `ScalingExecutorPlugin` natively supports multiple named instances chained together in a deterministic priority order, enabling composition of orthogonal concerns (resource gating, parallelism alignment, cost capping, audit, …) without coupling them inside a single plugin.

**NOTE**: This PR ensures the full alignment with the `[FLINK-39511] FLIP-514: Custom Evaluator plugin for Flink Autoscaler` PR (#1099).

## Brief change log

- **New SPI** - `org.apache.flink.autoscaler.ScalingExecutorPlugin<KEY, CTX>` with:
  - `priority()` (default `0`, lower runs first) for deterministic chain ordering.
  - `apply(Context, Map<JobVertexID, ScalingSummary>)` returning a (possibly modified) summary map; an empty map vetoes the scaling operation.
  - Immutable `Context` bundling the `JobAutoScalerContext`, the prefix-stripped per-instance `Configuration`, the evaluated metrics, and the job topology.

- **Discovery / selection split** - mirrors the `FlinkAutoscalerEvaluator` flow:
  - **Operator-startup discovery** (cached once) - `flink-kubernetes-operator/AutoscalerUtils#discoverCustomScalingExecutors(Configuration)` via the Flink `PluginUtils` mechanism; `flink-autoscaler-standalone/AutoscalerUtils#discoverCustomScalingExecutors(Configuration)` via Java `ServiceLoader`.
  - **Per-reconciliation selection** - `ScalingExecutor#resolveCustomScalingExecutors(Configuration)` matches the configured chain against the discovered SPI bag, sorts by `priority()`, and returns an unmodifiable ordered `Collection<Map.Entry<String, ScalingExecutorPlugin>>` carrying the user-chosen `<name>` for each resolved instance. This guarantees that per-job overrides via `FlinkDeployment.spec.flinkConfiguration` take effect on the next reconcile without restarting the operator.

- **Configuration** (metric-reporter idiom) in `AutoScalerOptions`:
  - `job.autoscaler.scaling.custom-executors` - list of named instances.
  - `job.autoscaler.scaling.custom-executor.<name>.class` - implementation FQN.
  - `job.autoscaler.scaling.custom-executor.<name>.<parameter>` - free-form per-instance options, delivered to the plugin via `Context.getConfiguration()` with the per-instance prefix stripped.
  - All keys honor the legacy `kubernetes.operator.`-prefixed form as a fallback; canonical wins on overlap.
  - Helpers: `customScalingExecutorClassKey`, `customScalingExecutorClassFallbackKey`, `customScalingExecutorClassOption`, `customScalingExecutorConfiguration`.

- **`ScalingExecutor` integration** - invokes `applyScalingExecutorPlugins` after the blocked check, after memory tuning, and after the resource-quota gate, but before persisting parallelism overrides. Approve / veto outcomes are surfaced as `ScalingApproved` / `ScalingVetoed` Kubernetes events with `messageKey = <reason>:<instanceName>` (so chained instances of the same plugin class do not de-duplicate). An INFO log line `Custom scaling executors resolved and sorted by priority: [...]` is emitted per reconciliation listing each `[name, class, priority]` entry in execution order.

- **Documentation** - new `## Custom Scaling Executors` section in `docs/content/docs/operations/plugins.md` covering the interface, SPI service file, packaging, YAML configuration, operator + standalone deployment with the corresponding discovery log lines, and the per-reconciliation chain log hint.

## Verifying this change

- **`flink-autoscaler/ScalingExecutorTest`** - extended with a dedicated nested `ScalingExecutorPluginTest` suite covering:
  - Approve / veto / modify / empty-map behaviour and chaining (approve→veto, veto-short-circuits, two-modify chain).
  - `priority()`-driven chain ordering using high / default / low priority instances registered out of order.
  - Approve / veto Kubernetes event emission with the correct `<reason>:<instanceName>` `messageKey`, including the no-plugins-registered case.
  - Legacy `kubernetes.operator.`-prefixed fallback for both the `scaling.custom-executors` list and the per-instance `<name>.class` key.
  - Per-instance scoped `Configuration` delivery (canonical prefix, legacy prefix, and canonical-wins-on-overlap semantics).
- **`flink-autoscaler-standalone/AutoscalerUtilsTest`** - verifies `ServiceLoader`-based discovery returns the registered `ScalingExecutorPlugin` instances.
- **`flink-kubernetes-operator/AutoscalerUtilsTest`** - verifies plugin-directory discovery via `PluginUtils`.

## Documentation

  - If yes, how is the feature documented? docs & JavaDocs


Dennis-Mircea:
This FLIP PR is now ready for the final review.

**NOTE**: If the #1099 PR will be merged before (and most probably it will) there will be some conflicts that will must be resolved afterwards.

Dennis-Mircea:
Raising the same comment as in #1099.

Coming back to this after stepping back to look at the broader module structure. Wanted to raise a design question before merge, because it affects what custom-plugin authors will compile against. This applies to #1099 PR as well.

### The asymmetry today

The operator's plugin SPIs are split across two modules:

| Module | SPIs it carries |
|---|---|
| `flink-kubernetes-operator-api` (thin) | `FlinkResourceListener` |
| `flink-kubernetes-operator` (implementation, heavy) | `FlinkResourceValidator`, `FlinkResourceMutator` |

A custom-validator or custom-mutator author has to depend on the heavy implementation module (which transitively pulls JOSDK, fabric8, the autoscaler, etc.) just to compile against one interface. A custom-listener author depends on a much thinner `-api` artifact. The asymmetry is historical, not designed.

### The question

This PR adds new plugin-style SPIs to `flink-autoscaler`, which is the implementation module. Would it be cleaner to land them in a dedicated `flink-autoscaler-api` module instead, so the user-facing answer to "what do I import to write an autoscaler plugin?" is a single thin artifact?

The plugin author's first question when implementing a new SPI is always "which module do I import?". Today that answer is inconsistent across SPIs and points at heavy implementation artifacts in most cases. Landing these new SPIs in a thin `flink-autoscaler-api` would set a cleaner precedent.

### Why now, not after merge

Once the new SPIs are in `flink-autoscaler`, the project implicitly commits to "autoscaler plugin SPIs live in the implementation module" as the pattern. A later restructuring becomes a multi-PR migration that breaks plugin-author coordinates a second time. The cost of asking the question now is a slight delay on this PR. The cost of asking it later is significantly higher.

### The broader implication

If we want to standardize on "plugin SPIs live in `-api` modules" going forward, that also implies:

- Moving `FlinkResourceValidator` and `FlinkResourceMutator` into `flink-kubernetes-operator-api` to restore symmetry with `FlinkResourceListener`.
- Creating `flink-autoscaler-api` carrying both the existing `JobAutoScaler` / `ScalingRealizer` / `AutoScalerStateStore` and the new SPIs from this PR.

That's a project-wide structural decision and would warrant a `[DISCUSS]` thread on dev@flink.apache.org before all of it lands. I do not want to block this PR with that broader conversation, but the immediate question (where the new SPIs in this PR should live) shapes the email and is worth aligning on first.

@gyfora, what do you think?

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
```diff
@@ -70,27 +77,43 @@ public class ScalingExecutor<KEY, Context extends JobAutoScalerContext<KEY>> {
...
+    public static final String SCALING_VETOED_REASON = "ScalingVetoed";
+    public static final String SCALING_APPROVED_REASON = "ScalingApproved";
+
     private static final Logger LOG = LoggerFactory.getLogger(ScalingExecutor.class);
 
     private final JobVertexScaler<KEY, Context> jobVertexScaler;
     private final AutoScalerEventHandler<KEY, Context> autoScalerEventHandler;
     private final AutoScalerStateStore<KEY, Context> autoScalerStateStore;
     private final ResourceCheck resourceCheck;
+    private final Collection<ScalingExecutorPlugin<KEY, Context>> discoveredScalingCustomExecutors;
```
gyfora: why not simply call it customExecutors?

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
```diff
@@ -459,6 +493,144 @@ private boolean checkIfBlockedAndTriggerScalingEvent(
         return !scaleEnabled || isExcluded;
     }
 
+    /**
+     * Apply pluggable scaling custom executors (SPI extension point).
+     *
+     * <p>The chain to apply is resolved per reconciliation against the per-job Configuration,
+     * mirroring custom metric evaluator plugin flow.
+     */
+    private Map<JobVertexID, ScalingSummary> applyScalingExecutorPlugins(
```
gyfora: rename to applyCustomExecutors

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
```diff
@@ -459,6 +493,144 @@ private boolean checkIfBlockedAndTriggerScalingEvent(
...
+                        SCALING_VETOED_REASON,
+                        String.format(
+                                "Scaling decision vetoed by scaling custom executor instance '%s' (class=%s).",
+                                instanceName, pluginClassName),
+                        SCALING_VETOED_REASON + ":" + instanceName,
+                        conf.get(SCALING_EVENT_INTERVAL));
+                return null;
+            }
+            LOG.info(
+                    "Scaling decision approved by scaling custom executor instance '{}' (class={}) with summaries: {}",
```
gyfora: I don't think that we should log this, this is pretty normal

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
```diff
@@ -459,6 +493,144 @@ private boolean checkIfBlockedAndTriggerScalingEvent(
...
+                    instanceName,
+                    pluginClassName,
+                    scalingSummaries.values());
+            autoScalerEventHandler.handleEvent(
+                    context,
+                    AutoScalerEventHandler.Type.Normal,
+                    SCALING_APPROVED_REASON,
```
gyfora: This should not be an event

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
```diff
@@ -459,6 +493,144 @@ private boolean checkIfBlockedAndTriggerScalingEvent(
...
+                            context, pluginConfiguration, evaluatedMetrics, jobTopology);
+            scalingSummaries = plugin.apply(pluginContext, scalingSummaries);
+            if (scalingSummaries == null || scalingSummaries.isEmpty()) {
+                LOG.info(
+                        "Scaling vetoed by scaling custom executor instance '{}' (class={}).",
+                        instanceName,
+                        pluginClassName);
+                autoScalerEventHandler.handleEvent(
+                        context,
+                        AutoScalerEventHandler.Type.Warning,
```
gyfora, line 539: Why is this a warning?

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutorPlugin.java
```diff
@@ -0,0 +1,120 @@
...
+        /** The autoscaler context for the current job. */
+        CTX autoScalerContext;
+
+        /**
+         * The prefix-stripped, per-plugin configuration for this invocation, populated from {@code
+         * job.autoscaler.scaling.custom-executor.<name>.<parameter>} entries (with the {@code
+         * kubernetes.operator.} legacy fallback honored). Never {@code null}; empty when the plugin
+         * has no per-instance options configured.
+         */
+        Configuration configuration;
```
gyfora, line 112: the autoscaler context already has configuration

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutorPlugin.java
```diff
@@ -0,0 +1,120 @@
...
+     * access to the autoscaler context, configuration, evaluated metrics, and job topology.
+     *
+     * @param <KEY> The job key.
+     * @param <CTX> Instance of {@link JobAutoScalerContext}.
+     */
+    @Value
+    class Context<KEY, CTX extends JobAutoScalerContext<KEY>> {
+
+        /** The autoscaler context for the current job. */
+        CTX autoScalerContext;
```
gyfora, line 104: maybe this should extend the autoscaler context instead -> maybe do the same in the evaluator plugin too?

gyfora (changes requested):

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
Dennis-Mircea: Because it cancels entirely the scaling decision of the autoscaler, and I see it more like a warning than an info event.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutorPlugin.java
Dennis-Mircea: I'd keep the wrapped `autoScalerContext` field rather than extend it. By not extending it you have the power of keeping generic and make it applicable for both operator autoscaler and standalone autoscaler. By extending it will introduce too much of unnecessary complexity.

Also, by not extending the CTX is better from the point of view of maintenance for every new change that the CTX may introduce.

On doing the same in the evaluator plugin, I'd leave it. The evaluator runs at a different phase (metric evaluation) and the `ScalingMetricEvaluator` doesn't even receive the actual autoscaler context from `JobAutoScalerImpl`.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutorPlugin.java
Dennis-Mircea: This field isn't the job configuration that autoscaler context contains, it's the per-instance, prefix-stripped plugin options, built from `job.autoscaler.scaling.custom-executor.<name>.<parameter>`. It's intentionally distinct from `autoScalerContext.getConfiguration()`, which is the shared job-level config.

It has to be instance-scoped because the executor supports multiple instances of the same plugin class registered under different `<name>`s, each receiving its own options.

I considered merging it into the job config (as the evaluator plugin does), but the evaluator is single-instance, so a merged view is unambiguous there. For a multi-instance chain I'd rather keep the Configuration explicit.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
Dennis-Mircea: Right, I removed the log and the event and let a single INFO log message instead only when the scalingSummaries differs. I also updated the placement of the general scaling event to be placed after the memory tuning and custom executors are applied.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
Dennis-Mircea: I removed it!

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
Dennis-Mircea: Addressed!

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
Dennis-Mircea: Yes, that's right, I missed this one in my previous refinements. I fixed it now!

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutorPlugin.java
gyfora: still I think we should try to extend the autoscaler context. And the same way we can merge it on top of the config as we did for the evaluator plugin. It's much cleaner to have a single context that extends what we already have instead of create a context that holds the other context and another config etc etc.

in addition we should do this also in https://github.com/apache/flink-kubernetes-operator/blob/main/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/metrics/ScalingMetricsEvaluatorPlugin.java (we can do in this PR or as a follow-up for the current operator release).

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
```diff
@@ -469,6 +475,11 @@ private boolean checkIfBlockedAndTriggerScalingEvent(
             Instant now) {
         var scaleEnabled = conf.get(SCALING_ENABLED);
         var isExcluded = CalendarUtils.inExcludedPeriods(conf, now);
+        if (scaleEnabled && !isExcluded) {
+            // Not blocked. The scaling report is emitted later, once the decision is final
+            // (after memory tuning, the max-resource check, and the custom scaling executors).
+            return false;
```
gyfora, line 481: sorry maybe it's a stopid question but why do we return early if we are not blocked? wouldn't this prevent us from ever reaching "autoScalerEventHandler.handleScalingEvent" in the normal scaling case?

Dennis-Mircea: Returning false let us proceed further and not act as "blocked", so memory tuning and scaling executors can be applied further + `autoScalerEventHandler.handleScalingEvent`. By returning true here it means that we are "blocked" and scaling execution is aborted.

gyfora: I mean compared to the previous logic the only difference seems to be that we don't trigger autoScalerEventHandler.handleScalingEvent. The previous logic also returned false at the end , this is what I don't get.

Dennis-Mircea: Compared to the previous change I removed the code block from `453` line (`message = SCALING_SUMMARY_HEADER_SCALING_EXECUTION_ENABLED`) which was equivalent with `autoScalerEventHandler.handleScalingEvent` that is now postponed / moved for later at line `180`.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutorPlugin.java
Dennis-Mircea: I created [FLINK-39959](https://issues.apache.org/jira/browse/FLINK-39959) to be tackled as follow up.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingExecutor.java
gyfora: ah makes sense, sorry

gyfora (approved):

PR #1086, opened 2026-04-13
[FLINK-39432] Add option to delete with session running session jobs

gyfora:

Add configuration to allow deletion of session clusters without blocking on session jobs. This allows the deletion and upgrade of clusters even in cases when the jobs cannot be deleted for some reason and may be desirable in some use-cases (where the session cluster upgrade is more important than the job health). The new config is disabled by default to preserve the existing behaviour

## Brief change log

- Introduce new config 
- New tests

## Verifying this change

Unit and manual testing

## Documentation

  - If yes, how is the feature documented? generated

### DIFF
--- a/docs/layouts/shortcodes/generated/dynamic_section.html
+++ b/docs/layouts/shortcodes/generated/dynamic_section.html
@@ -195,6 +195,12 @@
             <td>The interval before a savepoint trigger attempt is marked as unsuccessful.</td>
         </tr>
         <tr>
+            <td><h5>kubernetes.operator.session.block-on-session-jobs</h5></td>
+            <td style="word-wrap: break-word;">true</td>
+            <td>Boolean</td>
+            <td>Block FlinkDeployment deletion if managed jobs are running in the session cluster.</td>
+        </tr>
+        <tr>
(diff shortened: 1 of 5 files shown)


docs/layouts/shortcodes/generated/dynamic_section.html
```diff
@@ -194,6 +194,12 @@
             <td>Duration</td>
             <td>The interval before a savepoint trigger attempt is marked as unsuccessful.</td>
         </tr>
+        <tr>
+            <td><h5>kubernetes.operator.session.block-on-session-jobs</h5></td>
```
gaborgsomogyi: Purely from the config I can't really tell what it would do. Maybe `kubernetes.operator.session.deletion.block-on-jobs`?

gaborgsomogyi:
This failed, maybe flaky?
```
Error:    RestApiMetricsCollectorTest.testJmMetricCollection:187->lambda$testJmMetricCollection$4:189->assertMetricValueIs:319 
```

docs/layouts/shortcodes/generated/dynamic_section.html
gyfora: done

gaborgsomogyi (approved):

PR #1087, opened 2026-04-14
[hotfix] Fix docker build actions

gyfora:

The docker build pipelines are currently broken due to the use of actions that are no longer supported in the apache CI workflows.

## Brief change log

This PR replaces the existing actions with direct CLI commands for better support / transparency without changing the actual behaviour

### DIFF
--- a/.github/workflows/docker_push.yml
+++ b/.github/workflows/docker_push.yml
@@ -29,6 +29,7 @@ on:
     branches:
       - main
       - 'release-*'
+
 jobs:
   build_image:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


gyfora:
cc @SamBarker @csviri

.github/workflows/docker_push.yml
```diff
@@ -40,38 +41,80 @@ jobs:
...
+          if [[ "${GITHUB_REF}" == "refs/heads/main" ]]; then
+            TAGS+=("\"${IMAGE}:main\"")
+          fi
+
+          # Note: the original type=semver patterns are omitted here.
+          # The workflow triggers on 'release-*-rc*' tags, which are not valid
+          # semver (docker/metadata-action requires vX.Y.Z or X.Y.Z format), so
+          # the original action produced no semver tags either.
+
+          cat > "${BAKE_FILE}" <<EOF
```
SamBarker: I think this ends up copying the indentation from the `docker_push.yml` into the bake file which might be a bit strange for anyone reading the bake file to debug things. Using tabs and `<<-` would avoid that but might not be worth the change?

```suggestion
          cat > "${BAKE_FILE}" <<-EOF
```

SamBarker (approved):
LGTM

One possible nit but its cosmetics rather than functional

.github/workflows/docker_push.yml
gyfora: thanks @SamBarker this should now be fixed

gyfora: ah nevermind, we cannot use tabs in yaml. So keeping it as is

PR #1088, opened 2026-04-14
[FLINK-39938] FLIP-586: Composable Parallelism Alignment Modes for Flink Autoscaler

Dennis-Mircea:

The autoscaler currently hard-codes how it aligns a vertex's computed parallelism to the number of key groups or source partitions, using a single boolean adjust mode and an inline two-phase search inside `JobVertexScaler`. This change replaces that with a small set of named, composable alignment modes and an extensible plugin SPI, so the behavior is easy to reason about and operators can plug in their own alignment strategy without forking the autoscaler. Deployments that pinned the deprecated adjust mode keep their exact previous behavior, including the `ScalingLimited` event.

## Brief change log

  - Added a new `org.apache.flink.autoscaler.alignment` package: the `@Experimental` `AlignmentMode` SPI (an `isApplicable` gate plus an `align` method), `AlignmentContext` (a plain immutable inputs object exposing parallelism, partitions, ship strategies, the autoscaler context, evaluated metrics, topology, and a per-mode configuration), `BuiltInAlignmentMode` (`BALANCED` default, `EVENLY_SPREAD`, `OFF`), the deprecated `KeyGroupOrPartitionsAdjustMode` that preserves the old blocking behavior, and `ParallelismAligner` which holds the shared per-direction region search, the terminal fallbacks, mode resolution, and event emission.
  - Reworked `JobVertexScaler` so it no longer hard-codes the source/keyBy gate or the two-phase search. It now computes the clamped target parallelism and delegates to `ParallelismAligner`, which resolves the configured mode and applies it (or keeps the target when the mode does not apply).
  - Added the `scaling.alignment.mode` option (default `BALANCED`), plus the per-mode `scaling.alignment.mode.<name>.class` and `scaling.alignment.mode.<name>.<parameter>` keys. The old `scaling.key-group.partitions.adjust.mode` is deprecated, retained for backward compatibility, and superseded by the new option.
  - Made custom modes discoverable as plugins, mirroring the scaling-executor plugin SPI: `ServiceLoader` in the standalone autoscaler and Flink's `PluginManager` in the operator, selected by name and class.
  - Threaded the discovered modes from startup (`AutoscalerFactory` in the operator, `StandaloneAutoscalerEntrypoint` in standalone) through `ScalingExecutor` into `JobVertexScaler`.
  - Added an AlignmentMode plugin section to `plugins.md` and its Chinese mirror, and regenerated the autoscaler configuration reference.

## Verifying this change

  - `AlignmentModeTest` covers the built-in single-region search per scaling direction, the keep-the-target fallback, the legacy blocking behavior with its `ScalingLimited` event, applicability gating for non source / non keyBy vertices, and mode resolution (built-in by name, custom by class, deprecated key migration, and the default).
  - `JobVertexScalerTest` pins the deprecated legacy mode so every existing tuple reproduces the previous behavior exactly.
  - Discovery is verified in both runtimes: the standalone test loads a custom mode via `ServiceLoader`, and the operator test packages a custom mode into a plugin JAR and loads it via `PluginManager`.
  - The regenerated configuration reference is checked by `ConfigOptionsDocsCompletenessITCase`.

  - Core observer or reconciler logic that is regularly executed: yes (the autoscaler scaling decision path, with the previous behavior preserved for existing configurations)

## Documentation

  - If yes, how is the feature documented? docs and JavaDocs


flink-autoscaler/src/main/java/org/apache/flink/autoscaler/alignment/AlignmentMode.java
```diff
@@ -0,0 +1,122 @@
...
+ * behaviors are provided by {@link BuiltInAlignmentMode}. Custom implementations are discovered as
+ * plugins (via {@code ServiceLoader} in the standalone autoscaler, or Flink's {@code PluginManager}
+ * in the operator) and selected by name through {@code scaling.alignment.mode} plus {@code
+ * scaling.alignment.mode.<name>.class}.
+ *
+ * <p>An implementation may keep the computed target unchanged or adjust it, and decides for itself
+ * whether it applies to a given vertex (see {@link #isApplicable(Context)}).
+ */
+@Experimental
+public interface AlignmentMode {
```
gyfora: Maybe we should name this ParallelismAlignmentMode to be a bit more explicit , alignment itself is probably too general term to be intuitive.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/alignment/AlignmentMode.java
```diff
@@ -0,0 +1,122 @@
...
+    default boolean isApplicable(Context ctx) {
+        return ctx.getNumSourcePartitions() > 0
+                || ctx.getInputShipStrategies().contains(ShipStrategy.HASH);
+    }
+
+    /**
+     * Returns the parallelism to apply for the vertex described by {@code ctx}. Called only when
+     * {@link #isApplicable(Context)} returned {@code true}.
+     */
+    int align(Context ctx);
```
gyfora: Maybe call this: `adjustParallelism / computeParallelism / alignParallelism`

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/alignment/ParallelismAlignmentMode.java
```diff
@@ -0,0 +1,122 @@
...
+        Collection<ShipStrategy> inputShipStrategies;
+
+        /** The full autoscaler context (cluster client, job configuration, etc.). */
+        JobAutoScalerContext<?> jobContext;
+
+        /** The per-vertex evaluated metrics. */
+        Map<ScalingMetric, EvaluatedScalingMetric> evaluatedMetrics;
+
+        /** The job topology. */
+        JobTopology jobTopology;
```
gyfora, line 117: Feels like these should be a single context object, something like EvaluatedJobAutoScalerContext<..>

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
```diff
@@ -375,17 +376,65 @@ private static ConfigOptions.OptionBuilder autoScalerConfig(String key) {
...
                             .withFallbackKeys(
                                     oldOperatorConfigKey(
                                             "scaling.key-group.partitions.adjust.mode"))
                             .withDescription(
-                                    "How to adjust the parallelism of Source vertex or upstream shuffle is keyBy");
+                                    "Deprecated, use scaling.alignment.mode instead. How to adjust "
+                                            + "the parallelism of Source vertex or upstream partitioning is keyBy");
+
+    public static final ConfigOption<String> ALIGNMENT_MODE =
+            autoScalerConfig("scaling.alignment.mode")
```
gyfora: lets call this everywhere `scaling.parallelism-alignment.mode`

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/alignment/ParallelismAlignmentMode.java
```diff
@@ -0,0 +1,122 @@
...
+        int currentParallelism;
+
+        /** The clamped computed target parallelism, before alignment. */
+        int newParallelism;
+
+        /** The number of source partitions, or a non-positive value when not a source. */
+        int numSourcePartitions;
+
+        /** The vertex max parallelism (number of key groups). */
+        int maxParallelism;
```
gyfora, line 99: having these two separately forces every implementation to have an if logic for sources vs other vertices. Should we combine like in the ParallelismAligner `numKeyGroupsOrPartitions` ?

gyfora (changes requested):
Looks nice overall, but. a lot of code and new interfaces so let's take it a step at a time :)

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/alignment/ParallelismAlignmentMode.java
Dennis-Mircea: I looked into combining `numSourcePartitions` and `maxParallelism` into a single value, but I don't think it makes sense because both are needed:

- `isApplicable` distinguishes a partitioned source (`numSourcePartitions > 0`) from a keyBy (`HASH`). A single combined value can't express that, every applicable vertex would just have a positive divisor with no way to tell the two apart.
- The alignment cap is `min(numKeyGroupsOrPartitions, min(maxParallelism, parallelismUpperLimit))`. For a source, `numKeyGroupsOrPartitions = partitions != maxParallelism`, so with only the combined value you'd compute `min(partitions, upperLimit)` and lose the `maxParallelism` cap, letting a source align above its key-group limit (which Flink won't allow). You need both numbers to compute the cap.

The `ParallelismAligner.numKeyGroupsOrPartitions` is a derived quantity used internally where only the divisor matters, it's not a substitute for the two raw inputs the `Context` has to carry.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/config/AutoScalerOptions.java
Dennis-Mircea: Done!

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/alignment/ParallelismAlignmentMode.java
Dennis-Mircea: I'd say this thing is better to be tackled as part of [FLINK-39959](https://issues.apache.org/jira/browse/FLINK-39959).

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/alignment/AlignmentMode.java
Dennis-Mircea: Done!

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/alignment/AlignmentMode.java
Dennis-Mircea: Done!

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/alignment/ParallelismAlignmentMode.java
gyfora: fair

gyfora (approved):

PR #1093, opened 2026-04-19
[FLINK-39509][flink-kubernetes-webhook] Fix mutating webhook producing spurious object modifications on no-op patches

Dennis-Mircea:

The mutating webhook's `FlinkMutator` always performs a Jackson `convertValue` serialization round-trip (`HasMetadata` -> typed class -> back) on every CREATE/UPDATE admission request. When no `FlinkResourceMutator` actually modifies the resource, the round-tripped object can serialize differently from the original input (e.g., field ordering, null handling), causing the webhook to return a JSON patch to the API server even though nothing was logically changed.

This leads to inconsistent behavior where `kubectl patch` reports `patched` instead of `patched (no change)` for resources like `FlinkDeployment`, even when the patch contains no effective change (e.g., setting `restartNonce` to its current value). While this does not trigger an actual reconciliation (the operator correctly detects no spec diff), it is confusing and causes unnecessary `resourceVersion` bumps on the Kubernetes object.

## Brief change log

  - For each `mutate*` method in `FlinkMutator`, capture a snapshot of the typed object **before** running mutators via `mapper.valueToTree()`, then compare it against the state **after** mutators run
  - If nothing changed, return the **original** `HasMetadata` resource, bypassing the serialization round-trip entirely
  - Applied consistently to all three resource types: `FlinkDeployment`, `FlinkSessionJob`, `FlinkStateSnapshot`

## Verifying this change

Manually verified the change by:
  - Deploying the operator with the fix to a minikube cluster
  - Patching a `FlinkDeployment` with an identical `restartNonce` value and confirming `kubectl` now correctly reports `patched (no change)`
  - Patching a `FlinkSessionJob` with an identical `restartNonce` value and confirming the same `patched (no change)` output
  - Verifying that patches with actual changes (new `restartNonce` values) still correctly report `patched` and trigger reconciliation

Concretely, this change can be manually verified the change by deploying the `basic-session-deployment-and-job.yaml` example and patching with an already-applied `restartNonce` value:

**Before (without fix):**
```
$ kubectl patch flinkdeployment basic-session-deployment-example --type=merge -p '{"spec":{"restartNonce": 1}}'
flinkdeployment.flink.apache.org/basic-session-deployment-example patched
```

**After (with fix):**
```
$ kubectl patch flinkdeployment basic-session-deployment-example --type=merge -p '{"spec":{"restartNonce": 1}}'
flinkdeployment.flink.apache.org/basic-session-deployment-example patched (no change)
```

## Documentation

### DIFF
--- a/flink-kubernetes-webhook/src/main/java/org/apache/flink/kubernetes/operator/admission/mutator/FlinkMutator.java
+++ b/flink-kubernetes-webhook/src/main/java/org/apache/flink/kubernetes/operator/admission/mutator/FlinkMutator.java
@@ -66,9 +66,10 @@ public class FlinkMutator implements Mutator<HasMetadata> {
         return resource;
     }
 
-    private FlinkSessionJob mutateSessionJob(HasMetadata resource) {
+    private HasMetadata mutateSessionJob(HasMetadata resource) {
         try {
             var sessionJob = mapper.convertValue(resource, FlinkSessionJob.class);
+            var originalSessionJob = mapper.valueToTree(sessionJob);
             var namespace = sessionJob.getMetadata().getNamespace();
             var deploymentName = sessionJob.getSpec().getDeploymentName();
             var key = Cache.namespaceKeyFunc(namespace, deploymentName);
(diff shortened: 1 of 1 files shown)


gyfora:
can we please create a jira ticket and test for this?

Dennis-Mircea:
> can we please create a jira ticket and test for this?

Sure, I created [FLINK-39509](https://issues.apache.org/jira/browse/FLINK-39509). About the tests, these are covered as part of the https://github.com/apache/flink-kubernetes-operator/pull/1094 PR and [FLINK-39508](https://issues.apache.org/jira/browse/FLINK-39508) JIRA. I suggest to proceed with the current PR as-is first.

PR #1101, opened 2026-04-23
[FLINK-37766] FlinkSessionJob deletion blocked by finalizer

prshnt:
### What is the purpose of the change
Fix the operator getting stuck in a CANCELLING loop when a job has already reached a terminal state (e.g. FAILED, FINISHED) before the cancel request is processed. cancelJobOrError now returns a boolean indicating whether the cancel is actually pending (true) or the job was already terminated (false), allowing the reconciler to skip the async re-observe wait and proceed directly to cleanup.
Encountered the issue when running Flink jobs on our cluster. Upon searching, found the upstream tracking issue: https://issues.apache.org/jira/browse/FLINK-37766

### Brief change log
• Changed cancelJobOrError return type from void to boolean to distinguish between "cancel submitted, wait for async completion" vs "job already gone, proceed immediately"

• Extended isJobTerminated to also match Flink's "already reached another terminal state" error message (e.g. HTTP 400 BAD_REQUEST with that text), in addition to the existing HTTP CONFLICT check

• When the job is already missing or terminated during a STATELESS or CANCEL suspend, the operator no longer returns CancelResult.pending() — it falls through to CancelResult.completed()

### Verifying this change
This change adds tests and can be verified as follows:

• cancelErrorHandlingWithTerminalStateMessage: simulates a REST client returning HTTP 400 BAD_REQUEST with the "already reached another terminal state" message during a STATELESS cancel of a session job, and asserts that the job status transitions to FINISHED with the job ID cleared (rather than remaining stuck in CANCELLING)

• cancelSessionJobWithCancelModeAndTerminalStateMessage: same scenario but with CANCEL upgrade mode — asserts result.isPending() is false, reproducing the case where the finalizer was never removed, leaving the CR stuck in Terminating

• cancelApplicationJobWithCancelModeAndTerminalStateMessage: same scenario for application mode jobs with CANCEL upgrade mode

• Updated cancelSessionJobTest: asserts the job reaches FINISHED state (not CANCELLING) and the job ID is cleared when the job is already gone during a STATELESS cancel

### Does this pull request potentially affect one of the following parts:
• Dependencies: no
• Public API / CRD changes: no
• Core observer or reconciler logic that is regularly executed: yes — the cancel/suspend path in AbstractFlinkService

### Documentation
• Does this pull request introduce a new feature? no (bug fix)

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
@@ -401,7 +401,7 @@ public abstract class AbstractFlinkService implements FlinkService {
      * @return {@code true} if the job was already missing or terminated, and no further action is
      *     needed. {@code false} if cancellation was successfully initiated and is still pending
      *     (the caller should await completion before proceeding).
-     * @throws UpgradeFailureException if the job cannot be canceled due to an unexpected error, or
+     * @throws UpgradeFailureException if the job cannot be cancelled due to an unexpected error, or
      *     if the job is missing and {@code ignoreMissing} is set to {@code false}.
      */
     public boolean cancelJobOrError(
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
@@ -345,9 +345,11 @@ public abstract class AbstractFlinkService implements FlinkService {
                     }
                     break;
                 case CANCEL:
-                    cancelJobOrError(clusterClient, status, false);
-                    // This is async we need to return
-                    return CancelResult.pending();
+                    if (cancelJobOrError(clusterClient, status, false)) {
+                        // This is async we need to return
+                        return CancelResult.pending();
+                    }
+                    break;
             }
         }
         if (suspendMode.deleteCluster() || deleteCluster) {
@@ -370,9 +372,12 @@ public abstract class AbstractFlinkService implements FlinkService {
             switch (suspendMode) {
                 case STATELESS:
                 case CANCEL:
-                    cancelJobOrError(clusterClient, status, suspendMode == SuspendMode.STATELESS);
-                    // This is async we need to return and re-observe
-                    return CancelResult.pending();
+                    if (cancelJobOrError(
+                            clusterClient, status, suspendMode == SuspendMode.STATELESS)) {
+                        // This is async we need to return and re-observe
+                        return CancelResult.pending();
+                    }
+                    break;
                 case SAVEPOINT:
                     savepointPath = savepointJobOrError(clusterClient, status, conf);
                     break;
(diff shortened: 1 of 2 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
```diff
@@ -370,9 +370,12 @@ public CancelResult cancelSessionJob(
             switch (suspendMode) {
                 case STATELESS:
                 case CANCEL:
-                    cancelJobOrError(clusterClient, status, suspendMode == SuspendMode.STATELESS);
-                    // This is async we need to return and re-observe
-                    return CancelResult.pending();
+                    if (cancelJobOrError(
```
Dennis-Mircea: One concern on the scope: this PR only hardens the `FlinkSessionJob` cancel path (`AbstractFlinkService#cancelSessionJob` / the session-job finalizer), but `AbstractFlinkService#cancelJob`, used for the `FlinkDeployment` (application cluster) path, is left untouched and can hit the exact same failure mode described in [FLINK-37766](https://issues.apache.org/jira/browse/FLINK-37766):

Could we:
1. Extend the new "already terminal / not found" handling to `AbstractFlinkService#cancelJob`?
2. Add a unit test mirroring the new session-job test, but against `cancelJob`, with a mocked JM response returning the terminal-state error (and ideally a 404), asserting the finalizer completes cleanly.
3. Update the PR description (and ideally the JIRA) to make clear the fix covers both CR types?

prshnt: Good suggestion but I cannot modify the issue since I am not the one that created it. @gyfora  what would be your suggestion here? What is the general guideline to proceed in this case of scenario?

gyfora: I think this is a very nice improvement, we can and should of course fix the FlinkDeployment path as well. I can update the jira

prshnt: @gyfora Thank you for taking care of updating the JIRA. Then I can update this and can do it in this PR then.

gyfora: don't worry about the JIRA for now, I will update it before merging :)

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
```diff
@@ -383,14 +388,14 @@ public CancelResult cancelSessionJob(
         return CancelResult.completed(savepointPath);
     }
 
-    public void cancelJobOrError(
+    public boolean cancelJobOrError(
```
gyfora, line 407: Can we please add some javadocs here explaining the return param? Also I find it a bit strange to return true if the cancellation is pending and false if it is already done. Maybe we could swap the boolean values, I leave this up to you, as long as it's documented in the javadoc

prshnt: Yep, looking at it again does make sense I will swap the values and also have the javadoc ready.

prshnt: Done.

gyfora (approved):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
```diff
@@ -345,9 +345,11 @@ protected CancelResult cancelJob(
                     }
                     break;
                 case CANCEL:
-                    cancelJobOrError(clusterClient, status, false);
-                    // This is async we need to return
-                    return CancelResult.pending();
+                    if (!cancelJobOrError(clusterClient, status, false)) {
```
Dennis-Mircea, line 348: Don't we need here to have the same check as for session jobs: `suspendMode == SuspendMode.STATELESS`?

PR #1104, opened 2026-04-28
[hotfix] Validate jarURI in DefaultValidator

oscerd:
## What

Validate the `jarURI` field on **FlinkSessionJob** submission: reject malformed URIs, schemes outside an allowlist, and `http`/`https` URIs whose host resolves to loopback, link-local, site-local, wildcard or multicast addresses. FlinkDeployment is intentionally not validated, since application clusters reference a JAR shipped inside the image (e.g. `local://`) and the operator never fetches it.

## Config options

Both options are operator-level (`SECTION_SYSTEM`), resolved via `FlinkOperatorConfiguration` so a user-supplied `flinkConfiguration` in a CR cannot override them.

- `kubernetes.operator.user.artifacts.allowed-schemes` (`List<String>`, default `https`)
- `kubernetes.operator.user.artifacts.disallow-restricted-hosts` (`Boolean`, default `true`)

## Tests

- `mvn -pl flink-kubernetes-operator test` → 2133 / 0
- `mvn -pl flink-kubernetes-webhook test` → 105 / 0
- `mvn clean install -DskipTests` (full reactor) → BUILD SUCCESS

`DefaultValidatorTest` covers:
- `testJarUriSchemeValidation` and `testJarUriHostValidation` — direct unit tests on the static helper (allowed/disallowed schemes, malformed URIs, missing schemes, loopback/link-local/site-local/wildcard/multicast hosts, opt-out flag).
- `testSessionJobJarUriValidationUsesOperatorConfig` — verifies the validator reads the allowlist from the operator-level configuration and that a CR-supplied override of those keys is ignored.

### DIFF
--- a/docs/layouts/shortcodes/generated/dynamic_section.html
+++ b/docs/layouts/shortcodes/generated/dynamic_section.html
@@ -213,18 +213,6 @@
             <td>Create new FlinkStateSnapshot resources for storing snapshots. Disable if you wish to use the deprecated mode and save snapshot results to FlinkDeployment/FlinkSessionJob status fields. The Operator will fallback to legacy mode during runtime if the CRD is not found, even if this value is true.</td>
         </tr>
         <tr>
-            <td><h5>kubernetes.operator.user.artifacts.allowed-schemes</h5></td>
-            <td style="word-wrap: break-word;">"https";<wbr>"local"</td>
-            <td>List&lt;String&gt;</td>
-            <td>Comma separated list of URI schemes that are allowed for the job's jarURI. By default only 'https' and 'local' are allowed to prevent SSRF and local file disclosure via user-supplied URIs (e.g. 'http', 'file', 's3', 'hdfs', 'gs'). The 'local' scheme is preserved for application clusters that ship the JAR inside the image and is never fetched by the operator. Operators that need to fetch artifacts via other schemes (such as 's3' or 'hdfs') can extend this list. Scheme matching is case-insensitive.</td>
-        </tr>
-        <tr>
-            <td><h5>kubernetes.operator.user.artifacts.disallow-restricted-hosts</h5></td>
-            <td style="word-wrap: break-word;">true</td>
-            <td>Boolean</td>
-            <td>If enabled, jarURI hosts that resolve to loopback, link-local, site-local or wildcard addresses are rejected during validation. This prevents the operator from being coerced into issuing requests against cloud metadata services (e.g. 169.254.169.254) or other in-cluster services. Disable only if the operator legitimately needs to fetch from such addresses.</td>
-        </tr>
-        <tr>
             <td><h5>kubernetes.operator.user.artifacts.http.header</h5></td>
             <td style="word-wrap: break-word;">(none)</td>
             <td>Map</td>
--- a/docs/layouts/shortcodes/generated/dynamic_section.html
+++ b/docs/layouts/shortcodes/generated/dynamic_section.html
@@ -213,6 +213,18 @@
             <td>Create new FlinkStateSnapshot resources for storing snapshots. Disable if you wish to use the deprecated mode and save snapshot results to FlinkDeployment/FlinkSessionJob status fields. The Operator will fallback to legacy mode during runtime if the CRD is not found, even if this value is true.</td>
         </tr>
         <tr>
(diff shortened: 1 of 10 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultValidator.java
```diff
@@ -240,6 +247,11 @@ private Optional<String> validateJobSpec(
 
         Configuration configuration = Configuration.fromMap(confMap);
 
+        Optional<String> jarUriError = validateJarURI(job.getJarURI(), configuration);
```
gyfora: I don't really think that we should validate Application jarURIs here. I think session job validation makes sense because of the http download and local access on the operator pod but for applications these don't really matter.

docs/layouts/shortcodes/generated/dynamic_section.html
```diff
@@ -212,6 +212,18 @@
             <td>Boolean</td>
             <td>Create new FlinkStateSnapshot resources for storing snapshots. Disable if you wish to use the deprecated mode and save snapshot results to FlinkDeployment/FlinkSessionJob status fields. The Operator will fallback to legacy mode during runtime if the CRD is not found, even if this value is true.</td>
         </tr>
+        <tr>
+            <td><h5>kubernetes.operator.user.artifacts.allowed-schemes</h5></td>
+            <td style="word-wrap: break-word;">"https";<wbr>"local"</td>
```
gyfora: if we remove application validation then local can be removed as well

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
```diff
@@ -332,6 +333,34 @@ public static String operatorConfigKey(String key) {
...
+                            "Comma separated list of URI schemes that are allowed for the job's jarURI. "
+                                    + "By default only 'https' and 'local' are allowed to prevent SSRF and "
+                                    + "local file disclosure via user-supplied URIs (e.g. 'http', 'file', "
+                                    + "'s3', 'hdfs', 'gs'). The 'local' scheme is preserved for application "
+                                    + "clusters that ship the JAR inside the image and is never fetched by "
+                                    + "the operator. Operators that need to fetch artifacts via other "
+                                    + "schemes (such as 's3' or 'hdfs') can extend this list. "
+                                    + "Scheme matching is case-insensitive.");
+
+    @Documentation.Section(SECTION_DYNAMIC)
```
gyfora: These configs should go into SECTION_SYSTEM and be resolved in FlinkOperatorConfiguration otherwise the user would be able to override them from their CR's config

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultValidator.java
oscerd: Done in c38f8bd. Moved the call out of `validateJobSpec` so application clusters are no longer touched, and added it to `validateSessionJobOnly` which runs on every session job submission.

docs/layouts/shortcodes/generated/dynamic_section.html
oscerd: Done. `local` is gone from the default allowlist (default is now `https` only), and the option moved to the system section.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
oscerd: Good catch, done. Both options are now `SECTION_SYSTEM`, resolved via `FlinkOperatorConfiguration.fromConfiguration` (new `jarUriAllowedSchemes` / `jarUriDisallowRestrictedHosts` fields), and read by the validator from `configManager.getOperatorConfiguration()`. A new test in `DefaultValidatorTest#testSessionJobJarUriValidationUsesOperatorConfig` confirms a CR-supplied override of these keys is ignored.

gyfora (approved):

PR #1111, opened 2026-05-10
[FLINK-39646][examples] Update python example to Flink 2.2

Savonitar:

Bring `examples/flink-python-example/` in line with the rest of the examples directory, which targets Flink 2.2.

## Brief change log

  - **Dockerfile**: bump base image `flink:1.16` → `flink:2.2`
  - **python-example.yaml**: `flinkVersion: v1_20` → `v2_2` (matches every other example YAML).
  - **README.md**: refresh the PyFlink/Python compatibility note

## Verifying this change

This change has been verified manually on a local minikube cluster.
With Flink Kubernetes Operator 1.14.0 + cert-manager installed, `kubectl apply -f python-example.yaml` (with `imagePullPolicy: Never` for the local image) brings the FlinkDeployment to `STABLE / READY / RUNNING`. The JM logs confirm the operator invokes `PythonDriver` with the new `-pyclientexec /usr/bin/python3` and `pipeline.jars=local:///opt/flink/opt/flink-python-2.2.0.jar`. The TaskManager's `print` sink emits records from the `datagen` source continuously, with no restarts over a multi-hour run.

This change is **a documentation/example-only change** it does not modify any production code paths.

  - Dependencies (does it add or upgrade a dependency): (no, changes inside an example image only)

## Documentation

### DIFF
--- a/examples/flink-python-example/README.md
+++ b/examples/flink-python-example/README.md
@@ -46,11 +46,11 @@ user-defined arguments should be placed in the end. Check the [doc](https://nigh
 
 A working example would be:
 ```yaml
-args: ["-pyfs", "/opt/flink/usrlib/pythonjob/python_demo.py", "-pyclientexec", "/usr/bin/python3", "-py", "/opt/flink/usrlib/pythonjob/python_demo.py", "-myarg", "123"]
+args: ["-pyfs", "/opt/flink/usrlib/python_demo.py", "-pyclientexec", "/usr/bin/python3", "-py", "/opt/flink/usrlib/python_demo.py", "-myarg", "123"]
 ```
 But the following will throw exception:
 ```yaml
-args: ["-myarg", "123", "-pyfs", "/opt/flink/usrlib/pythonjob/python_demo.py", "-pyclientexec", "/usr/bin/python3", "-py", "/opt/flink/usrlib/pythonjob/python_demo.py"]
+args: ["-myarg", "123", "-pyfs", "/opt/flink/usrlib/python_demo.py", "-pyclientexec", "/usr/bin/python3", "-py", "/opt/flink/usrlib/python_demo.py"]
 ```
 
 ## Usage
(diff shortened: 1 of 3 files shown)


examples/flink-python-example/README.md
```diff
@@ -63,7 +63,10 @@ Dockerfile
 
 **Step 2**: Build docker image
 
-Check this [doc](https://nightlies.apache.org/flink/flink-docs-master/docs/deployment/resource-providers/standalone/docker/#using-flink-python-on-docker) for more details about building Pyflink image. Note, Pyflink 1.15.3 is only supported on x86 arch.  
+Check this [doc](https://nightlies.apache.org/flink/flink-docs-master/docs/deployment/resource-providers/standalone/docker/#using-flink-python-on-docker) for more details about building a PyFlink image. The included Dockerfile is based on `flink:2.2` and installs PyFlink 2.2.0 (which supports Python 3.9, 3.10, 3.11 and 3.12).
+
+Note: PyFlink's `pemja` dependency only ships pre-built Linux wheels for `x86_64`. On other architectures (e.g. Apple Silicon) the build falls back to compiling from source, which fails because `flink:2.2` ships a JRE-only image. Build for `linux/amd64` explicitly in that case (`docker build --platform linux/amd64 ...`).
```
Dennis-Mircea: Nice call out about `pemja` and `--platform linux/amd64`, but the code block right below still shows the plain `docker build` command without any concrete variant for non-x86 arch. I'd suggest phrasing it this way:

```suggestion
Note: PyFlink's `pemja` dependency only ships pre-built Linux wheels for `x86_64`. On other architectures (e.g. Apple Silicon) the build falls back to compiling from source, which fails because `flink:2.2` ships a JRE-only image, so on those hosts you must pass `--platform linux/amd64` to `docker build`.

```bash
# Uncomment when building for local minikube env:
# eval $(minikube docker-env)

# Uncomment on Apple Silicon / non-x86 hosts (pemja has no aarch64 wheel):
# DOCKER_BUILDKIT=1 docker build --platform linux/amd64 . -t flink-python-example:latest

# Comment out on Apple Silicon / non-x86 hosts:
DOCKER_BUILDKIT=1 docker build . -t flink-python-example:latest
```
```

File: examples/flink-python-example/README.md [hunk 2]
```diff
@@ -46,11 +46,11 @@ user-defined arguments should be placed in the end. Check the [doc](https://nigh
 
 A working example would be:
 ```yaml
-args: ["-pyfs", "/opt/flink/usrlib/pythonjob/python_demo.py", "-pyclientexec", "/usr/local/bin/python3", "-py", "/opt/flink/usrlib/pythonjob/python_demo.py", "-myarg", "123"]
+args: ["-pyfs", "/opt/flink/usrlib/pythonjob/python_demo.py", "-pyclientexec", "/usr/bin/python3", "-py", "/opt/flink/usrlib/pythonjob/python_demo.py", "-myarg", "123"]
```
Dennis-Mircea: While we're refreshing this example, can we also fix a pre‑existing inconsistency here? These two snippets reference `/opt/flink/usrlib/pythonjob/python_demo.py` (with a `pythonjob/` subdir), but both the Dockerfile (`ADD python_demo.py /opt/flink/usrlib/python_demo.py`) and `python-example.yaml` (`-py /opt/flink/usrlib/python_demo.py`) place the script directly under `/opt/flink/usrlib/`. Users following the README verbatim will get a `FileNotFoundException`. Easiest fix is to drop the `pythonjob/` segment from both args examples so all three files agree.

Savonitar: Good catch, thank you. fixed in follow up commit, both -pyfs and -py snippets now reference `/opt/flink/usrlib/python_demo.py`

examples/flink-python-example/README.md
Savonitar: Thanks for catching this. 
However, I took a slightly different approach: instead of the comment/uncomment toggle, I made `--platform linux/amd64` the default in the example:
1. It's a no-op on x86
2. on Apple Silicon Docker Desktop emulates amd64 transparently (verified end-to-end on minikube). 
The result is a single command that can be copy-pasted and that works on both archs. Updated the description. 

I'm happy to switch to your toggle suggestion if you'd prefer it. WDYT?

Dennis-Mircea (approved):

1996fanrui (approved):
Thanks for bumping flink example version, LGTM

Savonitar:
Thanks for the reviews and merge.

PR #1112, opened 2026-05-15
[FLINK-39688] Share a single PluginManager across operator and webhook startup

Dennis-Mircea:

The operator currently constructs a brand-new `PluginManager` four times during startup (one per consumer: metric reporters, validators, listeners, file systems), and the webhook constructs it twice more (validators, mutators). Every call rescans `/opt/flink/plugins`, builds a fresh isolated child `URLClassLoader` per plugin directory, and floods the startup logs with a duplicated `"Plugin loader with ID not found, creating it: <subdir>"` batch per call. This change consolidates discovery onto a single shared `PluginManager` per JVM, which removes the duplicate logs, drops redundant classloader allocations, and eliminates a latent classloader-identity hazard if a plugin class were ever loaded through two different SPIs.

## Brief change log

  - Added `OperatorPluginUtils` (in `flink-kubernetes-operator/utils/`) which owns the operator-wide plugin classloader policy. It exposes `enrichWithPluginParentFirstPatterns(Configuration)` and `createPluginManager(Configuration)`, and centralizes the `io.fabric8` / `com.fasterxml` parent-first patterns that previously lived in `ListenerUtils` only.
  - Added a 2-arg `(FlinkConfigManager, PluginManager)` overload to `ValidatorUtils.discoverValidators`, `ListenerUtils.discoverListeners`, `MutatorUtils.discoverMutators`, and a `(Configuration, PluginManager)` overload to `OperatorMetricUtils.initOperatorMetrics`. The existing single-arg overloads delegate to the new ones (creating a manager via `OperatorPluginUtils`) and are now annotated `@VisibleForTesting`, since after the refactor they have no production callers and exist only to keep test fixtures simple.
  - `FlinkOperator` constructor now creates one shared `PluginManager` via `OperatorPluginUtils.createPluginManager(baseConfig)` and threads it into all four consumers (`initOperatorMetrics`, `discoverValidators`, `discoverListeners`, `FileSystem.initialize`).
  - `FlinkOperatorWebhook` constructor mirrors the same pattern for `discoverValidators` and `discoverMutators`.
  - `ListenerUtils` no longer carries the `EXTRA_PARENT_FIRST_PATTERNS` constant or the `getListenerBaseConf` helper, the parent-first policy now applies process-wide via `OperatorPluginUtils`. Promoting it globally is intentional and safe: every plugin type (metric reporters, FS factories, validators, mutators, listeners) benefits from sharing the operator's fabric8 client and Jackson, avoiding duplicate informers and version skew.
  - Added a one-line INFO header before each `pluginManager.load(...)` call so the Flink-internal `"Plugin loader with ID found, reusing it: <subdir>"` batches are clearly attributed to a specific SPI: `"Loading FlinkResourceValidator implementations from plugin directory."` (`ValidatorUtils`), `"Loading FlinkResourceListener ..."` (`ListenerUtils`), `"Loading FlinkResourceMutator ..."` (`MutatorUtils`), and `"Initializing file system factories from plugin directory."` (`FlinkOperator`, before `FileSystem.initialize`). The metrics path already prints `"Initializing operator metrics using conf: ..."` and needs no extra header.

## Verifying this change

This change is already covered by existing tests, such as `ListenerUtilsTest`, `ValidatorUtilsTest`, `MutatorUtilsTest`, `OperatorMetricUtilsTest`, `FlinkOperatorTest`, `AdmissionHandlerTest`, and `FlinkStateSnapshotControllerTest` (the latter exercises the preserved single-arg overload). All pass locally. Runtime behavior was verified from the test logs: `FlinkOperatorTest` now emits a single `Initializing operator metrics ...` line followed directly by `FileSystem` initialization, with no repeated plugin scanning batches.

End-to-end verification on minikube against a real operator image with a populated `/opt/flink/plugins` confirms the same: the operator now logs exactly one `"Plugin loader with ID not found, creating it: <subdir>"` batch at startup (was 4 before this PR). The remaining `"Plugin loader with ID found, reusing it: <subdir>"` lines that still appear are emitted from inside Flink's `DefaultPluginManager.load(SomeSPI.class)` itself, once per registered loader per call, and are independent of how many `PluginManager` instances exist. They are out of scope for this PR (would require an upstream Flink log-level change), and operators that find them noisy can already silence them today via log4j:

```properties
logger.pluginmanager.name = org.apache.flink.core.plugin.DefaultPluginManager
logger.pluginmanager.level = WARN
```

## Documentation

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/OperatorPluginUtils.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/OperatorPluginUtils.java
@@ -42,7 +42,7 @@ public final class OperatorPluginUtils {
     private OperatorPluginUtils() {}
 
     /** Returns a copy of {@code baseConf} with the operator's parent-first patterns merged in. */
-    public static Configuration enrichWithPluginParentFirstPatterns(Configuration baseConf) {
+    private static Configuration enrichWithPluginParentFirstPatterns(Configuration baseConf) {
         var conf = new Configuration(baseConf);
         var patterns =
                 new ArrayList<>(
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/OperatorPluginUtils.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/OperatorPluginUtils.java
@@ -1,0 +1,67 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.kubernetes.operator.utils;
+
+import org.apache.flink.configuration.Configuration;
+import org.apache.flink.configuration.CoreOptions;
+import org.apache.flink.core.plugin.PluginManager;
+import org.apache.flink.core.plugin.PluginUtils;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Builds the single, process-wide {@link PluginManager} used by the operator and webhook for plugin
+ * discovery (metric reporters, file systems, validators, mutators, listeners).
+ *
+ * <p>Plugin classloaders delegate {@code io.fabric8} and {@code com.fasterxml} to the parent so all
+ * plugins share the operator's fabric8 client and Jackson, avoiding duplicate informers and version
+ * skew.
+ */
+public final class OperatorPluginUtils {
+
+    private static final List<String> EXTRA_PARENT_FIRST_PATTERNS =
+            List.of("io.fabric8", "com.fasterxml");
+
+    private OperatorPluginUtils() {}
+
+    /** Returns a copy of {@code baseConf} with the operator's parent-first patterns merged in. */
+    public static Configuration enrichWithPluginParentFirstPatterns(Configuration baseConf) {
+        var conf = new Configuration(baseConf);
+        var patterns =
+                new ArrayList<>(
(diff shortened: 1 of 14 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/OperatorPluginUtils.java
```diff
@@ -0,0 +1,67 @@
...
+ */
+public final class OperatorPluginUtils {
+
+    private static final List<String> EXTRA_PARENT_FIRST_PATTERNS =
+            List.of("io.fabric8", "com.fasterxml");
+
+    private OperatorPluginUtils() {}
+
+    /** Returns a copy of {@code baseConf} with the operator's parent-first patterns merged in. */
+    public static Configuration enrichWithPluginParentFirstPatterns(Configuration baseConf) {
```
spuru9:
```suggestion
    private static Configuration enrichWithPluginParentFirstPatterns(Configuration baseConf) {
```
Dont need it to be public as createPluginManager is the only caller.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/ValidatorUtils.java
```diff
@@ -36,13 +37,22 @@ public final class ValidatorUtils {
 
     private static final Logger LOG = LoggerFactory.getLogger(FlinkUtils.class);
 
+    @VisibleForTesting
     public static Set<FlinkResourceValidator> discoverValidators(FlinkConfigManager configManager) {
```
spuru9: This overload isn't called from any non-test code now. `@VisibleForTesting` keeps internal callers honest, but for anyone downstream (forks, vendored builds) the method is still public API on a published utility — `@Deprecated(forRemoval = false)` with a Javadoc pointer to the 2-arg version would actually signal intent to them. The three test sites (FlinkStateSnapshotControllerTest, Testing*Controller) could then migrate to the 2-arg form over time. WDYT?

Dennis-Mircea: `@Deprecated(forRemoval = false)` is the weaker form (discouraged, not scheduled for removal), so it actually contradicts the "for future removal" framing. `@VisibleForTesting` is already the canonical Flink signal for "internal, test-only", and `ValidatorUtils` is not part of any public API surface. I went a step further and deleted the single-arg overload outright, migrating the test call sites to the two-arg variant. Applied the same cleanup to `MutatorUtils` and `ListenerUtils` for consistency.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/OperatorPluginUtils.java
Dennis-Mircea: Fixed it!

spuru9 (approved):
LGTM

PR #1115, opened 2026-05-21
[FLINK-39691] Fix toString for AggregatedMetric

PavelZeger:
## What is the purpose of the change

This pull request fixes a typo in the AggregatedMetric#toString() implementation where the min field name was incorrectly rendered
as mim. The fix is applied consistently across the two copies of AggregatedMetric (in flink-autoscaler-standalone and
flink-kubernetes-operator) so their string representations remain identical.

## Brief change log

- Corrected the field label from mim= to min= in AggregatedMetric#toString() in flink-autoscaler-standalone.
- Applied the same correction to the copy of AggregatedMetric#toString() in flink-kubernetes-operator.

## Verifying this change

This change is a trivial rework / code cleanup without any new test coverage.

The change can additionally be verified as follows:

- `mvn spotless:check` passes for this change locally.
- `mvn clean verify` passes locally on the affected modules.
- Manually grepped the touched modules to confirm only the intended min= label remains:
```bash
grep -rn '"mim=' flink-autoscaler-standalone/src/main/java flink-kubernetes-operator/src/main/java
grep -rn '"min=' flink-autoscaler-standalone/src/main/java flink-kubernetes-operator/src/main/java
```

## Does this pull request potentially affect one of the following parts:

- Dependencies (does it add or upgrade a dependency): no
- The public API, i.e., is any changes to the CustomResourceDescriptors: no
- Core observer or reconciler logic that is regularly executed: no

## Documentation

- Does this pull request introduce a new feature? no
- If yes, how is the feature documented? not applicable

PR #1116, opened 2026-05-21
[FLINK-39692][Kubernetes-Operator] Include caught exceptions in warn/error log statements

PavelZeger:

This pull request fixes a LOG.error call site in StateSnapshotReconciler that swallowed the caught exception by formatting only a message string, without passing the Throwable to SLF4J. As a result the stack trace of savepoint-disposal failures was lost, making it harder to diagnose real operational failures (Flink REST client errors, IO errors, etc.). The change passes the exception as the last argument so SLF4J logs the full stack trace alongside the message.

## Brief change log

- `StateSnapshotReconciler#handleSnapshotCleanup` - pass the caught exception to LOG.error when failing to dispose a savepoint for a deployment, so the full stack trace is emitted.

## Verifying this change

This change is a trivial rework / code cleanup without any new test coverage. It only enriches an existing log statement with the caught exception; no control flow, public API, or behavior is changed.

The change has additionally been verified as follows:

- `mvn spotless:check` passes locally for the provided change.
- `mvn clean verify` passes locally on the affected module.

- The public API, i.e., is any changes to the CustomResourceDescriptors: no
- Core observer or reconciler logic that is regularly executed: no (logging-only change inside a catch branch in
StateSnapshotReconciler)

## Documentation

### DIFF
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/tuning/MemoryTuning.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/tuning/MemoryTuning.java
@@ -88,7 +88,7 @@ public class MemoryTuning {
         try {
             memSpecs = FLINK_MEMORY_UTILS.memoryProcessSpecFromConfig(config);
         } catch (IllegalConfigurationException e) {
-            LOG.warn("Current memory configuration is not valid. Aborting memory tuning.", e);
+            LOG.warn("Current memory configuration is not valid. Aborting memory tuning.");
             return EMPTY_CONFIG;
         }
 
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/tuning/MemoryTuning.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/tuning/MemoryTuning.java
@@ -88,7 +88,7 @@ public class MemoryTuning {
         try {
             memSpecs = FLINK_MEMORY_UTILS.memoryProcessSpecFromConfig(config);
         } catch (IllegalConfigurationException e) {
-            LOG.warn("Current memory configuration is not valid. Aborting memory tuning.");
+            LOG.warn("Current memory configuration is not valid. Aborting memory tuning.", e);
             return EMPTY_CONFIG;
         }
 
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/state/KubernetesAutoScalerStateStore.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/state/KubernetesAutoScalerStateStore.java
@@ -391,7 +391,7 @@ public class KubernetesAutoScalerStateStore
                 return IOUtils.toString(zi, StandardCharsets.UTF_8);
             }
         } catch (Exception e) {
-            LOG.warn("Error while decompressing scaling data, treating as uncompressed");
+            LOG.warn("Error while decompressing scaling data, treating as uncompressed", e);
             // Fall back to non-compressed for migration
             return compressed;
         }
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/state/KubernetesAutoScalerStateStore.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/state/KubernetesAutoScalerStateStore.java
@@ -391,7 +391,7 @@ public class KubernetesAutoScalerStateStore
                 return IOUtils.toString(zi, StandardCharsets.UTF_8);
             }
         } catch (Exception e) {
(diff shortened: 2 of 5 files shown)


Dennis-Mircea (changes requested):
The framing of this PR is wrong: A `LOG.warn(msg)` without the `Throwable` is not automatically a bug, in several of the call sites you are changing it is deliberate. Adding the stack trace blanket-style produces log spam on expected paths and obscures the intent of the original code. Each call site needs to be evaluated on its own merits.

Going site by site:
- **`KubernetesAutoScalerStateStore#decompress` (line 394)** - this is a known fallback path. The comment one line below literally says `// Fall back to non-compressed for migration`. During migration this catch will fire **on every annotation read** for resources stored under the old format. Today it logs one tidy `WARN`, after this PR it logs a full stack trace per annotation, which is exactly the kind of noise the original author was avoiding. Please drop this change.
- **`MemoryTuning#tune` (line 91)** - `IllegalConfigurationException` here is a user-config error. The user needs the message ("memory configuration is not valid") as it is already clear enough, not a parser stack trace from `MemoryProcessSpec`. Adding the throwable here adds 30+ stack frames of Flink internals on every reconcile while the user has invalid memory config.
- **`FlinkConfigManager` (line 265 & 270)** - same category as above, two catches for `NumberFormatException` / `IllegalArgumentException` on **user config keys**. The message already names the offending key. The exception type is the message essentially. A stack trace adds nothing diagnostic for a parse failure on a known string.
- **`IngressUtils#isVersionAtLeast` (line 370)** - catches `IllegalArgumentException` from `ModuleDescriptor.Version.parse`. The message already prints the unparseable `serverVersion` string.
- **`StateSnapshotReconciler` (line 177)** - this one is justified. Savepoint disposal failures are real operational failures where the upstream exception (Flink REST client, IO, etc.) is exactly what an operator needs to debug. This is the only change that I'd keep.

So out of 5 sites, 1 is clearly correct, 1 is clearly wrong (the decompression fallback), and 3 are noise. Please split the savepoint disposal fix into its own small PR with a focused JIRA, and drop the rest unless you can show a concrete debugging scenario each one improves.

The PR description has the same formatting issue as #1115.

PavelZeger:
> The framing of this PR is wrong: A `LOG.warn(msg)` without the `Throwable` is not automatically a bug, in several of the call sites you are changing it is deliberate. Adding the stack trace blanket-style produces log spam on expected paths and obscures the intent of the original code. Each call site needs to be evaluated on its own merits.
> 
> Going site by site:
> 
> * **`KubernetesAutoScalerStateStore#decompress` (line 394)** - this is a known fallback path. The comment one line below literally says `// Fall back to non-compressed for migration`. During migration this catch will fire **on every annotation read** for resources stored under the old format. Today it logs one tidy `WARN`, after this PR it logs a full stack trace per annotation, which is exactly the kind of noise the original author was avoiding. Please drop this change.
> * **`MemoryTuning#tune` (line 91)** - `IllegalConfigurationException` here is a user-config error. The user needs the message ("memory configuration is not valid") as it is already clear enough, not a parser stack trace from `MemoryProcessSpec`. Adding the throwable here adds 30+ stack frames of Flink internals on every reconcile while the user has invalid memory config.
> * **`FlinkConfigManager` (line 265 & 270)** - same category as above, two catches for `NumberFormatException` / `IllegalArgumentException` on **user config keys**. The message already names the offending key. The exception type is the message essentially. A stack trace adds nothing diagnostic for a parse failure on a known string.
> * **`IngressUtils#isVersionAtLeast` (line 370)** - catches `IllegalArgumentException` from `ModuleDescriptor.Version.parse`. The message already prints the unparseable `serverVersion` string.
> * **`StateSnapshotReconciler` (line 177)** - this one is justified. Savepoint disposal failures are real operational failures where the upstream exception (Flink REST client, IO, etc.) is exactly what an operator needs to debug. This is the only change that I'd keep.
> 
> So out of 5 sites, 1 is clearly correct, 1 is clearly wrong (the decompression fallback), and 3 are noise. Please split the savepoint disposal fix into its own small PR with a focused JIRA, and drop the rest unless you can show a concrete debugging scenario each one improves.
> 
> The PR description has the same formatting issue as #1115.

Thanks for the review @Dennis-Mircea .
All required fixes were addressed and the PR format was applied

Dennis-Mircea (changes requested):
Please also update the PR title to reflect the latest changes as well as the JIRA ticket.

Also, speaking about the PR title, `[logging]` is not a valid component tag. If you want to include a component, that will be `[kubernetes-operator]`.

PavelZeger:
> Please also update the PR title to reflect the latest changes as well as the JIRA ticket.
> 
> Also, speaking about the PR title, `[logging]` is not a valid component tag. If you want to include a component, that will be `[kubernetes-operator]`.

I've changed the title and the Jira ticket title as well

Dennis-Mircea (approved):

gyfora:
thanks @Dennis-Mircea , in general this is the kind of PR we should try to avoid as it provides minimal value and burns valuable reviewer time. Looks like something an AI would do :D

PR #1120, opened 2026-05-25
[FLINK-39752] Thread EventRecorder through FlinkResourceContext instead of holding it on the factory

Dennis-Mircea:

`FlinkResourceContextFactory` holds a single operator-scoped `EventRecorder`, which it uses when creating the `FlinkService`. The deployment controller reuses that same shared instance, but `registerSessionJobController` and `registerSnapshotController` each create their own `EventRecorder` via `EventRecorder.create(...)`. As a result, for those two controllers the recorder used by the reconciler and observer is a different instance from the one the `FlinkService` emits through.

This is harmless today because `EventRecorder` is a stateless dispatcher over the shared listeners and client, but it is a latent inconsistency that would become a real bug the moment any per-instance state is added to `EventRecorder`.

This PR removes the per-controller recorder creation so that the session-job and snapshot controllers reuse the single shared `EventRecorder` the factory already holds. Every controller and the `FlinkService` it drives now route through one operator-scoped recorder.

## Brief change log

  - Removed the local `EventRecorder.create(...)` in `FlinkOperator.registerSessionJobController` and `FlinkOperator.registerSnapshotController` so both reuse the shared `eventRecorder` field that is also passed to `FlinkResourceContextFactory`.

## Verifying this change

This change is already covered by existing tests, such as `FlinkDeploymentControllerTest`, `FlinkSessionJobControllerTest`, and `FlinkStateSnapshotControllerTest`. The behavior is unchanged because the shared and per-controller recorders are functionally identical, the change only collapses them to a single instance.

## Documentation

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/FlinkOperator.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/FlinkOperator.java
@@ -207,7 +207,6 @@ public class FlinkOperator {
 
     @VisibleForTesting
     void registerSessionJobController() {
-        var eventRecorder = EventRecorder.create(client, listeners);
         var metricManager =
                 MetricManager.createFlinkSessionJobMetricManager(baseConfig, metricGroup);
         var statusRecorder = StatusRecorder.create(client, metricManager, listeners);
@@ -240,7 +239,6 @@ public class FlinkOperator {
                 MetricManager.createFlinkStateSnapshotMetricManager(baseConfig, metricGroup);
         var statusRecorder =
                 StatusRecorder.createForFlinkStateSnapshot(client, metricManager, listeners);
-        var eventRecorder = EventRecorder.create(client, listeners);
         var reconciler = new StateSnapshotReconciler(ctxFactory, eventRecorder);
         var observer = new StateSnapshotObserver(ctxFactory, eventRecorder);
         var controller =


gyfora (changes requested):
Overall I don't really understand why we are moving the event record from the factory field to the getResource.

The eventRecorder is fixed and should be created once so there is not really a point in passing it always to getResourceContext, this is exactly the point of having the factory to avoid unnecessary object passing again and again.

If we want to ensure that we use a single event recorder then lets use the same in the controller that we pass to the context factory

Dennis-Mircea:
> Overall I don't really understand why we are moving the event record from the factory field to the getResource.
> 
> The eventRecorder is fixed and should be created once so there is not really a point in passing it always to getResourceContext, this is exactly the point of having the factory to avoid unnecessary object passing again and again.
> 
> If we want to ensure that we use a single event recorder then lets use the same in the controller that we pass to the context factory

Fair point on the threading. My intent was end-to-end ownership so each controller and the `FlinkService` it drives share the same recorder, and I leaned per-controller because that was already the more common shape as part of the existing implementation: session-job and snapshot controllers each create their own `EventRecorder` (only deployment reuses the factory's, and blue-green has none), and the `StatusRecorder` is per-controller too, so pairing a per-controller `EventRecorder` alongside it felt consistent.

That said, I agree with your approach. I'll keep the single `EventRecorder` on the factory and pass that same instance to every controller, dropping the `getResourceContext` / `FlinkResourceContext` plumbing. That gives one consistent recorder everywhere without passing it on every call, and cleans up the session-job and snapshot controllers that were creating their own.

Dennis-Mircea:
@gyfora I updated the PR and JIRA with the new context. Now it's a pretty straightforward and simple change.

PR #1121, opened 2026-05-25
[hotfix] Clean up FlinkBlueGreenDeploymentController

Dennis-Mircea:

`FlinkBlueGreenDeploymentController` had accumulated a handful of small but real issues:
- The logger was instantiated against `FlinkDeploymentController.class` (wrong category for every log line emitted by the blue/green path).
- The JavaDoc state machine and orchestration sections were collapsed into single paragraphs that did not render as lists.
- The `InformerEventSource` for `FlinkDeployment` was built inline in `prepareEventSources` instead of being centralized in `EventSourceUtils` like every other controller
- The `BlueGreenContext` was constructed inline in `reconcile` instead of going through `FlinkResourceContextFactory` like every other context type.
- A `logAndThrow(String)` helper existed that did not log and just wrapped `throw new RuntimeException(message)` without adding anything.

This PR fixes all of the above and also cleans up a few warnings in `FlinkResourceContextFactory`.

## Brief change log

  - Fixed the SLF4J logger category in `FlinkBlueGreenDeploymentController` (`FlinkDeploymentController.class` → `FlinkBlueGreenDeploymentController.class`).
  - Rewrote the controller-level JavaDoc to use proper `<h2>` headings and `<ol>` lists for the state machine and orchestration sections, with `{@link}` references to the relevant types.
  - Added `EventSourceUtils.getBlueGreenFlinkDeploymentInformerEventSource(...)` (parallels the existing `getBlueGreenIngressInformerEventSource`) and switched `prepareEventSources` to use it. The inline `InformerEventSourceConfiguration` / `InformerEventSource` / `Mappers` code in the controller is gone.
  - Added `FlinkResourceContextFactory.getBlueGreenContext(...)` (parallels `getFlinkStateSnapshotContext`) and switched both branches of `reconcile` to use it instead of constructing `BlueGreenContext` inline.
  - Removed `FlinkBlueGreenDeploymentController.logAndThrow(String)`. Replaced its three call sites in `FlinkBlueGreenDeployments.fromSecondaryResources` with direct `throw new RuntimeException(...)` statements.
  - In `FlinkResourceContextFactory`: collapsed the `instanceof FlinkDeployment` / `instanceof FlinkSessionJob` checks into pattern-matching `instanceof`, converted the deployment-mode `switch` into a switch expression, removed the dead `jobId` / `flinkDepJobId` / `flinkSessionJobId` locals, tightened the `getResourceContext` generic to `Context<CR>`, and added `@SuppressWarnings("unchecked")` for the down-cast.
  - Pruned the now-unused imports from the controller (`FlinkDeployment`, `InformerEventSourceConfiguration`, `InformerEventSource`, `Mappers`).

## Verifying this change

This change is already covered by existing tests, such as `FlinkBlueGreenDeploymentControllerTest`, `BlueGreenUtilsTest`, `BlueGreenIngressIT`, `FlinkDeploymentControllerTest`, `FlinkSessionJobControllerTest`, `FlinkStateSnapshotControllerTest`, `ApplicationReconcilerTest`, and `SessionJobReconcilerTest`.

  - Core observer or reconciler logic that is regularly executed: yes (control flow in `FlinkBlueGreenDeploymentController#reconcile` and `#prepareEventSources` was restructured. Behavior is unchanged, the call sites just go through `EventSourceUtils` and `FlinkResourceContextFactory.getBlueGreenContext` instead of inline construction.)

## Documentation

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeployments.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeployments.java
@@ -51,7 +51,7 @@ public class FlinkBlueGreenDeployments {
                 context.getSecondaryResources(FlinkDeployment.class);
 
         if (secondaryResources.isEmpty() || secondaryResources.size() > 2) {
-            FlinkBlueGreenDeploymentController.logAndThrow(
+            throw new RuntimeException(
                     "Unexpected number of dependent deployments: " + secondaryResources.size());
         }
 
@@ -63,13 +63,13 @@ public class FlinkBlueGreenDeployments {
 
             if (flinkBlueGreenDeploymentType == BlueGreenDeploymentType.BLUE) {
                 if (flinkBlueGreenDeployments.getFlinkDeploymentBlue() != null) {
-                    FlinkBlueGreenDeploymentController.logAndThrow(
+                    throw new RuntimeException(
                             "Detected multiple Dependent Deployments of type BLUE");
                 }
                 flinkBlueGreenDeployments.setFlinkDeploymentBlue(dependentDeployment);
             } else {
                 if (flinkBlueGreenDeployments.getFlinkDeploymentGreen() != null) {
-                    FlinkBlueGreenDeploymentController.logAndThrow(
+                    throw new RuntimeException(
                             "Detected multiple Dependent Deployments of type GREEN");
                 }
                 flinkBlueGreenDeployments.setFlinkDeploymentGreen(dependentDeployment);
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkResourceContextFactory.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkResourceContextFactory.java
@@ -20,15 +20,19 @@ package org.apache.flink.kubernetes.operator.service;
 import org.apache.flink.annotation.VisibleForTesting;
 import org.apache.flink.api.java.tuple.Tuple2;
 import org.apache.flink.kubernetes.operator.api.AbstractFlinkResource;
+import org.apache.flink.kubernetes.operator.api.FlinkBlueGreenDeployment;
 import org.apache.flink.kubernetes.operator.api.FlinkDeployment;
 import org.apache.flink.kubernetes.operator.api.FlinkSessionJob;
 import org.apache.flink.kubernetes.operator.api.FlinkStateSnapshot;
+import org.apache.flink.kubernetes.operator.api.status.FlinkBlueGreenDeploymentStatus;
 import org.apache.flink.kubernetes.operator.artifact.ArtifactManager;
 import org.apache.flink.kubernetes.operator.config.FlinkConfigManager;
+import org.apache.flink.kubernetes.operator.controller.FlinkBlueGreenDeployments;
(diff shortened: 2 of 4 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkResourceContextFactory.java
```diff
@@ -98,39 +102,40 @@ public FlinkStateSnapshotContext getFlinkStateSnapshotContext(
...
+            Context<FlinkBlueGreenDeployment> josdkContext,
+            FlinkBlueGreenDeployments deployments) {
+        return new BlueGreenContext(
+                bgDeployment, deploymentStatus, josdkContext, deployments, this);
+    }
+
+    @SuppressWarnings("unchecked")
     public <CR extends AbstractFlinkResource<?, ?>> FlinkResourceContext<CR> getResourceContext(
-            CR resource, Context josdkContext) {
+            CR resource, Context<CR> josdkContext) {
```
Savonitar: I like the idea, but will this change break https://github.com/apache/flink-kubernetes-operator/blob/main/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/bluegreen/BlueGreenDeploymentService.java#L289 which uses `Context<FlinkBlueGreenDeployment> josdkContext` and other callers?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeployments.java
```diff
@@ -51,7 +51,7 @@ static FlinkBlueGreenDeployments fromSecondaryResources(
                 context.getSecondaryResources(FlinkDeployment.class);
 
         if (secondaryResources.isEmpty() || secondaryResources.size() > 2) {
-            FlinkBlueGreenDeploymentController.logAndThrow(
+            throw new RuntimeException(
```
Savonitar, line 54: Could you please elaborate, why RuntimeException and not for example IllegalStateException (like in https://github.com/apache/flink-kubernetes-operator/blob/main/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/bluegreen/BlueGreenDeploymentService.java#L715)?

Just curious (not a concern).

Savonitar (commented):
Hi. Thanks for the cleanup ( especially the switch expression, nice compile-time safety improvement). 
I have a few questions below. Could u please take a look?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/FlinkResourceContextFactory.java
Dennis-Mircea: Good catch, you are right, that was a real break, I don't know how I missed that. I changed the type from CR to ?.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeployments.java
Dennis-Mircea: There was not a deep reason. The previous `logAndThrow(String)` helper just did `throw new RuntimeException(message)`, so inlining it kept the exact same exception type.

Savonitar (approved):
Thanks for addressing the feedback.
LGTM!

gyfora (approved):

PR #1123, opened 2026-05-26
[FLINK-39766][flink-autoscaler] Remove redundant AggregatedMetric local copy

Savonitar:

The flink-kubernetes-operator repository carries three identical local copies of https://github.com/apache/flink/blob/master/flink-runtime/src/main/java/org/apache/flink/runtime/rest/messages/job/metrics/AggregatedMetric.java 

It shadows the upstream class at the same fully-qualified name so call sites resolved to the local copy when compiling against pre 1.20 Flink.

## Brief change log

- Delete `flink-autoscaler/src/main/java/org/apache/flink/runtime/rest/messages/job/metrics/AggregatedMetric.java` 
- Delete `flink-kubernetes-operator/src/main/java/org/apache/flink/runtime/rest/messages/job/metrics/AggregatedMetric.java` 
- Delete `flink-autoscaler-standalone/src/main/java/org/apache/flink/runtime/rest/messages/job/metrics/AggregatedMetric.java`

## Verifying this change

This change is already covered by existing tests, e.g. by `ScalingMetricsTest`, `RestApiMetricsCollectorTest`, and `ScalingMetricCollectorTest`.

## Documentation


Dennis-Mircea (changes requested):
The local `AggregatedMetric.toString()` currently emits `min=`, but that's only because of two operator-side fixes:

- #1062 (merged) - fixed the typo in `flink-autoscaler/.../AggregatedMetric.java`.
- #1115 (open) - fixes the same typo in the operator + standalone copies.

The same typo (`mim=`) still ships in every released Flink - 1.20.1 (our current dep), 2.0.0, 2.1.0, 2.2.0, and is also still present on the `release-2.0`, `release-2.1`, `release-2.2`, and `release-2.3` branches. It's been fixed only on `master`, in your own commit [`0b77c11079`](https://github.com/apache/flink/commit/0b77c11079) / [apache/flink#28260](https://github.com/apache/flink/pull/28260), which is unreleased.

So if we merge this PR as-is today, `AggregatedMetric.toString()` falls back to the upstream class in `flink-runtime-1.20.1` and starts emitting `mim=` again in operator logs, undoing #1062 and pre-empting #1115. Our reconciler logs reference this string fairly often, so I'd rather not regress it.

Until this typo is fixed in the release the operator is actually using, this change should be closed and reopened once it can proceed.

Savonitar:
@Dennis-Mircea  good catch, you are right.
However it's worth to mention that two of the three local copies being deleted already have the `mim=` typo themselves, e.g. [`flink-kubernetes-operator/.../AggregatedMetric.java#L130`](https://github.com/apache/flink-kubernetes-operator/blob/main/flink-kubernetes-operator/src/main/java/org/apache/flink/runtime/rest/messages/job/metrics/AggregatedMetric.java#L130)

So removing those two wouldn't regress anything, the regression only matters for the autoscaler module. As for the actual usage: I originally raised this PR because I couldn't find any explicit `toString()` call.

But there is one implicit call in `ScalingMetricCollector.java` https://github.com/apache/flink-kubernetes-operator/blob/main/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/ScalingMetricCollector.java#L339

Thanks, agree we can postpone this PR.

Savonitar:
One option to unblock -> change ScalingMetricCollector.java from logging the whole map to something explicit to remove these local copies and improve code hygiene.

PR #1125, opened 2026-06-01
[FLINK-39814] Honor FlinkSessionJob savepoint dir in snapshots

lrsb:

`FlinkStateSnapshot`-triggered savepoints resolved their target directory from the parent
`FlinkDeployment`'s observe config, so a per-job `state.savepoints.dir` set on the
referenced `FlinkSessionJob` was silently dropped. This affects every operator-triggered savepoint when
`kubernetes.operator.snapshot.resource.enabled` is true (upgrade savepoints,
savepoint-on-deletion, manual triggers). Per-job `state.checkpoints.dir` is unaffected,
since Flink bakes it into the job's `CheckpointStorage` at submission.

## Brief change log

  - `FlinkStateSnapshotContext.referencedJobObserveConfig` now merges the referenced
    `FlinkSessionJob`'s `flinkConfiguration` via `FlinkConfigManager.getSessionJobConfig`,
    instead of only resolving the parent `FlinkDeployment`'s observe config
  - `StateSnapshotReconciler.triggerCheckpointOrSavepoint` uses
    `ctx.getReferencedJobObserveConfig()` to resolve the snapshot config; the
    `FlinkService` still comes from the `FlinkDeployment` context
  - Deployment-referenced snapshots are unaffected (fall back to the deployment observe
    config)

## Verifying this change

  - `FlinkStateSnapshotContextTest`: verifies `referencedJobObserveConfig` returns the
    per-job `state.savepoints.dir` for a `FlinkSessionJob` reference, and falls back to the
    deployment-level dir for a `FlinkDeployment` reference
  - Existing `FlinkStateSnapshotControllerTest` continues to pass (no regression)

## Documentation

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkStateSnapshotContext.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkStateSnapshotContext.java
@@ -65,7 +65,18 @@ public class FlinkStateSnapshotContext {
     }
 
     private Configuration referencedJobObserveConfig() {
-        return getConfigManager().getObserveConfig(getReferencedJobFlinkDeployment());
+        var flinkDeployment = getReferencedJobFlinkDeployment();
+        return getSecondaryResource()
+                .filter(FlinkSessionJob.class::isInstance)
+                .map(FlinkSessionJob.class::cast)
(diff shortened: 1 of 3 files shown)


Dennis-Mircea (approved):
The PR looks good. Thanks for fixing this!

gyfora (approved):

PR #1131, opened 2026-06-04
[FLINK-39838] Bump Flink version to 1.20.4

spuru9:

Bump the Flink version the operator builds against from `1.20.1` (Jan 2025) to `1.20.4` (Apr 2026) — the latest patch in the 1.20 line — to pick up the bug and security fixes accumulated across the 1.20.2, 1.20.3 and 1.20.4 releases. As a patch bump within the same minor it is source/binary compatible.

`flink.version` is the version the operator *builds* against; runtime support for the wider Flink range (2.0/2.1/2.2) is covered separately by the CI version matrix, so this does not change which Flink versions the operator can deploy.

## Brief change log

  - Bump `flink.version` `1.20.1` -> `1.20.4` in the root `pom.xml`
  - Align the autoscaling example image base from `flink:1.20.1-java17` to `flink:1.20.4-java17`
  - Align the e2e batch-job examples jar from `1.20.1` to `1.20.4`
  - Update the `flink-metrics-otel` comment that referenced the pinned `flink.version`

## Verifying this change

This change is already covered by existing tests / CI. Locally verified that `mvn clean install javadoc:javadoc -Pgenerate-docs` builds all modules successfully against Flink 1.20.4.

  - Dependencies (does it add or upgrade a dependency): **yes** (Flink `1.20.1` -> `1.20.4`, patch upgrade)

## Documentation

### DIFF
--- a/flink-autoscaler/src/test/java/org/apache/flink/autoscaler/RestApiMetricsCollectorTest.java
+++ b/flink-autoscaler/src/test/java/org/apache/flink/autoscaler/RestApiMetricsCollectorTest.java
@@ -164,13 +164,13 @@ class RestApiMetricsCollectorTest {
                                     new StandaloneClientHAServices(
                                             miniCluster.getRestAddress().get().toString()));
             var collector = new RestApiMetricsCollector<>();
-            Map<FlinkMetric, Metric> flinkMetricMetricMap = new HashMap<>();
-            // Metrics might not be available yet so retry the query until it returns results or the
-            // timeout reached.
+            // Metrics might not be available yet, and the JobManager exposes the slot metrics
+            // before the TaskManager has registered its slots (reporting 0). Retry the assertion
+            // until the slots are registered, or the timeout is reached.
             await().atMost(Duration.ofSeconds(60))
-                    .until(
+                    .untilAsserted(
                             () -> {
-                                final Map<FlinkMetric, Metric> results =
+                                final Map<FlinkMetric, Metric> flinkMetricMetricMap =
                                         collector.queryJmMetrics(
                                                 client,
                                                 Map.of(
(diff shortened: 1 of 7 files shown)


flink-autoscaler/src/test/java/org/apache/flink/autoscaler/RestApiMetricsCollectorTest.java
```diff
@@ -164,32 +164,29 @@ void testJmMetricCollection() throws Exception {
...
                                         collector.queryJmMetrics(
                                                 client,
                                                 Map.of(
                                                         "taskSlotsTotal",
                                                         FlinkMetric.NUM_TASK_SLOTS_TOTAL,
                                                         "taskSlotsAvailable",
                                                         FlinkMetric.NUM_TASK_SLOTS_AVAILABLE));
-                                flinkMetricMetricMap.putAll(results);
-                                return !results.isEmpty();
+                                assertThat(flinkMetricMetricMap)
```
spuru9, line 181: The 1.20.4 update leads to a failure in this test
As per analysis (by CC)
```
  FLINK-38703 — "ConcurrentModificationException in FineGrainedResourceManager during reporting metrics" — is the matching fix in 1.20.4 (it regressed in 1.20.3). Looking at the 1.20 backport (PR #27393), it
  changed the slot-count gauges (taskSlotsAvailable/taskSlotsTotal):
  - Before: the gauge computed the value live (getNumberFreeSlots()), which under the bug threw a ConcurrentModificationException while the cluster was still stabilizing — so the metric failed to serialize and
  effectively wasn't returned yet. That's exactly why the old "wait until non-empty" worked: it kept retrying through the unstable window.
  - After: the gauge reads a cached field (lastNumberFreeSlots/lastNumberRegisteredSlots) refreshed every second on the main thread, initialized to 0. No exception → the metric is now returned as 0 straight
  away → the old wait grabs the early 0.
```

spuru9:
@gyfora Have updated the NOTICE and the breaking test. Can you rerun the CI.

spuru9:
The CI has passed, if changes look ok. Please merge.

gyfora:
> The CI has passed, if changes look ok. Please merge.

You don't need to remind me, I get enough notifications :) but please run the maven build/tests next time before submitting a PR.

PR #1134, opened 2026-06-09
[FLINK-39890] Fix NaN in observed scaling coefficient when baseline rate is zero

lrsb:

When `job.autoscaler.observed-scalability.enabled` is `true`, the autoscaler throws a `NumberFormatException` while computing the observed scaling coefficient for any vertex with a zero true-processing-rate (idle / very low traffic). The exception aborts the entire scaling pass, so no parallelism overrides / resource requirements are applied and the job stays stuck at its deployed parallelism.

The root cause is in `AutoScalerUtils.optimizeLinearScalingCoefficient`: the denominator is `squaredSum * baselineProcessingRate`, but only `squaredSum` is guarded against zero. An idle vertex reports a finite true processing rate of `0.0` (passing the caller's `isNaN` guard), giving `baselineProcessingRate == 0.0` and `sum == 0.0`. The result is `alpha = 0.0 / 0.0 = NaN`, which `Math.max`/`Math.min` do not sanitize, so it reaches `BigDecimal.valueOf(NaN)` and fails.

This change guards the full denominator so a zero baseline (or zero `squaredSum`) falls back to linear scaling.

## Brief change log

  - Guard the complete denominator `squaredSum * baselineProcessingRate` in `AutoScalerUtils.optimizeLinearScalingCoefficient`, returning the linear-scaling fallback (`1.0`) when it is zero

## Verifying this change

  - Added `testCalculateScalingCoefficientWithZeroProcessingRateFallsBackToLinear` to `JobVertexScalerTest`, which builds a scaling history of vertices with a zero true-processing-rate and asserts `calculateObservedScalingCoefficient` returns the linear-scaling coefficient `1.0` instead of producing `NaN`

  - Core observer or reconciler logic that is regularly executed: yes (the observed scaling coefficient is computed on every scaling pass)

## Documentation

### DIFF
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/utils/AutoScalerUtils.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/utils/AutoScalerUtils.java
@@ -135,11 +135,12 @@ public class AutoScalerUtils {
             squaredSum += parallelism * parallelism;
         }
 
-        if (squaredSum == 0.0) {
+        double denominator = squaredSum * baselineProcessingRate;
+        if (denominator == 0.0) {
             return 1.0; // Fallback to linear scaling if denominator is zero
         }
 
-        double alpha = sum / (squaredSum * baselineProcessingRate);
+        double alpha = sum / denominator;
 
         return Math.max(lowerBound, Math.min(upperBound, alpha));
     }
--- a/flink-autoscaler/src/test/java/org/apache/flink/autoscaler/JobVertexScalerTest.java
+++ b/flink-autoscaler/src/test/java/org/apache/flink/autoscaler/JobVertexScalerTest.java
@@ -1307,6 +1307,24 @@ public class JobVertexScalerTest {
         assertEquals(1, linearReturnWithTwoScalingHistoryRecordScalingCoefficient);
     }
 
+    @Test
+    public void testCalculateScalingCoefficientWithZeroProcessingRateFallsBackToLinear() {
+        var currentTime = Instant.now();
+
+        var zeroProcessingRateHistory = new TreeMap<Instant, ScalingSummary>();
+        zeroProcessingRateHistory.put(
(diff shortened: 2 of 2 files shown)


Dennis-Mircea (commented):
Do you have a real reproduction for this? From I checked, in a real case scenario the `baselineProcessingRate` will never reach this code as `0`.

lrsb:
Yes, this is actually addressing prod issue we encountered. If you are asking for steps is can try to create a simple case.

Dennis-Mircea:
> Yes, this is actually addressing prod issue we encountered. If you are asking for steps is can try to create a simple case.

Thanks, I believe you hit a real `NumberFormatException` in prod. My point is that the value triggering it is `+Infinity`, not `0`.

An idle vertex doesn't report `TRUE_PROCESSING_RATE = 0.0`. In `ScalingMetricEvaluator.computeTprFromBusyTime`, `rate == 0` returns `Double.POSITIVE_INFINITY` ("nothing coming in, assume infinite processing power"), and the observed-TPR path does the same. So `baselineProcessingRate` becomes `+Infinity`, not `0`.

With that, your current guard doesn't catch it:
- `denominator = squaredSum * (+Infinity) = +Infinity`, so `denominator == 0.0` is false.
- the loop only skips `NaN`, not `Infinity`, so `sum` becomes `+Infinity` too.
- `alpha = +Infinity / +Infinity = NaN`, which `Math.max/Math.min` don't sanitize, so `BigDecimal.valueOf(NaN)` still throws.

So this patch would not fix the prod crash. The fix needs to handle non-finite values. Either guard the result (`if (!Double.isFinite(alpha)) return 1.0;`), or drop non-finite observations from the fit (`!Double.isFinite(...)` instead of `isNaN(...)` for both `baselineProcessingRate` and each `processingRate`).

Dennis-Mircea:
@lrsb are you still working on it?

PR #1142, opened 2026-06-19
[hotfix][api] Remove unused Diffable markers from spec/status types

Dennis-Mircea:

`Diffable` is a marker interface that `ReflectiveDiffBuilder` consults only to decide whether to recurse into a **nested** field (`Diffable.class.isAssignableFrom(field.getType())`). The root object of a diff is always processed unconditionally, so a root type's `Diffable`-ness is never checked.

This change removes those unused markers so `Diffable` only appears where the engine actually uses it.

## Brief change log

  - Removed `implements Diffable<...>` and the now-unused `Diffable` import from `AbstractFlinkSpec`.
  - Removed the same from `FlinkStateSnapshotSpec` and `FlinkStateSnapshotStatus`.
  - Kept `Diffable` on `JobSpec` and `TaskManagerSpec`, the nested fields the diff engine genuinely recurses into.

## Verifying this change

This change is already covered by existing tests, all of them passing already with this new change.

  - The public API, i.e., is any changes to the `CustomResourceDescriptors`: no (`Diffable` is an internal marker interface, not serialized; no CRD/schema impact)
  - Core observer or reconciler logic that is regularly executed: no (diff behavior is unchanged)

## Documentation


spuru9 (approved):
LGTM

gyfora (approved):

PR #1144, opened 2026-06-22
[FLINK-31849] Adds basic CRD validations

vsantwana:

This pull request moves simple, single-field spec validation from the operator's imperative DefaultValidator into declarative CRD-generator annotations, so these constraints are enforced by the Kubernetes API server at admission time rather than only during reconciliation. This gives users immediate, schema-level feedback on  `kubectl apply` and simplifies the validation logic in the operator.

## Brief change log

 - Annotated four spec fields with CRD-generator validation annotations: @Min(1) on JobManagerSpec.replicas and TaskManagerSpec.replicas, and @Required on FlinkDeploymentSpec.flinkVersion and IngressSpec.template
  -   Regenerated the flinkdeployments and flinkbluegreendeployments CRDs, which now carry minimum: 1 on the replicas fields and flinkVersion/template in their required lists
  - Removed the now-redundant imperative checks from DefaultValidator (including the validateTmReplicas method), keeping only the validation that genuinely cannot be expressed declaratively (e.g. the conditional parallelism fallback, the JM replicas > 1 requires HA cross-field rule, and the ingress URL parsing)

## Verifying this change
 - Added F`linkConfigurationYamlSupportTest#crdRejectsInvalidSpecValues`, which applies the regenerated CRD to an in-process API server (kubeapitest) and verifies that resources violating each new constraintare rejected at admission, while a fully specified deployment is accepted
  - Updated DefaultValidatorTest to drop the now-removed imperative checks; the remaining 41 validator tests pass, confirming no regression in the validation that stays in the operator

  - The public API, i.e., is any changes to the `CustomResourceDescriptors`: yes. 
  - Core observer or reconciler logic that is regularly executed: no (validation logic only; no observer/reconciler changes)

## Documentation

### DIFF
--- a/flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/spec/FlinkStateSnapshotSpec.java
+++ b/flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/spec/FlinkStateSnapshotSpec.java
@@ -20,6 +20,7 @@ package org.apache.flink.kubernetes.operator.api.spec;
 import org.apache.flink.annotation.Experimental;
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import io.fabric8.generator.annotation.Min;
 import lombok.AllArgsConstructor;
 import lombok.Data;
 import lombok.NoArgsConstructor;
@@ -46,6 +47,7 @@ public class FlinkStateSnapshotSpec {
      * Maximum number of retries before the snapshot is considered as failed. Set to -1 for
      * unlimited or 0 for no retries.
      */
+    @Min(-1)
     private int backoffLimit = -1;
 
     public boolean isSavepoint() {
--- a/flink-kubernetes-operator-api/src/test/java/org/apache/flink/kubernetes/operator/api/FlinkConfigurationYamlSupportTest.java
+++ b/flink-kubernetes-operator-api/src/test/java/org/apache/flink/kubernetes/operator/api/FlinkConfigurationYamlSupportTest.java
@@ -17,12 +17,17 @@
 
 package org.apache.flink.kubernetes.operator.api;
 
+import org.apache.flink.kubernetes.operator.api.spec.IngressSpec;
+import org.apache.flink.kubernetes.operator.api.utils.BaseTestUtils;
+
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
 import io.fabric8.kubeapitest.junit.EnableKubeAPIServer;
 import io.fabric8.kubernetes.api.model.GenericKubernetesResource;
+import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.fabric8.kubernetes.api.model.apiextensions.v1.CustomResourceDefinition;
 import io.fabric8.kubernetes.client.KubernetesClient;
+import io.fabric8.kubernetes.client.KubernetesClientException;
 import io.fabric8.kubernetes.client.dsl.NonDeletingOperation;
 import org.junit.jupiter.api.Test;
 
@@ -30,6 +35,7 @@ import java.io.File;
 import java.io.IOException;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 @EnableKubeAPIServer
 class FlinkConfigurationYamlSupportTest {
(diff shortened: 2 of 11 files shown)


vsantwana:
@gyfora I think we can expand it.

I just added single-field constraints that map cleanly to the CRD schema with zero behavior change for already-valid resources. The same approach can be extended incrementally:
**More single-field constraints**: additional @Min/@Max/@Pattern/@Size annotations, like 
 - FlinkDeploymentSpec.imagePullPolicy → @Pattern("Always|Never|IfNotPresent") (currently unvalidated free-form string; net-new constraint).
  - FlinkSessionJobSpec.deploymentName → @Pattern("[a-z]([-a-z\\d]{0,43}[a-z\\d])?") — the regex already exists in DefaultValidator.validateDeploymentName; this would move
  the session-job call to the schema (the metadata.name call stays, since metadata can't be annotated).

**@Required on always-mandatory fields**: e.g. serviceAccount and deploymentName. These are already mandatory in DefaultValidator, so the only change is moving rejection to admission time but I am not sure about this because it is a user-visible UX change


**Cross-field rules via CEL**:
 @ValidationRule (x-kubernetes-validations, K8s 1.25+) can express constraints like the snapshot "exactly one of checkpoint or savepoint" XOR, or immutability rules using transition rules (oldSelf). But this is very stretch goal.

flink-kubernetes-operator-api/src/test/java/org/apache/flink/kubernetes/operator/api/FlinkConfigurationYamlSupportTest.java
```diff
@@ -68,6 +73,52 @@ void upgradeCRDToYamlFlinkConfiguration() {
         assertThat(deployment.getSpec().getFlinkConfiguration().asFlatMap()).hasSize(5);
     }
 
+    /**
+     * Verifies that the validation constraints declared via CRD-generator annotations (FLINK-31849)
```
vsantwana: remove the JIRA ticket number

vsantwana:
We cannot ship this as long as have support for < 1.30.

Screenshot 2026-06-23 at 1 57 01 PM

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultValidator.java
```diff
@@ -154,7 +154,9 @@ private Optional<String> validateFlinkVersion(FlinkDeployment deployment) {
         var spec = deployment.getSpec();
         var version = spec.getFlinkVersion();
         if (version == null) {
-            return Optional.of("Flink Version must be defined.");
+            // Presence is enforced by the CRD schema (@Required on
+            // FlinkDeploymentSpec.flinkVersion); nothing further to validate here.
+            return Optional.empty();
```
Dennis-Mircea, line 159: The new validation is admission-only and does not cover the reconcile cycle. The new added constraints are enforced only at admission, and only if the updated CRD is installed, which is a manual step (Helm does not auto-upgrade `crds/`). `DefaultValidator`, however, runs on every reconcile as well, so removing these checks drops the runtime validation checks: after an operator bump where the CRD has not been replaced yet, an invalid spec reaching reconcile is no longer caught and hits deploy logic with a bad value.

flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/spec/FlinkStateSnapshotSpec.java
```diff
@@ -46,6 +47,7 @@ public class FlinkStateSnapshotSpec {
      * Maximum number of retries before the snapshot is considered as failed. Set to -1 for
      * unlimited or 0 for no retries.
      */
+    @Min(-1)
```
Dennis-Mircea, line 50: This doesn't look as a necessary validation constraint.

Dennis-Mircea (changes requested):
I'd say this PR touches a validation area that needs a more careful look, deeper investigation, and proper testing before it moves forward.

As it stands it feels a bit superficial, since it also misses a number of other fields that could be validated the same way, for example:

- `FlinkDeploymentSpec.serviceAccount` -> `@Required` (currently in `validateServiceAccount`, same pattern as `flinkVersion`).
- `FlinkSessionJobSpec.deploymentName` -> `@Required` + `@Pattern` (the `DEPLOYMENT_NAME_PATTERN` regex already exists).

Before marking this PR as ready for review, please make sure every single-field constraint in `DefaultValidator` that can be expressed declaratively is actually covered by an annotation. The checks that genuinely cannot move (cross-field or config-derived rules, e.g. the `replicas > 1` requires-HA rule, the parallelism fallback, ingress URL parsing) should be the only ones left imperative, and ideally called out as such.

PR #1145, opened 2026-06-23
[FLINK-39970] Retry incomplete JobManager deployment deletion

stellagai:

Kubernetes deployment deletion waits can fail or time out before the old JobManager deployment is fully removed. The operator currently logs these failures and continues reconciliation, which can submit a replacement cluster while the old deployment is still terminating and result in `AlreadyExists` errors.

## Brief change log

- Propagate non-404 errors while waiting for Kubernetes resources to be deleted.
- Retry reconciliation instead of creating a replacement cluster before deletion completes.
- Preserve the best-effort JobManager shutdown behavior before mandatory deployment deletion.
- Update deletion error and timeout tests.

## Verifying this change

This change added and updated tests and was verified with:

```bash
JAVA_HOME=/opt/homebrew/opt/openjdk@17/libexec/openjdk.jdk/Contents/Home \
  mvn -pl flink-kubernetes-operator -am \
  -DskipITs \
  -Dtest=AbstractFlinkServiceTest \
  -Dsurefire.failIfNoSpecifiedTests=false test
```

Tests run: 40, failures: 0, errors: 0, skipped: 0.

## Documentation

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
@@ -1243,9 +1243,10 @@ public abstract class AbstractFlinkService implements FlinkService {
                         Objects::isNull, timeout.toMillis(), TimeUnit.MILLISECONDS);
                 LOG.info("Completed {}", operation);
             } catch (KubernetesClientException kce) {
-                // We completely ignore not found errors and simply log others
+                // Not found means deletion completed. Other errors must abort this reconcile so a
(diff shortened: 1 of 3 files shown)


Dennis-Mircea (commented):
Thanks for the PR. The fix itself is correct, but this overlaps heavily with #1138 (FLINK-39953), which targets the same root cause in the same `deleteBlocking` wait catch and the same `shutdownJobManagersBlocking` caller.

gyfora:
I would prefer to close this in favor of the already merged solution that targets the timeout exceptions explicitly.

gyfora:
closing this in favor of https://github.com/apache/flink-kubernetes-operator/commit/8e567a5da7b22ebe95af4bfd1c36ee40cbba0b30

PR #1146, opened 2026-06-23
[FLINK-37448] Report actual TaskManager count in status.taskManager.replicas

vsantwana:

`status.taskManager.replicas` previously held a value computed from the spec at reconcile time (`ceil(parallelism / slots)`). This is inaccurate for deployments where the number of TaskManagers fluctuates dynamically rather than being fixed by configuration - most notably standalone clusters running in reactive mode.

This change makes the field report the **actual** number of TaskManagers registered with the running Flink cluster, sourced from the Flink REST API (`/taskmanagers`), for both application and session deployments.

## Brief change log

  - Added `FlinkService#getTaskManagerReplicas(Configuration)`, implemented in `AbstractFlinkService` via the existing `/taskmanagers` REST call; `getClusterInfo` now routes through it (single source of truth).
  - `AbstractFlinkDeploymentObserver#observeClusterInfo` now sets `status.taskManager` (label selector + actual replica count) whenever the JobManager is ready, covering both application and session clusters.
  - `ReconciliationUtils#getTaskManagerInfo` no longer derives the count from the spec; at reconcile time it only sets the label selector (replicas stay `0` until the cluster is observed) and still clears the info when not running.
  - Extracted the duplicated TaskManager label selector into FlinkUtils#getTaskManagerLabelSelector`.
  - Updated the `TaskManagerInfo.replicas` JavaDoc and the generated CRD eference docs to reflect the new semantics.

Note: this changes the semantics of `status.taskManager.replicas` from a spec-derived value to the observed cluster state.

## Verifying this change

  - `ApplicationObserverTest#observeReportsActualTaskManagerReplicas` verifies the
    status reflects the cluster's actual TaskManager count (independent of the
    spec-derived value) and tracks changes across observations.
  - `TestingFlinkService` gained an overridable `taskManagerReplicas` stub
    (defaults to a converged cluster, i.e. the spec-derived count).
  - Existing controller/observer/service/metrics tests updated/confirmed.

  - The public API, i.e., is any changes to the `CustomResourceDescriptors`: no (the field already exists; only its reported value changes)

## Documentation

  - Does this pull request introduce a new feature? no (behavior change to an existing status field)
  - If yes, how is the feature documented? JavaDocs / docs (CRD reference updated)

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/AbstractFlinkDeploymentObserver.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/AbstractFlinkDeploymentObserver.java
@@ -81,26 +81,20 @@ public abstract class AbstractFlinkDeploymentObserver
     private void observeClusterInfo(FlinkResourceContext<FlinkDeployment> ctx) {
         var flinkApp = ctx.getResource();
         try {
-            var observeConfig = ctx.getObserveConfig();
-            // Fetch the actual number of registered TaskManagers once and reuse it both for the
-            // cluster resource totals and the reported replica count, so a single /taskmanagers
-            // call backs a consistent snapshot.
-            var replicas = ctx.getFlinkService().getTaskManagerReplicas(observeConfig);
-            Map<String, String> clusterInfo =
+            var clusterInfo =
                     ctx.getFlinkService()
                             .getClusterInfo(
-                                    observeConfig,
-                                    flinkApp.getStatus().getJobStatus().getJobId(),
-                                    replicas);
-            flinkApp.getStatus().getClusterInfo().putAll(clusterInfo);
+                                    ctx.getObserveConfig(),
+                                    flinkApp.getStatus().getJobStatus().getJobId());
+            flinkApp.getStatus().getClusterInfo().putAll(clusterInfo.getInfo());
             logger.debug("ClusterInfo: {}", flinkApp.getStatus().getClusterInfo());
 
-            // Report the actual number of TaskManagers registered with the running cluster rather
-            // than a value derived from the spec, so the status reflects dynamically changing
-            // deployments (e.g. standalone reactive mode).
             var labelSelector =
                     FlinkUtils.getTaskManagerLabelSelector(flinkApp.getMetadata().getName());
-            flinkApp.getStatus().setTaskManager(new TaskManagerInfo(labelSelector, replicas));
+            flinkApp.getStatus()
+                    .setTaskManager(
+                            new TaskManagerInfo(
+                                    labelSelector, clusterInfo.getTaskManagerReplicas()));
         } catch (Exception e) {
             logger.warn("Exception while fetching cluster info", e);
         }
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/AbstractFlinkDeploymentObserver.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/AbstractFlinkDeploymentObserver.java
@@ -24,6 +24,7 @@ import org.apache.flink.kubernetes.operator.api.spec.JobSpec;
 import org.apache.flink.kubernetes.operator.api.spec.JobState;
 import org.apache.flink.kubernetes.operator.api.status.FlinkDeploymentStatus;
 import org.apache.flink.kubernetes.operator.api.status.JobManagerDeploymentStatus;
+import org.apache.flink.kubernetes.operator.api.status.TaskManagerInfo;
 import org.apache.flink.kubernetes.operator.controller.FlinkResourceContext;
 import org.apache.flink.kubernetes.operator.exception.DeploymentFailedException;
 import org.apache.flink.kubernetes.operator.exception.MissingJobManagerException;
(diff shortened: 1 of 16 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/AbstractFlinkDeploymentObserver.java
```diff
@@ -87,6 +88,14 @@ private void observeClusterInfo(FlinkResourceContext<FlinkDeployment> ctx) {
                                     flinkApp.getStatus().getJobStatus().getJobId());
             flinkApp.getStatus().getClusterInfo().putAll(clusterInfo);
             logger.debug("ClusterInfo: {}", flinkApp.getStatus().getClusterInfo());
+
+            // Report the actual number of TaskManagers registered with the running cluster rather
+            // than a value derived from the spec, so the status reflects dynamically changing
+            // deployments (e.g. standalone reactive mode). See FLINK-37448.
+            var replicas = ctx.getFlinkService().getTaskManagerReplicas(ctx.getObserveConfig());
```
Dennis-Mircea: By adding this implementation here, it will make a subsequent unnecessary REST call, as the `ctx.getFlinkService().getClusterInfo(...)` already does on REST call.. This should be integrated as part of `getClusterInfo(...)` method.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/deployment/AbstractFlinkDeploymentObserver.java
```diff
@@ -87,6 +88,14 @@ private void observeClusterInfo(FlinkResourceContext<FlinkDeployment> ctx) {
                                     flinkApp.getStatus().getJobStatus().getJobId());
             flinkApp.getStatus().getClusterInfo().putAll(clusterInfo);
             logger.debug("ClusterInfo: {}", flinkApp.getStatus().getClusterInfo());
+
+            // Report the actual number of TaskManagers registered with the running cluster rather
+            // than a value derived from the spec, so the status reflects dynamically changing
+            // deployments (e.g. standalone reactive mode). See FLINK-37448.
```
Dennis-Mircea: Too much unnecessary comments verbosity. This comment can be removed entirely.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/FlinkDeploymentControllerTest.java
```diff
@@ -317,9 +317,11 @@ public void verifyUpgradeFromSavepointLegacyMode(FlinkVersion flinkVersion) thro
         var jobs = flinkService.listJobs();
         assertEquals(1, jobs.size());
         assertEquals("s0", jobs.get(0).f0);
+        // The actual replica count is populated during observation from the running cluster; right
+        // after reconciliation (cluster not yet observed) it stays 0 (FLINK-37448).
```
Dennis-Mircea: No need for this comment.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
```diff
@@ -239,11 +237,12 @@ private static void updateLastReconciledJobSpec(
...
 
-    private static TaskManagerInfo getTaskManagerInfo(
-            String name, Configuration conf, JobState jobState) {
-        var labelSelector = "component=taskmanager,app=" + name;
+    private static TaskManagerInfo getTaskManagerInfo(String name, JobState jobState) {
         if (jobState == JobState.RUNNING) {
-            return new TaskManagerInfo(labelSelector, FlinkUtils.getNumTaskManagers(conf));
+            // The actual replica count is populated during observation from the running cluster
+            // (see AbstractFlinkDeploymentObserver#observeClusterInfo). Here we only set the label
+            // selector; the replica count stays 0 until the cluster is observed.
```
Dennis-Mircea: Unnecessary comment verbosity.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/ReconciliationUtils.java
```diff
@@ -239,11 +237,12 @@ private static void updateLastReconciledJobSpec(
...
+            return new TaskManagerInfo(FlinkUtils.getTaskManagerLabelSelector(name), 0);
```
Dennis-Mircea: This now resets replicas to 0 when RUNNING, and it stays 0 until the next observe populates it. In steady state this is fine, but right after every upgrade/spec change the scale subresource reports 0 until the next observe loop. This doesn't look ok.

Dennis-Mircea (changes requested):
This PR needs more test coverage as well. The new tests covers only the application observer. The change deliberately covers session clusters too, but there is no session-path test. Consider tightening the test coverage as well.

Dennis-Mircea (commented):
One more docs spot goes stale with this change. The docs were restructured in #1164, which added a **Status and Lifecycle** page describing this field. `docs/content/docs/custom-resource/status-and-lifecycle.md` still lists `taskManager` as "populated in application mode only", which is no longer accurate, you now set it in `AbstractFlinkDeploymentObserver` for session clusters too (and your new `SessionObserverTest` confirms it). Suggested row:

   ```
   | `taskManager` | `FlinkDeployment` | The TaskManager label selector and the actual number of TaskManagers registered with the running cluster, backing the Kubernetes scale subresource. Populated once the running cluster has been observed |
   ```

Please mirror the edit into `docs/content.zh/docs/custom-resource/status-and-lifecycle.md` as well (the zh tree is kept as a 1:1 mirror, same as you're already doing for `reference.md`).

gyfora (changes requested):
Hot @vsantwana , can you please address @Dennis-Mircea 's doc comment and then we can merge this :)

PR #1147, opened 2026-06-23
[FLINK-38049] Fix stateless job incorrectly using last-state restore when HA is enabled

lucasgameiroborges:

When HA is enabled and a stateless job is resubmitted (e.g. due to an unhealthy cluster), `resubmitJob` unconditionally overrode the upgrade mode to `LAST_STATE` and passed `requireHaMetadata=true` to `restoreJob`, ignoring the user-configured `STATELESS` upgrade mode. This caused the job to attempt a last-state restore from HA metadata instead of starting fresh.

## Brief change log

- `AbstractJobReconciler#resubmitJob`: skip the `LAST_STATE` mode override and the `requireHaMetadata` flag when the spec's upgrade mode is `STATELESS`
- Add `ApplicationReconcilerTest#testRestartUnhealthyStatelessJobWithHaEnabled` to reproduce the scenario: a stateless job with HA enabled is resubmitted after an unhealthy event with no HA metadata available — previously this would throw `UpgradeFailureException`, now it succeeds

## Verifying this change

New unit test `testRestartUnhealthyStatelessJobWithHaEnabled` in `ApplicationReconcilerTest` covers the fix. Existing `ApplicationReconcilerTest` and `ApplicationReconcilerUpgradeModeTest` suites (109 tests total) continue to pass.

## Does this pull request potentially affect one of the following areas

- Job lifecycle/upgrade: **yes** — affects resubmit path for stateless jobs when HA is active

_This is a fix for [FLINK-38049](https://issues.apache.org/jira/browse/FLINK-38049)._

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractJobReconciler.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/AbstractJobReconciler.java
@@ -607,13 +607,15 @@ public abstract class AbstractJobReconciler<
                 ctx.getResource().getStatus().getJobStatus().getSavepointInfo().getLastSavepoint();
         var lastSavepointKnown = upgradeStatePath != null || savepointLegacy != null;
 
-        if (requireHaMetadata) {
+        boolean statelessResubmit =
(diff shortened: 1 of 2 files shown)


Dennis-Mircea (commented):
@lucasgameiroborges are you still working on this?

PR #1149, opened 2026-06-30
[FLINK-39959] Improve the autoscaler context structure and propagation

Dennis-Mircea:

The autoscaler pipeline (metric collection, evaluation, scaling execution) passed its per-cycle data through long, growing parameter lists, and each stage reached into the previous stage's outputs in ad-hoc ways. The three plugin SPIs (FLIP-514 custom evaluator, FLIP-575 scaling executor, and FLIP-586 parallelism alignment) each defined their own context type that re-declared data already present on `JobAutoScalerContext` (configuration, evaluated metrics, job topology), so the same information existed in four shapes.

This change gives the autoscaler a single canonical context that is threaded through the whole cycle and enriched as it advances:

- A per-cycle `ScalingCycleState` carried by `JobAutoScalerContext` holds the working data of one scaling cycle.
- Each pipeline stage depends only on the context handed to it.
- All three plugin contexts become thin extensions of `JobAutoScalerContext` rather than parallel types.

## Brief change log

- Added `JobAutoScalerContext.ScalingCycleState`, the mutable per-cycle working state of the pipeline (cycle start instant, autoscaler metrics sink, collected metrics, scaling tracking and history, restart time, evaluated metrics). It is lazily created per context, has package-private setters, and is excluded from `toString`.
- Reworked `JobAutoScalerContext` to carry the cycle state and expose convenience accessors derived from it (`getJobTopology`, `getEvaluatedMetrics`, `getMetricsHistory`, `getRestartTime`). Added a protected copy constructor used by plugin contexts that shares the cycle state and replaces the configuration with the effective per-plugin one.
- Reworked the pipeline into three stages that each rely solely on the context: `ScalingMetricCollector.collect(ctx)`, `ScalingMetricEvaluator.evaluate(ctx)`, and `ScalingExecutor.execute(ctx)`. `runScalingLogic` in `JobAutoScalerImpl` becomes the cycle set-up plus those three calls.
- Moved the metric state store onto `ScalingMetricCollector` as a field, and the retained `lastEvaluatedMetrics` onto `ScalingMetricEvaluator` (now generic and stateful), so the scaling metric gauges keep reporting the latest values across cycles.
- Collapsed the evaluator entry points to a single `evaluate(Context)` (the production path that records results and registers gauges), backed by an internal `computeEvaluatedMetrics` worker that runs the custom evaluator plugin only when a context is present.
- Made all three autoscaler plugin contexts (`ScalingExecutorPlugin.Context`, `ScalingMetricsEvaluatorPlugin.Context`, and FLIP-586's `ParallelismAlignmentMode.Context`) extend `JobAutoScalerContext`. Each shares the canonical `ScalingCycleState`, exposes the effective per-plugin configuration through the inherited `getConfiguration()`, and adds only genuinely plugin-specific data: nothing extra for the executor, the in-progress evaluated vertex metrics and backlog flag for the evaluator, and the per-vertex alignment inputs (parallelism, key-group / partition counts, limits, ship strategies) plus the vertex's evaluated metrics for the alignment mode.
- Converted `ParallelismAlignmentMode.Context` from the standalone `@Value` object into the pattern above, dropping its wrapped `jobContext`, duplicated `jobTopology`, and separate `modeConfiguration` fields in favour of the inherited context, `getJobTopology()`, and `getConfiguration()`. `ParallelismAligner`, the built-in / legacy modes, and their tests were updated to the `Context<?>` signatures, and `JobVertexScaler` no longer threads the topology into `align(...)`.
- Dropped the vestigial `CTX` type parameter from `ScalingExecutorPlugin`, leaving `ScalingExecutorPlugin<KEY>` (propagated through the operator and standalone plugin discovery).
- Removed the redundant `delayedScaleDown` parameter that was threaded alongside the context.

## Verifying this change

This is an internal refactor and does not change autoscaler behavior.

- The full flink-autoscaler suite passes (257 tests), including the migrated tests for the collect / evaluate / execute stages and the FLIP-586 alignment tests.
- Added `ScalingMetricEvaluatorTest.testEvaluatorPluginContextExtendsCanonicalContext` and extended the executor's plugin-context test to assert the new guarantees: the plugin context is a `JobAutoScalerContext`, shares the canonical `ScalingCycleState` and metric group by identity, and returns the effective per-plugin configuration from `getConfiguration()`.
- Updated `AlignmentModeTest` to build `ParallelismAlignmentMode.Context` from a real parent context through the shared copy constructor.
- Downstream modules (flink-autoscaler-standalone, flink-kubernetes-operator) compile and pass checkstyle against the updated signatures.

## Documentation

- If yes, how is the feature documented? not applicable.

### DIFF
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerContext.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerContext.java
@@ -20,21 +20,33 @@ package org.apache.flink.autoscaler;
 import org.apache.flink.annotation.Experimental;
 import org.apache.flink.api.common.JobID;
 import org.apache.flink.api.common.JobStatus;
+import org.apache.flink.autoscaler.metrics.AutoscalerFlinkMetrics;
+import org.apache.flink.autoscaler.metrics.CollectedMetricHistory;
+import org.apache.flink.autoscaler.metrics.CollectedMetrics;
+import org.apache.flink.autoscaler.metrics.EvaluatedMetrics;
+import org.apache.flink.autoscaler.topology.JobTopology;
 import org.apache.flink.client.program.rest.RestClusterClient;
 import org.apache.flink.configuration.Configuration;
 import org.apache.flink.configuration.MemorySize;
 import org.apache.flink.configuration.TaskManagerOptions;
 import org.apache.flink.metrics.MetricGroup;
+import org.apache.flink.runtime.jobgraph.JobVertexID;
 import org.apache.flink.util.function.SupplierWithException;
 
+import lombok.AccessLevel;
 import lombok.Builder;
 import lombok.Getter;
-import lombok.RequiredArgsConstructor;
+import lombok.Setter;
 import lombok.ToString;
 
 import javax.annotation.Nullable;
 
+import java.time.Duration;
+import java.time.Instant;
+import java.util.Collections;
+import java.util.Map;
 import java.util.Optional;
+import java.util.SortedMap;
 
 /**
  * The job autoscaler context, it includes all details related to the current job.
@@ -42,9 +54,7 @@ import java.util.Optional;
  * @param <KEY> The job key.
  */
 @Experimental
-@RequiredArgsConstructor
 @ToString
-@Builder(toBuilder = true)
 public class JobAutoScalerContext<KEY> {
 
     /** The identifier of each flink job. */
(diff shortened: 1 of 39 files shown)


flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerContext.java
```diff
@@ -58,14 +68,61 @@ public class JobAutoScalerContext<KEY> {
...
+    /**
+     * Copy constructor used by plugin context subclasses. It copies all fields from the given
+     * context except the configuration, which is replaced by the provided one, so a plugin context
+     * exposes its effective per-plugin configuration through {@link #getConfiguration()}. It also
+     * shares the {@link ScalingCycleState}, so the derived context exposes the same per-cycle data
+     * (collected / evaluated metrics, topology, restart time) through the inherited accessors
+     * rather than duplicating it.
+     */
+    protected JobAutoScalerContext(
+            JobAutoScalerContext<KEY> autoScalerContext, Configuration configuration) {
```
gyfora: Feels like this constructor should call the overlayConfig instead of every caller individually calling it

gyfora (approved):

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerContext.java
Dennis-Mircea: Addressed!

PR #1150, opened 2026-07-01
[FLINK-38795] Support Flink 2.0+ HA in standalone mode

mdemierre:

When HA is was enabled in standalone deployment mode, the JobManager decorator passed the pod IP to the entrypoint using the `--host` CLI argument (application mode) and a bare positional argument (session mode). That `--host` argument was removed from the standalone entrypoints in Flink 2.0, breaking standalone HA deployments.

## Brief change log

Replace `--host` with the recommended dynamic config `-D jobmanager.rpc.address=$(POD_IP)` in both application and session modes.

This is a valid config in both Flink 1.x and 2.x, so no version detection is required.

## Verifying this change

- Existing unit tests, expanded with use of the real argument parser
- Existing e2e tests exercise this code path

## Documentation

### DIFF
--- a/flink-kubernetes-standalone/src/test/java/org/apache/flink/kubernetes/operator/kubeclient/decorators/CmdStandaloneJobManagerDecoratorTest.java
+++ b/flink-kubernetes-standalone/src/test/java/org/apache/flink/kubernetes/operator/kubeclient/decorators/CmdStandaloneJobManagerDecoratorTest.java
@@ -32,9 +32,7 @@ import org.junit.jupiter.api.Test;
 
 import java.util.List;
 
-import static org.hamcrest.MatcherAssert.assertThat;
-import static org.hamcrest.Matchers.contains;
-import static org.hamcrest.Matchers.containsInAnyOrder;
+import static org.assertj.core.api.Assertions.assertThat;
 
 /**
  * @link CmdStandaloneJobManagerDecorator unit tests
@@ -67,11 +65,11 @@ public class CmdStandaloneJobManagerDecoratorTest {
                 StandaloneKubernetesConfigOptionsInternal.ClusterMode.SESSION);
 
         FlinkPod decoratedPod = decorator.decorateFlinkPod(new FlinkPod.Builder().build());
-        assertThat(
-                decoratedPod.getMainContainer().getCommand(), containsInAnyOrder(MOCK_ENTRYPATH));
-        assertThat(
-                decoratedPod.getMainContainer().getArgs(),
-                containsInAnyOrder(CmdStandaloneJobManagerDecorator.JOBMANAGER_ENTRYPOINT_ARG));
+        assertThat(decoratedPod.getMainContainer().getCommand())
+                .containsExactlyInAnyOrder(MOCK_ENTRYPATH);
+        assertThat(decoratedPod.getMainContainer().getArgs())
+                .containsExactlyInAnyOrder(
+                        CmdStandaloneJobManagerDecorator.JOBMANAGER_ENTRYPOINT_ARG);
     }
 
     @Test
@@ -85,17 +83,16 @@ public class CmdStandaloneJobManagerDecoratorTest {
         configuration.set(SavepointConfigOptions.SAVEPOINT_PATH, "/tmp/savepoint/path");
 
         FlinkPod decoratedPod = decorator.decorateFlinkPod(new FlinkPod.Builder().build());
-        assertThat(
-                decoratedPod.getMainContainer().getCommand(), containsInAnyOrder(MOCK_ENTRYPATH));
-        assertThat(
-                decoratedPod.getMainContainer().getArgs(),
-                containsInAnyOrder(
+        assertThat(decoratedPod.getMainContainer().getCommand())
+                .containsExactlyInAnyOrder(MOCK_ENTRYPATH);
+        assertThat(decoratedPod.getMainContainer().getArgs())
+                .containsExactlyInAnyOrder(
                         CmdStandaloneJobManagerDecorator.APPLICATION_MODE_ARG,
                         "--allowNonRestoredState",
                         "--fromSavepoint",
                         "/tmp/savepoint/path",
                         "--job-classname",
-                        testMainClass));
+                        testMainClass);
     }
 
     @Test
@@ -105,11 +102,10 @@ public class CmdStandaloneJobManagerDecoratorTest {
                 StandaloneKubernetesConfigOptionsInternal.ClusterMode.APPLICATION);
 
         FlinkPod decoratedPod = decorator.decorateFlinkPod(new FlinkPod.Builder().build());
-        assertThat(
-                decoratedPod.getMainContainer().getCommand(), containsInAnyOrder(MOCK_ENTRYPATH));
-        assertThat(
-                decoratedPod.getMainContainer().getArgs(),
-                containsInAnyOrder(CmdStandaloneJobManagerDecorator.APPLICATION_MODE_ARG));
+        assertThat(decoratedPod.getMainContainer().getCommand())
+                .containsExactlyInAnyOrder(MOCK_ENTRYPATH);
+        assertThat(decoratedPod.getMainContainer().getArgs())
+                .containsExactlyInAnyOrder(CmdStandaloneJobManagerDecorator.APPLICATION_MODE_ARG);
     }
 
     @Test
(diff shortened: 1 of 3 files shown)


Dennis-Mircea (commented):
Thanks for opening this PR, @mdemierre! The FLINK-40047 seems to duplicate FLINK-38795 ("Do not use `--host` CLI param for standalone job"), which is the original ticket, already assigned to @ferenc-csaky, and whose comment proposes this exact fix. Worth linking the two and closing one as a duplicate.

Besides that the PR looks ok. I'd only strengthen the test coverage. Can we add real runtime test support to assert this?

mdemierre:
@Dennis-Mircea Thanks for the link, I didn't find it when I searched before opening mine 😅. I linked them. should I close mine and switch this PR to [FLINK-38795](https://issues.apache.org/jira/browse/FLINK-38795)?

For the runtime tests, what would be your suggestion? I see there is `e2e-tests/test_application_kubernetes_ha.sh` test script in the repo which includes standalone HA mode and runs with v2.0+ in CI.

The symptom on my side was an error of my job receiving the `--host` argument instead of Flink absorbing it. I assume the example job in the e2e test got this as well, but it didn't crash.

One thing I can do too is to feed the resulting arguments to the real parser classes of Flink. Not really rumtime but would prove the arguments are parseable and would have caught the --host issue.

mdemierre:
@Dennis-Mircea I have included the tests which use the real argument parsers

ferenc-csaky:
@mdemierre Thanks for the contribution. I'd prefer keeping the older ticket and closing the new one as duplicate. I'll do the Jira administration and assign the one i originally reported to you. Will also review this change later today.

flink-kubernetes-standalone/src/test/java/org/apache/flink/kubernetes/operator/kubeclient/decorators/CmdStandaloneJobManagerDecoratorTest.java
```diff
@@ -148,9 +165,97 @@ public void testApplicationKubernetesHAArgsAdded() {
...
                         "--test",
                         "123"));
     }
+
+    /**
+     * Verifies that the HA args produced for an application cluster are actually understood by the
+     * real Flink standalone application entrypoint parser and resolve to {@code
+     * jobmanager.rpc.address}. This guards against regressions where the generated args are removed
+     * or renamed in a future Flink version (as {@code --host} was in Flink 2.0).
+     */
```
ferenc-csaky: Let's remove the javadoc from all test cases. It's just noise.

flink-kubernetes-standalone/src/test/java/org/apache/flink/kubernetes/operator/kubeclient/decorators/CmdStandaloneJobManagerDecoratorTest.java
```diff
@@ -148,9 +165,97 @@ public void testApplicationKubernetesHAArgsAdded() {
...
+        FlinkPod decoratedPod = decorator.decorateFlinkPod(new FlinkPod.Builder().build());
+
+        // The leading APPLICATION_MODE_ARG ("standalone-job") is consumed by the Flink
+        // docker-entrypoint before the JVM, so it is not part of the entrypoint's own args.
+        Configuration parsed =
+                parseEntrypointConfig(
+                        new StandaloneApplicationClusterConfigurationParserFactory(),
+                        dropEntrypointToken(decoratedPod.getMainContainer().getArgs()));
+
+        assertEquals(
```
ferenc-csaky: Pls use AssertJ for any kind of assertions instead of JUnit asserts.

flink-kubernetes-standalone/src/test/java/org/apache/flink/kubernetes/operator/kubeclient/decorators/CmdStandaloneJobManagerDecoratorTest.java
```diff
@@ -148,9 +165,97 @@ public void testApplicationKubernetesHAArgsAdded() {
...
+
+    private static Configuration parseEntrypointConfig(
+            ParserResultFactory<? extends ClusterConfiguration> factory, List<String> args)
+            throws Exception {
+        // The standalone entrypoints require --configDir; the Flink docker-entrypoint always
+        // supplies it. Its value is irrelevant here since we assert on the parsed dynamic
+        // properties rather than the loaded flink-conf.yaml.
+        List<String> argsWithConfigDir = new ArrayList<>();
+        argsWithConfigDir.add("--configDir");
+        argsWithConfigDir.add("/opt/flink/conf");
```
ferenc-csaky: Since this won't be resolved let's give it a string like `dummy` to make it more explicit this path won't be resolved at all.

flink-kubernetes-standalone/src/test/java/org/apache/flink/kubernetes/operator/kubeclient/decorators/CmdStandaloneJobManagerDecoratorTest.java
```diff
@@ -148,9 +165,97 @@ public void testApplicationKubernetesHAArgsAdded() {
...
+    private static List<String> dropEntrypointToken(List<String> args) {
+        return args.subList(1, args.size());
+    }
+
+    private static Configuration parseEntrypointConfig(
+            ParserResultFactory<? extends ClusterConfiguration> factory, List<String> args)
+            throws Exception {
+        // The standalone entrypoints require --configDir; the Flink docker-entrypoint always
+        // supplies it. Its value is irrelevant here since we assert on the parsed dynamic
+        // properties rather than the loaded flink-conf.yaml.
```
ferenc-csaky: Let's shorten this to smth like `Adding configDir as it's required by the parser factory`. The rest is just noise.

flink-kubernetes-standalone/src/test/java/org/apache/flink/kubernetes/operator/kubeclient/decorators/CmdStandaloneJobManagerDecoratorTest.java
```diff
@@ -148,9 +165,97 @@ public void testApplicationKubernetesHAArgsAdded() {
...
+                parseEntrypointConfig(
+                        new StandaloneApplicationClusterConfigurationParserFactory(),
+                        dropEntrypointToken(decoratedPod.getMainContainer().getArgs()));
+
+        assertNull(parsed.get(JobManagerOptions.ADDRESS));
+    }
+
+    private static List<String> dropEntrypointToken(List<String> args) {
+        return args.subList(1, args.size());
+    }
```
ferenc-csaky: I think we should just integrate this to `parseEntrypointConfig`, it's only relevant there and would make the call side more concise. And would also move the explanation (in a shortened format) why the leading argument is removed from the list), and remove those comments from the test cases.

flink-kubernetes-standalone/src/test/java/org/apache/flink/kubernetes/operator/kubeclient/decorators/CmdStandaloneJobManagerDecoratorTest.java
```diff
@@ -96,6 +106,13 @@ public void testApplicationCommandAndArgsAdded() {
                         "/tmp/savepoint/path",
                         "--job-classname",
                         testMainClass));
+
+        // The full application command must be accepted by the real Flink standalone application
+        // entrypoint parser, i.e. --job-classname / --fromSavepoint / --allowNonRestoredState are
+        // valid, correctly-spelled options (parsing throws on an unrecognized option).
```
ferenc-csaky: Pls. remove this. it's just noise.

ferenc-csaky (changes requested):
The applied logic LGTM, added some comments for the test.

mdemierre:
> @mdemierre Thanks for the contribution. I'd prefer keeping the older ticket and closing the new one as duplicate. I'll do the Jira administration and assign the one i originally reported to you. Will also review this change later today.

Thanks @ferenc-csaky for the update and review. I'll rename the PR and commits to use the older Jira.

The e2e tests actually failed and it looks like it's related to my change as we can see "-d" in the address. I'll investigate.

mdemierre:
@ferenc-csaky I have cleaned up the commits to remove noisy comments. I switched this test file to use AssertJ only in a first refactoring commit. I also fixed the failing e2e test by separating `-D` and the value.

mdemierre:
The test failure seems to be a transient networking issue, but this time the HA standalone session tests passed, so the latest fix to separate the arguments looks OK.

ferenc-csaky (approved):
LGTM

PR #1153, opened 2026-07-06
[FLINK-39981] [Autoscaler] Correct dynamic source topology from active split metrics

lnbest0707:

Dynamic Kafka sources can remove clusters and splits while stale
`KafkaSourceReader...currentOffset` metric names remain visible to the autoscaler. The
legacy partition count can therefore stay above the source's actual active split count, leaving
empty source subtasks without a rescale or recovery action.

This change adds an opt-in topology correction path for
[`FLINK-39981`](https://issues.apache.org/jira/browse/FLINK-39981) using the
`DynamicKafkaSource.activeSplitCount` gauge added by
[`FLINK-39990`](https://issues.apache.org/jira/browse/FLINK-39990). The metric is optional:
disabled, missing, incomplete, or malformed data keeps the existing behavior. An exact all-zero
aggregate is retained only as a non-scaling signal that clears and re-arms a prior recovery latch.

The `N >= P` recovery branch depends on the connector recovery semantics merged in
[`apache/flink-connector-kafka#278`](https://github.com/apache/flink-connector-kafka/pull/278)
for [`FLINK-39980`](https://issues.apache.org/jira/browse/FLINK-39980). That change adopts
`SupportsSplitReassignmentOnRecovery`: restored active splits are reported back to the enumerator
and requeued for redistribution. Without that connector support, a same-parallelism restart could
restore the same skewed reader ownership, so operators should not rely on this recovery branch.

## Brief change log

  - Adds `job.autoscaler.dynamic-source.topology-correction.enabled`, defaulting to `false`.
  - Collects the optional active split gauge only when enabled and derives total active splits and
    minimum active splits from its aggregated value without replacing legacy
    `NUM_SOURCE_PARTITIONS`; an exact all-zero aggregate is retained only to clear a prior
    recovery latch and never becomes a scaling target.
  - After active splits remain below current parallelism (`N < P`) for the full metrics window,
    caps source parallelism without waiting for the ordinary scale-down delay.
  - When the active split gauge is valid, also uses it as the upper bound for later ordinary
    utilization scale-ups, so stale legacy partition metrics cannot scale the source back above
    its active split count.
  - Detects a persistent empty-reader assignment hole when `N >= P` and the minimum active split
    count is zero for the full metrics window, then requests one durable same-parallelism recovery;
    [`apache/flink-connector-kafka#278`](https://github.com/apache/flink-connector-kafka/pull/278)
    turns that recovery into active-split redistribution.
  - Adds a backward-compatible default realizer hook. The Kubernetes realizer uses an idempotent
    restart nonce for `LAST_STATE` jobs and reuses an already pending restart nonce.
  - Persists the recovery latch in Kubernetes and JDBC state stores, including old-state
    compatibility coverage.
  - Regenerates the autoscaler configuration documentation.

## Verifying this change

  - `mvn -q -pl flink-autoscaler -am test`
  - `mvn -q -pl flink-autoscaler-plugin-jdbc -am -Dtest=JdbcAutoScalerStateStoreTest -Dsurefire.failIfNoSpecifiedTests=false test`
  - `mvn -q -pl flink-kubernetes-operator -am -Dtest=KubernetesAutoScalerStateStoreTest,KubernetesScalingRealizerTest -Dsurefire.failIfNoSpecifiedTests=false test`
  - `mvn -q -pl flink-kubernetes-docs -am -DskipTests -Pgenerate-docs package`
  - `mvn -q spotless:check`

## Documentation

  - If yes, how is the feature documented? docs / JavaDocs

## AI assistance

This contribution was prepared with Codex (GPT-5); the author reviewed, adapted, and tested the
changes.

### DIFF
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
@@ -46,8 +46,10 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.OptionalInt;
 import java.util.SortedMap;
 
+import static org.apache.flink.autoscaler.config.AutoScalerOptions.DYNAMIC_SOURCE_TOPOLOGY_CORRECTION_ENABLED;
 import static org.apache.flink.autoscaler.config.AutoScalerOptions.MAX_SCALE_DOWN_FACTOR;
 import static org.apache.flink.autoscaler.config.AutoScalerOptions.MAX_SCALE_UP_FACTOR;
 import static org.apache.flink.autoscaler.config.AutoScalerOptions.OBSERVED_SCALABILITY_ENABLED;
@@ -57,6 +59,7 @@ import static org.apache.flink.autoscaler.config.AutoScalerOptions.SCALING_EVENT
 import static org.apache.flink.autoscaler.config.AutoScalerOptions.UTILIZATION_TARGET;
 import static org.apache.flink.autoscaler.config.AutoScalerOptions.VERTEX_MAX_PARALLELISM;
 import static org.apache.flink.autoscaler.config.AutoScalerOptions.VERTEX_MIN_PARALLELISM;
+import static org.apache.flink.autoscaler.metrics.ScalingMetric.ACTIVE_SOURCE_SPLIT_COUNT;
 import static org.apache.flink.autoscaler.metrics.ScalingMetric.EXPECTED_PROCESSING_RATE;
 import static org.apache.flink.autoscaler.metrics.ScalingMetric.MAX_PARALLELISM;
 import static org.apache.flink.autoscaler.metrics.ScalingMetric.NUM_SOURCE_PARTITIONS;
@@ -183,6 +186,17 @@ public class JobVertexScaler<KEY, Context extends JobAutoScalerContext<KEY>> {
             DelayedScaleDown delayedScaleDown) {
         var conf = context.getConfiguration();
         var currentParallelism = (int) evaluatedMetrics.get(PARALLELISM).getCurrent();
+        var activeSplitCount = getActiveSplitCount(conf, evaluatedMetrics);
+        if (activeSplitCount.isPresent() && activeSplitCount.getAsInt() < currentParallelism) {
+            return computeSourcePartitionCap(
+                    vertex,
+                    conf,
+                    evaluatedMetrics,
+                    currentParallelism,
+                    activeSplitCount.getAsInt(),
+                    delayedScaleDown);
+        }
+
         double averageTrueProcessingRate = evaluatedMetrics.get(TRUE_PROCESSING_RATE).getAverage();
         if (Double.isNaN(averageTrueProcessingRate)) {
             LOG.warn(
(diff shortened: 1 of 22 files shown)


flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
```diff
@@ -270,6 +279,80 @@ public ParallelismChange computeScaleTargetParallelism(
...
+
+        var activeSplitCountMetric = evaluatedMetrics.get(ACTIVE_SOURCE_SPLIT_COUNT);
+        if (activeSplitCountMetric == null
+                || !isPositiveWholeNumber(activeSplitCountMetric.getCurrent())
+                || activeSplitCountMetric.getCurrent() > Integer.MAX_VALUE) {
+            return ParallelismChange.noChange(currentParallelism);
+        }
+
+        int activeSplitCount = (int) activeSplitCountMetric.getCurrent();
+        if (activeSplitCount >= currentParallelism) {
```
bowenli86: Could we keep `activeSplitCount` as the upper bound after the first correction too?

For example, P=10 and N=5 gets corrected to 5. But once P=N, this falls through to regular scaling, which still uses stale `NUM_SOURCE_PARTITIONS`

Under load it can scale above 5 again, then get capped back on the next pass. Maybe clamp later scale ups to N too when the metric is valid

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
```diff
@@ -270,6 +279,80 @@ public ParallelismChange computeScaleTargetParallelism(
...
+        }
+
+        var activeSplitCountMetric = evaluatedMetrics.get(ACTIVE_SOURCE_SPLIT_COUNT);
+        if (activeSplitCountMetric == null
+                || !isPositiveWholeNumber(activeSplitCountMetric.getCurrent())
+                || activeSplitCountMetric.getCurrent() > Integer.MAX_VALUE) {
+            return ParallelismChange.noChange(currentParallelism);
+        }
+
+        int activeSplitCount = (int) activeSplitCountMetric.getCurrent();
```
bowenli86: Could we make N < P persist for a bit before bypassing the scale down delay?

During a DynamicKafkaSource metadata refresh, the enumerator sends `MetadataUpdateEvent` before the new enumerators start. Readers can briefly report a lower positive count while new splits are still in flight

One autoscaler poll in that window could cause an unnecessary downscale or restart. Maybe use the same stability idea as the N >= P path, or a shorter window

bowenli86:
@gyfora can you or anyone you know who has more expertise here can help review?

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
lnbest0707: Good catch — addressed in 234725a1. When the optional gauge is valid, ordinary utilization scaling now also clamps its upper bound to activeSplitCount, so after P=10/N=5 is corrected to P=5, stale legacy partition metrics cannot scale it back above 5. Missing or invalid gauges still use the legacy path. I also added tests for the steady-state cap and missing-gauge fallback.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobVertexScaler.java
lnbest0707: Good point — the previous full-window check only covered the N >= P && min == 0 recovery path. Addressed in 9bc4e470: N < P is now only exposed to the scaler after every retained metrics-window sample reports valid 0 < N < P. A transient or missing sample falls back to legacy behavior.

PR #1154, opened 2026-07-07
[FLINK-40092] Finalize BlueGreen transition after deleting previous deployment 

fqshopify:

This pull request removes an unnecessary extra reconciliation loop during `FlinkBlueGreenDeployment` cutover, and in doing so closes a window that could cause a spurious abort with no rollback target.

Previously, after the operator successfully deleted the previous (Blue) `FlinkDeployment`, it returned `noUpdate().rescheduleAfter(RETRY_DELAY_MS)` **without finalizing**, leaving the BlueGreen resource in `TRANSITIONING_TO_GREEN`. Finalization to the terminal `ACTIVE_GREEN` state only happened on a *later* reconciliation.

That extra loop is not just wasteful, it re-runs `monitorTransition` on Green while the transition is still considered in-flight, so if Green is observed **not ready** in that window it is routed to `shouldWeAbort` and Green is suspended. With Blue already deleted, there is no rollback target left so the terminal state is no pipelines running i.e. downtime!

As a result, once the previous deployment has been successfully deleted, the transition can and should be committed forward immediately. This change reuses the existing `finalizeBlueGreenDeployment` logic in the same reconciliation that deletes Blue, so the abort path is no longer reachable after cutover. 

## Brief change log

- Pass the target `FlinkBlueGreenDeploymentState` into `deleteDeployment`.
- Finalize the BlueGreen transition immediately after successful deletion of the previous `FlinkDeployment`.
- Preserve the existing retry behavior when deletion fails or is not acknowledged.
- Add a regression test verifying that the transition commits to `ACTIVE_GREEN` in the same reconciliation that deletes Blue.

## Documentation

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/bluegreen/BlueGreenDeploymentService.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/bluegreen/BlueGreenDeploymentService.java
@@ -608,7 +608,7 @@ public class BlueGreenDeploymentService {
         long deletionTimestamp = deploymentReadyTimestamp + deploymentDeletionDelayMs;
 
         if (deletionTimestamp < System.currentTimeMillis()) {
-            return deleteDeployment(currentDeployment, context);
+            return deleteDeployment(currentDeployment, context, nextState);
         } else {
             return waitBeforeDeleting(currentDeployment, deletionTimestamp);
         }
@@ -627,17 +627,20 @@ public class BlueGreenDeploymentService {
     }
 
     private UpdateControl<FlinkBlueGreenDeployment> deleteDeployment(
-            FlinkDeployment currentDeployment, BlueGreenContext context) {
+            FlinkDeployment currentDeployment,
+            BlueGreenContext context,
+            FlinkBlueGreenDeploymentState nextState) {
 
         boolean deleted = deleteFlinkDeployment(currentDeployment, context);
 
         if (!deleted) {
             LOG.info("FlinkDeployment '{}' not deleted, will retry", currentDeployment);
+            return UpdateControl.<FlinkBlueGreenDeployment>noUpdate()
+                    .rescheduleAfter(RETRY_DELAY_MS);
         } else {
             LOG.info("FlinkDeployment '{}' deleted!", currentDeployment);
+            return finalizeBlueGreenDeployment(context, nextState);
         }
-
-        return UpdateControl.<FlinkBlueGreenDeployment>noUpdate().rescheduleAfter(RETRY_DELAY_MS);
     }
 
     // ==================== Abort and Retry Methods ====================
--- a/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentControllerTest.java
+++ b/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentControllerTest.java
@@ -57,11 +57,13 @@ import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
 
 import java.io.IOException;
+import java.time.Duration;
 import java.time.Instant;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Stream;
 
 import static org.apache.flink.kubernetes.operator.api.spec.FlinkBlueGreenDeploymentConfigOptions.ABORT_GRACE_PERIOD;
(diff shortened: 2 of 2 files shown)


drossos (approved):
Great fix for the BG bug we found.

CC: @gyfora & @schongloo for 👀 and FYI

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentControllerTest.java
```diff
@@ -558,6 +558,69 @@ public void verifyFailureDuringTransition(FlinkVersion flinkVersion) throws Exce
...
+                FlinkBlueGreenDeploymentState.TRANSITIONING_TO_GREEN,
+                rs.reconciledStatus.getBlueGreenState());
+        assertEquals(2, getFlinkDeployments().size());
+
+        // Green becomes ready -> marks the ready timestamp and schedules deletion of Blue
+        simulateSuccessfulJobStart(getFlinkDeployments().get(1));
+        rs = reconcile(rs.deployment);
+
+        // Let the (zero) deletion delay elapse
+        Thread.sleep(10);
```
Dennis-Mircea: It's best to avoid using `sleep` here. Consider using an `await()` function instead.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentControllerTest.java
```diff
@@ -558,6 +558,69 @@ public void verifyFailureDuringTransition(FlinkVersion flinkVersion) throws Exce
         testTransitionToGreen(rs, customValue, null);
     }
 
+    @ParameterizedTest
+    @MethodSource("org.apache.flink.kubernetes.operator.TestUtils#flinkVersions")
+    public void verifyGreenNotAbortedWhenNotReadyAfterBlueDeleted(FlinkVersion flinkVersion)
```
Dennis-Mircea, line 566: The test asserts state/abort-timer/count but not the ingress flip to green, which is part of the cutover contract. I'd say it is worth adding.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/bluegreen/BlueGreenDeploymentService.java
```diff
@@ -627,17 +627,20 @@ private UpdateControl<FlinkBlueGreenDeployment> waitBeforeDeleting(
...
 
         boolean deleted = deleteFlinkDeployment(currentDeployment, context);
 
         if (!deleted) {
             LOG.info("FlinkDeployment '{}' not deleted, will retry", currentDeployment);
+            return UpdateControl.<FlinkBlueGreenDeployment>noUpdate()
+                    .rescheduleAfter(RETRY_DELAY_MS);
         } else {
             LOG.info("FlinkDeployment '{}' deleted!", currentDeployment);
+            return finalizeBlueGreenDeployment(context, nextState);
```
Dennis-Mircea, line 642: One concurrency gap to consider: The delete (child) and the finalize status patch (parent) are not atomic. If the reconcile crashes or the patch 409s after the delete is accepted but before finalize persists, recovery lands in Blue-gone + still `TRANSITIONING_TO_*` + abort timer set. The dispatch only checks Green readiness, not whether Blue still exists ([BlueGreenDeploymentService.java#L489](https://github.com/apache/flink-kubernetes-operator/blob/main/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/bluegreen/BlueGreenDeploymentService.java#L489)).

That still allows a rollback to an already-deleted Blue, just far more rarely than before. I'd close it structurally rather than rely on timing: once the previous deployment is absent during `TRANSITIONING_TO_*`, treat that as the point of no return and always finalize forward, never abort.

Dennis-Mircea (commented):
Nice change, and it's more than the description says: finalizing in the same reconcile that deletes Blue doesn't just save a loop, it closes a real correctness window. In the old flow, after Blue was deleted the transition stayed in `TRANSITIONING_TO_GREEN` with `abortTimestamp` set, so a later reconcile that saw a not-ready Green blip past the abort deadline would `abortDeployment` and roll back to an already-deleted Blue.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentControllerTest.java
fqshopify: Fixed

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentControllerTest.java
fqshopify: Fixed

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/bluegreen/BlueGreenDeploymentService.java
fqshopify: > The delete (child) and the finalize status patch (parent) are not atomic.

Very true. I like your suggestion: once the previous deployment is gone mid-transition, treat that as the point of no return and always finalize forward instead of aborting. 

I think that's a significant enough change to warrant it's own PR though so let's handle it separately after this one: https://github.com/apache/flink-kubernetes-operator/pull/1161

Dennis-Mircea (approved):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/bluegreen/BlueGreenDeploymentService.java
Dennis-Mircea: Sounds good!

PR #1159, opened 2026-07-10
[FLINK-40119] Reset deploymentReadyTimestamp on abort to prevent premature blue delete on next deploy

fqshopify:

When a Blue/Green transition is aborted, `abortDeployment` rolled the state back to `ACTIVE_BLUE` but left `deploymentReadyTimestamp` set to the value from the aborted transition (only the finalize paths reset it). On the **next** deployment, the delete-Blue gate in `shouldWeDelete` computes `deletionTimestamp = staleReadyTimestamp + deletionDelay`, which is already in the past, so Blue is deleted the instant Green first reports `RUNNING` - skipping the deletion delay entirely and causing downtime.

This pull request makes `abortDeployment` reset the transition markers the same way the finalize paths do, so a redeploy after an abort re-arms the delete-Blue deletion delay from scratch.

## Brief change log

- `abortDeployment` now resets `deploymentReadyTimestamp` to 0 on abort, so the next transition re-arms the delete-Blue delay instead of treating it as long expired.
- `abortDeployment` now also resets `abortTimestamp` to 0, making the abort teardown symmetric with `finalizeBlueGreenDeployment` / `finalizeSuspendedDeployment` (which already reset `deploymentReadyTimestamp`, `abortTimestamp`, and `savepointTriggerId`).
- Added a regression test exercising the full abort => redeploy lifecycle in `SAVEPOINT` upgrade mode.

## Verifying this change

This change added tests and can be verified using `mvn clean verify`

- Core observer or reconciler logic that is regularly executed: **yes** (the Blue/Green abort path in the reconciler)

## Documentation

- If yes, how is the feature documented? **not applicable** (bug fix)

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/bluegreen/BlueGreenDeploymentService.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/bluegreen/BlueGreenDeploymentService.java
@@ -689,6 +689,8 @@ public class BlueGreenDeploymentService {
                 getPreviousState(nextState, context.getDeployments());
         context.getDeploymentStatus().setBlueGreenState(previousState);
         context.getDeploymentStatus().setSavepointTriggerId(null);
+        context.getDeploymentStatus().setDeploymentReadyTimestamp(millisToInstantStr(0));
+        context.getDeploymentStatus().setAbortTimestamp(millisToInstantStr(0));
 
         var error =
                 String.format(
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/bluegreen/BlueGreenDeploymentService.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/bluegreen/BlueGreenDeploymentService.java
@@ -24,6 +24,7 @@ import org.apache.flink.kubernetes.operator.api.bluegreen.BlueGreenDeploymentTyp
 import org.apache.flink.kubernetes.operator.api.bluegreen.BlueGreenDiffType;
 import org.apache.flink.kubernetes.operator.api.lifecycle.ResourceLifecycleState;
 import org.apache.flink.kubernetes.operator.api.status.FlinkBlueGreenDeploymentState;
+import org.apache.flink.kubernetes.operator.api.status.FlinkBlueGreenDeploymentStatus;
 import org.apache.flink.kubernetes.operator.api.status.Savepoint;
 import org.apache.flink.kubernetes.operator.api.status.SavepointFormatType;
 import org.apache.flink.kubernetes.operator.api.status.SnapshotTriggerType;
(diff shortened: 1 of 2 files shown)


flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentControllerTest.java
```diff
@@ -558,6 +558,198 @@ public void verifyFailureDuringTransition(FlinkVersion flinkVersion) throws Exce
...
+                        TEST_DEPLOYMENT_NAME,
+                        TEST_NAMESPACE,
+                        flinkVersion,
+                        null,
+                        UpgradeMode.SAVEPOINT);
+
+        // Abort grace long enough to let Green be marked ready before any abort fires;
+        // a non-trivial deletion delay so the delete-Blue wait is observable.
+        var abortGracePeriodMs = 1500;
+        var deletionDelayMs = 1000;
```
Dennis-Mircea, line 657: Consider using `await()` instead of these long sleep intervals.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentControllerTest.java
drossos: I believe this breaks now because it is still assuming the previous behaviour where after abort timestamp won't be zero. Should probably do:

```suggestion
        assertEquals(0, instantStrToMillis(rs.reconciledStatus.getAbortTimestamp()));
```

That allows the test to pass with this new timer reset

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/bluegreen/BlueGreenDeploymentService.java
```diff
@@ -686,6 +686,8 @@ private UpdateControl<FlinkBlueGreenDeployment> abortDeployment(
                 getPreviousState(nextState, context.getDeployments());
         context.getDeploymentStatus().setBlueGreenState(previousState);
         context.getDeploymentStatus().setSavepointTriggerId(null);
+        context.getDeploymentStatus().setDeploymentReadyTimestamp(millisToInstantStr(0));
+        context.getDeploymentStatus().setAbortTimestamp(millisToInstantStr(0));
```
Dennis-Mircea: These three resets (+ `savepointTriggerId` above) are now duplicated in `finalizeBlueGreenDeployment`, `finalizeSuspendedDeployment`, and here. Could you extract a `resetTransitionMarkers(status)` helper and call it from all three, so they can't diverge again?

drossos (commented):
One comment about covering existing test case

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentControllerTest.java
fqshopify: Fixed

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/controller/FlinkBlueGreenDeploymentControllerTest.java
fqshopify: Fixed

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/bluegreen/BlueGreenDeploymentService.java
fqshopify: Fixed

drossos (approved):
Great stuff 👍 

CC: @gyfora / @schongloo as FYI on other blue-green fix

PR #1163, opened 2026-07-20
[FLINK-40195] helm: use static webhook certificate Secret name

applike-ss:
## Summary

Use the stable `flink-operator-webhook-server-cert` Secret name for the webhook certificate.

This avoids changing the Secret name during Helm upgrades and intentionally keeps the certificate name non-configurable.

## Changes

- Use `flink-operator-webhook-server-cert` directly in the cert-manager Certificate and controller Deployment.
- Remove the `webhook.certSecretName` value and helper template.
- Update English and Chinese documentation plus the deployment overview diagram.
- Add Helm unit coverage for the static name in both resources.

## Validation

- Helm unittest: certificate suite 5/5 tests passed.
- Helm unittest: deployment suite 25/25 tests passed.
- `helm lint helm/flink-kubernetes-operator --strict` passed.
- Render verification confirmed both resources keep the static name even when an obsolete `webhook.certSecretName` value is supplied.


helm/flink-kubernetes-operator/templates/_helpers.tpl
```diff
@@ -89,3 +89,10 @@ Supported values for .Values.logging.framework: "log4j2" (default), "logback".
 -Dlog4j.configurationFile=/opt/flink/conf/log4j-operator.properties
 {{- end -}}
 {{- end }}
+
+{{/*
+Create the name of the webhook certificate secret.
+*/}}
+{{- define "flink-operator.webhookCertSecretName" -}}
+{{- default (printf "%s-webhook-server-cert" .Release.Name) .Values.webhook.certSecretName }}
```
Dennis-Mircea: The default secret name changes from `webhook-server-cert` to `<release-name>-webhook-server-cert`. On `helm upgrade` of an existing install, cert-manager provisions a new secret under the new name and the deployment rolls to it, so it self-heals, but the old `webhook-server-cert` secret is left orphaned, and anything external that referenced the old name breaks silently.

Still, I do not see a valid reason to make this certificate configurable at this extent. I'd say using `flink-operator-webhook-server-cert` as the certificate name is sufficient to uniquely identify it within the namespace and to avoid any collision.

Dennis-Mircea (changes requested):
Thanks for opening the PR! The idea of this PR is not bad, as `webhook-server-cert` isn't that unique name. Still, if there isn't a mandatory requirement for having the cert manager using a dedicated secret name, I'd opt in for `flink-operator-webhook-server-cert` name.

Either way, please make sure to update the documentation accordingly, as the `webhook-server-cert` is currently referenced across multiple documentation pages.

helm/flink-kubernetes-operator/templates/_helpers.tpl
applike-ss: Updated and pushed the requested changes in a942d42e: the webhook certificate Secret is now statically named `flink-operator-webhook-server-cert`, the configurability and helper were removed, all documentation and diagram references were updated, and the diagram alignment was corrected. Could you please re-review the updated changes when you have a chance?

PR #1166, opened 2026-07-23
[FLINK-40162] Handle UpgradeFailureException in FlinkSessionJobController

milindl:

The state of the FlinkSessionJob (`status.jobStatus.state` and `status.state`) is stuck and not updated in case of an UpgradeFailureException (for instance, if the job ends quickly and the checkpointing interval is set, but state.checkpoints.dir is not configured).

## Brief change log

 - Moves  observer.observe() inside try-catch and add explicit catch for UpgradeFailureException, similar to FlinkDeploymentController.
 - `triggerErrorEvent` now accepts a reason similar to FlinkDeploymentController

## Verifying this change

 - testJobFinishedWithNonExternallyAddressableCheckpoint: Submits a session job, simulates it finishing with a checkpoint whose external pointer is   non-externally-addressable (checkpointing enabled but no state.checkpoints.dir), and verifies the job status correctly reflects FINISHED with the error recorded.
- Manual testing using minikube using the exact JIRA reproducer in the ticket (SleepJob with 5s delay, execution.checkpointing.interval: 2000, no state.checkpoints.dir):
  - FKO 1.15.0: job stuck at RUNNING/STABLE despite Flink REST API showing FINISHED
  - Patched FKO: job correctly transitions to FINISHED/STABLE with error recorded

  - Core observer or reconciler logic that is regularly executed: no (I think)
 
## Documentation


Dennis-Mircea (approved):
Thanks for opening this PR! The change looks good and it replicates well the same `FlinkDeploymentController` reconcile behavior.

PR #1168, opened 2026-07-30
[FLINK-40164] Complete suspending when the cancellation is observed

milindl:

Suspending a job through an asynchronous cancellation left the resource reporting the UPGRADING lifecycle state forever. An example that reproduces this exists in the ticket. 

I experienced it with FlinkSessionJobs (stateless or last-state upgrade mode, no other special settings), but even FlinkDeployments in application mode will suffer this (if `kubernetes.operator.job.upgrade.last-state.job-cancel.enabled` and `upgradeMode` is `last-state` for example).

Savepoint upgrades complete synchronously and are not affected.

## Brief change log

  - In the observer, after marking resource suspended, we check if a user has set the desired state of the job to SUSPENDED and whether it's UPGRADING, and if so, mark the spec as stable and set `ReconciliationState.DEPLOYED`. 
 
There is one related issue I did not solve: onTargetJobNotFound also calls markSuspended without completing the upgrade, so a stateless session job whose Flink job vanishes from the cluster mid-cancel (e.g. JM failover) will be stuck in UPGRADING.

## Verifying this change

  - Added unit tests for the observer.
  - Added tests to FlinkSessionJobControllerTest and FlinkDeploymentControllerTest. 
  - Manual verification by suspending a session job and observing the output of `kubectl get flinksessionjobs`.

  - Core observer or reconciler logic that is regularly executed: I'd say so yes for session jobs

## Documentation

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
@@ -350,6 +350,7 @@ public class JobStatusObserver<R extends AbstractFlinkResource<?, ?>> {
                             && JobStatus.CANCELLING.equals(previousJobStatus))) {
                 // The job was cancelled
                 markSuspended(resource);
+                finalizeSuspendedUpgrade(resource);
             }
 
             recordJobErrorIfPresent(ctx, clusterJobStatus);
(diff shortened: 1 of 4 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
```diff
@@ -373,6 +374,31 @@ private static void markSuspended(AbstractFlinkResource<?, ?> resource) {
...
+     * the suspended job state that this completion makes stable.
+     *
+     * @param resource The Flink resource whose cancellation was just observed.
+     */
+    private static void finalizeSuspendedUpgrade(AbstractFlinkResource<?, ?> resource) {
+        var reconciliationStatus = resource.getStatus().getReconciliationStatus();
+        var job = resource.getSpec().getJob();
+        if (reconciliationStatus.getState() != ReconciliationState.UPGRADING
+                || job == null
+                || job.getState() != JobState.SUSPENDED) {
```
gyfora: Can you explain why we have these job conditions here? Are they actually needed?

milindl: The first one is not needed, we would not get here with job == null. I will remove it.

The second one (covered by `testUpgradeNotCompletedWhenTargetStateStillRunning`) is needed if the situation is a change of the spec which is not suspension, something like a parallelism change.
It eventually works, but some of the state transitions to get there are not correct (the session job is marked suspended for a while). Which I thought we should avoid because they're changing the publicly visible behaviour.

gyfora (approved):

milindl:
Thanks for the approval @gyfora , could you merge it, too, as the tests look green?

PR #1169, opened 2026-08-05
[FLINK-39356] Fix NPE during FlinkStateSnapshot cleanup when status i…

nihar4276:
Problem:

A FlinkStateSnapshot deleted before the controller ever reconciled it has no status, since the status is only initialized in reconcile(). The cleanup path builds a FlinkStateSnapshotContext, which dereferences the status via getStatus().toBuilder(), so cleanup() threw a NullPointerException before entering its try block. The finalizer was never removed, leaving the resource stuck in terminating state and blocking namespace deletion, with the failed cleanup retried indefinitely.

Fix:

Return DeleteControl.defaultDelete() from cleanup() when the status is null. Such a snapshot was never triggered against Flink, so there is nothing to dispose of and the finalizer can be removed immediately.

Verification:

Added FlinkStateSnapshotControllerTest#testCleanupWithNullStatus, covering both a savepoint and a checkpoint whose status is null.

Verified manually on a local minikube cluster: a FlinkStateSnapshot deleted before the operator reconciles it now has its finalizer removed and the resource disappears, instead of hanging in terminating state.

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkStateSnapshotController.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkStateSnapshotController.java
@@ -89,6 +89,12 @@ public class FlinkStateSnapshotController
     @Override
     public DeleteControl cleanup(
             FlinkStateSnapshot flinkStateSnapshot, Context<FlinkStateSnapshot> josdkContext) {
+        if (flinkStateSnapshot.getStatus() == null) {
+            LOG.info(
+                    "Snapshot {} has no status, was never reconciled. Removing finalizer.",
+                    flinkStateSnapshot.getMetadata().getName());
+            return DeleteControl.defaultDelete();
+        }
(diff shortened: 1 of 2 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/controller/FlinkStateSnapshotController.java
```diff
@@ -89,6 +89,12 @@ public UpdateControl<FlinkStateSnapshot> reconcile(
     @Override
     public DeleteControl cleanup(
             FlinkStateSnapshot flinkStateSnapshot, Context<FlinkStateSnapshot> josdkContext) {
+        if (flinkStateSnapshot.getStatus() == null) {
+            LOG.info(
+                    "Snapshot {} has no status, was never reconciled. Removing finalizer.",
+                    flinkStateSnapshot.getMetadata().getName());
```
gyfora: I don't think we need this log, simply return delete

nihar4276: done @gyfora, removed it

gyfora (approved):

PR #1170, opened 2026-08-05
[FLINK-40246] Migrate logging stack to SLF4J 2.x and bump Logback to 1.5.34

spuru9:

`logback-core` 1.2.13 is the terminal release of an EOL line, so the CVEs filed against it will never be fixed there. This upgrades it to 1.5.34.

Logback and SLF4J have to move together: 1.2.13 binds only to SLF4J 1.7.x (`StaticLoggerBinder`), 1.5.x only to SLF4J 2.x (`ServiceLoader`). Bumping either alone leaves SLF4J unbound, and it falls back to a no-op logger instead of failing loudly. The Log4j2 binding becomes `log4j-slf4j2-impl` for the same reason.

`slf4j-api` 2.0.18 matches what JOSDK 5.5.0 and fabric8 7.8.0 already declare. Flink is unaffected: its jars only call the SLF4J API, which stays binary compatible for callers across 1.7 and 2.x.

### CVEs addressed

All six are against `ch.qos.logback:logback-core`.

| CVE | GHSA | Severity | Vulnerable range | First fixed |
| --- | --- | --- | --- | --- |
| CVE-2024-12798 | GHSA-pr98-23f8-jwxv | Medium | `< 1.3.15` (and `1.4.0`–`1.5.13`) | 1.3.15 / 1.5.13 |
| CVE-2024-12801 | GHSA-6v67-2wr5-gvf4 | Low | `< 1.3.15` (and `1.4.0`–`1.5.13`) | 1.3.15 / 1.5.13 |
| CVE-2025-11226 | GHSA-25qh-j22f-pwp8 | Medium | `< 1.3.16` (and `1.4.0`–`1.5.19`) | 1.3.16 / 1.5.19 |
| CVE-2026-1225 | GHSA-qqpg-mvqg-649v | Low | `< 1.5.25` | 1.5.25 |
| CVE-2026-9828 | GHSA-p47f-322f-whfh | Low | `<= 1.5.32` | 1.5.33 |
| CVE-2026-10532 | GHSA-jhq6-gfmj-v8fx | Low | `< 1.5.34` | 1.5.34 |

The JIRA proposes 1.5.25 as the target. CVE-2026-10532 raises the floor to 1.5.34.

## Brief change log

  - `slf4j-api` 1.7.36 -> 2.0.18, `logback-classic` / `logback-core` 1.2.13 -> 1.5.34
  - `log4j-slf4j-impl` -> `log4j-slf4j2-impl` across all modules, including the `maven-dependency-plugin` artifact item that stages the binding into the Docker image
  - `NOTICE` files: refreshed `slf4j-api` versions, and dropped `log4j-slf4j-impl` from the operator's since it became `provided` in FLINK-39501 and is no longer bundled
  - Removed the "Logging Library Version Overrides" hint from `docs/content{,.zh}/docs/operations/logging.md`: swapping the shipped logging jars is not a supported user operation

## Verifying this change

Covered by existing tests; the full suite passes locally (2216 in `flink-kubernetes-operator`, 105 in `flink-kubernetes-webhook`, plus the remaining modules).

The real gate is `e2e-tests/test_logback_logging.sh`, which asserts no duplicate-binding warning across Flink 1.20/2.0/2.1/2.2 in both native and standalone mode. It runs in the `e2e_logging` CI job. Draft until that has passed.

  - Dependencies (does it add or upgrade a dependency): **yes** - `slf4j-api`, `logback-classic`, `logback-core`, and `log4j-slf4j-impl` is replaced by `log4j-slf4j2-impl`

## Documentation

  - If yes, how is the feature documented? not applicable - existing logging documentation updated

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/content.zh/docs/operations/logging.md
+++ b/docs/content.zh/docs/operations/logging.md
@@ -115,11 +115,7 @@ Logback XML overrides replace the entire default configuration. Unlike Log4j2 `.
 
 ## Logging Library Version Overrides
 
-The operator ships with Logback 1.2.x and SLF4J 1.7.x. These versions are bundled in the Docker image and the SLF4J 1.7.x API is shaded into the operator JAR.
-
-{{< hint warning >}}
-Upgrading to Logback 1.4+/1.5+ or SLF4J 2.x is not supported. SLF4J 2.x uses a `ServiceLoader`-based binding mechanism that is incompatible with the SLF4J 1.7.x API shaded inside the operator. Replacing the JARs at runtime will result in `ClassNotFoundException: org.slf4j.impl.StaticLoggerBinder`.
-{{< /hint >}}
+The operator ships with Logback 1.5.x and SLF4J 2.0.x. These versions are bundled in the Docker image and the SLF4J 2.0.x API is shaded into the operator JAR.
 
 ## FlinkDeployment Logging Configuration
 
--- a/flink-kubernetes-operator/pom.xml
+++ b/flink-kubernetes-operator/pom.xml
@@ -123,8 +123,9 @@ under the License.
         </dependency>
 
         <!-- Logging: SLF4J API and Log4j2 implementation are shaded into the operator JAR.
-             Only the SLF4J bindings (log4j-slf4j-impl / logback-classic) are excluded
-             and shipped as separate JARs so the entrypoint can pick one. -->
+             Only the SLF4J bindings (log4j-slf4j2-impl / logback-classic + logback-core)
+             are excluded and shipped as separate JARs so the entrypoint can pick one.
+             SLF4J 2.x is required: JOSDK and the fabric8 client are compiled against it. -->
 
         <dependency>
             <groupId>org.slf4j</groupId>
@@ -134,7 +135,7 @@ under the License.
 
         <dependency>
             <groupId>org.apache.logging.log4j</groupId>
-            <artifactId>log4j-slf4j-impl</artifactId>
+            <artifactId>log4j-slf4j2-impl</artifactId>
             <version>${log4j.version}</version>
             <scope>provided</scope>
         </dependency>
@@ -318,7 +319,7 @@ under the License.
                         <!-- Log4j2 SLF4J binding (default logging framework) -->
                         <artifactItem>
                             <groupId>org.apache.logging.log4j</groupId>
-                            <artifactId>log4j-slf4j-impl</artifactId>
+                            <artifactId>log4j-slf4j2-impl</artifactId>
                             <version>${log4j.version}</version>
                             <outputDirectory>${project.build.directory}/log4j</outputDirectory>
                         </artifactItem>
--- a/flink-kubernetes-operator/pom.xml
+++ b/flink-kubernetes-operator/pom.xml
@@ -125,7 +125,11 @@ under the License.
         <!-- Logging: SLF4J API and Log4j2 implementation are shaded into the operator JAR.
              Only the SLF4J bindings (log4j-slf4j2-impl / logback-classic + logback-core)
              are excluded and shipped as separate JARs so the entrypoint can pick one.
-             SLF4J 2.x is required: JOSDK and the fabric8 client are compiled against it. -->
+             SLF4J 2.x is required: JOSDK and the fabric8 client are compiled against it.
+
+             The binding artifacts are declared provided purely so their versions are
+             resolved and recorded here; the maven-dependency-plugin below stages them
+             into the image from its own artifactItems, independently of this graph. -->
 
         <dependency>
             <groupId>org.slf4j</groupId>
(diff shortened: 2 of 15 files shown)


spuru9:
The migration only covers the operator module. operator-api and standalone pick up slf4j-api 1.7.36 transitively from Flink (flink-annotations and flink-kubernetes), but I've switched them to the 2.x  binding, so both now run on a NOP logger with all output silently discarded. No error, nothing in the test output. operator-api is published, so downstream consumers who don't pin slf4j would inherit it too.
                                                                                                                                                                                                                 
Pinning slf4j-api to fix that has a knock-on: autoscaler and plugin-jdbc have no binding of their own beyond the log4j-slf4j-impl 2.24.3 that flink-test-utils-junit drags in, so they'd go NOP as well. The operator module pulls the same old binding through flink-test-utils (no -junit).                                                                                                                               
  
Two options:
1. Global pin — what I've already got working. slf4j-api in root dependencyManagement, plus excluding the old binding from flink-test-utils / -junit in three modules and adding the 2.x binding at test scope in two. 
2. Update the required dependency in 1.20.x and then use those after the next release - that could be a major bump for a 1.x flink which we don't alter much and is just maintained.

What is you suggestion here.

spuru9:
cc: @gyfora @Dennis-Mircea

flink-kubernetes-operator/pom.xml
```diff
@@ -164,6 +165,13 @@ under the License.
             <scope>provided</scope>
         </dependency>
 
+        <dependency>
+            <groupId>ch.qos.logback</groupId>
+            <artifactId>logback-core</artifactId>
```
Dennis-Mircea: Is this actually needed, or is it redundant with the `maven-dependency-plugin` artifactItem?

docs/content.zh/docs/operations/logging.md
```diff
@@ -115,10 +115,10 @@ Logback XML overrides replace the entire default configuration. Unlike Log4j2 `.
 
 ## Logging Library Version Overrides
 
-The operator ships with Logback 1.2.x and SLF4J 1.7.x. These versions are bundled in the Docker image and the SLF4J 1.7.x API is shaded into the operator JAR.
+The operator ships with Logback 1.5.x and SLF4J 2.0.x. These versions are bundled in the Docker image and the SLF4J 2.0.x API is shaded into the operator JAR.
 
 {{< hint warning >}}
-Upgrading to Logback 1.4+/1.5+ or SLF4J 2.x is not supported. SLF4J 2.x uses a `ServiceLoader`-based binding mechanism that is incompatible with the SLF4J 1.7.x API shaded inside the operator. Replacing the JARs at runtime will result in `ClassNotFoundException: org.slf4j.impl.StaticLoggerBinder`.
+Downgrading to Logback 1.2.x or SLF4J 1.7.x is not supported. SLF4J 2.x locates the backend through a `ServiceLoader` (`org.slf4j.spi.SLF4JServiceProvider`), whereas Logback 1.2.x provides the older `org.slf4j.impl.StaticLoggerBinder`, which SLF4J 2.x ignores. Replacing the JARs at runtime does not raise an error: SLF4J falls back to a no-operation logger and all log output is silently discarded.
 {{< /hint >}}
```
Dennis-Mircea: I'd remove entirely this hint, as the logging library downgrade should not be something performed by the users.

flink-kubernetes-operator/pom.xml
spuru9: Removed.

docs/content.zh/docs/operations/logging.md
spuru9: Done

Dennis-Mircea: You're right about that this section is now just one sentence. After thinking a bit more on this I'd say that having the following content as part of this section will be the right call:
```
The operator's image bundles the SLF4J API (shaded into the operator JAR) and both logging frameworks: Log4j2 2.x by default, and Logback 1.5.x when `logging.framework=logback` is set. The bundled SLF4J API is 2.0.x.

These versions can be changed by rebuilding the image, for example to adopt a newer patch release before the operator does.
  
{{< hint warning >}}
Because the bundled SLF4J API is 2.x, a replacement framework must provide an SLF4J 2.x binding. A release
built for the older SLF4J 1.7 API will not bind, so the operator runs normally but produces no log output at all. Only Log4j2 2.19+ and Logback 1.3+ are compatible.
{{< /hint >}}
```

This new content includes the Log4j2 as well and let the users aware of both versioning management.

rmetzger:
Flink relies on slf4j 1x (I suspect). And afaik FKO has a dependency on Flink. Are we sure that any Flink code trying to log (when called from FKO) is still able to do so?

spuru9:
CI failure looks unrelated should be resolved on retry. 
```
[FATAL] Non-resolvable parent POM for org.apache.flink:flink-kubernetes-operator-parent:1.16-SNAPSHOT:
Could not transfer artifact org.apache:apache:pom:23 from/to central
(https://repo.maven.apache.org/maven2): authorization failed ... status: 403 Forbidden
```

rmetzger:
I manually restarted a failed task , it is now green: https://github.com/apache/flink-kubernetes-operator/actions/runs/31406205575/job/94061578994
but it the overall CI status is hard to assess -- can you just push an empty commit to trigger a new CI run?

spuru9:
> I manually restarted a failed task , it is now green: https://github.com/apache/flink-kubernetes-operator/actions/runs/31406205575/job/94061578994 but it the overall CI status is hard to assess -- can you just push an empty commit to trigger a new CI run?

Done

rmetzger (approved):
Change looks good, guarded by solid test coverage.

Merging.

PR #1173, opened 2026-08-11
[FLINK-40365][docs] Correct savepoint config requirements

dalelane:

Correct the docs to reflect the config requirements for savepoint upgrade mode

## Brief change log

doc update

## Verifying this change

updated docs to match my reading of the code as described in https://issues.apache.org/jira/projects/FLINK/issues/FLINK-40365

## Documentation

---

### DIFF
--- a/docs/content/docs/managing/job-management.md
+++ b/docs/content/docs/managing/job-management.md
@@ -109,7 +109,7 @@ How state is carried across the restart is controlled through `spec.job.upgradeM
 | Production Use         | Not recommended  | Recommended                            | Recommended                                |
 
 {{< hint info >}}
-When HA is enabled the `savepoint` upgrade mode may fall back to the `last-state` behavior in cases where the job is in an unhealthy state. This is why a checkpoint directory is required.
+When HA is enabled the `savepoint` upgrade mode may fall back to the `last-state` behavior in cases where the job is in an unhealthy state, restoring from the checkpoint directory. That fallback can be disabled with `kubernetes.operator.job.upgrade.last-state-fallback.enabled`.
 {{< /hint >}}
 
 With `FlinkStateSnapshot` resources enabled, the savepoint taken when suspending in `savepoint` mode is tracked as its own snapshot resource, as described under [Snapshot Management → Upgrade Triggering]({{< ref "docs/managing/snapshot-management#upgrade-triggering" >}}). A `last-state` upgrade creates no snapshot resource, it takes no new snapshot in the first place, reusing the latest checkpoint instead.
--- a/docs/content/docs/managing/job-management.md
+++ b/docs/content/docs/managing/job-management.md
@@ -100,16 +100,16 @@ When a referenced Kubernetes resource changes, the running job is not restarted 
 
 How state is carried across the restart is controlled through `spec.job.upgradeMode`, with three supported values: `stateless`, `last-state`, and `savepoint`. The setting controls both the stop and the restore mechanism, as detailed in the following table:
 
-|                        | Stateless        | Last State                             | Savepoint                                 |
-|------------------------|------------------|----------------------------------------|-------------------------------------------|
-| Config Requirement     | None             | Checkpointing Enabled                  | Checkpoint or savepoint directory defined |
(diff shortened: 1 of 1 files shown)


gyfora:
technically speaking last-state fallback can be disabled , in which case really only savepoint directory would be needed :)

dalelane:
@gyfora apologies if I've misread this and am wasting your time with a false alarm 

I was solely going off this bit: https://github.com/apache/flink-kubernetes-operator/blob/fdfcc695daba16e4f4a8c9a6c08b7f470787de00/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultValidator.java#L257-L263 

I thought this meant it'd reject a savepoint job with a null or whitespace checkpoints directory. Could you let me know what I missed please?

docs/content/docs/managing/job-management.md
```diff
@@ -100,16 +100,16 @@ When a referenced Kubernetes resource changes, the running job is not restarted
...
+|------------------------|------------------|----------------------------------------|--------------------------------------------|
+| Config Requirement     | None             | Checkpointing Enabled                  | Checkpoint and savepoint directory defined |
+| Job Status Requirement | None             | Job or HA metadata accessible          | Job Running                                |
+| Suspend Mechanism      | Cancel or delete | Cancel or delete (keeping HA metadata) | Cancel with savepoint                      |
+| Restore Mechanism      | Empty state      | Use HA metadata or the latest snapshot | Restore from savepoint                     |
+| Production Use         | Not recommended  | Recommended                            | Recommended                                |
 
 {{< hint info >}}
-When HA is enabled the `savepoint` upgrade mode may fall back to the `last-state` behavior in cases where the job is in an unhealthy state.
+When HA is enabled the `savepoint` upgrade mode may fall back to the `last-state` behavior in cases where the job is in an unhealthy state. This is why a checkpoint directory is required.
```
Dennis-Mircea: This new added sentence is not accurate and is exactly what @gyfora was referring to. The checkpoint directory is required for every non-stateless upgrade mode. I'd propose here better framing like the one below:
```suggestion
When HA is enabled the `savepoint` upgrade mode may fall back to the `last-state` behavior in cases where the job is in an unhealthy state, restoring from the checkpoint directory. That fallback can be disabled with `kubernetes.operator.job.upgrade.last-state-fallback.enabled`.
```

Dennis-Mircea (commented):
The table fix looks right to me, and only the hint addition should be a bit refined.

docs/content/docs/managing/job-management.md
dalelane: got it - that makes sense, thank you

Dennis-Mircea:
> @gyfora apologies if I've misread this and am wasting your time with a false alarm
> 
> I was solely going off this bit:
> 
> https://github.com/apache/flink-kubernetes-operator/blob/fdfcc695daba16e4f4a8c9a6c08b7f470787de00/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/validation/DefaultValidator.java#L257-L263
> 
> I thought this meant it'd reject a savepoint job with a null or whitespace checkpoints directory. Could you let me know what I missed please?

@dalelane this is a good finding, you didn't misread it. That check applies to any non-stateless upgrade mode, so a savepoint job without a checkpoints directory is rejected. The `or` in the table is wrong as things stand.

@gyfora is describing what is conceptually necessary, and with `kubernetes.operator.job.upgrade.last-state-fallback.enabled` set to false only the savepoint directory would be. The validator does not consult that flag, so in practice both are still required.

I'd say it is worth considering gating the checkpoint directory check on `kubernetes.operator.job.upgrade.last-state-fallback.enabled`, so savepoint mode does not demand config it will never use. The catch is that the flag is dynamic while validation happens once at admission, so a job validated with the fallback off would be left in an invalid state if it is turned back on later.@gyfora what do you think?

dalelane:
> I'd say it is worth considering gating the checkpoint directory check on `kubernetes.operator.job.upgrade.last-state-fallback.enabled`, so savepoint mode does not demand config it will never use

Makes sense - I'd be happy to take a run at that if it'd help to see what a fix like that would look like

PR #1175, opened 2026-08-12
[FLINK-40372][metrics] Unregister JOSDK resource metrics after cleanup

litiliu:

This pull request fixes [FLINK-40372](https://issues.apache.org/jira/browse/FLINK-40372).

`OperatorJosdkMetrics` caches resource-level metric groups and counters by `ResourceID`, but `cleanupDone` previously only incremented a cleanup counter. Metrics for deleted resources therefore remained registered and strongly referenced for the lifetime of the operator. Workloads that continuously create uniquely named resources, such as periodic `FlinkStateSnapshot` resources, could grow resource metric cardinality without bound.

This change removes the resource-specific metric caches and closes the resource metric group after JOSDK reports cleanup, unregistering the deleted resource's counters from the Flink metric registry.

## Brief change log

- Remove resource and resource-namespace metric group cache entries when cleanup completes.
- Remove cached counters belonging to the deleted resource scope.
- Close the resource metric group to unregister its counters.
- Extend the metric test listener and add coverage for cleanup, repeated cleanup, and same-name resource recreation.

## Verifying this change

- `OperatorJosdkMetricsTest`: 3 tests passed, covering cleanup, repeated cleanup, and metric re-registration for a recreated resource.
- Broader metrics test set: 41 tests passed across 8 metrics test classes.
- Spotless and Checkstyle checks passed.
- `git diff --check` passed.

A full `mvn clean verify` has not been run locally.

## Documentation

---

- [X] Yes (OpenAI Codex)

Generated-by: OpenAI Codex (GPT-5) Reviewed by Human

### DIFF
--- a/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/metrics/OperatorJosdkMetricsTest.java
+++ b/flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/metrics/OperatorJosdkMetricsTest.java
@@ -19,6 +19,7 @@ package org.apache.flink.kubernetes.operator.metrics;
 
 import org.apache.flink.configuration.Configuration;
 import org.apache.flink.kubernetes.operator.api.FlinkDeployment;
+import org.apache.flink.kubernetes.operator.api.FlinkStateSnapshot;
 import org.apache.flink.kubernetes.operator.config.FlinkConfigManager;
 import org.apache.flink.kubernetes.operator.controller.FlinkDeploymentController;
 import org.apache.flink.kubernetes.operator.exception.ReconciliationException;
@@ -36,6 +37,10 @@ import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 import java.util.Map;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.function.Consumer;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.fail;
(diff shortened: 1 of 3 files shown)


litiliu:
Hi @gyfora, could you please approve the two pending GitHub Actions workflows? Thanks!

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/metrics/OperatorJosdkMetricsTest.java
```diff
@@ -156,17 +156,13 @@ public boolean isLastAttempt() {
         assertEquals(1, getCount("Resource.Event"));
         assertEquals(1, getCount("Resource.Event.ADDED"));
 
-        operatorMetrics.cleanupDone(resourceId, metadata);
```
Dennis-Mircea, line 159: This line should not be deleted, and stay as is, because `cleanupDone` and `reconciliationFinished` are interconnected. By having this line deleted it suggests that the new change added as part of `OperatorJosdkMetrics#cleanupDone` is not enough, and after the `reconciliationFinished` the metrics are registered back by the JOSDK, so the PR main idea is discarded completed.

Dennis-Mircea (changes requested):
The direction is legitimate, unregistering per-resource metrics on delete is consistent with what the operator already does in `FlinkResourceContextFactory.cleanup().`

Two things before this can move forward:
1. First, as noted inline, the change does not achieve its goal today. This implementation must be well tested before opening a PR.
2. Second, could you expand the motivation in the JIRA? What cardinality growth was actually observed, and over what period? That helps size how much complexity is worth adding here, because doing this correctly also has to handle the kind collision in the ResourceID cache key, user-configurable scope formats, and concurrent access from the reconciler threads.

flink-kubernetes-operator/src/test/java/org/apache/flink/kubernetes/operator/metrics/OperatorJosdkMetricsTest.java
litiliu: Thanks for pointing this out. I agree that `cleanupDone` and the subsequent terminal callback must be tested together.
I have pushed a new version. 
The updated `testCleanupMetrics` now explicitly calls `cleanupDone(resourceId, metadata)` followed by both `reconciliationFinished(...)` and `reconciliationFailed(...)`, and asserts that the metric registry and resource-metrics cache remain empty. The concurrent tests also cover `cleanupDone` racing with each terminal callback.

I kept `testMetrics` focused on the normal callback path and restored its positive `Reconciliation.finished` assertion. The original `Reconciliation.cleanup` assertion cannot remain unchanged because the new contract is for `cleanupDone` to unregister the entire per-resource group immediately, so that counter no longer remains registered.

litiliu:
Thanks for the detailed review and for pointing these issues out, @Dennis-Mircea. I’ve added the reproduction logs and details of the suspected metric retention and memory leak to the JIRA. I really appreciate your guidance!
I have verified the latest version on our cluster, the heap memory lead didn't happen again. 
Looking forward to your review when you have time!

litiliu:
@gyfora Would you mind taking a look at this PR when you have time? Thanks!

PR #1178, opened 2026-08-12
[FLINK-40380] Document the runtime configuration and stop jobs from overriding operator config through it

Dennis-Mircea:

Follow-up to FLINK-35746, which added runtime configuration observation (reading a job's effective settings from the Flink REST API and layering them over the observed configuration).

Two gaps are addressed. The operator documentation still describes the previous configuration model, to the point of stating the operator cannot observe the running configuration. And the global job parameters taken from the REST response are applied unfiltered, so a job can override operator configuration for its own resource.

## Brief change log

**Documentation** (`docs/content{,.zh}/docs/internals/startup.md`)

- The Configuration section now lists four configurations rather than three, with a Runtime row describing what it is read from (the JobManager configuration, job execution and checkpoint config REST endpoints) and what it is used for.
- The Observe row states that the runtime configuration is layered on top, matching `FlinkResourceContext#getObserveConfig`.
- A short paragraph explains why the runtime configuration exists, since a spec is a request rather than a record of what the job ended up running with, followed by its lifecycle: fetched once per job, skipped for globally terminal jobs, cached per resource and job id, and falling back to spec-derived values while a fetch keeps failing.
- The closing warning previously said "None of these is the running configuration", which FLINK-35746 made untrue. It is replaced rather than removed, since the mapping only covers a subset of settings and only while the job runs, so a `config.yaml` baked into the image is still invisible to the operator.

**Filtering** (`FlinkRuntimeConfigurationUtils`)

- Global job parameters whose key is in the operator's own namespaces (`kubernetes.operator.` and `job.autoscaler.`) are no longer copied into the runtime configuration, and each dropped key is logged at WARN.
- The check uses the existing `KubernetesOperatorConfigOptions.K8S_OP_CONF_PREFIX` and `AutoScalerOptions.AUTOSCALER_CONF_PREFIX` constants so it follows the prefixes if they ever change.

The filter is deliberately limited to those two namespaces. The JobManager configuration endpoint legitimately contributes keys such as `rest.*`, so filtering more broadly by key name would risk dropping genuinely observed settings, whereas operator and autoscaler keys have no reason to arrive from a job's global parameters at all.

## Verifying this change

- `AbstractFlinkServiceTest#testMapJobConfigurationDropsOperatorControlledGlobalParameters` asserts that an operator key and an autoscaler key set as global job parameters are dropped, while an unrelated parameter and the mapped execution fields are kept.
- The existing `testMapJobConfigurationMapsAllExpectedFields` continues to pass, so ordinary parameters are unaffected.
- The full `flink-kubernetes-operator` suite passes (2240 tests).

- Core observer or reconciler logic that is regularly executed: yes, the runtime configuration mapping used by the job status observer

## Documentation

  - If yes, how is the feature documented? the Internals startup documentation is corrected to describe the runtime configuration added in FLINK-35746

---

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkRuntimeConfigurationUtils.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkRuntimeConfigurationUtils.java
@@ -17,11 +17,13 @@
 
 package org.apache.flink.kubernetes.operator.utils;
 
+import org.apache.flink.autoscaler.config.AutoScalerOptions;
 import org.apache.flink.configuration.CheckpointingOptions;
 import org.apache.flink.configuration.CoreOptions;
 import org.apache.flink.configuration.PipelineOptions;
 import org.apache.flink.configuration.StateBackendOptions;
 import org.apache.flink.configuration.StateChangelogOptions;
+import org.apache.flink.kubernetes.operator.config.KubernetesOperatorConfigOptions;
 import org.apache.flink.runtime.rest.messages.JobConfigInfo;
 import org.apache.flink.runtime.rest.messages.checkpoints.CheckpointConfigInfo;
 
@@ -131,6 +133,10 @@ public class FlinkRuntimeConfigurationUtils {
     /**
      * Extract execution configuration (parallelism, object-reuse, global job parameters) from a
      * {@link JobConfigInfo} REST response.
+     *
+     * <p>Global job parameters are set by the job itself and are merged over the observed
+     * configuration, so keys belonging to the operator's own namespaces are dropped. Otherwise a
+     * job could change how the operator manages it simply by declaring a matching global parameter.
      */
     public static Map<String, String> mapJobConfiguration(JobConfigInfo configurationInfo) {
         Map<String, String> jobConfig = new HashMap<>();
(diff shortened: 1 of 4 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkRuntimeConfigurationUtils.java
```diff
@@ -141,10 +147,30 @@ public static Map<String, String> mapJobConfiguration(JobConfigInfo configuratio
         jobConfig.put(
                 CoreOptions.DEFAULT_PARALLELISM.key(), String.valueOf(execInfo.getParallelism()));
         jobConfig.put(PipelineOptions.OBJECT_REUSE.key(), String.valueOf(execInfo.isObjectReuse()));
-        jobConfig.putAll(execInfo.getGlobalJobParameters());
+        execInfo.getGlobalJobParameters()
+                .forEach(
+                        (key, value) -> {
+                            if (isOperatorControlledKey(key)) {
+                                LOG.warn(
+                                        "Ignoring global job parameter '{}': the job cannot override operator configuration.",
```
gyfora: Let's remove this log, it's basically unnecessary and does not communicate anything interesting to the operator. Operator configs are not job configs anyways.

gyfora (commented):
Aside from the too verbose logging, this looks good!

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/FlinkRuntimeConfigurationUtils.java
Dennis-Mircea: I've removed it now.

gyfora (approved):

PR #1180, opened 2026-08-17
[FLINK-40400] Harden job artifact fetching in kubernetes operator

spuru9:

Harden and consolidate validation and fetching of the `FlinkSessionJob` `jarURI` so that
`http(s)` fetches consistently honor the operator's configured scheme and host policy.

## Brief change log

- Share the `jarURI` validation (scheme allowlist + host checks) between resource validation
  and the artifact fetcher via `JarUriValidationUtils`.
- `HttpArtifactFetcher` validates every redirect hop, follows only `http(s)` targets, and
  caps the number of redirects.
- Fetch-time policy is taken from the operator configuration; all resolved host addresses
  are checked.

## Verifying this change

Covered by `ArtifactManagerTest` and `DefaultValidatorTest`; also verified end-to-end on
Kubernetes.

Generated-by: Claude Code

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/artifact/HttpArtifactFetcher.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/artifact/HttpArtifactFetcher.java
@@ -96,53 +96,62 @@ public class HttpArtifactFetcher implements ArtifactFetcher {
             }
             conn.setRequestMethod("GET");
 
-            int status = conn.getResponseCode();
-            if (!isRedirect(status)) {
-                break;
-            }
-
-            String location = conn.getHeaderField("Location");
-            conn.disconnect();
-            if (location == null || location.isEmpty()) {
-                throw new IOException(
-                        "Received redirect (status "
-                                + status
-                                + ") from '"
-                                + currentUri
-                                + "' without a Location header");
-            }
-            if (++redirects > MAX_REDIRECTS) {
-                throw new IOException(
-                        "Too many redirects (>"
-                                + MAX_REDIRECTS
-                                + ") while fetching artifact from '"
-                                + uri
-                                + "'");
-            }
-            URL nextUrl;
+            // Release the connection on every path except the final (non-redirect) one, whose body
+            // is streamed below. This covers getResponseCode() and the redirect handling throwing.
+            boolean keepConnection = false;
             try {
-                nextUrl = new URL(currentUrl, location);
-            } catch (MalformedURLException e) {
-                throw new IOException(
-                        "Refusing to follow redirect from '"
-                                + currentUri
-                                + "' to '"
-                                + location
-                                + "': "
-                                + e.getMessage());
+                int status = conn.getResponseCode();
+                if (!isRedirect(status)) {
+                    keepConnection = true;
+                    break;
+                }
+
+                String location = conn.getHeaderField("Location");
+                if (location == null || location.isEmpty()) {
+                    throw new IOException(
+                            "Received redirect (status "
+                                    + status
+                                    + ") from '"
+                                    + currentUri
+                                    + "' without a Location header");
+                }
+                if (++redirects > MAX_REDIRECTS) {
+                    throw new IOException(
+                            "Too many redirects (>"
+                                    + MAX_REDIRECTS
+                                    + ") while fetching artifact from '"
+                                    + uri
+                                    + "'");
+                }
+                URL nextUrl;
+                try {
+                    nextUrl = new URL(currentUrl, location);
+                } catch (MalformedURLException e) {
+                    throw new IOException(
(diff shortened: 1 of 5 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/artifact/HttpArtifactFetcher.java
```diff
@@ -19,42 +19,135 @@
...
+            conn.setRequestMethod("GET");
 
-        conn.setRequestMethod("GET");
+            int status = conn.getResponseCode();
+            if (!isRedirect(status)) {
+                break;
+            }
 
-        String fileName = FilenameUtils.getName(url.getPath());
+            String location = conn.getHeaderField("Location");
```
dalelane: I think there is a risk of a connection leak here if any (non-redirect) errors happen before we get to `conn.disconnect()` - e.g. getResponseCode() throws an IOException

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/utils/JarUriValidationUtils.java
```diff
@@ -0,0 +1,93 @@
...
+
+        if (("http".equalsIgnoreCase(scheme) || "https".equalsIgnoreCase(scheme))
+                && disallowRestrictedHosts) {
+            String host = uri.getHost();
+            if (host == null || host.isEmpty()) {
+                return Optional.of("jarURI must include a host for http/https schemes");
+            }
+            InetAddress addr;
+            try {
+                addr = InetAddress.getByName(host);
```
dalelane: I'm not sure how paranoid we want to be here... but we could use `InetAddress.getAllByName(host)` and check all the InetAddress that match, not just the first one. (This is about catching addresses with more than one A record, where the first one might validate, but connections resolve to a different one)

Like I say... that might be overly paranoid. I think what you've got works, I'm just spit-balling about ways that someone determined could bypass it.

dalelane (approved):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/artifact/HttpArtifactFetcher.java
```diff
@@ -19,45 +19,149 @@
...
+                }
+            }
+        }
+
+        // Name the file from the original jarURI, not the redirect target, so a redirect can't
+        // change it (e.g. drop the .jar extension the JobManager upload requires).
+        String fileName = FilenameUtils.getName(originalUrl.getPath());
         File targetFile = new File(targetDir, fileName);
         try (var inputStream = conn.getInputStream()) {
             FileUtils.copyToFile(inputStream, targetFile);
```
Dennis-Mircea, line 162: Shouldn't we have here a download size cap as well? A malicious URL, or a redirect to one, can serve an enormous body and fill the operator's ephemeral storage.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/artifact/HttpArtifactFetcher.java
```diff
@@ -19,45 +19,149 @@
...
+                                + currentUri
+                                + "': "
+                                + validationError.get());
+            }
+
+            currentUrl = new URL(currentUri);
+            if (originalUrl == null) {
+                originalUrl = currentUrl;
+            }
+            conn = (HttpURLConnection) currentUrl.openConnection();
```
Dennis-Mircea, line 90: The connection here resolves the host again when it connects (`getResponseCode()` below), independent of the addresses `validateJarURI` already checked with `getAllByName`. So a rebinding DNS answer can hand the socket a different, internal address than the one that was validated. In practice the JVM DNS cache usually makes the two lookups agree, so the window is narrow unless caching is disabled, but pinning the validated IP for the connection (with the Host header preserved) is what closes it fully.

Also, shouldn't we have a timeout here? A jarURI (or a redirect target) pointing at a server that accepts the socket and then stalls hangs the fetch forever, and this runs on the reconcile thread, so it ties up a reconcile worker indefinitely.

spuru9:
@Dennis-Mircea We have a JIRA for Download size cap and Timeout for stalling host hangs reconcile thread as a seperate issue which is raised in FLINK-40402.
I can fold the pin validated IP as a part of that as well. 
Does that work?

Dennis-Mircea:
> FLINK-40402

Sure, it works. Do you plan to open a PR for that so we can include it as part of 1.16.0 operator release?

spuru9:
> Do you plan to open a PR for that so we can include it as part of 1.16.0 operator release?

Yes, just waiting on the PR to merge. As it would create conflict otherwise.

Dennis-Mircea (approved):

PR #1181, opened 2026-08-17
[FLINK-40401] Protect against large autoscaler config maps

spuru9:

Protect against large autoscaler config maps. Improperly packaged autoscaler config maps can cause the Flink Kubernetes operator to fail with an OutOfMemoryException. We should put a size limit.

## Brief change log

  - Added a bounded-read helper for decompressing autoscaler state (scaling history, scaling tracking, collected metrics) read from the `autoscaler-<app>` ConfigMap, capped at 20MB to match the existing YAML parser's own size limit
  - Oversized or corrupted compressed values are now discarded through the existing fallback/migration path instead of being fully decompressed into memory

## Verifying this change

  - Added `KubernetesAutoScalerStateStoreTest#testDiscardOversizedCompressedHistory`, which injects an oversized compressed ConfigMap entry directly and verifies it is discarded rather than causing unbounded memory growth

  - Core observer or reconciler logic that is regularly executed: no (only the autoscaler state store's read path)

## Documentation

---

- [X] Yes (Claude Code)

Generated-by: Claude Code

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/state/KubernetesAutoScalerStateStore.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/state/KubernetesAutoScalerStateStore.java
@@ -39,7 +39,6 @@ import org.apache.flink.shaded.jackson2.org.yaml.snakeyaml.LoaderOptions;
 
 import io.javaoperatorsdk.operator.processing.event.ResourceID;
 import lombok.SneakyThrows;
-import org.apache.commons.io.IOUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -49,6 +48,7 @@ import javax.annotation.Nullable;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
 import java.time.Instant;
 import java.util.Base64;
@@ -79,6 +79,11 @@ public class KubernetesAutoScalerStateStore
     @VisibleForTesting protected static final String DELAYED_SCALE_DOWN = "delayedScaleDown";
 
     @VisibleForTesting protected static final int MAX_CM_BYTES = 1000000;
+
+    /* Caps decompressed value size against gzip bombs. Legit values are a few MB at most
+     * (MAX_CM_BYTES compressed, ~2.5-4x ratio); this is well below the YAML loader's limit,
+     * which guards the parser, not decompression memory. */
+    @VisibleForTesting protected static final int MAX_DECOMPRESSED_BYTES = 8 * 1024 * 1024;
 
     protected static final ObjectMapper YAML_MAPPER =
             new ObjectMapper(yamlFactory())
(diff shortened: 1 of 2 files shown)


gyfora (approved):

spuru9:
Both the fails look unrelated. Might need a retrigger. 
Also seeing these time of maven timeouts in FKO more often than flink is there a way to resolve these flakinesss?

rmetzger:
Is the maven in CI configured with some retries for downloading dependencies? it seems CI is pretty unstable, retrying should improve the situation?
We could also look into adding an additional maven mirror?
I don't know if this mirror is still maintained? https://maven-central.storage.googleapis.com/README.md

spuru9:
Created https://issues.apache.org/jira/browse/FLINK-40406 for tracking. Can try retry first. Will pick after these PRs are resolved.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/state/KubernetesAutoScalerStateStore.java
```diff
@@ -388,15 +394,31 @@ private static String decompress(String compressed) {
         try {
             byte[] bytes = Base64.getDecoder().decode(compressed);
             try (var zi = new GZIPInputStream(new ByteArrayInputStream(bytes))) {
-                return IOUtils.toString(zi, StandardCharsets.UTF_8);
+                return readBounded(zi, MAX_DECOMPRESSED_BYTES);
             }
         } catch (Exception e) {
-            LOG.warn("Error while decompressing scaling data, treating as uncompressed");
+            LOG.warn("Error while decompressing scaling data, treating as uncompressed", e);
```
Dennis-Mircea: This log also fires on the normal migration path. Values stored before compression are not valid gzip, so `GZIPInputStream` throws and the code correctly falls back to treating them as uncompressed. Logging a full stack trace at WARN for that expected case is noise. Keep it without the exception, or attach it only when the value was expected to be compressed.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/state/KubernetesAutoScalerStateStore.java
```diff
@@ -80,6 +80,12 @@ public class KubernetesAutoScalerStateStore
 
     @VisibleForTesting protected static final int MAX_CM_BYTES = 1000000;
 
+    /* Caps the size of a single decompressed value so that a crafted/corrupted ConfigMap
+     * entry (e.g. a gzip bomb) cannot exhaust operator memory during decompression. Matches
+     * the YAML loader's code point limit below, since larger content would be rejected by
+     * the YAML parser anyway. */
+    @VisibleForTesting protected static final int MAX_DECOMPRESSED_BYTES = 20 * 1024 * 1024;
```
Dennis-Mircea: 20MB matches the YAML code point limit, not the heap. A legitimate value decompresses to at most ~4MB, because the compressed form is capped near 1MB (the ConfigMap limit plus the `MAX_CM_BYTES` trim) and the operator's own YAML compresses only ~2.5x to 4x (measured on real metric history), whereas a crafted or hijacked payload reaches its theoretical ~1032:1 maximum, which is the whole reason for a cap.

The default operator sets no heap limit, the peak per decode is ~32MB (ByteArrayOutputStream overshoot), and the cap bounds one value rather than the total, so concurrent poisoned CMs across the reconcile pool can still OOM under it. Deriving it from `Runtime.maxMemory()`, or dropping it near the ~4MB real ceiling, fits better than the parser limit.

Dennis-Mircea (commented):
It is worth noting that this PR only covers the gzip path for the compressed fields (`scalingHistory`, `scalingTracking` and `collectedMetrics`). What is also important is the `configOverrides` path. `configOverrides` is merged into `spec.flinkConfiguration` through `putAllFrom` with no key filter, so whoever controls it can plant arbitrary Flink settings, `env.java.opts.all` or a pod template among them. That only takes effect when `memory.tuning.enabled` is on, which is off by default, but still it is a valid case to consider, and is a place where we need to have some guards as well. Maybe we can tackle it as part of a different JIRA / PR.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/state/KubernetesAutoScalerStateStore.java
spuru9: As ~4MB is real ceiling. Dropping the value to 8MB with some headspace. Let me know if thats ok.

spuru9:
@Dennis-Mircea Have addressed your comments. Can you recheck.
Also can someone retrigger the CI. The last one passed and this change should not be blocking.

rmetzger (approved):
Change looks good -- thank you. Merging.

PR #1184, opened 2026-08-18
[FLINK-40414] Autoscaler state ConfigMap is treated as trusted input

Dennis-Mircea:

The autoscaler persists its state in a per-resource ConfigMap that is writable by any workload in the namespace (the Flink job Role grants configmaps write for Kubernetes HA), and the operator reads it back and acts on it every cycle. Two gaps let a poisoned entry affect the operator: the read paths are not uniformly fail-closed (one throws into the reconcile loop, one never discards the bad entry), and config overrides are applied to `spec.flinkConfiguration` without filtering, so a crafted entry could inject arbitrary Flink configuration (for example `env.java.opts` or a pod template) into a managed deployment when memory tuning is enabled. This hardens the whole state read boundary and filters config overrides on apply. It complements FLINK-40400 (artifact fetch SSRF) and FLINK-40401 (decompression bomb on the compressed state fields).

## Brief change log

  - `KubernetesAutoScalerStateStore`: all six state reads now route through a single `readState` wrapper that treats the stored value as untrusted and fails closed, discarding the offending entry and returning empty on any deserialization error. This fixes `getConfigChanges` (previously fell back to empty but left the poisoned entry to be re-read every cycle) and `getParallelismOverrides` (previously had no error handling, so a malformed value threw into the reconcile loop).
  - Added positive-integer value sanity for parallelism overrides: entries whose value is not a positive integer are dropped and logged.
  - `deserializeConfigOverrides` now propagates on bad input instead of catching and returning null, so the wrapper discards it uniformly with the other fields.
  - `MemoryTuning.TUNABLE_CONFIG_KEYS`: the allowlist of keys memory tuning is allowed to set, built from the 10 `taskmanager.memory.*` options it emits plus their fallback keys.
  - `JobAutoScalerImpl.dropNonTunableKeys`: filters both overrides and removals of the config changes read from the ConfigMap against that allowlist before `realizeConfigOverrides`, so only memory-tuning keys reach the deployment spec. Dropped keys are logged at WARN.
  - The compressed-field decompress path is intentionally untouched (owned by FLINK-40401). The compressed getters are only rerouted through the wrapper, which additionally turns the existing oversize fallback into a deliberate discard.

## Verifying this change

  - `KubernetesAutoScalerStateStoreTest#testDiscardMalformedConfigOverrides`: a poisoned config-overrides entry is discarded (removed), not silently re-read.
  - `KubernetesAutoScalerStateStoreTest#testParallelismOverridesFailClosedAndSanitized`: a malformed value returns empty instead of throwing, and a valid map keeps only positive-integer entries.
  - `JobAutoScalerImplTest#testConfigOverridesDropNonTunableKeys`: an injected `env.java.opts.all` override and a pod-template removal are dropped while the memory-tuning override is kept.
  - `MemoryTuningTest`: a drift guard asserting every key real tuning emits is contained in `TUNABLE_CONFIG_KEYS`, so the allowlist cannot silently fall behind the tuning logic.
  - Each new test was confirmed to fail when its corresponding fix is reverted in isolation. The full `flink-autoscaler` suite (290 tests) and the operator autoscaler package (40 tests) pass with checkstyle and spotless.

  - Core observer or reconciler logic that is regularly executed: yes, the autoscaler state read path and the config-override apply path run on every scaling cycle

## Documentation

---

### DIFF
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/JobAutoScalerImpl.java
@@ -26,6 +26,7 @@ import org.apache.flink.autoscaler.metrics.AutoscalerFlinkMetrics;
 import org.apache.flink.autoscaler.realizer.ScalingRealizer;
 import org.apache.flink.autoscaler.state.AutoScalerStateStore;
 import org.apache.flink.autoscaler.tuning.ConfigChanges;
+import org.apache.flink.autoscaler.tuning.MemoryTuning;
 import org.apache.flink.configuration.PipelineOptions;
 import org.apache.flink.util.Preconditions;
 
@@ -35,7 +36,9 @@ import org.slf4j.LoggerFactory;
 import java.time.Clock;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.stream.Collectors;
 
 import static org.apache.flink.autoscaler.config.AutoScalerOptions.AUTOSCALER_ENABLED;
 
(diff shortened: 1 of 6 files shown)


gyfora (approved):

gyfora (approved):

gyfora:
@Dennis-Mircea can you rebase please

Dennis-Mircea:
> @Dennis-Mircea can you rebase please

Done!

PR #1187, opened 2026-08-19
[FLINK-40443] Harden autoscaler metrics name parsing against complex regex

spuru9:

Simplify the Kafka/Pulsar partition metric name matching used by the autoscaler's `ScalingMetricCollector`, replacing the regex-based implementation with a plain string parser.

## Brief change log

  - Added `PartitionMetricNameParser` (`flink-autoscaler` `utils` package) implementing `parseKafkaPartitionKey`/`parsePulsarPartitionKey` as a string-based replacement for the previous regex-based matching.
  - Updated `ScalingMetricCollector.updateKafkaPulsarSourceNumPartitions` to use the new parser.
  - Minor cleanup: reuse the existing `JobTopology.isSource` helper instead of re-checking `getInputs().isEmpty()` inline.

## Verifying this change

  - Added `PartitionMetricNameParserTest`, covering the Kafka/Pulsar metric name shapes handled by the parser.
  - Existing `MetricsCollectionAndEvaluationTest#testKafkaPulsarNumPartitions` continues to pass unmodified, confirming behavior is preserved.

  - Core observer or reconciler logic that is regularly executed: yes (autoscaler metric collection path)

## Documentation

---

- [X] Yes (Claude Code)

Generated-by: Claude Code

### DIFF
--- a/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/utils/PartitionMetricNameParser.java
+++ b/flink-autoscaler/src/main/java/org/apache/flink/autoscaler/utils/PartitionMetricNameParser.java
@@ -1,0 +1,105 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.autoscaler.utils;
+
+/**
+ * Parses Kafka/Pulsar source-partition metric names to extract a stable per-partition key.
+ *
+ * <p>Uses plain string splitting/indexing rather than a regex, since metric names are fully
+ * job-controlled and this must stay linear-time regardless of input shape.
+ *
+ * <p>Parsing splits on the '.' delimiter and treats each name component as dot-free: Flink's metric
+ * query service replaces '.' within a component with '_' before joining, so a '.' in the
+ * identifiers returned by the REST metrics endpoint is always a real delimiter. Only correctness on
+ * legitimate input relies on this assumption; safety does not, since an unexpected shape merely
+ * yields null.
+ */
+public final class PartitionMetricNameParser {
+
+    private static final String KAFKA_CLUSTER = "kafkaCluster";
+    private static final String KAFKA_SOURCE_READER = "KafkaSourceReader";
+    private static final String TOPIC = "topic";
+    private static final String PARTITION = "partition";
+    private static final String CURRENT_OFFSET = "currentOffset";
+    private static final String PULSAR_CONSUMER = "PulsarConsumer";
+    private static final String NUM_MSGS_RECEIVED = "numMsgsReceived";
+    private static final String PULSAR_PARTITION_INFIX = "-partition-";
+
+    private PartitionMetricNameParser() {}
+
+    /**
+     * Returns {@code "<cluster>-<topic>-<id>"}, or {@code null} if this is not a Kafka
+     * currentOffset partition metric. The cluster segment is the literal "null" when absent; only
+     * distinct keys matter to callers.
+     */
+    public static String parseKafkaPartitionKey(String metricName) {
+        String[] parts = metricName.split("\\.", -1);
+        int n = parts.length;
+        // tail: KafkaSourceReader.topic.<topic>.partition.<id>.currentOffset
+        if (n < 6
+                || !parts[n - 6].equals(KAFKA_SOURCE_READER)
+                || !parts[n - 5].equals(TOPIC)
+                || !parts[n - 3].equals(PARTITION)
+                || !isAllDigits(parts[n - 2])
+                || !parts[n - 1].equals(CURRENT_OFFSET)) {
+            return null;
+        }
+        // optional ".kafkaCluster.<cluster>." precedes the tail; null when absent
+        String cluster = (n >= 8 && parts[n - 8].equals(KAFKA_CLUSTER)) ? parts[n - 7] : null;
+        return cluster + "-" + parts[n - 4] + "-" + parts[n - 2];
+    }
+
+    /**
+     * Returns {@code "<topic>-<id>"}, or {@code null} if this is not a Pulsar numMsgsReceived
+     * partition metric.
+     */
+    public static String parsePulsarPartitionKey(String metricName) {
+        String[] parts = metricName.split("\\.", -1);
+        int n = parts.length;
+        // tail: PulsarConsumer.<topic>-partition-<id>.<consumerHash>.numMsgsReceived
+        if (n < 3 || !parts[n - 1].equals(NUM_MSGS_RECEIVED)) {
+            return null;
+        }
+        for (int i = 0; i + 1 < n - 1; i++) {
+            if (!parts[i].equals(PULSAR_CONSUMER)) {
+                continue;
+            }
+            String segment = parts[i + 1]; // "<topic>-partition-<id>"
+            int idx = segment.lastIndexOf(PULSAR_PARTITION_INFIX);
+            if (idx <= 0) {
+                return null;
+            }
+            String id = segment.substring(idx + PULSAR_PARTITION_INFIX.length());
+            return isAllDigits(id) ? segment.substring(0, idx) + "-" + id : null;
+        }
+        return null;
+    }
+
+    private static boolean isAllDigits(String s) {
+        if (s.isEmpty()) {
+            return false; // matches \d+, which requires at least one digit
+        }
+        for (int i = 0; i < s.length(); i++) {
+            char c = s.charAt(i);
+            if (c < '0' || c > '9') { // ASCII only, matching regex \d
+                return false;
+            }
+        }
+        return true;
+    }
+}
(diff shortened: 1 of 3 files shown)


flink-autoscaler/src/test/java/org/apache/flink/autoscaler/utils/PartitionMetricNameParserTest.java
```diff
@@ -0,0 +1,145 @@
...
+import org.junit.jupiter.api.Test;
+
+import java.time.Duration;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertTimeoutPreemptively;
+
+/** Test for {@link PartitionMetricNameParser}. */
+public class PartitionMetricNameParserTest {
```
rmetzger, line 35: Can you remodel this test and include the old regex in here as well to make sure that the old regex and the new parser both produce exactly the same result?

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/utils/PartitionMetricNameParser.java
```diff
@@ -0,0 +1,102 @@
...
+
+package org.apache.flink.autoscaler.utils;
+
+/**
+ * Parses Kafka/Pulsar source-partition metric names to extract a stable per-partition key.
+ *
+ * <p>Uses plain string splitting/indexing rather than a regex, since metric names are fully
+ * job-controlled and this must stay linear-time regardless of input shape.
+ */
+public final class PartitionMetricNameParser {
```
rmetzger, line 32: are there e2e tests testing this against a real kafka / pulsar instance? I'm worried that this new code is breaking something

spuru9:
```
Old regex:
^.*?  (\.kafkaCluster\.(?<kafkaCluster>.+))?  \.KafkaSourceReader \.topic \.(?<kafkaTopic>.+) \.partition \.(?<kafkaId>\d+) \.currentOffset$

After split("."), the tail is always these 6 segments (indices from the end), plus an optional 2-segment cluster prefix:

┌─────────────────────────────────────────────────┬──────────────────────────────────────────────────────────────┬────────────────────────────────────────────────────────┐
│                   regex token                   │                          new parser                          │                        meaning                         │
├─────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────┤
│ ^.*?                                            │ (nothing — never inspected)                                  │ ignore everything before the tail (the operator scope) │
├─────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────┤
│ \.KafkaSourceReader                             │ parts[n-6].equals(KAFKA_SOURCE_READER)                       │ literal marker                                         │
├─────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────┤
│ \.topic                                         │ parts[n-5].equals(TOPIC)                                     │ literal marker                                         │
├─────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────┤
│ (?<kafkaTopic>.+)                               │ parts[n-4]                                                   │ the topic (extracted, not validated)                   │
├─────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────┤
│ \.partition                                     │ parts[n-3].equals(PARTITION)                                 │ literal marker                                         │
├─────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────┤
│ (?<kafkaId>\d+)                                 │ isAllDigits(parts[n-2])                                      │ partition id = digits                                  │
├─────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────┤
│ \.currentOffset$                                │ parts[n-1].equals(CURRENT_OFFSET)                            │ $ → must be the last segment                           │
├─────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────┤
│ (\.kafkaCluster\.(?<kafkaCluster>.+))?          │ n>=8 && parts[n-8].equals(KAFKA_CLUSTER) ? parts[n-7] : null │ the optional ? group, 2 segments ahead of the tail     │
├─────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────┤
│ kafkaCluster + "-" + kafkaTopic + "-" + kafkaId │ cluster + "-" + parts[n-4] + "-" + parts[n-2]                │ same key (cluster renders as "null" when absent)       │
└─────────────────────────────────────────────────┴──────────────────────────────────────────────────────────────┴────────────────────────────────────────────────────────┘

Pulsar

Old regex:
^.*  \.PulsarConsumer  \.(?<pulsarTopic>.+)-partition-(?<pulsarId>\d+)  \. .*  \.numMsgsReceived$

┌───────────────────────────────────────────────┬──────────────────────────────────────────────────────┬───────────────────────────────────────────────────────────────────────────────────────────────────┐
│                  regex token                  │                      new parser                      │                                              meaning                                              │
├───────────────────────────────────────────────┼──────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────┤
│ ^.*\.PulsarConsumer                           │ the for loop scanning for                            │ find the marker segment; anything before it is the ignored prefix                                 │
│                                               │ parts[i].equals(PULSAR_CONSUMER)                     │                                                                                                   │
├───────────────────────────────────────────────┼──────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────┤
│ (?<pulsarTopic>.+)-partition-(?<pulsarId>\d+) │ segment = parts[i+1], then                           │ topic-partition lives in one segment; lastIndexOf mirrors greedy .+ (prefers the last             │
│                                               │ lastIndexOf("-partition-") splits it                 │ -partition-)                                                                                      │
├───────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ (?<pulsarId>\d+)                              │ isAllDigits(id) where id = segment.substring(idx+11) │ id after -partition- must be digits                                                               │
├───────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ \..*\.numMsgsReceived$                        │ parts[n-1].equals(NUM_MSGS_RECEIVED) + loop bound    │ the $ anchor (last segment) + require room for a segment before it; the .* (consumer-hash) in     │
│                                               │ i+1 < n-1                               not checked, so any middle segments are accepted                                       │
├───────────────────────────────────────────────┼──────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────┤
│ pulsarTopic + "-" + pulsarId                  │ segment.substring(0, idx) + "-" + id    rops the -partition- infix)                                                            │
└───────────────────────────────────────────────┴──────────────────────────────────────────────────────┴───────────────────────────────────────────────────────────────────────────────────────────────────┘
```

flink-autoscaler/src/test/java/org/apache/flink/autoscaler/utils/PartitionMetricNameParserTest.java
spuru9: Done.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/utils/PartitionMetricNameParser.java
spuru9: No. Currently we dont have such test in CI.

rmetzger: Did you manually verify the parser against a real kafka / pulsar?

rmetzger (approved):
Change looks good to me, but I would prefer a review from somebody more familiar with the codebase as well.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/utils/PartitionMetricNameParser.java
spuru9: Yes — verified manually against real brokers, broadened well past the earlier topic-only check. Fresh minikube, operator built from this PR (`DEBUG` on `ScalingMetricCollector`), real Apache Kafka 3.9.1 (KRaft) + Pulsar 3.3.1 (standalone), OSS connectors kafka:3.3.0-1.20 / pulsar:4.2.0-1.20 on Flink 1.20. Per case I cross-checked broker ground truth → REST `/metrics` names → the operator's derived `numSourcePartitions`.

| # | Dimension | Setup | Expected | Detected |
|---|---|---|---|---|
| 1 | Multi-topic + higher scale + edge names | Kafka `events`(6) + `big24`(24) + `orders.eu.v1`(3, dotted) + `my-topic`(2, hyphen), parallelism 1 | 35 | 35 ✅ |
| 2 | Source parallelism > 1 | Kafka `par12`(12) at parallelism 4 (partitions spread across 4 subtasks) | 12 | 12 ✅ |
| 3 | Multiple source vertices in one job | one job with a Kafka source (`kmix`=5) **and** a Pulsar source (`pmix`=3) | 5 / 3 | Kafka vtx 5, Pulsar vtx 3 ✅ |
| — | Negative (non-partition source) | `StateMachineExample` | none | `numSourcePartitions=0`, no false positive ✅ |

Edge cases confirmed end-to-end:
- **Dot sanitization:** `orders.eu.v1` arrives as `…topic.orders_eu_v1.partition.N.currentOffset` (`.` in a component becomes `_` before joining), so the split-on-`.` parser gets a clean segment.
- **Hyphen vs `-partition-`:** `my-topic` / `orders-v2-partition-N` key off the right marker.
- **Parallelism > 1:** correct when aggregating `currentOffset` across subtasks.
- **Multiple sources:** each source vertex counted independently.

**Upfront caveat:** with `scaling.enabled=true` the parser computes `numSourcePartitions=4` every cycle, but I couldn't force a live scale-up on this single-node setup to show that count *binding* as the source's max-parallelism cap. That cap-application is pre-existing autoscaler logic anyway — this PR only computes the count.

spuru9: As discussed offline, created https://issues.apache.org/jira/browse/FLINK-40445 for adding the tests to CI

rmetzger: ok, nice, thank you

gyfora (approved):

rmetzger:
@spuru9 can you push an empty commit or something to get us a green ci?

PR #1189, opened 2026-08-20
[FLINK-40445] Autoscaler: integration test for Kafka/Pulsar source partition-metric detection

spuru9:

This pull request adds an integration test that verifies `ScalingMetricCollector` correctly derives `numSourcePartitions` for Kafka and Pulsar sources against real brokers. It is a follow-up to [FLINK-40443](https://issues.apache.org/jira/browse/FLINK-40443) / #1187, added per maintainer consensus on that PR as a separate, non-blocking PR. It does not touch the hardening fix itself.

## Brief change log

  - Adds `KafkaPulsarPartitionMetricsITCase`, covering: multi-topic summing, dotted and hyphenated topic names, source parallelism > 1, multiple source vertices in one job, and a non-Kafka/Pulsar source that must not report a partition count
  - Adds `MiniClusterJobDriver`, which runs the Flink job under test via a `MiniCluster` in its own JVM rather than in-process, because `flink-autoscaler` vendors a compatibility copy of `org.apache.flink.runtime.rest.messages.job.JobDetailsInfo` under `src/main/java` that shadows the real Flink class on this module's own classpath, which breaks a same-process JobManager's own REST handlers
  - Adds test-scoped dependencies to `flink-autoscaler/pom.xml` (`flink-connector-kafka`, `flink-connector-pulsar`, `testcontainers-kafka`, `testcontainers-pulsar`, `commons-codec`) and disables the failsafe module path (`useModulePath=false`), since the connector jars otherwise get placed on the module path as automatic modules and lose visibility into `flink-runtime`'s unnamed-module classpath classes
  - Wires `flink-autoscaler` into CI (`ci.yml`) with the same per-module `verify -Dit.skip=false` step already used for `flink-kubernetes-operator`, `flink-kubernetes-webhook`, and `flink-autoscaler-plugin-jdbc`, so the new ITCase actually runs instead of being silently skipped by the whole-reactor build

## Verifying this change

  - Added `KafkaPulsarPartitionMetricsITCase`, an integration test that spins up real Kafka and Pulsar brokers via Testcontainers and asserts `ScalingMetricCollector#getJobTopology` reports the correct `numSourcePartitions` for: multi-topic summing, dotted/hyphenated topic names, source parallelism > 1, multiple source vertices in one job, and a negative case (non-Kafka/Pulsar source reports no partition count)
  - `mvn -pl flink-autoscaler verify -Dit.skip=false` passes locally (300 unit tests + the new IT)
  - `mvn -pl flink-autoscaler checkstyle:check spotless:check` is clean
  - CI run on this PR is green on both the JDK 17 and JDK 21 matrix legs, including the new `Tests in flink-autoscaler` step: https://github.com/apache/flink-kubernetes-operator/actions/runs/32391143762

  - Dependencies (does it add or upgrade a dependency): yes, but test-scope only, confined to `flink-autoscaler/pom.xml` (`flink-connector-kafka`, `flink-connector-pulsar`, `testcontainers-kafka`, `testcontainers-pulsar`, `commons-codec`)

## Documentation

---

- [X] Yes (Claude Code)

### DIFF
--- a/flink-autoscaler/src/test/java/org/apache/flink/autoscaler/KafkaPulsarPartitionMetricsITCase.java
+++ b/flink-autoscaler/src/test/java/org/apache/flink/autoscaler/KafkaPulsarPartitionMetricsITCase.java
@@ -1,0 +1,367 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.autoscaler;
+
+import org.apache.flink.api.common.JobID;
+import org.apache.flink.api.common.JobStatus;
+import org.apache.flink.autoscaler.state.InMemoryAutoScalerStateStore;
+import org.apache.flink.autoscaler.topology.JobTopology;
+import org.apache.flink.client.program.rest.RestClusterClient;
+import org.apache.flink.configuration.Configuration;
+import org.apache.flink.metrics.groups.UnregisteredMetricsGroup;
+import org.apache.flink.runtime.highavailability.nonha.standalone.StandaloneClientHAServices;
+import org.apache.flink.runtime.rest.messages.job.JobDetailsInfo;
+
+import org.apache.kafka.clients.admin.Admin;
+import org.apache.kafka.clients.admin.AdminClientConfig;
+import org.apache.kafka.clients.admin.NewTopic;
+import org.apache.kafka.clients.producer.KafkaProducer;
+import org.apache.kafka.clients.producer.ProducerConfig;
+import org.apache.kafka.clients.producer.ProducerRecord;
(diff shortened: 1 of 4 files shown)


spuru9:
@rmetzger Can you check this. Follow up to an ask on earlier PR.

spuru9:
@Dennis-Mircea Can you help me a review.

flink-autoscaler/src/test/java/org/apache/flink/autoscaler/KafkaPulsarPartitionMetricsITCase.java
```diff
@@ -0,0 +1,367 @@
...
+            driverProcess.destroy();
+            if (!driverProcess.waitFor(10, TimeUnit.SECONDS)) {
+                driverProcess.destroyForcibly();
+            }
+        }
+        if (driverResultFile != null) {
+            Files.deleteIfExists(driverResultFile);
+        }
+        if (driverLogFile != null) {
+            Files.deleteIfExists(driverLogFile);
```
rmetzger: this will delete the log also in case of a failure -- keep it so that debugging is easier.

rmetzger: ideally when there's an error, print the entire log contents to the output?

flink-autoscaler/src/test/java/org/apache/flink/autoscaler/KafkaPulsarPartitionMetricsITCase.java
```diff
@@ -0,0 +1,367 @@
...
+                    return isOwnMainOutput;
+                });
+        entries.addAll(ownMainOutput);
+        return String.join(File.pathSeparator, entries);
+    }
+
+    /**
+     * Polls {@code resultFile} for the two lines {@link MiniClusterJobDriver} writes once ready.
+     */
+    private static void awaitDriverReady(Process process, Path resultFile) throws Exception {
```
rmetzger: use org.awaitility.Awaitility here as well?

rmetzger:
I agree with the approach and direction in this PR, happy to merge it once above items are addressed.

PR #1193, opened 2026-08-25
[FLINK-40465] Add version-independent classes to parse job config

mateczagany:

The operator currently uses `JobConfigInfo` of Flink 1.20 to deserialize the results of `/jobs/:jobid/config` since [FLINK-35746](https://issues.apache.org/jira/browse/FLINK-35746) but Flink 2.0 removed the required `execution-mode` field, which causes deserialization errors, repeated REST queries, then a timeout.

By adding a version-independent DTO to parse the job configuration, Flink 1.x and Flink 2.x responses could be parsed properly.

## Brief change log

- Add `CustomJobConfigInfo` and `CustomJobConfigHeaders` classes following existing naming patterns
- Remove and move some tests to `CustomJobConfigInfoTest`

## Verifying this change

- Added unit tests
- Verified that the error logs are no longer present when reconciling Flink 2.x jobs

## Documentation

---

- [x] Yes (please specify the tool below)


Dennis-Mircea (changes requested):
Thanks for opening this PR! As part of the operator, for the classes that are copied from the `flink-runtime` there is the following convention: those classes should be copied into the `flink-kubernetes-operator` module under the same fully qualified class name from the `flink-runtime` package. You can check the existing example for `org.apache.flink.runtime.rest.messages.JobExceptionsInfoWithHistory`.

By going with this convention, only the `JobConfigInfo` needs to be shadowed, as the `JobConfigHeaders`, from what I've checked, keeps the same implementation across the Flink versions.

Further, in order to shadow the `JobConfigInfo` correctly, we have to adapt the implementation to support the format of all the responses returned by different Flink versions. Currently, what you did as part of `CustomJobConfigInfo` will fully reject the old `execution-mode`. So, this has to be fixed as well.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/CustomJobConfigInfo.java
```diff
@@ -0,0 +1,61 @@
...
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+import java.util.Map;
+
+/** Custom Response for fetching job configuration in a multi-version compatible way. */
+@JsonIgnoreProperties(ignoreUnknown = true)
+@Data
+@NoArgsConstructor
+public class CustomJobConfigInfo implements ResponseBody {
```
gyfora: We could simply shadow/override the original package.classname for JobConfigInfo like we do for other similar classes. That would remove the flink service and custom header requirement.

mateczagany: Thank you @Dennis-Mircea and @gyfora I've pushed a new commit, it's become a lot simpler this way.

gyfora (approved):

Dennis-Mircea (approved):

PR #1194, opened 2026-08-26
[FLINK-40452] Cap exception fetching by size instead of line count

mbalassi:
JobStatusObserver capped JobManager exception text in K8s events byline count (stacktrace.split("\n"), keep first N lines). A single unbroken line of any size produces one array element, so it bypassed the cap entirely and could fill the operator heap.

 # Approach

  - Replaced `events.exceptions.stacktrace-lines` (line count, default 5)
  with `events.exceptions.stacktrace.max.length` (character count, default
  2048 — matches the two existing sibling options
  `exception.stacktrace.max.length` / `exception.field.max.length`; no
  config in the repo defaults to 4096, so 2048 was kept rather than
  raised).
  - Replaced the truncation itself: stacktrace.length() > 
  maxStackTraceLength + bounded substring, instead of splitting on \n.
  Also removes an unbounded 3-4x string-copy chain downstream of the old
  cap.
  - Old option removed outright (no value-preserving migration exists
  between line-count and character-count semantics).
  - Added a regression test (5000-char single line, no newlines) and
  updated existing tests, generated config docs, and hand-written
  event-format docs (both locales).

  # Testing

Updated existing unit test and added a new test to explicitly cover the bug scenario reported in FlINK-40452.

---

- [X] Yes (please specify the tool below)

Generated-by: Claude Code (Opus 5)

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkOperatorConfiguration.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkOperatorConfiguration.java
@@ -210,7 +210,8 @@ public class FlinkOperatorConfiguration {
                 operatorConfig.get(KubernetesOperatorConfigOptions.OPERATOR_EVENT_EXCEPTION_LIMIT);
         int reportedExceptionEventsMaxStackTraceLength =
                 operatorConfig.get(
-                        KubernetesOperatorConfigOptions.OPERATOR_EVENT_EXCEPTION_STACKTRACE_LINES);
+                        KubernetesOperatorConfigOptions
+                                .OPERATOR_EVENT_EXCEPTION_STACKTRACE_MAX_LENGTH);
 
         boolean manageIngress =
                 operatorConfig.get(KubernetesOperatorConfigOptions.OPERATOR_MANAGE_INGRESS);
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
@@ -780,12 +780,12 @@ public class KubernetesOperatorConfigOptions {
                             "How often to retrieve Kubernetes cluster resource usage information. This information is used to avoid running out of cluster resources when scaling up resources. Negative values disable the feature.");
 
     @Documentation.Section(SECTION_ADVANCED_RECONCILE)
-    public static final ConfigOption<Integer> OPERATOR_EVENT_EXCEPTION_STACKTRACE_LINES =
-            operatorConfig("events.exceptions.stacktrace-lines")
+    public static final ConfigOption<Integer> OPERATOR_EVENT_EXCEPTION_STACKTRACE_MAX_LENGTH =
+            operatorConfig("events.exceptions.stacktrace.max.length")
                     .intType()
-                    .defaultValue(5)
+                    .defaultValue(2048)
                     .withDescription(
-                            "Maximum number of stack trace lines to include in exception-related Kubernetes event messages.");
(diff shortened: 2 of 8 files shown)


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/FlinkOperatorConfiguration.java
```diff
@@ -210,7 +210,8 @@ public static FlinkOperatorConfiguration fromConfiguration(Configuration operato
                 operatorConfig.get(KubernetesOperatorConfigOptions.OPERATOR_EVENT_EXCEPTION_LIMIT);
         int reportedExceptionEventsMaxStackTraceLength =
                 operatorConfig.get(
-                        KubernetesOperatorConfigOptions.OPERATOR_EVENT_EXCEPTION_STACKTRACE_LINES);
```
Dennis-Mircea, line 213: Wouldn't be better/safer to first deprecate this option that was introduced as part of 1.12.0 operator release and remove it in further releases?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/config/KubernetesOperatorConfigOptions.java
```diff
@@ -780,12 +780,12 @@ public static String operatorConfigKey(String key) {
                             "How often to retrieve Kubernetes cluster resource usage information. This information is used to avoid running out of cluster resources when scaling up resources. Negative values disable the feature.");
 
     @Documentation.Section(SECTION_ADVANCED_RECONCILE)
-    public static final ConfigOption<Integer> OPERATOR_EVENT_EXCEPTION_STACKTRACE_LINES =
-            operatorConfig("events.exceptions.stacktrace-lines")
+    public static final ConfigOption<Integer> OPERATOR_EVENT_EXCEPTION_STACKTRACE_MAX_LENGTH =
+            operatorConfig("events.exceptions.stacktrace.max.length")
```
Dennis-Mircea, line 784: I'd say it's good to add a defensive support for negative values, and maybe consider `-1` value as unlimited length.

Dennis-Mircea (commented):
I also took a closer look at `JobStatusObserver` and I see that the `recordJobErrorIfPresent` method is not covered by this feature, and currently it can emit event messages of any size. I'd say will great to extend the max length support for this method as well.

Additionally, the `observeJobManagerExceptions` has a `TODO` in place as of now which brings into light another hot spot for this max cap feature. Currently, the history of exceptions is being retrieved via REST and can have a big size as well, so I'd say it is worth considering fixing that place as well and to wrap everything at once for `JobStatusObserver`.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/observer/JobStatusObserver.java
```diff
@@ -267,15 +268,14 @@ private void emitJobManagerExceptionEvent(
...
+            if (stacktrace.length() > maxStackTraceLength) {
                 eventMessage
+                        .append(stacktrace, 0, maxStackTraceLength)
                         .append("... (")
-                        .append(lines.length - maxStackTraceLines)
-                        .append(" more lines)");
+                        .append(stacktrace.length() - maxStackTraceLength)
+                        .append(" more characters)");
+            } else {
+                eventMessage.append(stacktrace);
```
spuru9, line 278: nit optional: we can simply this to 
```
          if (stacktrace != null && !stacktrace.isBlank()) {                                                                                                                                                     
              getSubstringWithMaxLength(stacktrace, maxStackTraceLength).ifPresent(eventMessage::append);                                                                                                        
              if (stacktrace.length() > maxStackTraceLength) {                                                                                                                                                   
                  eventMessage                                                                                                                                                                                   
                          .append("... (")                                                                                                                                                                       
                          .append(stacktrace.length() - maxStackTraceLength)                                                                                                                                     
                          .append(" more characters)");                                                                                                                                                          
              }                                                                                                                                                                                                  
          }
```

spuru9 (commented):
LGTM post the addressing the comments from Dennis.

@mbalassi Can you take a look at them.

PR #1195, opened 2026-08-26
[FLINK-40474] Autoscaler parallelism overrides serialized in the operator's YAML dialect break job submission for Flink 1.x jobs

Dennis-Mircea:

This PR fixes the autoscaler writing `pipeline.jobvertex-parallelism-overrides` in the standard YAML flow form (`{vertexA: '2', ...}`), which jobs on Flink versions lower than 2.0 cannot parse, failing every full submission after a scaling (upgrades, last-state restores, session job submits). The overrides are now always serialized in the legacy dialect, which every Flink version parses.

## Brief change log

- `KubernetesScalingRealizer`, `KubernetesAutoScalerStateStore`, `JdbcAutoScalerStateStore` and `ConfigChanges` serialize overrides with an explicit `standardYaml=false`.
- New `FlinkConfMountDecorator#normalizeMapTypedOptions` re-renders string-valued overrides in the target version's dialect, applied when writing the cluster config file and the session job jar-run body.
- Docs diagrams now show the legacy form `<vertex-id1>:<new-parallelism>,...` (rendered text and embedded draw.io sources).

## Verifying this change
The implementation is covered by unit test for each portion of the code that was touched. In addition, these changes were tested against a live autoscaling deployment, and they were demonstrated to be fully valid.

## Documentation

---


gyfora:
we do not need to upgrade the diagrams to be honest, definitely not all of them :D

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/tuning/ConfigChanges.java
```diff
@@ -39,7 +39,8 @@ public class ConfigChanges {
     @Getter private final Set<String> removals = new HashSet<>();
 
     public <T> ConfigChanges addOverride(ConfigOption<T> configOption, T value) {
-        overrides.put(configOption.key(), ConfigurationUtils.convertValue(value, String.class));
+        overrides.put(
+                configOption.key(), ConfigurationUtils.convertValue(value, String.class, false));
```
gyfora, line 43: instead of hardcoding false here, why don't we decide based on the flink version?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/kubeclient/decorators/FlinkConfMountDecorator.java
```diff
@@ -249,4 +251,20 @@ public String getFlinkConfFilename() {
     public static boolean useStandardYamlConfig(FlinkVersion flinkVersion) {
         return flinkVersion != null && flinkVersion.isEqualOrNewer(FlinkVersion.v2_0);
     }
+
+    /**
+     * Re-parses Map-typed options that arrive already serialized to a String (e.g. the autoscaler
+     * parallelism overrides stored in the CR spec) and sets them back as typed values.
+     */
+    public static void normalizeMapTypedOptions(Configuration source, Configuration target) {
+        if (!source.containsKey(PipelineOptions.PARALLELISM_OVERRIDES.key())) {
```
gyfora: This is a bit of a strange logic to single out a config key based on the autoscaler logic. Would be great if the autoscaler set it correctly already

gyfora (commented):
Something doesn't feel right here. The autoscaler should set the overrides in the correct yaml format depending on the flink version / config that we have for this job not go back and forth.

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/tuning/ConfigChanges.java
Dennis-Mircea: Good question. I'll try to explain a bit in detail about this one.

I decided to go this way mainly because users may need to downgrade the operator from 1.16.0 back to 1.15.0 (maybe because something did not go as expected). If this is not set to false and we go with the standard YAML parser, which is the default in 1.16.0, then once downgraded to 1.15.0, the legacy YAML parser will fail to read the standard YAML format.

The above argument is the reason for setting the flag to false everywhere in this PR. This place concretely is not problematic, because it only handles memory tuning keys, which are sizes and fractions, not Lists or Maps, which are the types that serialize differently under the standard YAML format. I set it to false here as well just to align with the other places and to avoid confusion about the overall approach.

From the implementation point of view, there is also a limitation: the `flink-autoscaler` module does not depend on the `flink-kubernetes-operator` modules, so `FlinkVersion` is not visible at this layer, and the standalone autoscaler has no such notion at all.

I think it is safer to proceed this way for now, for at least the next couple of releases. Once support for Flink below 2.0 is dropped and downgrades to 1.15.0 are no longer a concern, we can move to standard YAML everywhere and drop this. What do you say?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/kubeclient/decorators/FlinkConfMountDecorator.java
Dennis-Mircea: The autoscaler sets it in legacy format based on my previous argument of having the operator downgrades covered. This method makes the deployed configuration properties, which are serialized differently by the standard YAML parser and owned by the operator,  independent of what the stored value looks like, whether it came from a manual user edit, or if it got tampered. This is more like a security safeguard.

Dennis-Mircea:
> we do not need to upgrade the diagrams to be honest, definitely not all of them :D

I updated them because I figured out that when I exported them at first from draw.io they were part of a multi-page doc, and they were embedding the entire source document. The only actual change was made to autoscaler-parallelism-pipeline.svg to correctly reflect the new format.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/kubeclient/decorators/FlinkConfMountDecorator.java
gyfora: Is this anyways necessary? I though the YAML parser handles both the old and new format values

flink-autoscaler/src/main/java/org/apache/flink/autoscaler/tuning/ConfigChanges.java
gyfora: sounds good

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/kubeclient/decorators/FlinkConfMountDecorator.java
Dennis-Mircea: Yep, that's true. The only case that I thought about was also a downgrade that may happen, but still the operator is writing the YAML in legacy format now, so it will not be a problem anymore. I'll revert this.

gyfora (approved):

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/state/KubernetesAutoScalerStateStore.java
```diff
@@ -307,7 +307,7 @@ private static SortedMap<Instant, CollectedMetrics> deserializeEvaluatedMetrics(
     }
 
     private static String serializeParallelismOverrides(Map<String, String> overrides) {
-        return ConfigurationUtils.convertValue(overrides, String.class);
+        return ConfigurationUtils.convertValue(overrides, String.class, false);
     }
 
     private static Map<String, String> deserializeParallelismOverrides(String overrides) {
```
mateczagany, line 313: If the operator is configured with a legacy `flink-conf.yaml`, state written by older operator versions in standard YAML form would fail to parse as `standardYaml` would be set to false in `GlobalConfiguration`.

By setting `standardYaml` to true for `convertValue` calls inside the deserialization, it would be able to parse standard YAML, and fallback to the legacy parser if that fails.

Quite the edge case, as one would need to revert from a standard YAML config to a legacy one, but I think the fix is quite simple.

Dennis-Mircea: The operator always mounts the config.yaml.

PR #1198, opened 2026-08-28
[hotfix] Remove redundant autoscaler Plugin discovery

Dennis-Mircea:

The purpose of this PR is to remove the redundant discovery for both FlinkDeployment and FlinkSessionJob, and instead of this to adopt just one global discovery.

## Brief change log

- Plugin discovery was moved out from `AutoscalerFactory` to the FlinkOperator constructor.
- All 3 current plugins are wrapped inside the `AutoscalerPlugins` class, which makes extensibility easier in the future for new plugins that may be added.

## Verifying this change

This modification is already covered by the existing tests and was also tested against a real autoscaling example.

## Documentation

---


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/AutoscalerPlugins.java
```diff
@@ -0,0 +1,60 @@
...
+
+/** The autoscaler SPI implementations discovered from the operator's plugin manager. */
+@Getter
+public class AutoscalerPlugins {
+
+    private final Collection<ScalingMetricsEvaluatorPlugin> evaluators;
+    private final Collection<ScalingExecutorPlugin<ResourceID>> scalingExecutors;
+    private final Collection<ParallelismAlignmentMode> alignmentModes;
+
+    private AutoscalerPlugins(
```
spuru9: Because this constructor is `private`, any test that wants to instantiate `AutoscalerPlugins` (or test `AutoscalerFactory` with
  mock/stub plugins) is forced to spin up a real `PluginManager`:
```java
    AutoscalerPlugins.discover(OperatorPluginUtils.createPluginManager(new Configuration()))
```

  Consider:

  1. Making the constructor public (or package-private / `@VisibleForTesting`) and adding null-safe fallbacks (Collections.emptyList()).
  2. Providing an empty() factory method:

        public static AutoscalerPlugins empty() {
            return new AutoscalerPlugins(
                    Collections.emptyList(), Collections.emptyList(), Collections.emptyList());
        }

  This will significantly simplify unit testing across the codebase

spuru9 (approved):
LGTM +1

gyfora (approved):

PR #1199, opened 2026-08-28
[FLINK-40467] Randomise application result store paths

dalelane:

Flink 2.3 gives every application-mode execution an ApplicationResultStore, which defaults to
`{high-availability.storage-path}/application-result-store/{ha.cluster-id}`. That path is consistent across deployments (and not included in HA cleanup) so entries written by a terminated application outlive the upgrade that replaced it.

SessionDispatcherLeaderProcess recovers dirty results unkeyed, so the replacement cluster picks up the previous deployment's entry, and Dispatcher#maybeSubmitApplicationInApplicationMode then deliberately submits nothing when execution.shutdown-on-application-finish is false. This results in healthy JobManagers with no jobs.

## Brief change log

This commit makes the storage path unique on each cluster launch so nothing stale can be recovered.

## Verifying this change

Added a unit test that recreates the scenario

## Documentation


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
```diff
@@ -273,6 +285,24 @@ private static void setRandomJobResultStorePath(Configuration effectiveConfig) {
         }
     }
 
+    private static void setRandomApplicationResultStorePath(Configuration effectiveConfig) {
+        if (effectiveConfig.contains(HighAvailabilityOptions.HA_STORAGE_PATH)) {
```
james-kan-shopify, line 289: Hi! I've encountered the same bug, and have been working in a fork regarding this. Thanks for opening up a PR here! 

Suggesting to consider also setting `application-result-store.delete-on-commit=false`, mirroring`setRandomJobResultStorePath`.

`maybeSubmitApplicationInApplicationMode` skips submission on a clean result entry, but per its javadoc that lookup only works when `DELETE_ON_COMMIT=false`. Since the operator sets `shutdown-on-application-finish=false`, a terminal app's JM lives on until the shutdown TTL, and a restart in that window re-bootstraps it: `main()` re-runs, and multi-job apps fail outright. FLINK-27569 one level up.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
```diff
@@ -273,6 +285,24 @@ private static void setRandomJobResultStorePath(Configuration effectiveConfig) {
...
+            // The application result store shares the HA storage path, which outlives
+            // the HA metadata deleted on upgrade. Dirty entries are recovered unkeyed,
+            // so a terminal entry left behind by a previous deployment makes the
+            // replacement cluster skip submitting the new job.
+            // Giving each deployment a unique path makes sure that nothing stale is
+            // recovered.
+            effectiveConfig.set(
```
james-kan-shopify, line 296: Curious, if you think it is worthwhile to gate this behind a version check? Since this is unique to jobs post 2.3, we'd be unnecessarily creating entries in storage. Given that this fix addresses what may be a cleanup race condition under the hood, but isn't directly resolving the race, it may be ideal to minimize it to 2.3 onwards?

james-kan-shopify (commented):
thanks for opening this up! happen to encounter this and attempting to patch the same thing here.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
dalelane: I didn't think this was necessary - the config option was only added in FLINK-38975 for Flink 2.3, so in earlier versions of Flink, isn't this non-existent config value just going to be silently ignored?

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/ApplicationReconciler.java
dalelane: I do think that's worth fixing, but IMO it's a separate enough issue that it'd be better in a separate pull request (maybe a separate Jira bug issue too)

gyfora: If we need delete on commit false here as well same way as the job result store then we should fix it in the same PR / issue

dalelane: @gyfora I'm happy to do that if you're okay with the approach in the code here 

James fairly highlighted the resource leak implication of the approach I took [on the dev@ list](https://lists.apache.org/thread/2gs2xzzmh8371rnc2993o9bvjsm4n6bd) where I said:

> re: resource leak
> Yeah, I agree that what I'm proposing in FLINK-40467 doubles down on the deal made in FLINK-27573, and maybe that's not a good thing. I used a random UUID because that's what I saw setRandomJobResultStorePath do, but in hindsight making directories anonymous makes the admin cleanup task harder as you can't derive the deployment from a directory name to know what is safe to remove. Perhaps the operator should record the path it generates in status - this would let us automate cleaning up the path for previous launches?

if you want to discuss that aspect first, I'll hold off on adding to the PR. But if you're okay with the approach, I'll add the delete on commit

gyfora: I think we have the exact same result leak already for the job result store logic.  Not setting this config means that we have an inconsistent behaviour between different flink versions. We can consider how to avoid the resource leak as the next step but for now I think it makes sense to follow the currently established approach

dalelane: added in a0fe7fbbb9c76fdd3eba62c186f2fab3ea31f2cf

Dennis-Mircea (commented):
Thanks for the fix, @dalelane! I posted a detailed analysis on the dev thread (https://lists.apache.org/thread/tyw45k5ww3yzfmfkmhz241nm82swlrsm) arguing that the current issue, unlike FLINK-27569, is an ordering problem (the operator tears the JM down before Flink deletes the dirty application result) and can be solved without introducing a second permanent leak. I would suggest holding this PR until the direction is agreed.

gyfora:
Can you please rebase on main to pull in the e2e fix?

PR #1200, opened 2026-09-01
FLINK-40531 Serialize parallelism overrides in a version-portable format

nadberezny:
Flink can write map-valued config settings in two styles: the old a:1,b:2 and the newer {a: '1', b: '2'}. Every Flink version can read the old style; only newer parsers can read the new one.

The autoscaler lets Flink choose the style automatically, and that choice depends on which config file the operator itself was started with. So if the operator runs with config.yaml, it writes parallelism overrides in the new style into every FlinkDeployment it scales. Any JobManager still reading a legacy flink-conf.yaml then fails to start, because it cannot parse the value. The operator itself reports no problem — only the job breaks.

It also cannot recover by itself: the autoscaler reuses the previous override string whenever the overrides haven't changed, so once a bad string is written it stays there.

Fix: always write the old style, which works on all Flink versions, and rewrite the string if the existing one is unreadable.


Dennis-Mircea (changes requested):
This was already tackled as part of #1195.

PR #1201, opened 2026-09-01
[FLINK-40535] Standalone session clusters are redeployed instead of scaled in-place

Dennis-Mircea:

This PR fixes the 2 broken code pieces part of `StandaloneFlinkService#scale`:
- Wrong `JobSpec` extraction.
- Wrong `Integer` comparison.

Besides the above, the PR also fixes the redundant k8s deployment querying and better models the scaling tests that are backing this method. 

## Brief change log
- `JobSpec` is extracted now via `ctx.getResource().getSpec().getJob()`.
- `Integer` comparison is done now via `!Objects.equals(actualReplicas, desiredReplicas)`.
- `deployment.get()` is performed only once instead of twice.
- `scaling` tests are now parameterized with a proper mechanism of correctly setting the replica, instead of wrongly keep them on `null`.

## Verifying this change

Those changes can be verified by the existing tests.

## Documentation

---


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/StandaloneFlinkService.java
```diff
@@ -179,17 +180,17 @@ public boolean scale(FlinkResourceContext<?> ctx, Configuration deployConfig) {
         var name = StandaloneKubernetesUtils.getTaskManagerDeploymentName(clusterId);
         var deployment =
                 kubernetesClient.apps().deployments().inNamespace(namespace).withName(name);
-
-        if (deployment == null || deployment.get() == null) {
+        var deploymentCR = deployment.get();
+        if (deploymentCR == null) {
```
spuru9, line 184: Minor nit: `Deployment` is a standard built-in Kubernetes resource (`apps/v1`), rather than a "CR" (Custom Resource). Naming it `tmDeployment` or `deploymentResource` would be slightly more accurate, though this is purely stylistic.

spuru9 (approved):
LGTM. Just one optional nit.

PR #1202, opened 2026-09-02
[FLINK-40542] Block session cluster deletion when running jobs cannot be determined

spuru9:

`SessionReconciler.getNonTerminalJobs` returned `Set.of()` on any query failure, and `cleanupInternal` treats an empty set as "safe to delete". So a session `FlinkDeployment` with an unreachable JobManager got deleted anyway during cleanup, stopping unmanaged jobs without a checkpoint.

This makes the guard fail closed: an indeterminate result blocks deletion instead of assuming the cluster is empty. Per review feedback, it also bounds this and several other unbounded blocking REST calls in the operator and autoscaler with their configured client timeouts, so a hung JobManager can't stall them indefinitely.

## Brief change log

  - `getNonTerminalJobs` now returns `Optional<Set<JobID>>`; empty means "could not determine", distinct from a present empty set ("no jobs").
  - `cleanupInternal` blocks deletion and emits `CleanupFailed` on the indeterminate case, rescheduling until the JobManager is reachable — mirroring the block-on-session-jobs branch. Also closes the `RestClusterClient` via try-with-resources and bounds the query with the Flink client timeout.
  - A session cluster that never deployed is deleted immediately instead of blocking forever, mirroring `ApplicationReconciler`/`SessionJobReconciler`. Also null-guards the jobs list and restores the interrupt flag on `InterruptedException`.
  - Same bounded-timeout fix applied to `AbstractFlinkService#fetchSavepointInfo/#fetchCheckpointInfo/#fetchCheckpointStats/#populateStateSize` (`kubernetes.operator.flink.client.timeout`) and `ScalingMetricCollector#queryAggregatedMetricNames`, `RestApiMetricsCollector#queryAggregatedVertexMetrics/#queryJmMetrics/#queryAggregatedTmMetrics` (`job.autoscaler.flink.rest-client.timeout`).

Related: fail-open was introduced with block-on-unmanaged-jobs (FLINK-28648); the new block lifts automatically so it doesn't reintroduce the deadlock fixed in FLINK-39618. The existing opt-out (`block-on-unmanaged-jobs`/`block-on-session-jobs` = false, FLINK-39432) still allows force-deleting a permanently unreachable cluster.

## Verifying this change

  - `SessionReconcilerTest`: `testCleanupBlocksWhenNonTerminalJobsCannotBeDetermined` and `testCleanupDeletesImmediatelyWhenNeverDeployed`.
  - Manually verified on minikube: deleting a session cluster with the JobManager down is blocked with a `CleanupFailed` event and completes once reachable.
  - The additional timeout-bounding changes are mechanical (no happy-path behavior change) and covered by the existing `AbstractFlinkServiceTest`, `ScalingMetricCollectorTest`, `RestApiMetricsCollectorTest`, `JobAutoScalerImplTest` suites.

  - Dependencies: no
  - Public API / `CustomResourceDescriptors`: no
  - Core observer or reconciler logic: yes (session cleanup, checkpoint/savepoint fetch, autoscaler metric collection)

## Documentation

  - New feature? no

---

- [X] Yes (Claude Code)

Generated-by: Claude Code (Claude Opus 4.8)

### DIFF
--- a/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/SessionReconciler.java
+++ b/flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/SessionReconciler.java
@@ -44,6 +44,7 @@ import org.slf4j.LoggerFactory;
 
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
 /**
@@ -143,39 +144,55 @@ public class SessionReconciler
 
     // Detects jobs which are not in globally terminated states
     @VisibleForTesting
-    Set<JobID> getNonTerminalJobs(FlinkResourceContext<FlinkDeployment> ctx) {
+    Optional<Set<JobID>> getNonTerminalJobs(FlinkResourceContext<FlinkDeployment> ctx) {
         LOG.debug("Starting nonTerminal jobs detection for session cluster");
-        try {
-            // Get all jobs running in the Flink cluster
-            var flinkService = ctx.getFlinkService();
-            var clusterClient = flinkService.getClusterClient(ctx.getObserveConfig());
+        var flinkService = ctx.getFlinkService();
+        try (var clusterClient = flinkService.getClusterClient(ctx.getObserveConfig())) {
             var allJobs =
                     clusterClient
                             .sendRequest(
                                     JobsOverviewHeaders.getInstance(),
                                     EmptyMessageParameters.getInstance(),
                                     EmptyRequestBody.getInstance())
-                            .get()
+                            .get(
+                                    ctx.getOperatorConfig().getFlinkClientTimeout().toMillis(),
+                                    TimeUnit.MILLISECONDS)
                             .getJobs();
 
-            // running job Ids
+            if (allJobs == null) {
+                return Optional.of(Set.of());
+            }
+
             Set<JobID> nonTerminalJobIds =
                     allJobs.stream()
                             .filter(job -> !job.getStatus().isGloballyTerminalState())
                             .map(JobDetails::getJobId)
                             .collect(Collectors.toSet());
 
-            return nonTerminalJobIds;
+            return Optional.of(nonTerminalJobIds);
         } catch (Exception e) {
             LOG.warn("Failed to detect nonTerminal jobs in session cluster", e);
-            return Set.of();
+            if (e instanceof InterruptedException) {
+                Thread.currentThread().interrupt();
+            }
+            return Optional.empty();
         }
     }
 
     @Override
     public DeleteControl cleanupInternal(FlinkResourceContext<FlinkDeployment> ctx) {
-        var sessionJobs = ctx.getJosdkContext().getSecondaryResources(FlinkSessionJob.class);
         var deployment = ctx.getResource();
+        var status = deployment.getStatus();
+
+        if (status.getReconciliationStatus().isBeforeFirstDeployment()) {
+            LOG.info("Session cluster was never deployed, deleting immediately");
+            var conf = ctx.getDeployConfig(deployment.getSpec());
+            ctx.getFlinkService()
+                    .deleteClusterDeployment(deployment.getMetadata(), status, conf, true);
+            return DeleteControl.defaultDelete();
+        }
+
+        var sessionJobs = ctx.getJosdkContext().getSecondaryResources(FlinkSessionJob.class);
 
         boolean blockOnSessionJobs =
                 ctx.getObserveConfig()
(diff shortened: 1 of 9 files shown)


Dennis-Mircea (changes requested):
While I do agree with the idea of this PR, the fix is too narrow, as there are multiple other REST service/autoscaler methods that must benefit from the same fix that you did here. 

Currently, there are the following operator-related methods that are not having the `kubernetes.operator.flink.client.timeout` set:
- `AbstractFlinkService#populateStateSize`
- `AbstractFlinkService#fetchCheckpointStats`
- `AbstractFlinkService#fetchCheckpointInfo`
- `AbstractFlinkService#fetchSavepointInfo`

Further, there are the following autoscaler-related methods that should have the `job.autoscaler.flink.rest-client.timeout` set:
- `ScalingMetricCollector#queryAggregatedMetricNames`
- `RestApiMetricsCollector#queryAggregatedVertexMetrics`
- `RestApiMetricsCollector#queryJmMetrics`
- `RestApiMetricsCollector#queryAggregatedTmMetrics`

spuru9:
> While I do agree with the idea of this PR, the fix is too narrow, as there are multiple other REST service/autoscaler methods that must benefit from the same fix that you did here.

Should I add those to this PR?

Dennis-Mircea:
> > While I do agree with the idea of this PR, the fix is too narrow, as there are multiple other REST service/autoscaler methods that must benefit from the same fix that you did here.
> 
> Should I add those to this PR?

Yes, and update the JIRA and PR title and description.

spuru9:
@Dennis-Mircea Have updated the PR and JIRA.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/SessionReconciler.java
```diff
@@ -143,39 +144,55 @@ private void recoverSession(FlinkResourceContext<FlinkDeployment> ctx) throws Ex
...
 
-            return nonTerminalJobIds;
+            return Optional.of(nonTerminalJobIds);
         } catch (Exception e) {
             LOG.warn("Failed to detect nonTerminal jobs in session cluster", e);
-            return Set.of();
+            if (e instanceof InterruptedException) {
+                Thread.currentThread().interrupt();
+            }
+            return Optional.empty();
```
Dennis-Mircea, line 178: I took a closer look on this one, and here by returning the `Optional.empty()` we cannot differentiate between a broken session JobManager and an unreachable session JobManager. Every failure lands in the same catch, so further, `cleanupInternal` has no way to tell a JobManager that is mid-restart from one that is crashlooping, in ImagePullBackOff, misconfigured, or whose Deployment was removed by hand. If we have a broken session JobManager (with or without unmanaged Flink session jobs), we end up in an endless loop with no cleanup, where the finalizer is never removed, the CR stays in Terminating, and the broken JM Deployment is never torn down.

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/reconciler/deployment/SessionReconciler.java
```diff
@@ -143,39 +144,55 @@ private void recoverSession(FlinkResourceContext<FlinkDeployment> ctx) throws Ex
...
         }
     }
 
     @Override
     public DeleteControl cleanupInternal(FlinkResourceContext<FlinkDeployment> ctx) {
-        var sessionJobs = ctx.getJosdkContext().getSecondaryResources(FlinkSessionJob.class);
         var deployment = ctx.getResource();
+        var status = deployment.getStatus();
+
+        if (status.getReconciliationStatus().isBeforeFirstDeployment()) {
```
Dennis-Mircea, line 187: This guard here doesn't seem to me to help and to tackle a broken session JobManager case. A broken session JobManager can most probably be deployed already (aka `lastReconciledSpec != null`) and this guard be bypassed entirely.

Dennis-Mircea (changes requested):
After taking a deeper look at this one, it feels to me that we are touching a too sensitive place in order to achieve something that was never promised. The main question that I have here is: Why do we want to complicate ourselves on this existent stable flow just to turn the unmanaged session jobs check into a hard guarantee? Why would someone deploy an unmanaged session job with the expectation for having it "managed"? The operator does not give that expectation anywhere else, because an upgrade-type spec change already deletes and recreates the session cluster without checking unmanaged jobs at all. The check was introduced in FLINK-28648 as best-effort, and returning an empty set on failure was consistent with that.

So, here we are trying to resolve partially, not even fully, something that was never guaranteed by the operator. In order to see if this makes sense to be adopted fully within the operator I'd recommend opening a discussion thread where we can brainstorm this.

Speaking about the rest of the changes, I'd resume this PR/changes only for timeout resolution for REST API calls.

PR #1206, opened 2026-09-07
[FLINK-40583] Filter operator keys on the remaining runtime config paths

Dennis-Mircea:

This PR fixes the possible exchange paths of the operator and autoscaler properties (`kubernetes.operator.*` and `job.autoscaler.*`) between the operator and the Flink cluster, which may break the current expected behavior or may expose unnecessary sensitivity.

## Brief change log

- `AbstractFlinkService` - Dropped the operator keys on the JM REST response and on the session job submission, reusing `isOperatorControlledKey` from #1178 instead of a second copy of the check.
- `FlinkRuntimeConfigurationUtils` - Made the `isOperatorControlledKey` public so both paths share one definition

## Verifying this change

This change can be verified with the current test and against a real live Flink cluster.

## Documentation

---


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/service/AbstractFlinkService.java
spuru9: minor: Now that isOperatorControlledKey is public, removeOperatorConfigs still has its own inline copy of the exact same prefix check right next to it — worth deduplicating since that's the stated point of the change.

```
    @VisibleForTesting
    protected static Configuration removeOperatorConfigs(Configuration config) {
        Configuration newConfig = new Configuration(config);
        for (String key : config.keySet()) {
            if (FlinkRuntimeConfigurationUtils.isOperatorControlledKey(key)) {
                newConfig.removeKey(key);
            }
        }
        return newConfig;
    }
```
(Note: this makes the K8S_OP_CONF_PREFIX static import unused — remove that line too.)

spuru9 (commented):
Looks good, one minor optional cleanup.

gyfora (approved):

gyfora:
can you please rebase for the latest e2e fixes?

Dennis-Mircea:
> can you please rebase for the latest e2e fixes?

Done!

PR #1207, opened 2026-09-07
[FLINK-40587] Recreating a FlinkDeployment throws an autoscaler exception

Dennis-Mircea:

This PR addresses the FLINK-40587 bug by:
- Making the ConfigMap cache retrieval rely on the resource UID comparison, so later updates will not be made to a stale ConfigMap that will raise the same exception as presented in the JIRA ticket.
- Taking over the ConfigMap left behind by the previous resource, by switching its ownerReference to the current resource and clearing its data, so the Kubernetes garbage collector will not delete it together with the autoscaler data written in the meantime.
- Recreating the ConfigMap from the in-memory state when the flush finds it already deleted, instead of failing with a 404 and losing the autoscaler cycle.

The data of the previous ConfigMap is dropped rather than taken over, because that state belongs to a resource that is gone and reusing it would make the result depend on whether the operator or the garbage collector gets there first.

## Brief change log

- `ConfigMapStore#getConfigMap` - compare the resource UID on cache access, so a recreated resource re-reads from Kubernetes instead of reusing the view of a collected ConfigMap.
- `ConfigMapView constructor` - capture the UID of the resource the view is created for, once, so the cache check does not re-fire on every access while the old ConfigMap is still in place.
- `ConfigMapView#takeOverIfLeftByPreviousOwner` - switch the `ownerReference` to the current resource and clear the data of a ConfigMap left behind by a previous one, in memory, persisted at flush time.
- `ConfigMapView#flush` - on a 404 from the update, drop the `resourceVersion` and `uid` and create the ConfigMap from the in-memory state. Any other failure is rethrown.
- `ConfigMapStore#flush` - treat a 404 that still reaches this point, meaning that the create itself failed, as expected, invalidate the cache, and log at WARN. Every other failure is still logged at ERROR and rethrown, so genuine problems keep raising `AUTOSCALERERROR`

## Verifying this change

The change can be verified by the newly added tests and by having a live FlinkDeployment with autoscaler enabled, over which delete and re-deploy are performed.

## Documentation

---


flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/state/ConfigMapStore.java
```diff
@@ -93,9 +102,19 @@ public void removeInfoFromCache(ResourceID resourceID) {
...
+    private static boolean isNotFound(Exception e) {
+        return e instanceof KubernetesClientException
+                && ((KubernetesClientException) e).getCode() == HttpURLConnection.HTTP_NOT_FOUND;
+    }
+
     private ConfigMapView getConfigMap(KubernetesJobAutoScalerContext jobContext) {
-        return cache.computeIfAbsent(
-                jobContext.getJobKey(), (id) -> getConfigMapFromKubernetes(jobContext));
+        var ownerUid = jobContext.getResource().getMetadata().getUid();
+        return cache.compute(
```
spuru9, line 112: major: cache.compute re-checks the owner UID on every call, not just once. If the old CM hasn't been GC'd yet after a recreate, every getConfigMap() call in the same reconcile sees a mismatch and re-fetches from k8s, discarding whatever was put() earlier in that same pass (since flush() only runs once at the end). E.g. put("x", ...) then put("y", ...) before the recreate's old CM is gone → x is silently lost. 

Might need the "is this cache entry stale" decision made once per ConfigMapView, not re-derived from the live owner ref on every access.

spuru9 (commented):
Thanks for the change.

Have a comment. Can you verify

flink-kubernetes-operator/src/main/java/org/apache/flink/kubernetes/operator/autoscaler/state/ConfigMapStore.java
Dennis-Mircea: This comment made me think one more time about the whole logic here, so I took a step back and reassessed everything one more time.

So, the assessment of the existent code within the operator is the following:
- The resourceRetriever (which talks directly with the K8s API server) is called on the first GET and cached afterwards, and on flush for any updated key for a different owner reference.
- When the actual resource was recreated, it didn’t depend on whether it was out of or mid-autoscaler reconciliation cycle, and the flush failed with an exception, discarding all the data that was updated within that cycle, and triggering a new recreation of the ConfigMap (based on clearing the cache) on the next cycle. Also, if the ConfigMap takes more than 1 cycle to be entirely deleted, it is continuously writing to a stale ConfigMap that will be deleted anyway.
- This approach, besides the actual exception and error event emitted, is also losing at least one autoscaler cycle of computations, which does have importance even if we are in the autoscaler metric stabilization window.

The assessment of the previous implementation that I did was the following:
- The resourceRetriever is called on every GET, and on flush for any updated key for a different owner reference. It is not cached because, when building the ConfigMapView it will see it as already existent and will use the existent ConfigMap data, which references the old owner reference, which causes no cache hit while the ConfigMap is not garbage-collected, bringing in place the following scenarios:
    - if the previous old ConfigMap is still in place for the whole cycle and gets recreated until the next one, we have the same data discarding/loss as before + no error + multiple Gets.
    - If the previous old ConfigMap is getting deleted mid-cycle, then we are at risk of discarding just a portion of the data (just some key-value pairs) while preserving the other, and this brings real inconsistency
    - If the previous old ConfigMap is getting deleted instantly before the cycle

So, even with the suggestion that you brought above, this will still cache the old ConfigMap data with the new owner reference while the old reference ConfigMap still being in place, which can still cause the flush at the end to fail with a 404. This will invalidate the cache and will trigger a new recreation of the ConfigMap (based on clearing the cache) on the next cycle. This will keep the same data discarding logic as before. Still, this approach is losing at least one autoscaler cycle of computations, so is not good enough.

Based on the above, I took a step further and I updated the implementation with the following logic:
- On each GET, if the owner reference is different, then get the existing ConfigMap (if it wasn’t garbage collected already) and take it over by updating the owner reference and clearing the data which is no longer relevant, by doing all this computation in memory and by updating the cache, which will no longer trigger a miss on the owner reference.
- At the flush time, try to update the physical ConfigMap if still present with the in-memory snapshot, if not, create it from scratch with the data collected so far.

To sum up, with the above updated approach, we will no longer be at risk of data inconsistency, we will have determinism, and we will not lose any autoscaler cycle of computations. I updated the PR content + description to reflect this.

gyfora (approved):

gyfora:
can you please rebase for the latest e2e fixes?

Dennis-Mircea:
> can you please rebase for the latest e2e fixes?

Done!

PR #1210, opened 2026-09-16
[FLINK-38798] Publish a Helm provenance file for the operator chart

Admaing:

Closes [FLINK-38798](https://issues.apache.org/jira/browse/FLINK-38798).

The released Helm chart archive (`flink-kubernetes-operator-<OPERATOR-VERSION>-helm.tgz`) was published
without a Helm provenance file, so users had no way to check the origin and the integrity of the chart
with the standard Helm tooling. The release directory already carries a GPG `.asc` signature for the
archive, but `.asc` is not the format `helm verify` / `helm install --verify` consume.

This change makes the release script publish `flink-kubernetes-operator-<OPERATOR-VERSION>-helm.tgz.prov`
next to the archive, so the chart can be verified with `helm verify`, `helm pull --verify` and
`helm install --verify`.

## Brief change log

- `tools/releasing/create_source_release.sh` now creates a Helm provenance file for the packaged chart,
  clear-signed with the same release key that already signs the `.asc` files.
- `SHA256SUM` is resolved alongside the existing `SHASUM`.
- The Helm operations page documents how to verify the released chart (en + zh).

## Verifying this change

There is no automated test coverage for the release script, so this was verified end to end manually.

Ran `tools/releasing/create_source_release.sh` with a test RSA release key, then:

- `helm verify --keyring flink.gpg flink-kubernetes-operator-9.9.9-helm.tgz` prints
  `Chart Hash Verified`;
- served the release directory over HTTP, added it with `helm repo add`, and
  `helm pull --verify --keyring flink.gpg localtest/flink-kubernetes-operator` succeeds as well. This
  exercises the real repository path, including the `files:` entry name matching the archive name that is
  actually downloaded;
- `helm repo index` produces a populated `index.yaml` whose `digest` matches the sha256 of the archive;
- appending a byte to the archive makes `helm verify` fail with `sha256 sum does not match`; removing the
  `.prov` file fails with `could not load provenance file`; an unrelated keyring fails with
  `signature made by unknown entity`;
- the Apache license header is still present in `Chart.yaml` inside the archive, so FLINK-27747 does not
  regress.

The documented keyring recipe was verified as well: `gpg --import` writes a keybox on GnuPG 2.1+, which
Helm cannot read (`openpgp: invalid data: tag byte does not have MSB set`), so the docs use a second
`gpg --export` step to produce a legacy binary keyring.

Note on the implementation: `helm package --sign` cannot be used here, because `helm package` rewrites
`Chart.yaml` and drops the ASF license header (see FLINK-27747), which is why the release script packages
the chart with `tar`. The provenance file is therefore produced with `gpg --clearsign`, using the message
format Helm expects.

## Documentation


Admaing:
Hi @gyfora, this PR is ready for review.

Closes [FLINK-38798](https://issues.apache.org/jira/browse/FLINK-38798) — it makes the release script publish a Helm provenance file (`.tgz.prov`) next to the chart, so users can run `helm verify` / `helm pull --verify` / `helm install --verify`.

- Docs + release script only, no runtime impact.
- Verified end to end with a throwaway signing key: `helm verify` and `helm pull --verify` both succeed, and tampered archives / missing `.prov` / unrelated keyrings are all rejected.
- CI on the latest commit (`7ee6be55`) is still `action_required` — it needs a maintainer to approve the workflow run.

Could you take a look when you get a chance? Thanks!

